PR #49, opened 2018-03-19
Remove PHP code in favor of JS

felixheidecke:
Remove all PHP Code from the frontend in favor of a pure JS implementation


DeepDiver1975:
this still uses php in make run. we might want to switch to some node server?????

DeepDiver1975:
screenshot from 2018-03-20 12-31-47

PR #72, opened 2018-04-04
Add eslintrc file

PVince81:
@felixheidecke please review

Note that you might need to install eslint + eslint-plugin-vue globally.
Then you can run `eslint App.vue` to check syntax.

Not sure how common IDEs do it (in vim it just runs eslint on CLI)

Not sure if we should provide these eslint packages in package.json, especially that we don't have a package.json for phoenix but for every app individually.


.eslintrc.json
```diff
@@ -0,0 +1,31 @@
+{
+    "env": {
+        "browser": true,
+        "es6": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:vue/essential"
```
felixheidecke: I'd recommend the essential preset. Hope it's not too forgiving for @PVince81 ?!

patrickjahns:
@felixheidecke 
please consider adding it to https://github.com/owncloud/coding-standard so we can deliver it as a standard to all our repositories via a npm (?) package

PVince81:
@felixheidecke please take over

PR #83, opened 2018-04-13
Add boot-loader for apps

felixheidecke:
Apps ship with `boot.js` to register themselves and navigation elements.


felixheidecke:
This is the interesting part:

https://github.com/owncloud/phoenix/blob/app-loader/apps/market/js/boot.js

config.json
```diff
@@ -0,0 +1,6 @@
+{
```
DeepDiver1975: we didn't want to check this one in. please adopt the sample config file

index.html
```diff
@@ -107,7 +107,11 @@
 		</div>
 	</main>
 
-	<script src="core/uikit/dist/js/uikit.min.js"></script>
-	<script src="core/js/core.bundle.js" defer></script>
+    <script src="core/uikit/dist/js/uikit.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/vue/2.5.16/vue.js"></script>
```
DeepDiver1975: cdn? why? nogo. think of offline instances and about privacy

config.json.sample
```diff
@@ -1,18 +1,6 @@
 {
 	"backendUrl": "http://localhost/owncloud",
-	"apps": {
```
DeepDiver1975: why drop apps? why own apps.json? One config to rule them all - right?

You wanted to reduce requests ;-)

src/default.js
```diff
@@ -34,14 +39,16 @@ OC = new Vue({
...
-				this.apps = apps;
-				deferred.resolve();
-			});
-
-			return deferred;
+			$.when(
+				$.getJSON('/config.json', (config) => {
+					this.config = config;
+				}),
+				$.getJSON('/apps.json', (apps) => {
```
DeepDiver1975: why two requests?

src/default.js
```diff
@@ -52,26 +59,31 @@ OC = new Vue({
 		 */
 
 		_setupApps () {
-			var deferred = $.Deferred();
 
-			_.map(this.apps, (app, name) => {
+			_.forEach(this.apps, (app, i) => {
 
-				requirejs([`${this.appPath}/${name}/js/boot.js`], ( boot ) => {
+				// TODO: Find better var name for 'foo'
```
DeepDiver1975: bar?

PR #120, opened 2018-05-15
Implement file browsing

OshanIvantha:
## Changes made

- Implement the navigation within directories
- Add back navigation
- Implement file loading on page reload

### DIFF
--- a/apps/files/src/components/Files-App.vue
+++ b/apps/files/src/components/Files-App.vue
@@ -122,7 +122,7 @@ export default {
 			this.$route.push()
 		},
 		toggleFavorite (item) {
-			this.files[item].stared = (this.files[item].stared) ? false : true;
+			this.files[item].stared = (!this.files[item].stared);
 		},
 		routerLink (item) {
 			this.$router.push({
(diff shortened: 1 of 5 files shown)


apps/files/src/components/Files-App.vue
```diff
@@ -1,257 +1,339 @@
...
+                    pathSplit = pathSplit.slice(1, pathSplit.length - 1);
+                    this.path = pathSplit
+                }
+
+                // List all files
+                OC.$client.files.list(absolutePath).then(files => {
+                    // Remove the root element
+                    files = files.splice(1);
+
+                    let delay = Math.random() * (1000 - 200) + 200;
```
felixheidecke: I wrote this to demonstrate a delay + the loading animation. We can gladly get rid of that :-)

apps/files/src/components/Files-App.vue
```diff
@@ -1,257 +1,339 @@
...
+                            });
+                        });
+
+                        this.self = files.self;
+
+                        this.loading = false;
+
+                        this.resetSelect();
+                    }, delay)
+                }).catch(error => {``
```
felixheidecke: Whats up with the ticks?

apps/files/src/components/Files-App.vue
```diff
@@ -1,257 +1,339 @@
...
+            },
+            typeOfFile(showHidden) {
+                showHidden = (typeof showHidden !== 'undefined') ? showHidden : false;
+                return _.filter(this.files, 'extension')
+            }
+        }
+    }
 </script>
 
 <style lang="less">
```
felixheidecke: I'd recommend against "larger" styling inside the vue files and rather have that move to a separate Less file, but we can do this in the future.

apps/files/src/components/Files-App.vue
OshanIvantha: Oh typo I guess.

apps/files/src/components/Files-App.vue
OshanIvantha: They were there initially. I will start separating them.

PR #126, opened 2018-05-22
Improve app-loading and extensions

felixheidecke:
## App-Loading

Return most basic app info on setup instead of a separate object. (see boot.js)
Less code, more control!

## Extending

Providing and requesting extension points no matter the loading order:

### OC.$extend.request:

```
OC.$extend.request('appId', 'topbar-right').then( payload => {
	// Do magic when extension-point is ready
});
```

### OC.$extend.provide:

Provide an extension point from within an app, referencing a method that runs when an extension point is requested.

`OC.$extend.provide('appId', 'topbar-right', this.extendTopbarRight);`


src/default.js
```diff
@@ -103,34 +94,34 @@ OC = new Vue({
...
-					this.apps[i] = _.assignIn(defaults, foo.info);
-
-					// inject self
-					foo.setup(foo).then(() => {
-						if (this.apps.length === ++i) {
-							this.$bus.emit('phoenix:apps-setup')
+				requirejs([this.appJS(app.id, 'boot')], ( a ) => {
+
+					a.setup().then((appInfo) => {
+						if (app.id != appInfo.id) {
```
PVince81: use strict equality

in general if you use a linter like jslint, jshint or eslint it should tell you about this

PR #127, opened 2018-05-22
Material icons node dependency

felixheidecke:
Remove link to fonts.googleapis.com in favor of `material-icons` node dependency

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,16 @@ node_modules: package.json package-lock.json
 core/css/uikit.%.css: src/themes/%.less node_modules
 	node_modules/less/bin/lessc src/themes/$*.less core/css/uikit.$*.css --relative-urls
 
+core/css/material-icons.css: src/themes/icons.scss node_modules
+	node node_modules/.bin/sass src/themes/icons.scss core/css/material-icons.css --no-source-map
+
+core/fonts: node_modules
+	mkdir core/fonts
+
+core/fonts/MaterialIcons-Regular.%: core/fonts node_modules
+	cp node_modules/material-icons/iconfont/MaterialIcons-Regular.* core/fonts/
+
+
 core/js/core.bundle.js: node_modules
 	npm run build
 
@@ -26,11 +36,11 @@ core/js/core.bundle.js: node_modules
 # core
 #
 .PHONY: core
-core: core/js/core.bundle.js core/css/uikit.owncloud.css
+core: core/js/core.bundle.js core/css/uikit.owncloud.css core/css/material-icons.css core/fonts/MaterialIcons-Regular.%
 
 .PHONY: clean-core
 clean-core:
-	rm -rf core/js/core.bundle.js.* core/css/uikit.*.css
+	rm -rf core/js/core.bundle.js.* core/css/uikit.*.css core/fonts core/css/material-icons.css
 	rm -Rf node_modules
 
 #
@@ -52,4 +62,3 @@ clean-app-%:
 .PHONY: run
 run: build
 	php -S "$(SERVER_HOST)"
-
--- a/Makefile
+++ b/Makefile
@@ -17,13 +17,13 @@ node_modules: package.json package-lock.json
 	npm install && touch node_modules
 
 core/css/uikit.%.css: src/themes/%.less node_modules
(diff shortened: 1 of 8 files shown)


felixheidecke:
@DeepDiver1975 can you help move [the copy process](https://github.com/owncloud/phoenix/pull/127/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R12) to a nice method in the Makefile? … Pleeeeeeeease 😁

PVince81:
next, you need to split `npm install` in its own rule:
```
core-deps:
    npm install
```

then change core rule to just depend on the other rules
```
core: core-deps compile-icons compile-theme
```

DeepDiver1975:
@felixheidecke conflicting

PVince81:
Fixes https://github.com/owncloud/phoenix/issues/228

Makefile
```diff
@@ -19,18 +19,28 @@ node_modules: package.json package-lock.json
 core/css/uikit.%.css: src/themes/%.less node_modules
 	node_modules/less/bin/lessc src/themes/$*.less core/css/uikit.$*.css --relative-urls
 
+core/css/material-icons.css: src/themes/icons.scss node_modules
+	node node_modules/.bin/sass src/themes/icons.scss core/css/material-icons.css --no-source-map
```
PVince81: is the `node` in front of it required ? I thought one can just use directly whatever is in `.bin` (it might have a shebang)

PVince81: you can replace "core/css/material-icons.css" with `$@` which is an alias for the target rule if you like

Makefile
```diff
@@ -19,18 +19,28 @@ node_modules: package.json package-lock.json
...
+
+core/fonts: node_modules
+	mkdir core/fonts
+
+core/fonts/MaterialIcons-Regular.%: core/fonts node_modules
+	cp node_modules/material-icons/iconfont/MaterialIcons-Regular.* core/fonts/
```
PVince81: if you use the placeholder `%` in the rule name, you need to also use it somewhere here in this command as `$*`.

DeepDiver1975: well - I basically want all MaterialIcons-Regular.* to be copied over

what would be the best rule to write this?

_still asking myself if this is the right approach - can't webpack just grab that?_

PVince81: yeah, I suspect webpack should be able to grab and even compile some stuff directly without needing extra Makefile rules.

Makefile
```diff
@@ -26,11 +36,11 @@ core/js/core.bundle.js: node_modules
 # core
 #
 .PHONY: core
-core: core/js/core.bundle.js core/css/uikit.owncloud.css
+core: core/js/core.bundle.js core/css/uikit.owncloud.css core/css/material-icons.css core/fonts/MaterialIcons-Regular.%
```
PVince81: this might work only by chance

the "%" placeholder here will never have a value

I suggest just using "core/fonts" as dependency and have all the copying stuff done inside the "core/fonts" rule

DeepDiver1975:
why not use https://www.npmjs.com/package/vue-material-design-icons ?

DeepDiver1975:
> why not use https://www.npmjs.com/package/vue-material-design-icons ?

has hint for webpack - https://gitlab.com/robcresswell/vue-material-design-icons#tips

michaelstingl:
https://www.npmjs.com/package/vue-material-design-icons is MIT, but they're using https://github.com/Templarian/MaterialDesign , which is using [SIL OPEN FONT LICENSE](https://github.com/Templarian/MaterialDesign/blob/master/LICENSE). Does this work for us?

DeepDiver1975:
> https://www.npmjs.com/package/vue-material-design-icons is MIT, but they're using https://github.com/Templarian/MaterialDesign , which is using [SIL OPEN FONT LICENSE](https://github.com/Templarian/MaterialDesign/blob/master/LICENSE). Does this work for us?

@hodyroff

felixheidecke:
> why not use https://www.npmjs.com/package/vue-material-design-icons ?

The same reason why I don't fancy vueify right now. These would have to be compiled into each app instead of the theme.

DeepDiver1975:
Can't we Deal with this by some intelligent bundling?

PR #135, opened 2018-05-28
Implement file downloading

OshanIvantha:
This will only start working after the PR #66 of js-owncloud-client is merged

### DIFF
--- a/apps/files/src/components/File-Details-Buttonrow.vue
+++ b/apps/files/src/components/File-Details-Buttonrow.vue
@@ -6,7 +6,7 @@
         i.material-icons(class=a.class)= a.icon
 
     mixin iconButton(a)
-        button.uk-button(class='uk-button-' + a.buttonType, class=a.class, type="button").uk-flex.uk-flex-middle
+        button.uk-button(class='uk-button-' + a.buttonType, class=a.class, type="button", @click=a.click).uk-flex.uk-flex-middle
             if a.text
                 +icon({icon: a.icon, class: 'uk-margin-small-right'})
                 span= a.text
@@ -16,7 +16,7 @@
     // --- Vue component -----------------------------------------------
 
     .uk-button-group.uk-margin-medium-top
-        +iconButton({class: 'uk-button-small', icon: 'file_download', buttonType: 'primary', text: 'Download' })
+        +iconButton({class: 'uk-button-small', icon: 'file_download', buttonType: 'primary', text: 'Download', click: 'download()' })
         +iconButton({class: 'uk-button-small', icon: 'more_horiz', buttonType: 'default'})
         div(uk-dropdown="mode: click")
             .uk-margin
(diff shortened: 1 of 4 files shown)


felixheidecke:
@OshanIvantha ping me if you need help resoling conflicts here.

felixheidecke:
Well … downloading works, but not as expected. Can the download stream already start instead of loading the blob into the browser and then trigger the download?

Currently, I have to wait for the browser to load the file.

_I hope my words didn't confuse you_

apps/files/src/components/File-Details-Buttonrow.vue
```diff
@@ -40,10 +40,31 @@
...
             return {
                 fileAction: null
             }
+        },
+        methods: {
+            download() {
+                console.log("cat")
+                console.log(this.files)
+                for(let i = 0; i < this.files.length; i++){
+                    OC.$client.files.getFileContents(this.files[i].path).then(res => {
```
DeepDiver1975: This is Just wrong. Let the browser handle the download.

apps/files/src/components/File-Details.vue
```diff
@@ -1,131 +1,130 @@
 <template lang="pug">
-	// material icon mixin
```
DeepDiver1975: Please fix the whitespace Messer. Thx

apps/files/src/components/File-Details-Buttonrow.vue
OshanIvantha: Oh sorry I've missed some debug code.. 
Can you clarify on "Let the browser handle the download" please? I thought this is how we allow the download to happen through browser.

DeepDiver1975: Basically an anchor tag with the url to the File should be enough.

OshanIvantha:
@felixheidecke I don't think the js-owncloud-client API supports download streaming. 
https://github.com/owncloud/js-owncloud-client

apps/files/src/components/File-Details-Buttonrow.vue
OshanIvantha: That is what I have done. I have to call the *getFileContents* multiple times when the user has selected multiple files to be downloaded. As its current state, js-owncloud-client API doesn't support downloading multiple files with a single call. And then I let the browser handle each response using and anchor tag. 

If I am still not approaching that correctly, please advise.

apps/files/src/components/File-Details.vue
OshanIvantha: Fixed. File-Details.vue represents the same indentation in the master branch except for the lines I changed.

felixheidecke:
> I don't think the js-owncloud-client API supports download streaming.

Hmm … we need a solution for this. @DeepDiver1975 is there any other way to request a file via direct link + token?

felixheidecke:
Let's merge this for now and open a new PR concerning single and multi download.

DeepDiver1975:
> Let's merge this for now and open a new PR concerning single and multi download.

the url to be used for downloading is created here https://github.com/owncloud/js-owncloud-client/blob/a9e48343b140dd1f6fb0678be44f0f590392ccde/owncloud/fileManagement.js#L78

and yes - looks like there is a method missing to expose this link ... lets add this ...
open question: how do we pass the credentials/token?

DeepDiver1975:
oh ... seems like the solution is similar to what we already have implemented ... http://blog.bguiz.com/2014/07/03/file-download-with-http-request-header/

Issue #141, opened 2018-05-30
Ignore package-lock.json files


felixheidecke:
I suggest ignoring the package-lock.json files altogether. They do not make sense as we pre-compile and pack Applications anyway instead of having the user compile it for themselves.

PVince81:
@felixheidecke we need them for the build script that will build these pre-compiled packages so we have consistent lib versions and deterministic builds.

felixheidecke:
I need to look into suppress the creation of a package-lock.json file in a node development environment since it makes no sense here. For production, sure … locking packages make sense.

scramb:
As recommend in https://github.com/npm/npm/issues/18007
They suggest using [npm-merge-driver](https://npm.im/npm-merge-driver) package which tells git how to fast-forward merge the conflicts in package-lock. The other way to go is using yarn instead of npm. Since the yarn-lock file is not json, this feels easier to handle imo.

DeepDiver1975:
we anyhow wanted to switch to yarn ....

scramb:
@DeepDiver1975 should i add yarn to https://github.com/owncloud/phoenix/pull/286 or shall i prepare a new one with the whole process and yarn in another PR?

DeepDiver1975:
I'm fine putting this into #286 - THX

tempelgogo:
merged implicit with yarn migration https://github.com/owncloud/phoenix/pull/286

PR #159, opened 2018-06-15
acceptance tests on drone

individual-it:
run the UI test on drone and delete `.travis.yml`


individual-it:
@phil-davis it was updated when I run `make` locally so I committed it

DeepDiver1975:
> Is package-lock.json added intentionally?

I think we are not yet using this properly - see #141

PR #163, opened 2018-06-19
Single Vue Instance

felixheidecke:
Fixes https://github.com/owncloud/phoenix/issues/94

## Things to be done:

- [x] Add vuex and vue-router to Phoenix base
- [x] move demo-app to expose a vue component
- [x] remove "old" helper and extend classes
- [x] Try to teach vue-router to dynamically extend routes
- [x] Use events instead of watchers to allow for single- and unwatching functionality
- [x] Move config loading to the store
- [x] Employ webpack to create smaller chunks
- [x] Add VueTidyRoutes to collect routes before instantiating vue
- [x] Fix resolve errors on chaning routes
- [x] Adapt files app to new structure
- [x] Fix `@TODO`s
- … _work in progress_


felixheidecke:
Currently looking for a way to dynamically extend routes instead of defining them ahead of time.

PVince81:
In general I think "ahead of time" should be acceptable, because: if you'd write a separate vue webapp, as far as I understand, you anyway need to define your routes upfront before booting the instance.
So here in Phoenix you could have a callback (extension point) from the main router where apps will register their routes.

felixheidecke:
Not giving up dynamic route extending. Looking @ [vue-tidyroutes](https://github.com/edgarnadal/vue-tidyroutes) to move route models to their respective components. Let's see if this will do it :grimacing:

felixheidecke:
The Idea is to load the app bundle from the server as it is needed, not ahead of time. The router needs to have the component model ready as they are beeing passed to vue upon instatiation. So I would be rather nice not to load all the bundles before mount :-|

I fear that I need to further separate bundles using webpack so I can it's async import function as it seems to work hand in hand with async vue components. Giving it a try.

felixheidecke:
https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components

PVince81:
> So I would be rather nice not to load all the bundles before mount :-|

makes sense

Unless we split the bundle to have a "routes.js" part containing only the route definitions upfront.

We have something similar in the backend where a separate PHP file contains only the routes. Not sure if that makes sense.

felixheidecke:
Ha … I just had the best Idea ever. Event prevents ugly waiting and one less source for race conditions. Just wait a little bit.

felixheidecke:
The current state implements a way that allows us to load apps and plugins easily using the webpack chunking. This makes the extra boot.js files obsolete and bundles are loaded directly whereas chunks are loaded on demand.

@OshanIvantha @PVince81 please check out and test using `config.json.sample` as your config file.
You can switch between applications in the menu and see a cloud plugin in the upper righthand corner.

If you like the approach, I'll start to adapt the current state of the files app as well as iron out a couple of "ugly" parts to be able to merge this.

OshanIvantha:
@felixheidecke I get an error when I try to run this. 

screen shot 2018-06-28 at 12 01 53 am

apps/files/src/components/File-Details.vue
```diff
@@ -84,7 +84,7 @@
 		},
 		watch : {
 			file () {
-				OC.$bus.emit('files:file-details-update', this.file)
+				OC.$events.emit('files:file-details-update', this.file)
```
PVince81: do we still need these events ?

apps/servstat/src/app.js
```diff
@@ -0,0 +1,56 @@
...
+	computed : {
+		server () {
+			return this.$root.config.server;
+		}
+	}
+}
+
+const ip = {
+	name : "ServstatIp",
+	template : "<h1>Backend URL lautet {{ server }}</h1>",
```
PVince81: sudden German ?

src/default.js
```diff
@@ -1,288 +1,81 @@
...
+		}];
 
-		/**
-		 * Boot an application
-		 *
-		 * @param obj app with appId as key
-		 * @return Promise
-		 */
+		// @TODO: Need better implementation than [a, b, c, d, e, …]
+		_.forEach([a, b, c, d, e], (app) => {
```
PVince81: can you fix this right away ? at first glance it is very unclear what the letters are about

src/mixins/pluginHelper.js
```diff
@@ -0,0 +1,21 @@
...
+
+		/**
+		 * Path to application file
+		 *
+		 * @param exp Extensionpoint
+		 * @param plugin Plugin
+		 */
+
+		getPlugin (ext) {
+			return _.find( this.$root.plugins, ['extend', ext]);
```
PVince81: I wonder if this should be `this.$plugins` ? that would imply that accessing plugins after initialization is desired. probably not useful, so maybe keep it as is

apps/servstat/src/app.js
felixheidecke: Woops

apps/files/src/components/File-Details.vue
felixheidecke: Haven't touched the files app jet. (Except for some global find and replace stuff)

src/mixins/pluginHelper.js
felixheidecke: I'll try to prototype them, but right now these are given to the root element upon instantiation so in order to access them, we need to address `this.$root`.

DeepDiver1975:
@felixheidecke conflicts

individual-it:
I have the same problem as @OshanIvantha `127.0.0.1:35718 [404]: /apps/[object%20Object]/js/[object%20Object].bundle.js - No such file or directory` 
how did you fix it?

OshanIvantha:
@individual-it 
Using only `make` doesn't work.
You need to run `npm install` and `npm run build` on *demo* and *serverstat* apps.

individual-it:
@OshanIvantha then that might be the reason why the tests fail, need to adjust https://github.com/owncloud/phoenix/blob/master/.drone.yml#L60

felixheidecke:
@individual-it please check out and test using config.json.sample as your config file.

individual-it:
@felixheidecke OK got it running, but the UI changed quite a bit, so i will need to change the tests accordingly

individual-it:
tests works fine locally but fail on drone. see https://drone.owncloud.com/owncloud/phoenix/54/3
any idea?

individual-it:
OK test are modified and passing now

Issue #150, opened 2018-06-07
File List feature


PVince81:
Ticket to track requirements and progress on the file list feature.
A task is done when the file list was wired up already to the APIs (no dummy)

- [x] list files
- [ ] file actions: see https://github.com/owncloud/phoenix/issues/149
- [x] delete single file
- [x] delete multiple files
- [x] select/deselct multiple files for bulk actions
- [x] select/deselct all files for bulk actions
- [x] rename single file
- [x] create new folder
- [ ] create new file (TODO: ticket for extensibility of file formats like text editor, etc, like NewFileMenu in core) https://github.com/owncloud/phoenix/issues/274
- [x] move selected entries to subfolder with drag and drop
- [x] drag to breadcrumb to move to parent folder
- [x] breadcrumbs
- [x] file download
- [ ] file upload (with and without chunking) https://github.com/owncloud/phoenix/issues/67
- [ ] progress bar for operations
- [ ] spinners / state for all operations
- [x] sidebar (to be tracked in detail in a separate ticket)

@OshanIvantha @felixheidecke please update accordingly and link to tickets and PRs

PVince81:
https://github.com/owncloud/phoenix/issues/275

PR #165, opened 2018-06-27
File browser

OshanIvantha:
Implement moving files by drag and drop (#150)

- move selected entries to subfolder with drag and drop
- drag to breadcrumb to move to parent folder


apps/files/src/components/Files-App.vue
```diff
@@ -249,6 +253,17 @@
 			},
 			isChecked(item) {
 				return _.includes(this.selected, item);
+			},
+			onDrop(dropLocation, dropData, dragData, event) {
+                if(dropLocation === 'file-list' && dropData.type === 'folder'){
```
felixheidecke: Indent

felixheidecke:
This PR needs some adjustment to work with the new approach. Moving the store to the global store ect.

OshanIvantha:
@felixheidecke These implementations are reapplied in #216 
The issue of using vue-drag-drop plugin without attaching  it to the Vue instance has yet to be resolved.

PR #230, opened 2018-09-18
Improve makefiles

PVince81:
Don't repeat unneeded operations.

Only fetch deps when needed, not for every `make run`.

Please test if this works on your side as well.
`make clean` then `make run` fetches deps.
Next time `make run` directly starts the server, no npm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ SERVER_HOST=0.0.0.0:8300
 
 apps=files
 all_apps=$(addprefix app-,$(apps))
+core_bundle=core/js/core.bundle.js
 
 all: build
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,28 +13,37 @@ build: core $(all_apps)
 .PHONY: clean
 clean: clean-core $(addprefix clean-app-,$(apps))
 
+node_modules: package.json package-lock.json
```
PVince81: this rule tells "the target of this build is the 'node_modules' folder". so if this folder does not exist it will run it.

also, it will run it if the mtime of "package.json" or "package-lock.json" is more recent than the target "node_modules". so every time you change one of these, it will rerun this rule.

Makefile
```diff
@@ -12,28 +13,37 @@ build: core $(all_apps)
...
+	npm install && touch node_modules
```
PVince81: `touch` here makes sure that the mtime of the folder is the most recent as the folder might be created much earlier, this is only for accuracy

Makefile
```diff
@@ -12,28 +13,37 @@ build: core $(all_apps)
...
+
+core/css/uikit.%.css: src/themes/%.less node_modules
+	node_modules/less/bin/lessc src/themes/$*.less core/css/uikit.$*.css --relative-urls
+
+core/js/core.bundle.js: node_modules
```
PVince81: to be able to generate "core/js/core.bundle.js" we require  "node_modules" to exist, else we can't run npm.
this will trigger the "node_modules" rule above

Makefile
```diff
@@ -12,28 +13,37 @@ build: core $(all_apps)
...
+
 #
 # core
 #
 .PHONY: core
-core:
-	npm install
-	node_modules/less/bin/lessc src/themes/owncloud.less core/css/uikit.owncloud.css --relative-urls
-	npm run build
+core: core/js/core.bundle.js core/css/uikit.owncloud.css
```
PVince81: building core requires the core bundle and the uikit css file.

"uikit.owncloud.css" matches the rule "uikit.%.css" which replaces % (and $*) with "owncloud".
if we have more such files in the future and require them in the core build, they'd also match automatically

PR #232, opened 2018-09-23
Move drag-drop init before attempt to load config and init app

julijane:
Move the registration of the drag and drop components so that it is ensured that this is done before the app is initialized.


DeepDiver1975:
@julijane feel free to join our public chat channel https://talk.owncloud.com/channel/phoenix - looking forward to see you there

PR #241, opened 2018-09-25
Use ES2017 async/await pattern

julijane:
This change replaces the func().then(...) constructs with the async/await pattern introduced by ES2017. 

While it is not too obvious in this change, this makes the code more readable and more intuitively understandable.  A good example here is the login code where we do two asynchronous API calls in succession which in the original required nested "then".

This will not cause compatibility problems as async/await is syntactic sugar and will for the time being be transpiled by babel into more traditional code.


felixheidecke:
Haven't tested, but if it works its finde by me :-)

PR #255, opened 2018-10-03
l10n integration and core annotations

scramb:
Added l10n support by adding vue-gettext and easygettext dependencies.


src/default.js
```diff
@@ -29,6 +29,20 @@ import VueRouter   from 'vue-router';
...
+
+import GetTextPlugin from 'vue-gettext';
+import translations from '../l10n/translations.json'
+
+Vue.use(GetTextPlugin, {
+  availableLanguages: {
+  en_GB: 'British English',
+  de_DE: 'German',
+  },
+  defaultLanguage: 'en_GB',
```
DeepDiver1975: Generally speaking we consider en-US the default
the current language is selected from the browser settings. is that already in place?

src/default.js
```diff
@@ -29,6 +29,20 @@ import VueRouter   from 'vue-router';
 Vue.use(VueEvents);
 Vue.use(VueRouter);
 
+// --- Gettext ----
+
+import GetTextPlugin from 'vue-gettext';
+import translations from '../l10n/translations.json'
+
+Vue.use(GetTextPlugin, {
+  availableLanguages: {
```
DeepDiver1975: can we extract available languages from the translations.json?
otherwise we have to touch always two locations when adding a language.

tempelgogo: This information is hardcoded in the l10n/Makfile currently.
We could change this behaviour as soon as we use webpack as building system, because then this information is inside the javascript scope.

At this point it's a little tricky, because the vue-gettext plugin want's this information to map available locales to language names& to json content string.
We could simply parse all available languages from the json file on requiring it, but this wouldn't be a fast & clean solution.

src/default.js
tempelgogo: No, currently it's hardcoded english language as default.

I wasn't sure if we should take some language settings from the ownCloud Backend, but for sure, we will respect the browser settings for now.

DeepDiver1975: lets start off with the browser setting. THX

src/default.js
DeepDiver1975: ok - let's keep this for a future PR. THX

DeepDiver1975:
- when running make translations the generated translations.json looks different de <-> de_DE

screenshot from 2018-10-04 16-00-10

src/default.js
```diff
@@ -29,6 +29,28 @@ import VueRouter   from 'vue-router';
...
+<<<<<<< HEAD
```
DeepDiver1975: merge conflict :see_no_evil:

scramb: i tried to squash all those commits into the last 3 but this seems like it didn't work out well...

DeepDiver1975: worst case: start a new branch ;-)

scramb: okay will do next time

DeepDiver1975: please resolve it in here or recreate. let me know if you need support.
I honestly don't want to merge this in in the current state ;-)

PR #285, opened 2018-10-18
added eslint & initial fix

scramb:
See https://github.com/owncloud/phoenix/issues/268

added eslint with standard.js style // reuse of the existing config

small initial fix - rules can be tightened


scramb:
@DeepDiver1975 unknown pipeline error.

PR #302, opened 2018-11-02
Feature/oauth2 rebased

DeepDiver1975:
Requires:
- core master or stable10 branch
- oauth master

screenshot from 2018-10-11 13-15-44

screenshot from 2018-10-08 12-46-19

screenshot from 2018-10-08 12-46-29

screenshot from 2018-10-08 12-46-46


tempelgogo:
Login works, but the template explodes after login.
To be more accurate, the navBar hides & the console throws `TypeError: "_vm.user is undefined"`

DeepDiver1975:
> strange - it does did for me ....

ah - there is some initialization missing - this got lost because of the change to the auth storage ...

PR #303, opened 2018-11-05
refs #302 use oAuth config from vuex

tempelgogo:
* adjusted sample config, containing the oAuth api url & token
* inject config into vuex
* use oAuth config from vuex
* mutation names adopted

### DIFF
--- a/src/default.js
+++ b/src/default.js
@@ -73,6 +73,9 @@ Vue.component('drop', Drop);
                 redirect : to => arguments[0].navItems[0].route
             }];
 
+            // inject custom config into vuex base state
+            store.state.config.auth = config.data.auth
+
             for (let app of arguments) {
                 if (app.routes) routes.push(app.routes);
                 if (app.plugins) plugins.push(app.plugins);
--- a/src/services/auth.js
+++ b/src/services/auth.js
@@ -1,21 +1,25 @@
 import Axios from 'axios'
 import { VueAuthenticate } from 'vue-authenticate'
 
-const vueAuthInstance = new VueAuthenticate(Axios, {
-  providers: {
-    oauth2: {
-      name: 'ownCloud',
-      clientId: 'Cxfj9F9ZZWQbQZps1E1M0BszMz6OOFq3lxjSuc8Uh4HLEYb9KIfyRMmgY5ibXXrU',
-      url: 'http://deepdiver:8080/index.php/apps/oauth2/api/v1/token',
-      authorizationEndpoint: 'http://deepdiver:8080/index.php/apps/oauth2/authorize',
-      responseType: 'token',
-      responseParams: {
-        code: 'code',
-        clientId: 'client_id',
-        redirectUri: 'redirect_uri'
+export function initVueAuthenticate (config) {
+  if (config) {
+    return new VueAuthenticate(Axios, {
+      providers: {
+        oauth2: {
+          name: 'ownCloud',
+          clientId: config.clientId || '',
+          url: config.apiUrl || '',
+          authorizationEndpoint: config.authUrl || '',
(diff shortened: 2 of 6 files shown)


src/store/auth.js
```diff
@@ -1,59 +1,67 @@
...
+  },
+  getters: {
+    isAuthenticated: state => {
+      return state.isAuthenticated
+    }
   },
-
   mutations: {
-    isAuthenticated (state, payload) {
+    IS_AUTHENTICATED(state, payload) {
```
DeepDiver1975: Is this some vuex naming convention? Uppercase functions look :ghost:

src/services/auth.js
```diff
@@ -1,21 +1,25 @@
...
-      clientId: 'Cxfj9F9ZZWQbQZps1E1M0BszMz6OOFq3lxjSuc8Uh4HLEYb9KIfyRMmgY5ibXXrU',
-      url: 'http://deepdiver:8080/index.php/apps/oauth2/api/v1/token',
-      authorizationEndpoint: 'http://deepdiver:8080/index.php/apps/oauth2/authorize',
-      responseType: 'token',
-      responseParams: {
-        code: 'code',
-        clientId: 'client_id',
-        redirectUri: 'redirect_uri'
+export function initVueAuthenticate (config) {
+  if (config) {
```
DeepDiver1975: in case config is missing we need to render some error page or something ....

src/services/auth.js
```diff
@@ -1,21 +1,25 @@
...
+    return new VueAuthenticate(Axios, {
+      providers: {
+        oauth2: {
+          name: 'ownCloud',
+          clientId: config.clientId || '',
```
DeepDiver1975: I think there are no reasonable default - if not set -> :boom:

src/store/auth.js
```diff
@@ -1,59 +1,67 @@
...
           displayname: res['display-name'],
           email: Object.keys(res.email).length === 0 ? '' : res.email
         })
       })
     },
 
-    authenticate (context, payload) {
+    authenticate(context, payload) {
+      // TODO shouldn't we use vue-authenticate global & inject it into vue ?
+      vueAuthInstance = initVueAuthenticate(context.rootState.config.auth)
```
DeepDiver1975: context.rootState.config.auth is undefined ....

screenshot from 2018-11-05 14-43-07

src/default.js
```diff
@@ -73,6 +73,9 @@ Vue.component('drop', Drop);
                 redirect : to => arguments[0].navItems[0].route
             }];
 
+            // inject custom config into vuex base state
+            store.state.config.auth = config.data.auth
```
DeepDiver1975: ah - config.data is a string - this is not yet parsed json .... because of that auth does not work ...

DeepDiver1975: local issue with my config.json -> ignire

src/store/auth.js
tempelgogo: internal usage:
https://github.com/serv3r/phoenix/blob/master/apps/files/src/store.js
external recommendation:
https://vuejs.org/v2/style-guide/#Non-flux-state-management-use-with-caution

src/services/auth.js
tempelgogo: Is there already a route or component for the error page?

src/services/auth.js
tempelgogo: https://github.com/owncloud/phoenix/pull/303#discussion_r230759679

src/services/auth.js
DeepDiver1975: we don't have something like this - I'll open a issue for this ... THX

PR #306, opened 2018-11-06
Fix webUI accepectance tests to run tests locally

paurakhsharma:
# Description
Fixes webUI accepectence tests to run tests locally. 
Now, there will be no need to run `selenium-server` separately.

### DIFF
--- a/tests/acceptance/stepDefinitions/loginContext.js
+++ b/tests/acceptance/stepDefinitions/loginContext.js
@@ -6,8 +6,6 @@ Given(/^the user has browsed to the login page$/,
 		const loginPage = client.page.loginPage();
 		return loginPage
 		.navigate()
-		.waitForElementVisible('@openLoginDialogButton', 1000)
-		.click('@openLoginDialogButton')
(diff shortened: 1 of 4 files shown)


tests/acceptance/stepDefinitions/loginContext.js
```diff
@@ -6,8 +6,6 @@ Given(/^the user has browsed to the login page$/,
 		const loginPage = client.page.loginPage();
 		return loginPage
 		.navigate()
-		.waitForElementVisible('@openLoginDialogButton', 1000)
```
DeepDiver1975: why are those not needed any more?

paurakhsharma: Because, currently as soon as the user visits phoenix link, the login dialog box opens. That line waits for login button on top-right to be visible and tries to click on it.

DeepDiver1975: ok - well this will change soon anyway ... THX

PR #307, opened 2018-11-06
Fix acceptance tests for oauth login

DeepDiver1975:
files list is not opening because of a current issue we are facing within the router and the auth state ...

PR #310, opened 2018-11-07
oAuth2 fix - still need to reload page or authenticate 2 times

scramb:
acceptance tests worked with local tweeks should be working in drone aswell...


.drone.yml
```diff
@@ -83,6 +83,7 @@ services:
       - npm run build
       - cd /var/www/owncloud/phoenix
       - cp tests/drone/config.json config.json
+      - cp tests/ tests/
```
DeepDiver1975: ??????

tests/acceptance/features/login.feature
```diff
@@ -9,7 +9,9 @@ Feature: login users
 
 	Scenario: admin login
 		Given the user has browsed to the login page
-		When the user logs in with username "admin" and password "admin" using the webUI
+		When the user clicks the authenticate button
```
DeepDiver1975: this is from #307 - please rebase on that PR and let this PR merge into #307 - THX

PR #311, opened 2018-11-07
Feature serv3r oauth2 rebased

scramb:
made a little cleanup on the store and config init


tests/drone/wait-for-phoenix.sh
```diff
@@ -0,0 +1,7 @@
+curl -s phoenix:8300 > /dev/null
```
DeepDiver1975: we dont need this file .... I'm using a built-in script now ...

scramb: removed it - your right. I missed that in rebase

DeepDiver1975:
Tested - works - awesome!

PR #319, opened 2018-11-12
Introducing Themability & Vuetify

### LINKED ISSUE #228: Don't load anything from the Internet
We should bundle everything:
- material icons, etc

Currently it connects to remote sites like google for downloading these.

@felixheidecke

scramb:
## Description
Introduced themability with vuetify and built example loginPage. Theme will be selected as before from `config.json` by name. `themeName.json` in `phoenix/themes` folder will be loaded to vuex store to make it accesible by every component.
Material-Icons and Vuetify get bundled with the inline styles by webpack to one core.css file.

 ## Related Issue
- closes #228
- closes #127 
- refs #1  

 ## Motivation and Context
This PR could resolve all in #1 requested features and will help us to build a straight forward themeable design.

 ## How Has This Been Tested?
- developmentMode:
    - `yarn run watch fe.serv3r.cloud:8300` on localhost connected against ownCloud X running on a dedicated machine.
    - changed colors from ownCloud blue to yellow and back
- dist-output:
    -  `dist` output run with python as web-server connected against same ownCloud X
    - changed colors from ownCloud blue to yellow and back

 ## Screenshots:
2018-11-14-104652_1712x1006_scrot
2018-11-14-104549_1710x1007_scrot

 ## Types of changes
- [x] Technical debt
- [ ] Tests

 ## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised:


DeepDiver1975:
Theming works really nice

screenshot from 2018-11-13 17-00-56

PVince81:
@scramb mind reformatting the PR description ?

The PR subject should be a text that could be included as is in a changelog.

The first post of the PR should explain what the PR is about, and can include "fixes XYZ", etc.

From reading the code it seems to load the theme JSON but not sure what it does with it and why it works. A summary would be nice. Thanks!

@DeepDiver1975 introduce PR template ?

scramb:
> @scramb mind reformatting the PR description ?
> 
> The PR subject should be a text that could be included as is in a changelog.
> 
> The first post of the PR should explain what the PR is about, and can include "fixes XYZ", etc.
> 
> From reading the code it seems to load the theme JSON but not sure what it does with it and why it works. A summary would be nice. Thanks!
> 
> @DeepDiver1975 introduce PR template ?

I adjusted title and description of the PR

DeepDiver1975:
github templates are here #322

PVince81:
Will it work with separate apps ? Does the theme need to be recompiled for every app or can it be used ?

src/pages/error.vue
```diff
@@ -1,28 +1,29 @@
...
-        </div>
-    </div>
+  <div>
+    <v-layout class="background primary">
+      <v-img
+      v-if="configuration.theme.logo.big"
+      :src="configuration.theme.logo.big"
+      width="50%"
+      :aspect-ratio="1.96">
+      <v-flex class="theme--light grey lighten-2 pa-5 ma-3 center-dialog" xs4>
```
PVince81: "pa-5" and "ma-3" are these hard-coded values ?

what if we want to change the theme to have a different padding for these kind of components ?
might make more sense to have something like "pa-small", "pa-medium", "pa-large"

src/pages/error.vue
```diff
@@ -31,19 +32,30 @@ export default {
...
 </script>
 
 <style scoped="true">
-    .background {
-        height: 100%;
-    }
+  .background{
+    height: "100%";
+  }
+  .center-dialog {
```
PVince81: shouldn't this be a generic CSS class for centering ?

src/pages/error.vue
scramb: https://vuetifyjs.com/en/layout/spacing - this are the spacing helpers - they can be used like {property}{direction}-{size}

src/pages/error.vue
scramb: actually it should be centered by justify-content-center but this didn't work as expected. In this case i could trace back the problem to uikit - so this can be resolved as soon as we can get rid of uikit.

scramb:
> Will it work with separate apps ? Does the theme need to be recompiled for every app or can it be used directly ?

The theme is injected into vue instance and can be used from the apps aswell. Since we're loading the config from vuex we don't need to recompile

src/pages/error.vue
PVince81: mind adding a "FIXME" comment in the code so that the next passersby will not wonder ?

src/pages/error.vue
PVince81: As I understand, in some cases one needs to use spacing helpers on top of existing predefined padding/margin defaults. But here I'm surprised that we have to use it that often.

So spacings are not defined as part of components ? Does `v-btn` have default paddings as part of the theme or do I need to explicitly set them every time I define a button ?

src/pages/error.vue
scramb: added FIXME

src/pages/error.vue
scramb: We could set this globally and override if needed i adjusted this

PVince81: as far as I understand, UIkit should not interfere with existing elements as it's "opt-in", you need to explicitly set the classes. So I don't think it''s a side effect.

You guys have decided to go with "pa-2" for whatever reason, maybe it was visually fitting. If possible it would be good to bring this to the default as soon as possible as I'm afraid that the markup becomes unmaintainable if we start adding such exceptions everywhere.

I'm fine merging this PR first then please address this as next task.

PR #411, opened 2018-12-14
Improve security by detecting expired oauth token in vue-authenticate wrapper

tempelgogo:
## Description

improve security by detecting invalid oauth token.

improve dataflow in vuex auth wrapper, derive default auth state to provide fallback values instead of re-init the vue-authenticate lib inside getters

## Motivation and Context
security fixes & cleanup in auth service wrapper

## How Has This Been Tested?

- login
- delete vue auth cookie
- reload

## Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

## Checklist:

- Code changes
## (Possible) Tasks:
- [ ] re-implement showNotification
- [ ] add Notification for logout case

### DIFF
--- a/src/store/user.js
+++ b/src/store/user.js
@@ -1,58 +1,47 @@
 'use strict'
 import initVueAuthenticate from '../services/auth'
+import router from '../router/'
 
 let vueAuthInstance
 
 const state = {
+  token: '',
   displayname: null,
   email: null,
   isAuthenticated: false
 }
 
 const actions = {
-  login (context, payload) {
-    if (vueAuthInstance !== undefined) {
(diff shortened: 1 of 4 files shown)


src/store/user.js
```diff
@@ -1,58 +1,47 @@
 'use strict'
 import initVueAuthenticate from '../services/auth'
+import router from '../router/'
```
DeepDiver1975: router in state? I did not expect that ... well ... ;-)

DeepDiver1975:
Nice improvement! THX

PR #421, opened 2018-12-17
Bugfix/some low hanging fruits

DeepDiver1975:
## Description

Some low hanging fruits have been addressed:
- when hovering the bread crumb a pointer is shown
- upload progressbar has no fixed width now - was giving trouble with longer file names
- fix logout even in case no user was logged in

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/store/user.js
+++ b/src/store/user.js
@@ -18,6 +18,11 @@ const actions = {
       context.commit('SET_USER', state)
       // force redirect to login page after logout
       router.push({ name: 'login' })
+    }).catch(() => {
+      // reset user to default state
+      context.commit('SET_USER', state)
+      // force redirect to login page after logout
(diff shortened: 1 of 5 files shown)


src/store/user.js
```diff
@@ -18,6 +18,11 @@ const actions = {
       context.commit('SET_USER', state)
       // force redirect to login page after logout
       router.push({ name: 'login' })
+    }).catch(() => {
+      // reset user to default state
+      context.commit('SET_USER', state)
```
tempelgogo: assuming logout() returns false; What to do ?
we could pull this out of this promise, because actually the logout action doesn't need any return value...

PR #430, opened 2018-12-20
[files] Fix error handling on upload and success handling in sub folder

DeepDiver1975:
## Description
1. Error handling on upload was not properly implemented
2. Successful upload into a subfolder was buggy as well

## How Has This Been Tested?
1. try to upload into a readonly folder or upload where the user has no quota - error notification shall pop up
2. try uploading into a subfolder. progess bar shall disappear and file shall be visible

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/FileUpload.js
+++ b/apps/files/src/FileUpload.js
@@ -8,7 +8,7 @@ class FileUpload {
 
   upload (file, additionalData = {}) {
     let xhr = new XMLHttpRequest()
-    xhr.responseType = 'json'
+    xhr.responseType = 'text'
 
     // Headers
     xhr.open(this.type, this.url + encodeURIComponent(file.name), true)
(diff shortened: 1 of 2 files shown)


apps/files/src/FileUpload.js
```diff
@@ -8,7 +8,7 @@ class FileUpload {
 
   upload (file, additionalData = {}) {
     let xhr = new XMLHttpRequest()
-    xhr.responseType = 'json'
+    xhr.responseType = 'text'
```
PVince81: is "text" for the case where the body is empty ?

DeepDiver1975: by setting 'text' the response body holds the xml response as plain text - we can later try to parse and understand it.

in case you set 'json' the response is empty because the body has wrong format

PVince81: ah, of course... it's XML and not JSON

Issue #482, opened 2019-01-10
CI/CD


DeepDiver1975:
### Expected behaviour
- each pull request shall be built and deployed
- each pr is testable on it's own
- after opening a PR the link to the test instance will be put into a comment

scramb:
I got some deployment ready, which is tbd. 
I'm still not in knowledge of the deployment process as is.
 I would rather suggest deploying on a PR basis with docker container to keep the maintenance afford as low as possible. The deployment to the Registry is no big issue; The trouble came up with spawning the Container on a different machine, then the deployment agent. In my opinion the easiest way to provide this functionality is to create a script(allready is created but needs to be adopted to fit to infrastructure) and alllow only the usage of this script and bind it to a ssh-key. 
This would provide the ability to spawn docker container of the same image with a different tag. The cleanup needs to be done by weekly cronjob or similar.
@DeepDiver1975 @tboerger any improvements on this?

DockerScript as example:
```
#!/bin/bash

export DRONE_PULL_REQUEST=$SSH_ORIGINAL_COMMAND

docker run --label traefik.docker.network=web --label traefik.enable=true --label traefik.basic.frontend.rule=${DRONE_PULL_REQUEST}.owncloud.team --label traefik.basic.port=8300 --label traefik.basic.protocol=https --expose=8300  --network="web" --network="default" --rm owncloud/phoenix:${DRONE_PULL_REQUEST} 
```

DeepDiver1975:
I'm totally out of this topic - the servers is @tboerger 's realm ...

PVince81:
we already have CI in place now with acceptance tests.

nice would be to have JS tests as well, separate topic: https://github.com/owncloud/phoenix/issues/2026

closing as done

tboerger:
CD is not done... we have not built any PR deployment so far.

PVince81:
do we want CD on master maybe ? daily ? or is weekly enough together with the weekly tarball release ? @LukasHirt @DeepDiver1975

tboerger:
we already got CD for the master branch at phoenix.owncloud.com, but AFAIK the others also would like to see a lightweight environment for every pull request that gets automatically destroyed later on.

DeepDiver1975:
Yes - the idea was to have all PRs deployed - but I honestly question the effort on this.
In 1 year of development we maybe had 2 or 3 PRs where this would have been valuable.

This can also be done in direct 1-to-1 sessions between dev and who-ever-whats-to-review

[...]

PR #499, opened 2019-01-15
Add simply Dockerfile - config.json has to be mounted

DeepDiver1975:
## Description
One docker image to rule them all .... well - Phoenix

## Related Issue
https://github.com/owncloud/storage/issues

## How Has This Been Tested?
- within docker-compose in https://github.com/owncloud/storage

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -69,6 +69,12 @@ pipeline:
       - cd /var/www/owncloud/phoenix
       - yarn run acceptance-tests-drone
 
+  build-docker-image:
+    image: toolhippie/docker:latest
+    pull: true
+    commands:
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
> looks good, but we should think about deploying those somewhere maybe only to a private registry and only those pushed to master to a public one? Just thinking in favor of #482 - so we could deploy them somewhere and just prune away all unnecessary images every week.

for the time being I only need this in a customer docker-compose scenario where I expect the dev to run docker-composer build ..... 

any future steps are to be discussed.

Issue #172, opened 2018-07-05
Search files


PVince81:
- [ ] filter files in current folder based on file name pattern
- [ ] search in files in multiple folders: integrate with https://github.com/owncloud/documentation/issues/4241 to provide the ability to search for files in multiple folders

It is not clear yet whether we still want to have a global search field at the top right which is supposed to search more than just files (also calendars, contacts) or have another field in the files app for files-only search.

DeepDiver1975:
- needs change in js-owncloud-client

DeepDiver1975:
>     * filter files in current folder based on file name pattern

filtering can be done pure client side since we hold the full file list of the current folder in the client



>     * search in files in multiple folders: integrate with https://github.com/owncloud/documentation/issues/4241 to provide the ability to search for files in multiple folders

this is to be done by using OwnCloud.files.search

DeepDiver1975:
There will be two issues on their own. one for filter and one for search

PR #513, opened 2019-01-17
Filter search term in open folder

### LINKED ISSUE #517: Filter files locally in the current loaded folder
# Scenario
```gherkin
Given a user has various files in a folder
When the user filters for <criteria>
Then only files/folders matching the <criteria> are displayed
 Examples:
    | criteria | 
    | mime-type is txt |
    | file is hidden |
    | file name contains 'abc' |
    | ... |
```

# Technical
- only client side operation
-- no server interaction

# UX
- we already have the filter button in th eupper right corner
- let's just reuse it

tempelgogo:
## Description
The second commit is the offline-search functionality for open folder.
To be able to reuse the SearchBar, i've designed some webpack alias magic to use inside app's build systems, visible in the first commit of this PR.

## Related Issue

- refs #172 
- close #517

## Motivation and Context
We need some way Interface in Phoenix which apps can use to import Models, Services, Mixins, etc...
One possible solution is to outsource them & publish them on npm, so we could import them in phoenix same as in apps. This may be an idea if our pattern library is available & grown, but we need some way to use them right now. To avoid proliferation of incorrect settings, i've created a little helper script located in phoenix root directory. Apps can wrap their webpack config up & it get's enriched with some nice path aliases.
I could adopt some more helper & mixin code if we got some interface for loading them in apps...

## Screenshots:
filenamefilter

## Types of changes

## Checklist:

- [X] Code changes
- [X] Documentation ticket referenced

## Open tasks:

- [x] adopt UI. it's a filter function 
- [ ] talk about the new path aliases

### DIFF
--- a/apps/files/src/components/FilesApp.vue
+++ b/apps/files/src/components/FilesApp.vue
@@ -21,7 +21,6 @@
         <oc-dialog-prompt :oc-active="createFile" v-model="newFileName"
                           ocTitle="Create new file ..." @oc-confirm="addNewFile" @oc-cancel="createFile = false; newFileName = ''"></oc-dialog-prompt>
                           <v-flex xs2>
-                            <search-bar @search="onSearch" :value="searchQuery" />
                           </v-flex>
           <v-menu offset-y >
           <v-progress-circular
@@ -94,11 +93,17 @@
           </span>
         </v-flex>
         <v-menu transition="scale-transition">
-          <v-btn slot="activator" flat><v-icon large>filter_list</v-icon></v-btn>
+          <v-btn slot="activator" flat @click="focusFilenameFilter"><v-icon large>filter_list</v-icon></v-btn>
           <v-list>
             <v-list-tile v-for="(filter, fid) in filters" :key="fid">
               <v-list-tile-title v-text="filter.name"></v-list-tile-title>
               <v-checkbox v-model="filter.value"></v-checkbox>
+            </v-list-tile>
+            <v-list-tile>
+              <v-list-tile-title>
+                <span v-translate>Name</span>
+              </v-list-tile-title>
+              <search-bar @search="onFilenameFilter" :value="fileFilterQuery" ref="filenameFilter" autofocus />
             </v-list-tile>
           </v-list>
         </v-menu>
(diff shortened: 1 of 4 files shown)


DeepDiver1975:
Nice start. THX
Searching in the currently folder (pure Client Side) is what i would Name "Filter". This is Something we need as Well but in a different context.

Search is What we consider a Server Side functionality and can search all Files of the User with the füll folder tree.

This is why the js-owncloud-client was extended with the search function.

tempelgogo:
> Nice start. THX
> Searching in the currently folder (pure Client Side) is what i would Name "Filter". This is Something we need as Well but in a different context.
> 
> Search is What we consider a Server Side functionality and can search all Files of the User with the füll folder tree.
> 
> This is why the js-owncloud-client was extended with the search function.

@DeepDiver1975 I've seen your search method, thanks for implementing it. It's the following task, but we still have to discuss some things about that.

But the first part of issue #172 is named `filter files in current folder based on file name pattern`, so it belongs to this context in my opinion.
Besides that i would like to use a reuseable SearchBar for your new search method, so we can use the same look&feel in apps. I felt like the mechanism to load them was missing, so i've done this example to inject path aliases into files-app webpack. Files App is now able to load from 'oc_components/Spinner', for example.
The magic is visible in https://github.com/owncloud/phoenix/pull/513/commits/4487d870e0c085ed48435dc296c77418e3fe91e8
But this way, we abstract all path's under our control, and can change them in phoenix to influence the next app builds.

It wasn't the best idea to put the alias-loader into this "search for open folder" named PR, but could you take a look at it anyway, please :) ?
Is this workaround a possible solution to expose our components to apps ?

DeepDiver1975:
> But the first part of issue #172 is named `filter files in current folder based on file name pattern`, so it belongs to this context in my opinion.

you are absolutly right - my fault - I should have taken more care here.
I'll split up the issue in two tickets.

DeepDiver1975:
@tempelgogo my approach on registering components and make them usable in apps: https://github.com/owncloud/phoenix/commit/36146aee750c431de600fe5d0755099047fbe91d#diff-9e4fc1110432e662806df44bc5b846a5

PR #549, opened 2019-01-24
Show nothing if the file size is unknown

DeepDiver1975:
## Description
In some cases the folder size needs to be computed in the backend and takes time (e.g. files external the user has to move into all folders).
Instead of showing 'Pending ...' which has no meaning to the user we simply show nothing.

## How Has This Been Tested?
- :hand: 

screenshot from 2019-01-24 10-21-04

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/mixins.js
+++ b/apps/files/src/mixins.js
@@ -6,7 +6,7 @@ export default {
   filters: {
     fileSize (int) {
       if (int < 0) {
-        return 'Pending ...'
+        return ''
       }
       if (isNaN(int)) {
         return '???'


hodyroff:
ok. Lets see what the feedback from the users looks like, they might tell us we should never show a size on folders (seems like they are used to this from windows file browser ...).

PR #553, opened 2019-01-25
Update to babel 7 and add browserlist to get support for Edge

DeepDiver1975:
## Description
- [x] Update to babel 7
- [x] introduce browserslist

## Related Issue
- #546 

## Screenshots
edge-works

## How Has This Been Tested?
- [x] :hand: with edge
- [ ] :hand: with ie

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] fix logo to be displayed in edge
- [ ] fix ie - error below

PR #567, opened 2019-01-30
@babel/core7 and @babel/polyfill, add Browserslist

scramb:
Originial PR #553 

Basically the same, lost the trace and did it again.


DeepDiver1975:
From my understanding you are missing one important point: 
https://github.com/owncloud/phoenix/pull/553/files#diff-659af4641d8d19d971c143583c5b437eR1

This is required from my understanding to pull in the proper polyfills ....

Anyway - I don't understand why you are redoing the while PR ...

scramb:
im not sure if this is correct as far as i got while reading in all of those polyfill things `@babel/polyfill` should include those and they get wrapped into bundle by babel-loader. I'll  close this one if we can fiddle this mess i made in your PR. I just got mad since i messed up your PR and wanted to fix this so badly.

DeepDiver1975:
I got one step closer in ma PR by removing axios which does not work on IE11 ..... rest should be fine afaik ....

scramb:
then i'll close this one - and we discuss the core-js thing further since this should be included by the right use of `@babel/polyfill`

PR #568, opened 2019-01-30
add yarn update and bump packages

scramb:
## Description

Bumped some dependencies and added a script `yarn update` which now updates core and app dependencies

## Related Issue

- Fixes #563
- Fixes #564
- Fixes #565
- Fixes #566 

## Motivation and Context

Easier to update without merging 100 of PR's for single dependencies

## How Has This Been Tested?

- :hand: 

## Types of changes

## Checklist:

- [x] Code changes


update.js
```diff
@@ -0,0 +1,41 @@
+var fs = require('fs')
+var chalk = require('chalk')
+var resolve = require('path').resolve
+var join = require('path').join
+var cp = require('child_process')
+var os = require('os')
+
+if(fs.existsSync('config.json')) {
```
DeepDiver1975: I See this really complicated. The build being dependent on a runtime config File feels so wrong.

Just Walk into all sub folders within apps

scramb: if i just walk into every app folder this would break build on apps that may be WIP or broken atm which still are in my working dir. this way i can exclude them to the minimum required to drone config.

DeepDiver1975: Still .... Im not convinced. Had enough issues already ....

PR #570, opened 2019-01-31
remove race condition in acceptance tests

freestate:
## Description
improve acceptance-tests

## Related Issue
about 70% of all drone build fails...

## Motivation and Context
force-push to restart the pipeline builds is no solution!
The race condition in our acceptance tests was really time-expensive for all of our developers, but i may fixed it.

## How Has This Been Tested?

- 42 times `yarn acceptance-tests`
- 23 times `yarn acceptance-tests` on @tempelgogo notebook

## Types of changes

- [X] Technical debt
- [X] Tests

## Checklist:
- [X] Acceptance tests modified

### DIFF
--- a/tests/acceptance/features/webUILogin/login.feature
+++ b/tests/acceptance/features/webUILogin/login.feature
@@ -12,6 +12,5 @@ Feature: login users
 		When the user clicks the authenticate button
 		And the user logs in with username "admin" and password "admin" using the webUI
 		And the user authorizes access to phoenix
-		And the user browses to the files page
 		Then the files table should be displayed
 		And the files table should not be empty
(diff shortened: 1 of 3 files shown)


tempelgogo:
@freestate could you please change the `fileRows` selector to match only the first occurence of a row?
Maybe via `nth-of-type()` ?

tests/acceptance/features/webUILogin/login.feature
```diff
@@ -12,6 +12,5 @@ Feature: login users
 		When the user clicks the authenticate button
 		And the user logs in with username "admin" and password "admin" using the webUI
 		And the user authorizes access to phoenix
-		And the user browses to the files page
```
tempelgogo: This line is interesting because it's not mentioned in the `loginContext.js`, if I re-insert it, the race condition re-appears. So I'm fine with removing it.

Issue #476, opened 2019-01-10
Conflict handling on upload


PVince81:
The old frontend had a conflict dialog whenever the user uploaded a file into a folder that already had such file in there.

The conflict dialog let the user pick between "keep current (don't overwrite)", "overwrite" or "auto-rename".

The dialog itself also handled multiple file uploads, so the user could solve conflicts for multiple files at a time.

In the context of Phoenix we need to think whether such pattern is still needed or whether there are better UX patterns to cover such use cases.

@DeepDiver1975

DeepDiver1975:
This has to be handled under different scenarios.

- Upload new file
- Upload new version of one file
- Upload multiple files (all new?)
- Upload a full folder

^ the conflict resolution is different under each of these scenarios

PVince81:
in the the old UI it was the same handler for all these

PVince81:
also note, the autorename, if we still want it, must be implemented on the client side as well.
in the old code we renamed then pinged the server until the 409 Conflict (or 412 precondition failed?) disappears

DeepDiver1975:
```gherkin
Scenario: User upload a new file
  Given the file hello.txt does not exist
  When the user uploads a file named hello.txt
  Then the file is uploaded and appears in the file list

Scenario: User uploads a file which already exists
  Given the file hello.txt already exists
  When the user uploads a file named hello.txt
  Then a notification is displayed that the file already exists and cannot be uploaded
  And the file will not be uploaded

Scenario: User uploads a file which already exists in the server but is not visible in the client
  Given the file hello.txt already exists on the server
  And the file does not exist on the client
  When the user uploads a file named hello.txt
  Then the file is uploaded to the server
  But the server response with 409/Conflict because the file exists

**Comment by scramb** [comment]
```
Scenario: User uploads a file which already exists in the server but is not visible in the client
  Given the file hello.txt already exists on the server
  And the file does not exist on the client
  When the user uploads a file named hello.txt
  Then the file is uploaded to the server
  But the server response with 409/Conflict because the file exists
```
not sure how this scenario will happen - i'll take care of the 2nd scenario, the first scenario should work as described

DeepDiver1975:
> not sure how this scenario will happen -

Steps to reproduce this:
1. open phoenix in browser tab A
2. open phoenix in browser tab B
3. upload hello.txt in tab A
4. DO NOT RELOAD TAB B
5. upload hello.txt in tab B
6. get conflict ...

[...]

PR #582, opened 2019-01-31
add conflict handling for existing files

scramb:
## Description

Add error handling for file upload. 
If file exists on Client and Server decline upload. Open notification with error message.

## Related Issue

- Refs #476

## How Has This Been Tested?

- :hand: 

## Types of changes

## Checklist:

- [x] Code changes

### DIFF
--- a/apps/files/src/components/FileUpload.vue
+++ b/apps/files/src/components/FileUpload.vue
@@ -11,6 +11,7 @@
 <script>
 import FileUpload from '../FileUpload.js'
 import { mapActions, mapGetters } from 'vuex'
+import { find } from 'lodash'
 
 export default {
   props: {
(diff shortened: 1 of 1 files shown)


apps/files/src/components/FileUpload.vue
```diff
@@ -34,18 +35,27 @@ export default {
...
       let files = e.target.files || e.dataTransfer.files
       if (!files.length) return
       for (let file of files) {
-        this.upload(file)
+        let exists = find(this.files, ['name', file.name])
+        if(!exists){
+          this.upload(file)
+        } else {
+          this.showNotification({
+            title: `Upload for ${file.name} failed - File allready exists`,
```
DeepDiver1975: We need to handle Translations here

apps/files/src/components/FileUpload.vue
```diff
@@ -52,7 +52,7 @@ export default {
           this.upload(file)
         } else {
           this.showNotification({
-            title: `Upload for ${file.name} failed - File allready exists`,
+            title: `${this.$gettext('Upload for')} ${file.name} ${this.$gettext('failed - File allready exists')}`,
             type: 'error'
```
DeepDiver1975: Interpolation required. Position of the variable in the string is different depending on the language

scramb:
little problem with interpolation this is not supported by easygettext which we use to extract. I'll take a look if i can maybe get this upstream a bit faster

DeepDiver1975:
Rebasing right now .... will merge right away ...

PR #588, opened 2019-02-02
close #516 search for files

### LINKED ISSUE #516: Search files
# Scenario
```gherkin
Given a user has files in a folder sturcture
When the user searches for hello
The the user will see all files with hello in the file name anywhere in the file structure
```

# Technical
- use js-owncloud-client to search on the server OwnCloud.files.search
- more information: https://doc.owncloud.org/server/developer_manual/webdav_api/search.html#search-files
- default behavior is to search for file names - if e.g. search engine apps are installed this can search the file content as well
- for now we ignore the filter-files api

# UX
- one search bar to rule them all ????
- ???

tempelgogo:
## Description
search for files in files app via ownCloud search.
It uses the search-bar component, which get's enriched with auto-trim functionality in this PR....
Because most search feature's don't accept empty strings as search pattern.

I had to hide LastEdit date & FileSize, because ownCloud does not return them in search.
Is this intended behaviour ?

I've implemented some regex workaround to generate the file extension from name, but i'm not able to differentiate between folder or file; So i use the File icon as fallback, even for folders.
My workaround enables the use of the custom file type icons.

## Related Issue

- Fixes #516

## Motivation and Context
global search is a nice thing

## How Has This Been Tested?

- open Files app
- type 'md'
- see markdown files

## Screenshots
filesearch

## Types of changes

## Checklist

- [X] Code changes

## Open tasks

Remove leading slash in file path ?

The current UX is miserable, the global search does not change the breadcrumb / path, the local folder search is hidden in filters, etc. We should discuss this further!

@DeepDiver1975 the init bug is still available, as discussed in talk #phoenix room.
Search bar has autofocus, so if you type fast enough, your search will crash.
Do you prefer to use your double-init-workaround or should I open an issue in  [js-owncloud-client](https://github.com/owncloud/js-owncloud-client) ? I prefer the second option.


DeepDiver1975:
Lets discuss this on monday.

DeepDiver1975:
We need to address following issues in js-owncloud-client:
- [ ] fix path
- [ ] add option to define properties which shall be returned
- [ ] fix the user id issue

To be addressed in phoenix:
- search only after the user pressed enter
- add a button to do the same ?

@tempelgogo

DeepDiver1975:
To fix the path please use this function: https://github.com/owncloud/phoenix/blob/7aee7fb20999fd741642a66c4bf1c49a9d2ff722/apps/files/src/store.js#L11

DeepDiver1975:
>     * add option to define properties which shall be returned

https://github.com/owncloud/js-owncloud-client/pull/157

DeepDiver1975:
>     * fix the user id issue

https://github.com/owncloud/js-owncloud-client/pull/158

@tempelgogo

DeepDiver1975:
@tempelgogo client is updated in master - https://github.com/owncloud/phoenix/pull/592

tempelgogo:
replaced by #607

PR #607, opened 2019-02-05
global file search

### LINKED ISSUE #516: Search files

tempelgogo:
## Description
search for files in files app via ownCloud search.
It uses the search-bar component, which get's enriched with auto-trim functionality in this PR....
Because most search feature's don't accept empty strings as search pattern.

I had to hide LastEdit date & FileSize, because ownCloud does not return them in search.
Is this intended behaviour ?

I've implemented some regex workaround to generate the file extension from name, but i'm not able to differentiate between folder or file; So i use the File icon as fallback, even for folders.
My workaround enables the use of the custom file type icons.

## Related Issue

- Fixes #516
- replace pr #588 

## Motivation and Context
global search is a nice thing

## How Has This Been Tested?

- open Files app
- type 'md'
- see markdown files

## Screenshots
filesearch

## Types of changes

## Checklist

- [X] Code changes

The current UX is miserable, the global search does not change the breadcrumb / path, the local folder search is hidden in filters, etc. We should discuss this further!

@DeepDiver1975 the init bug is still available, as discussed in talk #phoenix room.
Search bar has autofocus, so if you type fast enough, your search will crash.
Do you prefer to use your double-init-workaround or should I open an issue in  [js-owncloud-client](https://github.com/owncloud/js-owncloud-client) ? I prefer the second option.


apps/files/src/components/FilesApp.vue
```diff
@@ -236,7 +235,23 @@ export default {
       this.fileUploadProgress = progressTotal / this.inProgress.length
       return this.fileUploadProgress
     },
-
+    onSearch (searchQuery) {
+      this.searchQuery = searchQuery
+      // do not search for empty strings
+      if (!searchQuery) return
+      this.$client.files.search(searchQuery).then((searchedFiles) => {
```
DeepDiver1975: Dav props need to be addee here ...

tempelgogo: I wasn't sure about using them, because actually excluding them doesn't change anything.
I could omit `{http://owncloud.org/ns}size`, because in my case, the folder size (which is why we're using this prop, isn't it?) isn't visible anyway.
But I've used the default props from your test including the size prop.

We could provide some default list size via `state.config.defaultListSize` or something, but before then I don't know what's the best result size to use; So i prefer to use your hardcoded default value via providing `null` as second parameter.

DeepDiver1975:
Looks good for results in the same folder .

In case of results which live in sub folders we need to tweak this. The bahvior is currently a bit undeterministic. ...

DeepDiver1975:
we need to work on the breadcrumbs ...

screenshot from 2019-02-07 16-33-46

DeepDiver1975:
screenshot from 2019-02-08 11-44-42

tempelgogo:
i missed the ui refactoring. the state was completely rewritten, so i've rewritten my PR.
superseded by https://github.com/owncloud/phoenix/pull/666

PR #633, opened 2019-02-07
Use If-Not-Match: * to prevent overwrite of new uploaded file

DeepDiver1975:
## Description
We need to prevent data loss at any time. It is not allowed to overwrite a file silently.

## Related Issue
- refs #476 

## How Has This Been Tested?
- create a file on the server - e.g. via the old webui 
- do not reload phoenix
- create the same file in phoenix
- see an error poping up

## Screenshot
Screenshot from 2019-04-23 16-56-24

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:
- [ ] Add owncloud server version check
- [ ] Add acceptance tests
- [x] update js-owncloud-client
- [ ] add header check on reupload (once implemented in the ui)

### DIFF
--- a/apps/files/src/components/FilesTopBar.vue
+++ b/apps/files/src/components/FilesTopBar.vue
@@ -109,6 +109,8 @@ export default {
     },
     headers () {
       return {
+        // will trigger 412 precondition failed if a file already exists
+        'If-None-Match': '*',
         'Authorization': 'Bearer ' + this.getToken
       }
     },


DeepDiver1975:
>  But I guess this wasn't goal of this PR - right?

exactly - renaming is a different operation and has different ways to address this ....

Issue #433, opened 2018-12-21
Phoenix Components and Screen Layout


DeepDiver1975:
Proposed Screen Layout:
![desktop](https://user-images.githubusercontent.com/1005065/50350765-6e6a0b00-0540-11e9-92ab-e380ecd39d51.png)

- Phoenix top bar which is fixed and is not scrollable
  -  contains a hamburger menu which will trigger the app menu
  - contains the themable logo
  - contains the themable product name
  - contains list of icons which are injected by apps (like notifications)
  - last element is the user avatar and the user name
- app menu - triggered by the hamburger menu
  - lists apps
  - lists admin settings (if admin)
  - lists user settings
  - logout button
- app content area
  - this is the area which is full under control of the current loaded app - own component oc-content
  - this area can be used by apps at they need/wanr
  - proposed elements which will be provided by phoenix as components are.
    - app menu bar - second app specific menu bar - not scrollable - oc-app-toolbar
    - left nav bar - used for over all nagivation of the app - files tree, list of calendars, mail boxes ... oc-app-navigation-bar
    - right side bar - used for individual opeations of a selected element in the main area - oc-app-sidebar
    - footer - fixed footer which can hold information as well as icon elements - oc-app-footer
    - main content area - scroll able - all menu bars and the top bar will not follow the scroll - oc-app-content

DeepDiver1975:
UX ideas on sidebar and file action placement - take from box

![screenshot from 2019-01-10 13-05-24](https://user-images.githubusercontent.com/1005065/50967469-8518d780-14d8-11e9-82aa-fa60615b0e54.png)
![screenshot from 2019-01-10 13-05-08](https://user-images.githubusercontent.com/1005065/50967470-85b16e00-14d8-11e9-840d-ab58995edb90.png)

DeepDiver1975:
- document components in pattern lib using vuepress(?)

michaelstingl:
Seems https://vueds.com/ is the most interesting tool for us. (uses using custom version of Vue Styleguidist, https://vue-styleguidist.github.io/)

Here is a comparison:
https://blog.hichroma.com/storybook-vs-styleguidist-2bd93d6dcc06

scramb:
I think we should create a repository for vue-design-system as soon as possible. So that we can add upcoming components right here. If we approach the matter more or less clever, we can completely do without vuetify and work with our own component or pattern library in a short time. In addition, this puts us in the position of completely controlling components and processing them as required. 

[...]

PR #663, opened 2019-02-14
Introduce OcAppSidebar, OcAppNavbar and OcAppContent

scramb:
## Description

This Commit will Introduce OcAppSidebar component with a reference implementation in files-app.
As befor OcAppSidebar can be used as base component and can be extended by slots.
All slots are inspirated by current Sidebar appearance and are open for discussion!

#### OcAppSidebar
* Position: Right-hand side of app content
* Width: 20vw
* Slots:
    * icon - left upper icon
    * title - title
    * action - right upper action slot, can be disabled
    * headerContent - row below actual header
    * content - the full content area
    * footer - footer area not used now

#### OcAppContent introduced
* Position: centered 
* Width: All available space
* Slots: Content - is there more we need?

#### OcAppNavbar introduced
* Position: Left-hand side of app content
* Width: same as sidebar
* Slots: Content - more tbd

## Related Issue

- Refs #433 

## Motivation and Context

Most of the existing UI is based on vuetify which make us depend on a framework. We want to offer developers the possible chance to use base components like OcAppTopbar and OcAppSidebar to get easy layout and positioning inside phoenix app.

## How Has This Been Tested?

- :hand: 

## Screenshots:
screenshot_2019-02-14 owncloud phoenix

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes

## ToDo:
- [ ] define layout for OcAppNavbar
- [ ] define usage for OcAppNavbar

## Important Note:
merge #664 before this

### DIFF
--- a/apps/files/src/components/FileDetails.vue
+++ b/apps/files/src/components/FileDetails.vue
@@ -1,55 +1,46 @@
 <template>
-  <v-navigation-drawer
-  v-model="drawer"
-  class="grey lighten-2"
-  right
-  floating
-  permanent
-  style="width:100%"
-  >
-  <v-layout column>
-    <v-layout primary row>
-      <v-flex>
-        <v-icon color="white" class="pl-4" medium>folder</v-icon>
-      </v-flex>
-      <v-flex white--text align-self-center>
-        <span class="subheading"> {{ getTabName }} </span>
-      </v-flex>
-    </v-layout>
-    <v-layout primary row>
-      <v-spacer />
-      <v-btn @click.native="deleteSelectedFiles" flat><v-icon color="white" medium>delete</v-icon></v-btn>
-      <v-spacer />
-      <v-btn @click.native="downloadFiles" v-if="items.length <= 1" flat><v-icon color="white" medium>cloud_download</v-icon></v-btn>
-      <v-btn disabled v-else flat><v-icon color="white" medium>archive</v-icon></v-btn>
-      <v-spacer />
-    </v-layout>
-    <v-tabs
-      v-model="active"
-      color="primary lighten-5"
-      dark
-      slider-color="yellow"
-    >
-      <v-tab
-        v-for="tab of fileSideBars"
-        :key="tab.name"
-        ripple
+  <oc-app-side-bar>
+    <template slot="title">
+      <span>{{ getTabName }}</span>
+    </template>
+    <template slot="headerContent">
+      <v-layout primary row>
+        <v-spacer />
+        <v-btn @click.native="deleteSelectedFiles" flat><v-icon color="white" medium>delete</v-icon></v-btn>
+        <v-spacer />
+        <v-btn @click.native="downloadFiles" v-if="items.length <= 1" flat><v-icon color="white" medium>cloud_download</v-icon></v-btn>
+        <v-btn disabled v-else flat><v-icon color="white" medium>archive</v-icon></v-btn>
+        <v-spacer />
+      </v-layout>
+    </template>
+    <template slot="content">
+      <v-tabs
+        v-model="active"
+        color="primary lighten-5"
+        dark
+        slider-color="yellow"
       >
-        {{ tab.name }}
-      </v-tab>
-      <v-tab-item
-        v-for="tab of fileSideBars"
-        :key="tab.name"
-      >
-        <component :is="tab.component"></component>
-      </v-tab-item>
-    </v-tabs>
-</v-layout>
-</v-navigation-drawer>
+        <v-tab
+          v-for="tab of fileSideBars"
+          :key="tab.name"
+          ripple
+        >
+          {{ tab.name }}
+        </v-tab>
+        <v-tab-item
+          v-for="tab of fileSideBars"
+          :key="tab.name"
+        >
+          <component :is="tab.component"></component>
+        </v-tab-item>
+      </v-tabs>
+    </template>
+  </oc-app-side-bar>
 </template>
 
 <script>
 import Mixins from '../mixins'
+import OcAppSideBar from 'oc_components/OcAppSideBar.vue'
 import { mapActions, mapGetters } from 'vuex'
 
 export default {
@@ -64,6 +55,7 @@ export default {
     }
   },
   components: {
+    OcAppSideBar
   },
   methods: {
     ...mapActions('Files', ['deleteFiles']),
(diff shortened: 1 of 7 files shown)


apps/files/src/components/FileDetails.vue
```diff
@@ -1,55 +1,46 @@
...
-      color="primary lighten-5"
-      dark
-      slider-color="yellow"
-    >
-      <v-tab
-        v-for="tab of fileSideBars"
-        :key="tab.name"
-        ripple
+  <oc-app-side-bar>
+    <template slot="title">
```
DeepDiver1975: For my understanding: why are we going for this template/slot approach?

Vuetify uses explicit components which can be added here.

What I have in mind:

```html
<oc-app-side-bar>
 <oc-app-sidebar-title>{{ getTabName }} </oc-app-sidebar-title>
  <oc-side-bar-header>
   <v-layout>.....</v-layout>
  </oc-side-bar-header>
  <oc-side-bar-content>
    <v-layout>.....</v-layout>
  </oc-side-bar-content>
</oc-side-bar>
```

This is the cool stuff I like about vuetify: the compilation of components - this is what I'd like to have with our components as well.

DeepDiver1975: This can be done in a followup PR ;-)

apps/files/src/components/FilesApp.vue
```diff
@@ -1,24 +1,21 @@
   <template>
-    <v-content>
-      <v-container id="files-app" fluid pa-0>
-      <v-layout row fill-height>
-        <v-flex :class="{'xs12': selectedFiles.length === 0, 'xs6': selectedFiles.length > 0 }" pa-0 fill-height>
+    <div class="oc-app" id="files-app">
+      <oc-app-content>
```
DeepDiver1975: I'd expect oc-app-content to be the root element here

src/Phoenix.vue
```diff
@@ -62,6 +62,24 @@ export default {
 </script>
 
 <style>
+  .oc-app {
```
DeepDiver1975: that should be in the components - right?

src/Phoenix.vue
```diff
@@ -77,7 +95,6 @@ export default {
   }
   .sidebar { grid-area: sidebar;
             background-color: brown;
-            width: 300px;
```
DeepDiver1975: I need this width for another project :man_facepalming:

DeepDiver1975:
> All slots are inspirated by current Sidebar appearance and are open for discussion!

no idea if fixed slots are the way to go. I'd prefer to give the developers the full content area and let them place what ever content they want inside. There are for sure limitation what to use ... but generally speaking: use any vuetify component.

src/Phoenix.vue
scramb: we should adjust the value other ways, then setting it to fixed 300px - not sure what the exact requirement looks like - i'll check on that

src/Phoenix.vue
scramb: That should be in a Global Stylesheet tbh.

apps/files/src/components/FilesApp.vue
scramb: We got some problems here, in this approach oc-app-content and oc-app-sidebar both are child components to oc-app which for now is a css-class and needs to be adopted

scramb:
> > All slots are inspirated by current Sidebar appearance and are open for discussion!
> 
> no idea if fixed slots are the way to go. I'd prefer to give the developers the full content area and let them place what ever content they want inside. There are for sure limitation what to use ... but generally speaking: use any vuetify component.

Same as in the first comment - The fixed slots give as the chance to controll app styles to be within phoenix design. Other way around - what are benefits of oc-app-sidebar component if it's actually only a div with some predefined css tied to it's wrapper. To define every slot as a component which is allowed inside the rootComponent will take alot of effort AND would be alot easier if defined somewhere clean to not blow up the project folder; 
```
/src/components/OcApp/OcAppContent/OcAppSidebar
                                                              /OcAppSidebar/OcAppSidebarHeader
                                                              /OcAppNavbar
                                                              /OcAppContentContent
```
etc. doesn't look anygood to me

src/Phoenix.vue
DeepDiver1975: isn't it the job of the component to take care of template, script and style? A component is an isolated unit -> all in one. This is also the power of the vue files from my understanding.

global is :hankey:

src/Phoenix.vue
DeepDiver1975: well - the requirment is; I need at least ~ 280px on the side bar to display all necessary information properly :man_shrugging:

apps/files/src/components/FilesApp.vue
DeepDiver1975: > Same as in the first comment - The fixed slots give as the chance to controll app styles to be within phoenix design.

from my understanding the common design comes from the individual properly designed components and not by a globally defined stylesheet which creates a mess of css classes.

In this case we could have stayed with the original uikit apporach ....

scramb: > In this case we could have stayed with the original uikit apporach ....

I just tried to keep things newly build as far away from vuetify as possible to give us the chance to eventually migrate to a own pattern library easier - which was not part of the discussion as vuetify came into play

src/Phoenix.vue
scramb: global is shit.
but this is the only way i can controll the app behavior for all parts in focus of positon, size and layout. we can break it down to the core to make use of the components - which is actually part of a pattern library aswell imo...

src/Phoenix.vue
scramb: now i get your point - is this a requirement for EVERY device -screensize

DeepDiver1975: as of now - yes - in the future on a mobile screen we need a concept where the sidebar consumes the full screen ..... later ...

apps/files/src/components/FileDetails.vue
```diff
@@ -85,12 +77,16 @@ export default {
   },
   computed: {
     ...mapGetters(['getToken', 'fileSideBars']),
-
     getTabName () {
       if (this.items.length === 0) {
         return ''
       }
-      return (this.items.length > 1) ? this.$gettext('Multiple Files') : this.items[0].name
+      // return (this.items.length > 1) {
```
DeepDiver1975: unnecessary comment ..

src/Phoenix.vue
```diff
@@ -12,7 +12,7 @@
       <top-bar></top-bar>
       <side-menu></side-menu>
       <router-view v-if="this.$route.matched[0].components.appTopbar" class="appTopbar" name="appTopbar"></router-view>
-      <router-view style="margin-top: 115px" name="appContent"></router-view>
+      <router-view class="" style="margin-top: 115px" name="appContent"></router-view>
```
DeepDiver1975: empty class?

PR #696, opened 2019-02-18
close #665 fix file filters

### LINKED ISSUE #665: Local File Filter dysfunctional
Local File Filter broken since https://github.com/owncloud/phoenix/pull/623 due to component splitting.

## Related Issue:
- #662 local folder search broken

## Description:
prepared in #666

## ToDo:
- [ ] use event's instead of v-model & inject value into vuetify `input-value`
- [ ] test all filters

tempelgogo:
## Description
moved filter into `fileFilters.js`, load in store.
Split v-model from checkbox to catch event manually & inject value into v-checkbox `input-value` prop

## Related Issue

- Fixes #665

## Motivation and Context
filters were hardcoded before, but we need them at a few places...
abstract them in state.

## How Has This Been Tested?

- create folder, directory & hidden files
- check & uncheck file filters

design has not changed, It's just working again

## Types of changes

## Checklist:

- [X] Code changes

### DIFF
--- a/apps/files/src/fileFilters.js
+++ b/apps/files/src/fileFilters.js
@@ -1,0 +1,13 @@
+export const fileFilters = [{
+  name: 'Files',
+  tag: 'file',
+  value: true
+}, {
+  name: 'Folders',
+  tag: 'folder',
+  value: true
+}, {
+  name: 'Hidden',
+  tag: 'hidden',
+  value: false
+}]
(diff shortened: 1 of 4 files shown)


apps/files/src/fileFilters.js
```diff
@@ -0,0 +1,13 @@
+export const fileFilters = [{
+  name: 'Files',
```
DeepDiver1975: still no translation :see_no_evil: ;-)

tempelgogo: `vue-gettext` has no global available getter AFAIK; so i needed to wrap it into a real SingleFileComponent, thus get the ViewModel access via `this`
I don't want to pollute the TopBar, so I've created a new File.

If you like it, i could squash it.

PR #736, opened 2019-02-22
WIP: Introduce Global Event Bus

scramb:
## Description

:construction: 
Introducing a global EventBus to handle Global actions like file-download. file-progress or even Notifications (tbd)

## Motivation and Context

Since things like upload/download progress or the notifications don't need to be persisted in state.

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes

### DIFF
--- a/src/services/eventbus.js
+++ b/src/services/eventbus.js
@@ -1,0 +1,31 @@
+import Vue from 'vue'
+import store from '../store'
+
+export const EventBus = new Vue()
+
+function fetchDownload (file) {
+  const url = this.$client.files.getFileUrl(file.path)
+  let anchor = document.createElement('a')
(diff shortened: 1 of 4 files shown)


DeepDiver1975:
please remove the dependency bump commit - THX

src/services/eventbus.js
```diff
@@ -0,0 +1,31 @@
+import Vue from 'vue'
+import store from '../store'
+
+export const EventBus = new Vue()
+
+function fetchDownload (file) {
```
DeepDiver1975: Having the real operation on the event bus seems wrong to me.

The bus is for pure communication. Senders and receivers live outside

PR #737, opened 2019-02-22
WIP: introduce postcss

scramb:
## Description

:construction: 
Introduce postcss and some postcss-plugins like postcss-nested and autoprefixer to handle nested CSS and auto-prefixing browser-vendor-prefixes.

## Types of changes

## Checklist:

- [x] Code changes


scramb:
Right. We should move this to the next week and implement it in a pattern lib right away.

PR #739, opened 2019-02-22
Feature - Add Download-Feedback

### LINKED ISSUE #526: File download gives no user feedback ....
### Steps to reproduce
0. open dev tools, network tab, observe ...
1. Upload a bigger file (~400MB)
2. Click on download icon
3. There is no user feedback until the file is fully downloaded

### Expected behaviour
- the should be immediate user feedback - best would be to popup the save dialog ...

### Solution Baseline - minimal requirements - works in all supported browsers
As soon as the user hits the download button the button will mutate into a small spinning wheel to give user feedback. 
It will spin until the download is finished.
This is a simplistic approach which will work on older browsers as well.
In the future we can add more advanced features.
We do not show any progress but only an endless spinner to tell the user that there is something going on

### Future Solutions for browsers with more capabilities - Not part of this ticket
- Use stream API (not fully supported anywhere - https://caniuse.com/#feat=streams)
- Use service workers (not on IE - https://caniuse.com/#feat=serviceworkers)

scramb:
## Description

File-Download now adds file to progress and shows indeterminated loading.

## Related Issue

- Fixes #526 

## How Has This Been Tested?

- :hand: 

## Types of changes

## Checklist:

- [x] Code changes

## Open tasks:

- [ ] add ReadableStream API to get current progress


DeepDiver1975:
>     * we did talk about a spinner which is replacing the download button - what was driving against this?

one can trigger hundreds of downloads this way

screenshot from 2019-02-22 16-30-23

screenshot from 2019-02-22 16-30-54

DeepDiver1975:
download triggered from buttom action causes js errors in console

screenshot from 2019-02-22 16-33-20

DeepDiver1975:
I can still trigger a download many times in a row and there is still no spinner on the file list. Please adjust as discussed. THX

DeepDiver1975:
still error when download is triggered from action panel and sidebar

scramb:
* As i merged master, i don't see any download button in the sidebar anymore.
* In Chrome 72 im able to download files via FileActionTab and via Button in FilesList.
* In Firefox the download is not working anymore as reported above.
* Download is now locked while downloading

scramb:
I found an article descriping an similar problem, the solutions are relative paths to be used for firefox.
What is kinda flaw in the download logic at all.
https://stackoverflow.com/questions/30444250/download-attribute-not-working-in-firefox-but-working-in-chrome

Issue #377, opened 2018-12-05
Doesn't load files/folder when owncloud-server is not in apache root dir


paurakhsharma:
# Description
Phoenix client doesn't load files/folders if the owncloud server is not in the apache root dir.
But, if the owncloud-server is in apache root dir everything works fine.

# Steps to reproduce
1) Configure Apache to point `www` dir
2) Clone owncloud core insider `www` so that server is in `www/core`
3) Access owncloud server through `localhost/core`
4) Change phoenix config in `config.json` as:
`{
	"server" : "http://localhost/core",
	"theme": "owncloud",
	"version": "0.1.0",
	"apps" : [
		"files"
	],
	"auth": {
		"clientId": "j3VLpTvmRSxqNGJWxAmn4JA3Nu9BoFbL7EoM8zAcDiPfkBCYaXqiSTiW2CFbM1LP",
		"apiUrl": "http://localhost/core/index.php/apps/oauth2/api/v1/token",
		"authUrl": "http://localhost/core/index.php/apps/oauth2/authorize"
	}
}
`

5) Access Phoenix and login using oauth2
6) See that the files/folder are not loaded in the webUI (Note: was able to see them in owncloud-server frontend)

![image](https://user-images.githubusercontent.com/20378877/49507926-9eab7b80-f8a9-11e8-9794-8ca4de9556a1.png)

![image](https://user-images.githubusercontent.com/20378877/49508027-d31f3780-f8a9-11e8-9fff-587f4ffa9501.png)

c.c @individual-it @phil-davis @DeepDiver1975

paurakhsharma:
> Which OC version do you use ? oc/core master branch cloned via git ?

Yes

DeepDiver1975:
this issue needs to be fixed in js-owncloud-client

PR #751, opened 2019-02-26
enable history mode

### LINKED ISSUE #573: No fragment in the OpenID Connect redirect uri allowed
According to https://tools.ietf.org/html/rfc6749#section-3.1.2 it is not allowed to have a fragment in the redirect uri.

We need to change this asap to ensure compatibility with any OpenID Connect Providers out there

THX @labkode for the hint

### LINKED ISSUE #606: Enable history mode in vue-router
### Description
as mentioned in #573, we need to to change the phoenix application to history mode to get rid of the hash in the url.

`/#/login` -> `/login`
`/#/files/list/Open_Data` -> `/files/list/Open_Data`

It's supported within our target platform: https://caniuse.com/#feat=history

### Expected behaviour
`/files/list/Open_Data`

### Actual behaviour
`/#/files/list/Open_Data`

tempelgogo:
## Description
need to add `base` to config & change return url in oc.
setting base url in html document (derived  from config value) allows us to use relative paths, even if phoenix is running in a subdirectory.

## Related Issue

- Fixes #606
- Fixes #573
- refs #377

## Motivation and Context
- use HistoryAPI
- respect  OpenID Specs

## How Has This Been Tested?

- login works
- acceptance tests

## Types of changes

- [X] Technical debt
- [ ] Tests

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,12 +1,13 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+	<base href="<%= htmlWebpackPlugin.options.baseUrl %>" />
 	<meta charset="UTF-8">
 	<title>ownCloud Phoenix</title>
 	<meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no">
 	<meta name="theme-color" content="#375f7E"/>
 	<link rel="manifest" href="manifest.json">
-  <script src="node_modules/requirejs/require.js"></script>
+  <script src="/require.js"></script>
 	<noscript>
 		<style>
 			#enable-js-banner {
--- a/.drone.yml
+++ b/.drone.yml
@@ -41,7 +41,7 @@ pipeline:
       - php occ log:manage --level 0
       - php occ config:list
       - php occ a:e oauth2
-      - php occ oauth2:add-client Phoenix Cxfj9F9ZZWQbQZps1E1M0BszMz6OOFq3lxjSuc8Uh4HLEYb9KIfyRMmgY5ibXXrU 930C6aA0U1VhM03IfNiheR2EwSzRi4hRSpcNqIhhbpeSGU6h38xssVfNcGP0sSwQ http://phoenix:8300/#/oidc-callback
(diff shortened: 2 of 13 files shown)


index.html
```diff
@@ -1,12 +1,13 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+	<base href="<%= htmlWebpackPlugin.options.baseUrl %>" />
```
DeepDiver1975: I did not yet get why we need this

DeepDiver1975: @tempelgogo

tempelgogo: The [base](https://developer.mozilla.org/de/docs/Web/HTML/Element/base) is a relict from a time when HTML was used to build websites, not Web Applications.
It enable's relative urls in our web document, by prefixing them with the base url.
This case especially matters on deep links in history mode, 'cause they do not reflect the real path. Take these paths as example, imagine we are re-loading the page:
- `http://fe.serv3r.cloud:8300/files/list/home`
- `http://fe.serv3r.cloud:8300/files`

Fulfilling the requirement to deploy phoenix in a subdirectory makes this even worse:
- `http://fe.serv3r.cloud:8300/phoenix/files`
- `http://fe.serv3r.cloud:8300/demos/vue/phoenix/master-branch/files`

Everything beginning at `files/` is getting rewritten to `index.html` with corresponding params, we don't know which part of the url is *real*, apps don't know their *deployment base path* like `http://fe.serv3r.cloud:8300/demos/vue/phoenix/master-branch/files.bundle.js`

Where to find the index bundle ?

Webpack as our script bundler is basically able to **traverse through the import/require dependency tree**, which is awesome to tree-shake code duplicates away or adjust import paths.
Unfortunately our webpack isn't able to do so, due to our dynamic require.js-based app loading system.  
There isn't even a guaranty that all apps are available while bundling… so how to modify them ?

This problem has already occurred in many places such as cross-app linkage, including the polyfills, …  
I am not aware of any other **least-effort** solution, which fulfills **all our current requirements**, especially that apps must support dynamic installation.

JavaScript was never made for this funky stuff…

tempelgogo: The `[html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin#options)` (which we are currently using) has support for an *base* attribute.
Unfortunately I couldn't convince it to work. So I've injected it old-school styled manually... which has some disadvantages, because webpack may put code before it, which we do't want it to; *base* url has to be set before it's used first.
Finally it's performing better as before.

DeepDiver1975: So this means that after deployment the index.html has to be manipulated.
No Idea If this is acceptable....

tempelgogo:
Management is not interested in this approach.

PR #840, opened 2019-03-27
Add Avatar to show image given userid

tomneedham:
Created a new component to show an avatar given the userid. 

 - loading spinner whilst we try to get it
 - avatar shown if available
 - placeholder icon if not

todo
 - [x] server url
 - [x] global avatar config?


src/components/OcAvatar.vue
```diff
@@ -0,0 +1,51 @@
...
+      /**
+       * Set to object URL when loaded, or on failure, icon placeholder is shown
+       */
+      avatarSource: '',
+      loading: true
+    }
+  },
+  mounted: function () {
+    let headers = new Headers()
+    let url = 'http://cloud.local/remote.php/dav/avatars/'+this.userid+'/128.png'
```
tomneedham: TODO

DeepDiver1975: fixed

src/components/Top-Bar.vue
```diff
@@ -7,7 +7,7 @@
     </oc-topbar-item>
 
     <oc-topbar-item slot="right">
-      <oc-icon name="account_circle" color="white"></oc-icon>
+      <ocAvatar class="uk-margin-small-right" :userid="user.displayname" />
```
tomneedham: @DeepDiver1975 from what I found, we dont have the userid in the user object here - thoughts?

DeepDiver1975: should be user.id

DeepDiver1975: ah - missing in vuex

Screenshot from 2019-03-28 11-52-25

src/components/Avatar.vue
```diff
@@ -0,0 +1,56 @@
...
+      loading: true
+    }
+  },
+  mounted: function () {
+    if (!this.enabled) {
+      return;
+    }
+    let headers = new Headers()
+    let instance = this.$root.config.server || window.location.origin
+    let url = instance + '/remote.php/dav/avatars/' + this.userid + '/128.png'
```
DeepDiver1975: at time of mounting userid can not yet be set - better move this into a watcher on the property and load upon change

src/components/Top-Bar.vue
DeepDiver1975: pushed a commit to add the is - now having troubles with the initial binding - see my comment on watching the prop .....

PR #846, opened 2019-04-01
Add avatar component

tomneedham:
New version of https://github.com/owncloud/phoenix/pull/840

 - [ ] Watch still not working


tomneedham:
Spent far too long on this trying to figure it out. Tried a watcher, and to update from there. Tried ` this.$store.watch` inside the mounted of the topbar. 

Essentially, in the TopBar, we need to pass in the userid into the avatar. But this must be updated when the user state becomes available. So how to pass in a vuex state variable into a component prop, and update it as / when it changes?

PVince81:
from my understanding you shouldn't need any watcher, it should already work automagically, given that user id is in a store already and should be observable.

@felixheidecke any more clues ?

DeepDiver1975:
It works With the Display name - right? So where is the difference?

PR #913, opened 2019-04-09
add collaboration status

butonic:
A collaboration status component to sharing sidebar
- [x] requires value passing to ocRadio: https://github.com/owncloud/owncloud-design-system/pull/187
- [ ] needs wiring up data model for shares / collaborators, for now dummy data is left in

### DIFF
--- a/apps/files/src/components/OcCollaborationStatus.vue
+++ b/apps/files/src/components/OcCollaborationStatus.vue
@@ -1,0 +1,199 @@
+<template>
+  <div class="oc-collaboration-status">
+    <oc-accordion uk-accordion="collapsible: false; multiple: true">
+      <oc-accordion-item class="uk-open" title="Owner(s)">
+        <template slot="content">
+          <oc-user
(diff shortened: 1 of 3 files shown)


apps/files/src/components/OcCollaborationStatus.vue
```diff
@@ -0,0 +1,199 @@
...
+      this.shares = this.shares.filter(function (s){
+        return s != share
+      })
+      */
+    }
+  }
+}
+</script>
+
+<docs>
```
DeepDiver1975: no docs needed in here

butonic: I left them in to see what kind of objects need to be passed in as props. Could be removed after logic has been added ... then again ... an example to use this component is good to have, isn't it?

DeepDiver1975: I see no need - the component will be used as is within phoenix - this will change anyhow over time.

PR #938, opened 2019-04-11
Use oc-nav-bar to implement Top-Bar

DeepDiver1975:
Screenshot from 2019-04-11 16-22-46

### DIFF
--- a/src/components/Top-Bar.vue
+++ b/src/components/Top-Bar.vue
@@ -1,14 +1,21 @@
 <template>
-  <oc-topbar variation="primary">
-    <oc-topbar-logo icon="menu" title="Files" slot="left" @click="toggleSidebar(!isSidebarVisible)"></oc-topbar-logo>
-    <oc-topbar-item slot="title">
-      <oc-img :src="configuration.theme.logo.big" style="height: 60px" />
-    </oc-topbar-item>
-    <oc-topbar-item slot="right">
-      <Avatar class="uk-margin-small-right" :userid="user.id" />
-      <span>{{ user.displayname }}</span>
-    </oc-topbar-item>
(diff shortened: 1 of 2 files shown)


felixheidecke:
I suggest dropping sticky for now. The "outer shell" around the app including the topbar should remain @ a fixed 100vh where as app should scroll. So Topbar will remain stationary and "sticky" can and should be used inside the apps only.

src/components/Top-Bar.vue
```diff
@@ -1,14 +1,21 @@
...
-    <oc-topbar-logo icon="menu" title="Files" slot="left" @click="toggleSidebar(!isSidebarVisible)"></oc-topbar-logo>
-    <oc-topbar-item slot="title">
-      <oc-img :src="configuration.theme.logo.big" style="height: 60px" />
-    </oc-topbar-item>
-    <oc-topbar-item slot="right">
-      <Avatar class="uk-margin-small-right" :userid="user.id" />
-      <span>{{ user.displayname }}</span>
-    </oc-topbar-item>
-  </oc-topbar>
+  <oc-navbar tag="header" class="oc-topbar" :sticky="true" :offset="0">
```
felixheidecke: Let's drop sticky here as it's not intended to be used with the topbar

DeepDiver1975: > I suggest dropping sticky for now. The "outer shell" around the app including the topbar should remain @ a fixed 100vh where as app should scroll. So Topbar will remain stationary and "sticky" can and should be used inside the apps only.

I assume this requires some scss magic - if so - please open an issue in ods - THX

DeepDiver1975: stiicky killed for now

PR #973, opened 2019-04-15
No need to createObjectURL if the request failed

DeepDiver1975:
Fixes this issue:
Screenshot from 2019-04-15 16-26-13

### DIFF
--- a/src/components/Avatar.vue
+++ b/src/components/Avatar.vue
@@ -38,16 +38,19 @@ export default {
       headers.append('Authorization', 'Bearer ' + this.getToken)
       fetch(url, { headers })
         .then(response => {
-          if (response.status === 404) {
-            this.avatarSource = ''
-            this.loading = false
-          } else {
+          if (response.ok) {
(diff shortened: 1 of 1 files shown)


src/components/Avatar.vue
```diff
@@ -38,16 +38,19 @@ export default {
...
-        .then(blobby => {
+        .then(blob => {
           this.loading = false
-          this.avatarSource = window.URL.createObjectURL(blobby)
+          this.avatarSource = window.URL.createObjectURL(blob)
+        })
+        .catch(error => {
+          this.avatarSource = ''
+          this.loading = false
+          console.log('There has been a problem with your fetch operation: ', error.message)
```
felixheidecke: Shouldn't this be a `console.warn()`?

DeepDiver1975: well - usually this will fail with a 404 in case no avatar is defined for a user.
Nothing to warn about this fact from my poc.

PR #1017, opened 2019-04-26
make clean all apps

butonic:
I wanted to build a docker container ... but it kept pulling in hundreds of mb, even after I did a `make clean`. This PR adds missing Makefiles for all apps and as a result correctly deletes the `node_modules` and `dist` folders.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 SERVER_HOST=0.0.0.0:8300
 
-apps=files
+apps=files markdown-editor pdf-viewer
 all_apps=$(addprefix app-,$(apps))
 core_bundle=dist/core/core.bundle.js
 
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
we have yarn commands to do all this - no need for Makefiles

yarn install-all
yarn build-all
yarn clean-all
yarn update-all

butonic:
ok ... so get rid of the Makefile?

DeepDiver1975:
well - wee need to keep the l10n stuff after https://github.com/owncloud/phoenix/blob/d3fc85419fedc674a972af2fd2dfc8a6cb492176/Makefile#L54 ....

DeepDiver1975:
I'm closing this as discussed above.

PR #1018, opened 2019-04-26
add .dockerignore

butonic:
I may want to build a docker container while having built phoenix locally. Adding this .dockerignore file prevents docker from copying all the node_modules folders.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+node_modules
+apps/*/node_modules


.dockerignore
```diff
@@ -0,0 +1,4 @@
+.git
+node_modules
+owncloud-design-system/node_modules
```
DeepDiver1975: not really needed anymore - right?

butonic: removed

Issue #583, opened 2019-02-01
Sharing SideBar


DeepDiver1975:
This issue is about defining how the file sharing side bar shall look like and how it shall work.

@pmaier1 let's chat ;-)
@michaelstingl your input is valuable here as well
Andreas is required here as well.

THX

PVince81:
@felixheidecke please share your mock ups / click dummy

felixheidecke:
The sharing mock ups / click dummy was integrated into the old Phoenix version and was about to be extracted into it's own repo: https://github.com/owncloud/files_sharing_fe

I revived an old version of [classic Phoenix](https://github.com/owncloud/phoenix/tree/classic) in here, which shows the concept of a multi-stage sharing process where one could choose between quick share and using more options. Wasn't perfect, but a first draft.

DeepDiver1975:
> I revived an old version of [classic Phoenix](https://github.com/owncloud/phoenix/tree/classic) in here, which shows the concept of a multi-stage sharing process where one could choose between quick share and using more options. Wasn't perfect, but a first draft.

thx - I'll try to revive this and paste screenshots ... unless you have some left @felixheidecke

felixheidecke:
PS: mockups were to be developed as [design guidelines](https://github.com/owncloud/enterprise/issues/2791) evolved to make sure UX would follow global principles.

felixheidecke:
Some old screenshots https://cloud.owncloud.com/index.php/s/MCqRCZGGqicdBva

DeepDiver1975:
![phoenix_classic_share_001](https://user-images.githubusercontent.com/1005065/52936249-e21b6f00-335b-11e9-9943-143b7c4886b0.png)
![phoenix_classic_share_002](https://user-images.githubusercontent.com/1005065/52936248-e21b6f00-335b-11e9-9829-8d14976ecfae.png)
![phoenix_classic_share_003](https://user-images.githubusercontent.com/1005065/52936247-e182d880-335b-11e9-8149-92ad184d2168.png)
![phoenix_classic_share_004](https://user-images.githubusercontent.com/1005065/52936246-e182d880-335b-11e9-9feb-d463beebca12.png)

DeepDiver1975:
related to #1186 

we don't need two issues -> closing this one

PR #1102, opened 2019-05-09
Add group sharing

DeepDiver1975:
## Description
Sharing with groups is now possible

## Related Issue
- refs #583 

## How Has This Been Tested?
- :hand: 

Screenshot from 2019-05-09 12-03-58

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/pageObjects/filesPage.js
+++ b/tests/acceptance/pageObjects/filesPage.js
@@ -58,6 +58,45 @@ module.exports = {
     },
     /**
      *
+     * @param {string} fileName
+     */
+    openSharingDialog: function (fileName) {
+      const shareBtnSelector = this.getFileRowSelectorByFileName(fileName) +
+        this.elements['shareButtonInFileRow'].selector
+
+      return this.initAjaxCounters()
+        .waitForFileVisible(fileName)
+        .useXpath()
+        .moveToElement(this.getFileRowSelectorByFileName(fileName), 0, 0)
+        .click(shareBtnSelector)
+        .waitForElementVisible('@sharingSideBar')
+        .useCss()
+    },
+    /**
+     *
+     * @param {string} input
+     */
+    enterAutoComplete: function (input) {
+      return this.initAjaxCounters()
+        .waitForElementVisible('@sharingAutoComplete')
+        .setValue('@sharingAutoComplete', input)
+    },
+    getShareAutocompleteItemsList: function () {
+      const itemsList = []
+      const page = this
+      return this
+        .waitForElementVisible('@sharingAutoCompleteDropDownElements')
+        .api.elements('css selector', this.elements['sharingAutoCompleteDropDownElements'].selector, (result) => {
+          result.value.forEach(function (value) {
+            page.api.elementIdText(value.ELEMENT, (text) => {
+              itemsList.push(text.value)
+            })
+          })
+        })
+        .then(() => itemsList)
+    },
+    /**
+     *
      * @param {string} fromName
      * @param {string} toName
      */
@@ -274,6 +313,10 @@ module.exports = {
       selector: '//button[@aria-label="Delete"]',
       locateStrategy: 'xpath'
     },
+    shareButtonInFileRow: {
+      selector: '//button[@aria-label="Share"]',
+      locateStrategy: 'xpath'
+    },
     deleteFileConfirmationBtn: {
       selector: '#oc-dialog-delete-confirm'
     },
@@ -337,6 +380,19 @@ module.exports = {
     markedFavoriteInFileRow: {
       selector: '//span[contains(@class, "oc-star-shining")]',
       locateStrategy: 'xpath'
+    },
+    sharingSideBar: {
+      selector: '#oc-files-sharing-sidebar'
+    },
+    sharingAutoComplete: {
+      selector: '#oc-sharing-autocomplete .oc-autocomplete-input'
+    },
+    sharingAutoCompleteDropDown: {
+      selector: '#oc-sharing-autocomplete .oc-autocomplete-suggestion-list'
+    },
+    sharingAutoCompleteDropDownElements: {
+      selector: '#oc-sharing-autocomplete .oc-autocomplete-suggestion'
     }
+
   }
 }
--- a/tests/acceptance/stepDefinitions/provisioningContext.js
+++ b/tests/acceptance/stepDefinitions/provisioningContext.js
@@ -23,6 +23,19 @@ function initUser (userId) {
   return fetch(client.globals.backend_url + '/ocs/v2.php/cloud/users/' + userId, { method: 'GET', headers: headers })
 }
 
+function createGroup (groupId) {
+  const body = new URLSearchParams()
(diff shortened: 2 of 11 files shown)


DeepDiver1975:
We have corb all over the place 🤷
We need to have a closer look ... One day.

I'll have a look at the File Sharing issue asap ...

DeepDiver1975:
>     * Error in render: "TypeError: e.toLowerCase is not a function"

can you please debug this? I cannot reproduce this @LukasHirt - THX

DeepDiver1975:
@individual-it @paurakhsharma please review my nightwatch gymnastics - THX
@LukasHirt tests are running properly - please try to reproduce your issue - THX

tests/acceptance/pageObjects/filesPage.js
```diff
@@ -56,6 +56,71 @@ module.exports = {
...
+    enterAutoComplete: function (input) {
+      return this.initAjaxCounters()
+        .waitForElementVisible('@sharingAutoComplete')
+        .setValue('@sharingAutoComplete', input)
+    },
+    /**
+     *
+     * @param {string} pattern
+     */
+    sharingAutoCompleteShallContainElementsWith: function (pattern) {
```
paurakhsharma: Please replace shall with should

tests/acceptance/pageObjects/filesPage.js
```diff
@@ -56,6 +56,71 @@ module.exports = {
...
+          })
+        })
+
+      return this
+    },
+    /**
+     *
+     * @param {string} pattern
+     */
+    sharingAutoCompleteShallNotContainElementsWith: function (pattern) {
```
paurakhsharma: Please replace shall with should

tests/acceptance/stepDefinitions/provisioningContext.js
```diff
@@ -41,3 +54,26 @@ Given('the quota of user {string} has been set to {string}', function (userId, q
   )
     .then(res => httpHelper.checkStatus(res, 'Could not set quota.'))
 })
+
+Given('these users have been created with default attributes but not initialized:', function (dataTable) {
```
paurakhsharma: Currently, we have the same method for `user has been created with default attributes`, `user has be created but not initialized`, `user has been created`.
Later we have to implement proper methods for them.

tests/acceptance/stepDefinitions/sharingContext.js
```diff
@@ -25,3 +25,20 @@ Given('user {string} has shared file/folder {string} with user {string}', functi
...
+  return client.page.filesPage().enterAutoComplete(input)
+})
+
+Then('all users and groups that contain the string {string} in their name should be listed in the autocomplete list on the webUI', function (pattern) {
+  return client.page.filesPage().sharingAutoCompleteShallContainElementsWith(pattern)
+})
+
+Then('the users own name should not be listed in the autocomplete list on the webUI', function () {
+  // TODO: where to get the current user from
+  return client.page.filesPage().sharingAutoCompleteShallNotContainElementsWith('admin')
```
paurakhsharma: We have to implement something similar to this https://github.com/owncloud/core/blob/master/tests/acceptance/features/lib/OwncloudPage.php#L284
And remove this hard-coded name.

That can be done in other PR

DeepDiver1975: Shouldn't the current logged in user be stored in some variable somewhere?

tests/acceptance/pageObjects/filesPage.js
DeepDiver1975: well - I did choose this method naming because it is not the filePage which should/shall do something - but the sharingAutoComplete.

I'm happy to get reasonable proposals - THX

tests/acceptance/stepDefinitions/sharingContext.js
paurakhsharma: Current admin username is stored in `client.globals.backend_admin_username` but,
That doesn't work for all users.

This step should work for all users. So, we should get the current user using the webUI.
But In core in some API tests we set the current user using 
the behat step `"As user "<user_name>"` and set username as a global variable.
https://github.com/owncloud/core/blob/master/tests/acceptance/features/bootstrap/BasicStructure.php#L676

But I don't think this is a good idea for a webUI test. Because when the user logs out we have to remember to reset this variable.

tests/acceptance/pageObjects/filesPage.js
paurakhsharma: I suggested this because the behat step says `should be listed in the autocomplete list on the webUI`.
Also, shall denote in the future.

LukasHirt:
> @individual-it @paurakhsharma please review my nightwatch gymnastics - THX
> @LukasHirt tests are running properly - please try to reproduce your issue - THX

Couldn't reproduce - don't know what it caused. Works fine now.

tests/acceptance/pageObjects/filesPage.js
phil-davis: In a way this is just a discussion about "legal" words. "should" has a bit more "wiggle-room", "shall" means it has to. "shall" is a more archaic word in ordinary modern English usage, it mostly happens in legal documents and sometimes in system specifications that are written in a "legal contract" way.
In Gherkin step text we have been using "should" - that makes the step more readable in ordinary English. The code underneath has asserts that impose a "shall" condition.
Take your pick, but be consistent.

individual-it: I would even go a step further and keep as much assertions as possible in the context.
have a function `getSharingAutoCompleteUsers()` or similar and do the assertion in the context, I think that is the correct place to do it. Reason: you might want to get the users from an other context and do some other assertion with it. the "Page" just does date input/output and the decision if the behaviour is correct or not is done externally (by the user or the context)

individual-it:
@paurakhsharma please take over for fixing the acceptance tests

paurakhsharma:
@individual-it I have made the required changes. Please have a look

tests/acceptance/stepDefinitions/sharingContext.js
```diff
@@ -31,12 +32,29 @@ When('the user types {string} in the share-with-field', function (input) {
 })
 
 Then('all users and groups that contain the string {string} in their name should be listed in the autocomplete list on the webUI', function (pattern) {
-  return client.page.filesPage().sharingAutoCompleteShallContainElementsWith(pattern)
+  return client.page.filesPage().getShareAutocompleteItemsList()
+    .then(itemsList => {
+      itemsList.forEach(item => {
+        if (!item.toLocaleLowerCase().includes(pattern.toLocaleLowerCase())) {
```
individual-it: lets not convert this to lowercase, we can test how usernames that were created with upper-cases do act

individual-it:
@paurakhsharma please rebase and squash acceptance test commits into one

DeepDiver1975:
THX 👍

PR #1266, opened 2019-06-06
Added automatic lint after commit

LukasHirt:


ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUISharingInternalUsers</strong> failed. Please find the screenshots inside ...</summary><p>





</p></details>

package.json
```diff
@@ -95,5 +97,16 @@
...
-  ]
+  ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{vue,js}": [
+      "eslint --fix",
```
DeepDiver1975: better use yarn lint-fix?
Just to be sure the same config and file selection applies?

DeepDiver1975: design-system uses pretty-quick https://github.com/owncloud/owncloud-design-system/blob/33422423e959e0ccb8d6fa6c627304d10dac0d35/package.json#L28

LukasHirt: I tried to use `yarn lint --fix` but it didn't work. Not sure why. 🤷‍♂ 
Yes, it uses the same.
Pretty-quick is using prettier instead of lint and I think the rules there are a little bit different or am I just getting it wrong?

DeepDiver1975: > Pretty-quick is using prettier instead of lint and I think the rules there are a little bit different or am I just getting it wrong?

I have no idea to be honest. :man_shrugging:

PR #1332, opened 2019-06-19
parse viewer role properly

butonic:
No idea why I made viewer role consist of read, write and delete ... has been a while since I fell deeply in love with bitfcking ... oh wait ... that never happened ...


butonic:
running the web ui tests failed .. seems to be an unrelated problem AFAICT

PR #1381, opened 2019-06-28
Feature/breadcrumbs include home

felixheidecke:
## Description

Add the home directory to the breadcrumbs for easier implementation.

## Related Issue

Works properly with https://github.com/owncloud/owncloud-design-system/pull/375

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FilesAppBar.vue
+++ b/apps/files/src/components/FilesAppBar.vue
@@ -145,14 +145,15 @@ export default {
       }]
 
       if (this.$route.params.item) {
-        this.$route.params.item.split('/').forEach((element, index) => {
-          index++ // First item is home
+        let absolutePath = this.$route.params.item
+        const pathSplit = absolutePath.split('/').filter((val) => val)
+        for (let i = 0; i < pathSplit.length; i++) {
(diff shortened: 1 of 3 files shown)


apps/files/src/components/FilesAppBar.vue
```diff
@@ -114,18 +114,30 @@ export default {
...
+        to: '/files/list'
+      }]
+
+      if (this.$route.params.item) {
+        this.$route.params.item.split('/').forEach((element, index) => {
+          index++ // First item is home
+          breadcrumbs.push({
+            index,
+            text: element,
+            to: `${breadcrumbs[index - 1].to}/${element}`
```
felixheidecke: As we are exploding a URI coming from the router, I think we can forgo re-encoding it as it was done before.

DeepDiver1975: it has to be encoded - otherwise routing will not work - please bring back - THX

PR #1401, opened 2019-07-02
Update dependencies 20190702

DeepDiver1975:


package.json
```diff
@@ -42,40 +42,40 @@
...
-    "chromedriver": "^75.0.0",
+    "chromedriver": "^75.0.1",
     "copy-webpack-plugin": "^5.0.3",
     "core-js": "3.1.4",
     "css-loader": "^3.0.0",
     "cucumber": "cucumber/cucumber-js#issue-727-retry",
     "cucumber-pretty": "^1.5.2",
     "easygettext": "^2.7.0",
-    "eslint": "5.16.0",
+    "eslint": "6.0.1",
```
LukasHirt: Build when running `yarn watch-all` still fails because of the eslint.

DeepDiver1975: let's nail eslint to < 6 - THX - will take care

DeepDiver1975: updated ...

LukasHirt: Works fine now, thanks!

PR #1596, opened 2019-07-23
Disable caching for all Phoenix deliverables

DeepDiver1975:
##  Description
New caching setting ensure that all assets are not cached at all.
Pro: always get latest frontend deliverables
Con: more server load

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,12 +1,13 @@
 
 # BEGIN Caching
-<ifModule mod_headers.c>
-    <filesMatch "\\.(ico|jpg|jpeg|png|gif|ttf|otf|woff|woff2|eot|svg)$">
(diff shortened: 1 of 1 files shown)


LukasHirt:
Would maybe make sense to use files fingerprinting and then enforce no-cache only on the index.html? That way we could save a little bit from the server.

DeepDiver1975:
That will hardly Work With the dynamically loaded Apps in config.json

LukasHirt:
> That will hardly Work With the dynamically loaded Apps in config.json

Right, damn

PR #1738, opened 2019-08-08
Tests for collaborators roles

LukasHirt:
## Description
Added feature file with test for collaborator roles

## Motivation and Context
As seen in #1718 bug in collaborators role part can quite easily slip the current tests. To keep the tests also clean I'd say it makes sense to create this scenario in separate feature (this way the actual scenarios can be also with simple three steps Given, when, then). Also later we will need to check for custom role as well and also the combination can/cannot share.

## How Has This Been Tested?

- test environment: 🤖 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/webUISharingInternalUsers/checkUsersRole.feature
+++ b/tests/acceptance/features/webUISharingInternalUsers/checkUsersRole.feature
@@ -1,0 +1,16 @@
+Feature: Checking users role after he is added as collaborator
+  As a user
+  I want to assign specific roles to users
+  So that those users have only permissions the user wants them to have
+
+  Background:
+    Given these users have been created with default attributes:
+      | username |
+      | user1    |
+      | user2    |
+    And user "user1" has logged in using the webUI
+
+  Scenario: share a file with another user and assign him role Viewer
+    Given the user shares file "lorem.txt" with user "User Two" as "Viewer" using the webUI
+    When the user opens the share dialog for file "lorem.txt" using the webUI
+    Then user "User Two" should be listed as "Viewer" in the collaborators list on the webUI


tests/acceptance/features/webUISharingInternalUsers/checkUsersRole.feature
```diff
@@ -0,0 +1,16 @@
...
+
+  Background:
+    Given these users have been created with default attributes:
+      | username |
+      | user1    |
+      | user2    |
+    And user "user1" has logged in using the webUI
+
+  Scenario: share a file with another user and assign him role Viewer
+    Given the user shares file "lorem.txt" with user "User Two" as "Viewer" using the webUI
```
individual-it: If the sharing itself is something you want to test make it a "When" step
If its only a set-up step, leave it as "Given" but use the API (do not mention `using the webUI` and use present-perfect tense. Similar to existing steps like `user {string} has shared file/folder {string} with user {string}`

See also https://doc.owncloud.com/server/10.0/developer_manual/core/acceptance-tests.html#writing-a-given-step for the difference between Given and When steps

Most likely we want to have at least 2 tests here:
1. Given a file was shared -> When the user checks the share dialog -> Then the correct type of collaborators should be displayed
2. When a file was shared with an other collaborator of type X using the WebUI-> Then the correct permissions should be set (check by querying the API)

Scientifically correct there should be 2 scenarios, but that would nearly double the test-execution time, so I would suggest to be programmatic and combine both

```
When the user shares file "lorem.txt" with user "User Two" as "Viewer" using the webUI
And the user opens the share dialog for file "lorem.txt" using the webUI
Then user "User Two" should be listed as "Viewer" in the collaborators list on the webUI 
And the permissions of the last received share of "user2" should be "read,download"

LukasHirt: Then i'd say let's use the Given + api. For sharing we have already tests so we don't really need to take care of that here.

individual-it: what do you think about this approach: https://github.com/owncloud/phoenix/pull/1761
basically integrate this checks into the other sharing tests

LukasHirt: I personally don't like that the scenarios are starting getting really big but I totally understand the time saving reasons there. Closing this PR and leaving it for yours :wink:

PR #1895, opened 2019-09-04
Add download feedback

### LINKED ISSUE #526: File download gives no user feedback ....

### LINKED ISSUE #1808: file download does not work on password protected public shares
### Steps to reproduce
1. create a public link share of a folder
2. try to download a file from within that folder via public link

### Expected behaviour
file should be downloaded

### Actual behaviour
file cannot be downloaded

### Client configuration
**Browser:**
Chrome

#### Browser log
```
Failed to load resource: the server responded with a status of 401 (Unauthorized)
core.bundle.js:75 Error: Unauthorized
    at core.bundle.js:75```

LukasHirt:
## Description
Use request in file download method, add the file that is being downloaded into progress queue and disable file actions if download action is in progress.

## Related Issue

- Refs #526
This is a temporary solution since we wanted to rework the progress handler anyway and it also doesn't solve the issue with download being done in the background so just referencing the issue instead of closing it.
- Fixes #1808 

## Motivation and Context
When user downloads a big file he could start triggering download multiple times since he had no feedback until the file was downloaded in the background. With request, we can track the progress and add the file into the progress queue. To also prevent triggering the download again it's better to disable actions for the file that is being currently downloaded.

## How Has This Been Tested?

- test environment: Manually
1. Download a small file
2. Download a big file
3. Create public link
4. Download a file shared with public link
5. Create a password-protected public link
6. Download a file shared with a password-protected link

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FileList.vue
+++ b/apps/files/src/components/FileList.vue
@@ -36,7 +36,15 @@
               </oc-table-cell>
               <oc-table-cell :class="{ 'uk-visible@s' : _sidebarOpen }" class="uk-position-relative">
                 <div class="uk-button-group uk-margin-small-right" :class="{ 'uk-visible@m' : !_sidebarOpen, 'uk-visible@xl' : _sidebarOpen  }">
-                  <oc-button v-for="(action, index) in actions" :key="index" @click.stop="action.handler(item, action.handlerData)" :disabled="!action.isEnabled(item)" :icon="action.icon" :ariaLabel="action.ariaLabel" />
+                  <oc-button
+                    v-for="(action, index) in actions"
+                   :key="index"
+                   @click.stop="action.handler(item, action.handlerData)"
+                   :disabled="!action.isEnabled(item) || $_actionInProgress(item)"
+                   :icon="action.icon"
+                   :ariaLabel="action.ariaLabel"
+                   :uk-tooltip="$_disabledActionTooltip(item)"
+                  />
                 </div>
                 <oc-button
                   :id="'files-file-list-action-button-small-resolution-' + index"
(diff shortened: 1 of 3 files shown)


PVince81:
In OC core I remember that we had to introduce a magic cookie to detect whenever the download had started. Not sure why we don't need this approach here any more.

Ref: https://github.com/owncloud/core/blob/master/apps/files/ajax/download.php#L56 and https://github.com/owncloud/core/blob/master/apps/files/js/files.js#L313

Maybe it was about cases where the download is opened in a separate tab/window.

In the case of Phoenix we likely cannot work with such cookies, so not much point in trying to reproduce the case.

LukasHirt:
> In OC core I remember that we had to introduce a magic cookie to detect whenever the download had started. Not sure why we don't need this approach here any more.

I'm not really sure what was the reason to implement this in old UI or how this could be beneficial to Phoenix since we don't use redirection to any download URL but directly trigger the download of a file with `download` attribute (apart from ie11). Can you pls clarify? @PVince81

LukasHirt:
There is also `content-disposition: attachment` which should be able to allow the cross-origin download. Haven't yet worked with it so can't really say if it would fit our use case.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

PVince81:
@LukasHirt I don't think this is a concern for now, let's move forward.

PR #2025, opened 2019-09-23
Add same id for main file list and shared with pages

dpakach:
## Description

Change the id of files list on "Shared with Me" page to `files-list` from `shared-with-list`

## Motivation and Context

To make all the files list components consistent so that it would make it easier to run automated tests. There is already id `files-list` in favorites page.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/components/Collaborators/SharedFilesList.vue
+++ b/apps/files/src/components/Collaborators/SharedFilesList.vue
@@ -84,7 +84,7 @@ export default {
 </script>
 
 <template>
-  <oc-table middle divider class="oc-filelist" id="shared-with-list" v-if="!loadingFolder">
+  <oc-table middle divider class="oc-filelist" id="files-list" v-if="!loadingFolder">
     <oc-table-group>
       <oc-table-row>
         <oc-table-cell type="head" class="uk-text-truncate" v-translate>Name</oc-table-cell>


LukasHirt:
Would it maybe make sense to use class here instead and preserve the view specific ids? I guess in situation when we want to check for some view specific columns (e. g. "Shared from" or "Status") would the id be useful. What do you think @individual-it ?

dpakach:
> Would it maybe make sense to use class here instead and preserve the view specific ids? I guess in situation when we want to check for some view specific columns (e. g. "Shared from" or "Status") would the id be useful. What do you think @individual-it ?

I think that would be good too. As long as we have consistency and we can use simple selectors for tests, then it easier.

individual-it:
what about using the same id everywhere and having different `aria-label` to describe it in detail?

PVince81:
using the same id in multiple places of the DOM is asking for trouble

I already lost some time trying to debug an issue where Selenium was not able to find an element. The reason was that there were two elements in the DOM with the same id, so Selenium ignored it.

Let's use another attribute like class or aria-label for look-up, but avoid duplicating ids.

dpakach:
> using the same id in multiple places of the DOM is asking for trouble
> 
> I already lost some time trying to debug an issue where Selenium was not able to find an element. The reason was that there were two elements in the DOM with the same id, so Selenium ignored it.
> 
> Let's use another attribute like class or aria-label for look-up, but avoid duplicating ids.

Ok sounds good :+1: . In that case we also need to fix the favorites page because it uses same id as the main file list.

dpakach:
Looks like favorites page and all files page use same component. So closing this for now.

PR #2079, opened 2019-09-26
In case of auth errors (access denied, invalid token or cannot refres…

DeepDiver1975:
## Description
In case of auth errors (access denied, invalid token or cannot refresh token) the state is cleared - otherwise there is no way to get out of this

## How Has This Been Tested?
- login via oauth to owncloud
- wait until the token times out (30min afaik)
- access denied page pops up
- go back to login page
- login page shall appear (without this patch the accessDenied page will show up)

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/services/auth.js
+++ b/src/services/auth.js
@@ -92,6 +92,9 @@ export function initVueAuthenticate (config) {
       logout () {
         return mgr.signoutRedirect()
       },
+      clearLoginState () {
+        return mgr.removeUser()
+      },
       mgr: mgr,
       events () {
         return mgr.events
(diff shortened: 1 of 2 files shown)


PVince81:
I happened to be stuck in such loop so I checked out this branch and the problem went away :+1:

PVince81:
@DeepDiver1975 please evaluate if acceptance testing is possible and if yes raise a ticket for it

PR #2117, opened 2019-10-01
Fix password handling in public links

felixheidecke:
## Description

- Immediate icon change if the link is protected (or not)
- Allow removing the password from the form.

## Related Issue

- Fixes https://github.com/owncloud/phoenix/issues/1999
- Fixes https://github.com/owncloud/phoenix/issues/2000

## Motivation and Context
See issues

## How Has This Been Tested?

Manual test

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FileLinkForm.vue
+++ b/apps/files/src/components/FileLinkForm.vue
@@ -59,7 +59,10 @@
       </div>
       <div class="uk-width-1-1 uk-width-3-5@m">
         <label class="uk-form-label" for=""><span v-translate>Password</span><em v-if="$_passwordEnforced" class="uk-margin-small-left">(<span v-translate>required</span>)</em></label>
-        <input type="password" autocomplete="new-password" class="uk-input" :class="{ 'uk-form-danger': !$_passwordIsValid }" v-model="password" :placeholder="hasPassword && password === null? '********' : placeholder.password" id="oc-files-file-link-password" />
+        <div class="uk-position-relative">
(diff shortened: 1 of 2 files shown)


apps/files/src/components/FileLinkForm.vue
```diff
@@ -59,7 +59,10 @@
       </div>
       <div class="uk-width-1-1 uk-width-3-5@m">
         <label class="uk-form-label" for=""><span v-translate>Password</span><em v-if="$_passwordEnforced" class="uk-margin-small-left">(<span v-translate>required</span>)</em></label>
-        <input type="password" autocomplete="new-password" class="uk-input" :class="{ 'uk-form-danger': !$_passwordIsValid }" v-model="password" :placeholder="hasPassword && password === null? '********' : placeholder.password" id="oc-files-file-link-password" />
+        <div class="uk-position-relative">
+          <input type="password" autocomplete="new-password" class="uk-input" :class="{ 'uk-form-danger': !$_passwordIsValid }" v-model="password" :placeholder="hasPassword && password === null? '********' : placeholder.password" id="oc-files-file-link-password" />
+          <div uk-close v-if="!$_passwordEnforced && hasPassword && password!=''" class="uk-position-small uk-position-center-right oc-cursor-pointer" @click="password=''" :uk-tooltip="$_passwordRemoveText" />
```
PVince81: seems the `_passwordEnforced` condition does not do anything.

I tried enforcing password only for the "Editor" role and the cross was still visible to discard the password.

Maybe the `_passwordEnforced` function is not working ? As I can see it's not using proper bitmasks...

PVince81:
Tested, all works except for the password enforced bit.

Still, even without that, an error message is properly displayed in case I try to save a link without password while password is enforced. So merging this even without the check would be fine by me for now...

PR #2129, opened 2019-10-02
Rename and delete will be disallowed in case the parent folder has no…

DeepDiver1975:
… permissions fot these two operations.

## Description
In some setups (Samba) the permission of the parent folder has a higher priority to disallow rename and delete then the child item itself.

## How Has This Been Tested?
- needs to be verified on acl environment

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/components/FileList.vue
+++ b/apps/files/src/components/FileList.vue
@@ -40,7 +40,7 @@
                     v-for="(action, index) in actions"
                    :key="index"
                    @click.stop="action.handler(item, action.handlerData)"
-                   :disabled="!action.isEnabled(item) || $_actionInProgress(item)"
+                   :disabled="!action.isEnabled(item, parentFolder) || $_actionInProgress(item)"
                    :icon="action.icon"
                    :ariaLabel="action.ariaLabel"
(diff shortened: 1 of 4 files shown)


apps/files/src/components/FileList.vue
```diff
@@ -128,7 +128,7 @@ export default {
     Mixins
   ],
   name: 'FileList',
-  props: ['fileData', 'starsEnabled', 'checkboxEnabled', 'dateEnabled'],
+  props: ['fileData', 'starsEnabled', 'checkboxEnabled', 'dateEnabled', 'parentFolder'],
```
LukasHirt: Why using prop instead of getter from state directly in component?

DeepDiver1975: Just following the existing patter - files are pushed in via props - just doing it the same way for the parent folder.

I didn't want to start a refactor now to get rid of all the props :man_shrugging:

LukasHirt: I see. Thanks for info

PR #2177, opened 2019-10-08
Handle errors while deleting and renaming files

DeepDiver1975:
## Description
Delete and rename operations did not handle error conditions at all before this PR.
While it is UX-wise far from good - this is better then nothing.

The whole error handling flow as well as the user confirmation flow is under review at the moment.

As say: better then nothing

## How Has This Been Tested?
- Lock a file using webdav locks (cadaver or manual insert into database)
- try to delete or rename
- see error notification

Screenshot from 2019-10-08 14-17-53
Screenshot from 2019-10-08 14-17-39

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/store/actions.js
+++ b/apps/files/src/store/actions.js
@@ -446,7 +446,7 @@ export default {
       context.commit('ADD_FILE', _buildFile(file))
     }
   },
-  deleteFiles (context, { files, client, publicPage }) {
+  deleteFiles (context, { files, client, publicPage, $gettext, $gettextInterpolate }) {
     const promises = []
     for (const file of files) {
       let p = null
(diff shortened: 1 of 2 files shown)


apps/files/src/store/actions.js
```diff
@@ -459,7 +459,16 @@ export default {
...
       }).catch(error => {
-        console.log('error: ' + file.path + ' not deleted: ' + error)
+        let translated = $gettext('Error while deleting "%{file}"')
+        if (error.statusCode === 423) {
+          translated = $gettext('Error while deleting "%{file}" - the file is locked')
+        }
+        const title = $gettextInterpolate(translated, { file: file.name }, true)
+        context.dispatch('showMessage', {
+          title: title,
+          desc: error.message,
```
LukasHirt: This will result in two "locked" messages, won't it? One in title, one in desc

PR #2179, opened 2019-10-08
Download feedback spinner

LukasHirt:
## Description
Use spinner to show that item has action in progress and remove download from upload menu

## Motivation and Context
Upload menu was now restructured and the download doesn't fit there anymore and since actions in general and their feedback is going to be reworked (already were some mockups for this but haven't been worked on it since) and quick solution is needed now it doesn't make too much sense to me to spend now a lot of effort to try to keep them together in a nice way in the upload menu.

## How Has This Been Tested?

- test environment: Manually
1. Download any file

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FileList.vue
+++ b/apps/files/src/components/FileList.vue
@@ -27,6 +27,12 @@
                 <oc-file @click.native.stop="item.type === 'folder' ? navigateTo(item.path.substr(1)) : openFileActionBar(item)"
                         :name="$_ocFileName(item)" :extension="item.extension" class="file-row-name" :icon="fileTypeIcon(item)"
                         :filename="item.name" :key="item.id"/>
+                <oc-spinner
+                  v-if="$_actionInProgress(item)"
+                  size="small"
+                  :uk-tooltip="$_disabledActionTooltip(item)"
+                  class="uk-margin-small-left"
+                />
               </oc-table-cell>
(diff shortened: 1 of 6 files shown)


apps/files/src/components/FileList.vue
```diff
@@ -210,11 +217,16 @@ export default {
       }
 
       return null
+    },
+
+    $_disabledActionSpinnerTooltip (item) {
+      const message = this.$gettext('There is currently an action in progress for this %{itemType}')
+      return this.$gettextInterpolate(message, { itemType: item.type })
```
DeepDiver1975: item.type is either file or folder - this shall be translated as well.

LukasHirt: Right - fixed 👍

apps/files/src/components/FileList.vue
```diff
@@ -200,21 +207,32 @@ export default {
...
+
+        return this.$gettext('Another action is currently in progress for this file')
       }
 
       return null
+    },
+
+    $_disabledActionSpinnerTooltip (item) {
+      if (item.type === 'folder') {
+        return this.$gettext('There is currently an action in progress for this folder')
```
DeepDiver1975: can we reuse the message as above? We shall try to reuse existing messages as far as possible. This will reduce workload for translators.

LukasHirt: Sure, I'll adjust it.

LukasHirt: Done

PR #2198, opened 2019-10-09
Reset search value on clear action

### LINKED ISSUE #2120: [QA] "X" icon in search input does not delete text
* Click into the search input field
* type any text
* click on the "X" icon

The "X" icon disappears. The text remains
Expected result: The text should be deleted.

LukasHirt:
## Description
If a user clicks on the "x" icon next to the text inside of search bar the value gets now deleted.

## Related Issue

- Fixes #2120

## How Has This Been Tested?

- test environment: Manually
1. Enter some text into search bar
2. Click on the "x" icon
3. See icon disappear and the text being deleted
4. Enter some text into search bar and hit enter
5. Repeat step 2. after search has been completed.
6. Repeat with mobile search

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FilesAppBar.vue
+++ b/apps/files/src/components/FilesAppBar.vue
@@ -20,7 +20,16 @@
         </span>
       </div>
       <div v-if="!publicPage()" class="uk-width-auto uk-visible@m">
-        <oc-search-bar @search="onFileSearch" :value="searchTerm" :label="searchLabel" :loading="isLoadingSearch" :button="false"/>
+        <oc-search-bar
+          @search="onFileSearch"
+          :value="searchTerm"
+          :label="searchLabel"
+          :loading="isLoadingSearch"
+          :button="false"
+          @clear="onSearchClear"
+          :key="searchBarKey"
+          ref="globalSearchBar"
+        />
       </div>
       <div class="uk-width-auto">
         <div class="uk-button-group" :key="actionsKey">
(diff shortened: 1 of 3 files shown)


apps/files/src/components/FilesAppBar.vue
```diff
@@ -479,6 +496,12 @@ export default {
       setTimeout(() => {
         self.linkCopied = false
       }, 1000)
+    },
+
+    onSearchClear () {
+      this.resetSearch()
+      // Forces update of search bar
+      this.searchKey++
```
PVince81: searchBarKey ? if this worked for you without this correction then maybe the key is not needed ?

please add a comment for "searchBarKey" to explain why it's needed at all

LukasHirt: It is needed since we are setting empty string in store when resetting search which doesn't trigger properly the re-render of components. Changing key is a good way how to force the component to re-render. I'll add the comment.

PVince81: yes, understood, but in the top of the file you called it `searchBarKey` https://github.com/owncloud/phoenix/pull/2198/files#diff-09c6ca55e7ff0a167829a169b4a0d1aaR116

or is that yet another property ?

PVince81:
is the dependencies update intended in this PR ?

LukasHirt:
> is the dependencies update intended in this PR ?

Yes, it is. It brings a new event in the search bar component `onClear`

PVince81:
CI restarted

PR #2225, opened 2019-10-11
Handle remove from progress in its own mutation

LukasHirt:
## Description
Move remove from progress into its own method, use id instead of name to find the right progress and call the action on requests loadend event.

## How Has This Been Tested?

- test environment: Manually
1. Upload one file
2. Upload multiple files
3. Upload folder

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/FileUpload.js
+++ b/apps/files/src/FileUpload.js
@@ -1,11 +1,12 @@
 class FileUpload {
-  constructor (file, path, url, headers = {}, onProgress = () => {}, type = 'POST') {
+  constructor (file, path, url, headers = {}, onProgress = () => {}, type = 'POST', onUploadEnded = () => {}) {
     this.file = file
     this.path = path
     this.url = url
     this.headers = headers
     this.onProgress = onProgress
     this.type = type
+    this.onUploadEnded = onUploadEnded
   }
 
   upload (options = {}) {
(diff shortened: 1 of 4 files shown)


apps/files/src/FileUpload.js
```diff
@@ -31,10 +32,14 @@ class FileUpload {
     xhr.upload.addEventListener('progress', (e) => {
       this.onProgress(e, this.file)
     }, false)
+
     const promise = new Promise((resolve, reject) => {
       xhr.onload = e => {
         xhr.status >= 200 && xhr.status < 400 ? resolve(e) : reject(new Error(xhr.statusText))
       }
+      xhr.onloadend = e => {
```
PVince81: is this polyfilled for IE and co ? see https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onloadend

PVince81: AFAIK using onload should be enough here ?

LukasHirt: Tested it in ie11 - works.

The good thing about onloadend is that it's triggered also in case it failed.

PVince81: ok, so if it works it means we have a polyfill in place.

note that chrome also doesn't support this property, mind testing there as well ?

LukasHirt: Already tested. Works in IE11, FF, chrome, safari and edge.

PR #2325, opened 2019-10-25
Adjusted changelog and version to 0.2.5

LukasHirt:


CHANGELOG.md
```diff
@@ -4,6 +4,89 @@ All notable changes to this project will be documented in this file.
...
+### Added
+- Test multiple collaborators on a single share https://github.com/owncloud/phoenix/pull/2081
+- Run tests on internet explorer https://github.com/owncloud/phoenix/pull/2082
+- Draw.io app integration https://github.com/owncloud/phoenix/pull/2083
+- Await trashbin deletion requests in acceptance tests https://github.com/owncloud/phoenix/pull/2093
+- Check that required permissions are displayed for files https://github.com/owncloud/phoenix/pull/2098
+- Logout step for acceptance tests https://github.com/owncloud/phoenix/pull/2100
+- New file menu entries for different file types https://github.com/owncloud/phoenix/pull/2111
+- Drone starlark https://github.com/owncloud/phoenix/pull/2112
+- Properly reset capabilities on logout https://github.com/owncloud/phoenix/pull/2116
```
PVince81: that should belong under "Fixed" ?

CHANGELOG.md
```diff
@@ -4,6 +4,89 @@ All notable changes to this project will be documented in this file.
...
+- Disable drag and drop for ie11 https://github.com/owncloud/phoenix/pull/2128
```
PVince81: this would be for the "Removed" section and maybe add a quick text "because of compatibility issues" or something

CHANGELOG.md
```diff
@@ -4,6 +4,89 @@ All notable changes to this project will be documented in this file.
...
+- Rename and delete will be disallowed in case the parent folder has no permissions fot these two operations https://github.com/owncloud/phoenix/pull/2129
+- Acceptance tests for deleting shared files and deleting shares https://github.com/owncloud/phoenix/pull/2140
+- Reject promise if invalid response during permalink check in acceptance tests https://github.com/owncloud/phoenix/pull/2141
+- Acceptance test for checking delete button state when prohibiting deleting in share https://github.com/owncloud/phoenix/pull/2146
```
PVince81: note that for core I usually would not add any entries that only add tests, not sure if we should that the same here

CHANGELOG.md
LukasHirt: Will move it there. I originally thought it fits here because the reset wasn't there at all. But yeah, make sense to have it in Fixed. Thanks

CHANGELOG.md
```diff
@@ -4,6 +4,54 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## [0.2.5]
+### Added
+- Run tests on internet explorer https://github.com/owncloud/phoenix/pull/2082
+- Draw.io app integration https://github.com/owncloud/phoenix/pull/2083
+- Logout step for acceptance tests https://github.com/owncloud/phoenix/pull/2100
```
PVince81: also remove test entries such as these ?

Issue #2324, opened 2019-10-25
Skip links


marcus-herrmann:
# Why skip links are useful

> The main content is not usually the first thing on a web page. Keyboard and screen reader users generally must navigate a long list of navigation links, sub-lists of links, corporate icons, site searches, and other elements before ever arriving at the main content. This is particularly difficult for users with some forms of motor disabilities ([Source](https://webaim.org/techniques/skipnav/))

Imagine you are a keyboard only user, already know the overall structure of the website or web app, have recognized which items repeat on every page or view and you only want to navigate to the main content. Without a shortcut to the main content (that is a skip link) you have to tab through all focusable items of the header, the navigation and all other items coming before the main content. It is best practice (and also a guideline specified by the WCAG to offer means "for bypassing repetitive blocks of content". Skip links work best in conjunction with ARIA landmarks (`main`, `nav`, `[role="search"]`, `aside`, `footer` and so on) because landmarks in combination with a good headline structure and skip links offer screen reader users multiple ways of navigating to functionality or content directly (or giving them them the opportunity to get a quick overview of the website).

# How to implement them

To make most out of skip links, place them directly as the first element of the `<body>`. With this (and given there is no `tabindex` tinkering) they should be the first item of the tabindex (meaning: the first item that will get focus in the document once you hit the tab key).

Skip links don't have to be visible all the time, just on focus is fine. This can be solved in the following manner:

    a#skiptocontent {
    	position: absolute;
    	top: -40px;
    	left: 0px;
    }
    
    a#skiptocontent:focus {
    	top: 0;
    }

To use a visibility-hidden class instead of positioning the non-focused skip link off-canvas is also fine – as long as all the CSS properties that make it hidden in the first place (`position`, `clip`, `height`, `width`, `overflow` and so on) are reset **properly.**

Regarding wording, a simple "Skip to [target]" works best, so for example "Skip to main content". In my opinion it is better to call out to where the link interaction leads, instead of what exactly the links skips over.

While the link itself could be a component implementing one or more skip links means to, if necessary, add the `id` attribute to the skip targets – such such a component would not work in isolation.

# Good examples in the wild

Just visit the site and, before doing anything, hit your keyboards TAB key:

- [https://www.deque.com/](https://www.deque.com/)
- [https://developer.mozilla.org/](https://developer.mozilla.org/)
- [https://tatianamac.com/](https://tatianamac.com/)
- [https://www.gov.uk/](https://www.gov.uk/)

[...]

PR #2326, opened 2019-10-25
Add skip to component, id attribute for <main>

marcus-herrmann:
### Adds skipTo functionality
Unfortunately vue-router has problem with the usual approach to implent skip link functionality, with links `<a href="#main">Skip to main content</a>`.  Issues around this are open for years ([#1668 being the main and original one](https://github.com/vuejs/vue-router/issues/1668). But none of the workarounds listed there are actually moving focus to the container with the "#"-provided id (they are rather meddling with window.history or scrolling things into visible viewport, both of which hasn't reliable influence on the keyboard focus). So I solved the skipTo functionality with a button, which is not ideal, but at least a comprehensible way (compared to changes in the router itself). Once/if this behaviour will be resolved we could easily swap the `<skip-to>` component with a link.

### Why no component in ODS?
Several reasons:
* The skipTo component can't be properly presented in the Design System since it starts invisible
* The component is not independent from the DOM structure
* I do not expect the CSS to be ever reused somewhere else.

## Related Issue
#2324.

- Fixes <issue_link>

## Motivation and Context
[Reasoning](https://github.com/owncloud/phoenix/issues/2324#issue-512457351)

## How Has This Been Tested?
Manually, only.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/src/components/SkipTo.vue
+++ b/src/components/SkipTo.vue
@@ -1,0 +1,51 @@
+<template>
+  <button class="skip-button" @click="skipToTarget"><slot></slot></button>
+</template>
+
+<script>
+export default {
+  props: {
+    /*
+    * The element to focus and to skip to
+    */
+    target: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    targetElement () {
+      return document.getElementById(this.target)
+    }
+  },
+  methods: {
+    skipToTarget () {
+      // Make targetElement programmatically focusable
+      this.targetElement.setAttribute('tabindex', '-1')
+
+      // Focus and scroll into view
+      this.targetElement.focus()
+      this.targetElement.scrollIntoView()
+    }
+  }
+}
+</script>
+
+<style scoped>
+.skip-button {
+  position: absolute;
+  top: -100px;
+  left: 0;
+  z-index: 1;
+  -webkit-appearance: none;
+  border: none;
+  background-color: #ffffff;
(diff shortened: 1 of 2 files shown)


src/components/SkipTo.vue
```diff
@@ -0,0 +1,51 @@
...
+      // Focus and scroll into view
+      this.targetElement.focus()
+      this.targetElement.scrollIntoView()
+    }
+  }
+}
+</script>
+
+<style scoped>
+.skip-button {
```
PVince81: would it be possible to use UIKit's 'uk-invisible` class instead ?

src/components/SkipTo.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <button class="skip-button" @click="skipToTarget"><slot></slot></button>
```
PVince81: depending how often we're going to need this, maybe we can implement this in the design system ?

marcus-herrmann: If the interface stays like this (the navigation being hidden away off-screen initially anyway) only one instance of a skip link (the one going to main) is necessary, in my opionion. Regarding other thoughts on SkipTo and ODS, see "Why no component in ODS?" above.

PR #2408, opened 2019-11-07
WIP public link separate token from path

PVince81:
This was an attempt based on https://github.com/owncloud/phoenix/pull/2239 to separate the token from the path in public link. However this seems to be a rabbit hole that needs more attention later.

I'm leaving it in this PR for now for later reference / recycling.


kulmann:
Thanks for leaving this here! Has been solved "already" during the last 1,5 years. Closing.

PR #2542, opened 2019-11-25
Add focus management for uploads

marcus-herrmann:
Regarding: progressbar, to listwrapper on complete

## Description
Improves upload user flow for screen reader users by making them programmatically aware of 
* the interstitial interface state of an upload in progress
* the end of mentioned upload progress

## Related Issue

- Fixes https://github.com/owncloud/owncloud-design-system/issues/536
- Hard dependency on https://github.com/owncloud/owncloud-design-system/pull/556

## Motivation and Context
While visual users can comprehend that an (larger) upload is running and get visual clues about its end, screen readers stay silent during the progress. This PR tackles this issue by adding two steps:
1. When the upload starts programmatic focus will be sent to the progressbar. In the related ODS PR `OcProgress` is improved with aria states/information about the upload so screen reader can pick its state up. In order to do so the progressbar has to have focus.
2. Once the upload has finished screen reader will be informed about the completion via a live region and focus will be sent to the container that wraps the file lists (where the newly uploaded files then will recide).

## How Has This Been Tested?
- No automatic tests (related to my issues with the overall test setup)
- Manual tests with two tools
  - Tracking of the focus in the browser's console via the following JavaScript `document.addEventListener('focusin', () => {console.log(document.activeElement);});`
  - [NerdeRegion Chrome extension for a visible output of live regions](https://chrome.google.com/webstore/detail/nerderegion/lkcampbojgmgobcfinlkgkodlnlpjieb)

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes

## Open tasks:

- [ ] https://github.com/owncloud/owncloud-design-system/pull/556 has to be merged prior to this PR

### DIFF
--- a/apps/files/src/components/FilesApp.vue
+++ b/apps/files/src/components/FilesApp.vue
@@ -3,7 +3,7 @@
       <files-app-bar />
       <upload-progress v-show="inProgress.length" class="uk-padding-small uk-background-muted" />
       <oc-grid class="uk-height-1-1 uk-flex-1 uk-overflow-auto">
-        <div class="uk-width-expand uk-overflow-auto uk-height-1-1" @dragover="$_ocApp_dragOver" :class="{ 'uk-visible@m' : _sidebarOpen }">
+        <div ref="filesListWrapper" tabindex="-1" class="uk-width-expand uk-overflow-auto uk-height-1-1" @dragover="$_ocApp_dragOver" :class="{ 'uk-visible@m' : _sidebarOpen }">
           <oc-loader id="files-list-progress" v-if="loadingFolder"></oc-loader>
           <trash-bin v-if="$route.name === 'files-trashbin'" :fileData="activeFiles" />
           <shared-files-list v-else-if="sharedList" :fileData="activeFiles" @sideBarOpen="openSideBar" />
@@ -142,7 +142,11 @@ export default {
       this.renameDialogErrorMessage = this.validateFileName(value)
     }
   },
-
+  created () {
+    this.$root.$on('upload-end', () => {
+      this.delayForScreenreader(() => this.$refs.filesListWrapper.focus())
+    })
+  },
   computed: {
     ...mapGetters('Files', [
       'selectedFiles', 'activeFiles', 'dropzone', 'loadingFolder', 'highlightedFile', 'currentFolder', 'inProgress',
--- a/apps/files/src/components/UploadProgress.vue
+++ b/apps/files/src/components/UploadProgress.vue
@@ -10,6 +10,7 @@
         icon="expand_more"
       />
       <translate
+        id="upload-label"
         :translate-n="count"
         translate-plural="Uploading %{ count } items"
       >
@@ -21,13 +22,16 @@
         class="uk-width-1-1 uk-width-medium@s"
       >
         <oc-progress
+          ref="progressbar"
           :value="totalUploadProgress"
           :max="100"
+          aria-labelledby="upload-label"
           class="uk-width-expand uk-margin-remove-bottom"
         />
         <span>
           {{ totalUploadProgress | roundNumber}} %
         </span>
+        <oc-hidden-announcer level="assertive" :announcement="announcement" />
       </oc-grid>
     </oc-grid>
     <oc-drop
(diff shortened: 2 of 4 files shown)


apps/files/src/components/FilesApp.vue
```diff
@@ -77,6 +78,14 @@ export default {
       console.info('trace', arguments)
     },
 
+    onUploadStart () {
+      this.progressbarFocussed = true
```
PVince81: what if the user tabs away, do you reset this ? seems the flag is not really useful, at least not in this form. maybe you intended to use it elsewhere ?

apps/files/src/components/FilesApp.vue
```diff
@@ -1,9 +1,9 @@
   <template>
     <div id="files" class="uk-flex uk-flex-column">
-      <files-app-bar />
-      <upload-progress v-show="inProgress.length" class="uk-padding-small uk-background-muted" />
+      <files-app-bar @upload-start="onUploadStart" @upload-end="onUploadEnd" />
```
PVince81: indeed, it starts to feel like we need some event bus to globally detect such events like uploads

apps/files/src/components/FilesApp.vue
marcus-herrmann: The flag is only meant to be watched if it once changes from `false` to `true`, so it doesn't really matter if it's set to true for too long. As said in the chat, hacky solution, and as you are thinking out loud further below, maybe a refactor using an EventBus is better.

apps/files/src/components/FilesApp.vue
marcus-herrmann: On it.

marcus-herrmann: Event bus established here: https://github.com/owncloud/phoenix/pull/2542/files#diff-c900a0e34358d44a55381e5e63de50caR70

In use for example here: https://github.com/owncloud/phoenix/pull/2542/files#diff-eec759a01d0f06a12081c5e8285c0d85R68

src/phoenix.js
```diff
@@ -65,6 +65,10 @@ Vue.use(VueMeta, {
 Vue.component('drag', Drag)
 Vue.component('drop', Drop)
 
+// --- Event Bus ----
+
+Vue.prototype.$bus = new Vue()
```
PVince81: @LukasHirt can you confirm that this makes sense ?

LukasHirt: TBH I'd rather see event bus avoided. @marcus-herrmann is it necessary here? Would it be much trouble covering this with state? If needed to stay, pls make the bus a constant and import it only where it's needed.

apps/files/src/components/UploadProgress.vue
```diff
@@ -9,25 +9,30 @@
...
         icon="expand_more"
       />
-      <translate
-        :translate-n="count"
-        translate-plural="Uploading %{ count } items"
-      >
-        Uploading %{ count } item
-      </translate>
+      <span id="upload-label">
+        <translate
```
LukasHirt: Translate tag is a span in default. You can assign the id directly and avoid one extra tag.

src/phoenix.js
marcus-herrmann: Covering it with state was my first approach, I talked with @PVince81 about that and we agreed that this approach was overengeenering/not what the store was meant to be. Keep in mind that there would be more occurances for this, to send for example focus across the interface/components that are not close to each other.

marcus-herrmann: @LukasHirt Here's the first approach mentioned above: https://github.com/owncloud/phoenix/pull/2624

apps/files/src/components/UploadProgress.vue
marcus-herrmann: Thx. Fixed.

src/phoenix.js
PVince81: @LukasHirt are we ok with the bus then ?

PVince81: now thinking, I remember seeing that it's possible to access the root component using `this.$root`, would it be possible to use that component as the bus instead ?

marcus-herrmann: This appears to work as well. PR updated

PVince81:
Rebased

PR #2535, opened 2019-11-25
Test: Use path helpers in acceptance tests

skshetry:
## Description

This introduces some path helpers (`resolve`, `normalize` and `join`) to use in our acceptance tests.

#### Why not use some npm lib?
This is because, we have some specific requirements (such as there's no difference between `/path` and `path`, and we will always love to work with the one without slashes and not being tied to os filesystem,i.e only lexical path transformation).

## Related Issue

- Related to #2499 

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #2571, opened 2019-11-26
Test: Set skeletondirectory when running tests

### LINKED ISSUE #2568: Test: Set skeletondirectory when running tests
We already set skeletondirectory automagically on core. It'd be great to have it set here as well.

skshetry:
## Description

This PR sets `skeletondirectory` when running tests. Sadly, `client` is not ready when`BeforeAll` runs, so I have to run this in `Before`.

## Related Issue

- Fixes #2568

## Motivation and Context

Make it easier to run tests for everyone, and one less thing for me to remember

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/helpers/config.js
+++ b/tests/acceptance/helpers/config.js
@@ -4,6 +4,10 @@ const occHelper = require('./occHelper')
 const { join } = require('./path')
 const { difference } = require('./objects')
 const _ = require('lodash')
+const pLimit = require('p-limit')
+
+// run 10 promises at once at max
+const limit = pLimit(10)
 
 let config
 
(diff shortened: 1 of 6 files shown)


skshetry:
Since this does not affect CI, I request all reviewers to make sure this works when they checked out this branch locally. :)

skshetry:
@dpakach, `core` sets whatever it needs to write on the core, and if someone tries to run acceptance tests, it will swiftly shift to the required skeleton directory, so, that won't be much problem (except for polluting folders for new users). But, I don't have strong opinion towards any, but would prefer to not reset the values at the end for simplicity.

tests/acceptance/setup-local.js
```diff
@@ -1,13 +1,37 @@
 import { setDefaultTimeout, After, Before } from 'cucumber'
-import { createSession, closeSession, startWebDriver, stopWebDriver } from 'nightwatch-api'
+import { createSession, closeSession, startWebDriver, stopWebDriver, client } from 'nightwatch-api'
+import join from 'join-path'
```
individual-it: shouldn't we use `join` from `helpers/path`? see #2535

skshetry: I was just lazy to import from `./helpers/path`, as the behavior is same. Fixing it.

tests/acceptance/helpers/config.js
```diff
@@ -0,0 +1,115 @@
...
+  const newConfig = await getConfigs()
+
+  const appConfig = _.get(newConfig, 'apps')
+  const systemConfig = _.get(newConfig, 'system')
+
+  const initialAppConfig = _.get(config, 'apps')
+  const initialSysConfig = _.get(config, 'system')
+
+  await rollbackSystemConfigs(
+    difference(systemConfig, initialSysConfig),
```
individual-it: feels like I would give the function both configs and do the diff inside the rollback function

tests/acceptance/helpers/config.js
```diff
@@ -0,0 +1,122 @@
+const fetch = require('node-fetch')
+const httpHelper = require('./httpHelper')
+const occHelper = require('./occHelper')
+const { join } = require('./path')
+const { difference } = require('./objects')
+const _ = require('lodash')
+const pLimit = require('p-limit')
+
+// run 10 promises at once at max
+const limit = pLimit(10)
```
individual-it: why that?

skshetry: After your previous review, I made changing all those configurations concurrent. Just tried to future-proof and not overwhelm our computers. I know 10 is already a lot.

PR #2613, opened 2019-12-03
[wip] GRPC client

diocas:
Hi,

This PR is a work in progress and the idea is to start the discussion on how to provide a GRPC client for Phoenix. This discussion was started here: https://github.com/owncloud/owncloud-sdk/pull/384

## Motivation and Context
At CERN we want to test the feasibility of using GRPC between REVA and the webapp. As it was agreed with Owncloud, to accomplish this the owncloud-sdk should define a very clear interface, usable by Phoenix, and it should be modular in order to easily replace the current webdav implementation with something else. But this is not the case at the moment.

## Description
This PR strips Phoenix completely of webdav. This is the only way we could provide a common interface for any access type! Phoenix should only know about the sdk interfaces/objects and nothing else.
I would like to know what information is required by the UI so that we can translate this into the methods and objects that sdk should provide.

## How Has This Been Tested?
As mentioned, this PR is a work in progress and things are not guaranteed to work. I did basic testing using the interface. With the GRPC client, for now only the listing of folders is implemented.


labkode:
@felixboehm @DeepDiver1975 please give feedback

apps/files/src/components/PublicLinks/FilesDrop.vue
```diff
@@ -105,15 +105,7 @@ export default {
...
-      const properties = this.davProperties.concat([
-        this.$client.publicFiles.PUBLIC_LINK_ITEM_TYPE,
-        this.$client.publicFiles.PUBLIC_LINK_PERMISSION,
-        this.$client.publicFiles.PUBLIC_LINK_EXPIRATION,
-        this.$client.publicFiles.PUBLIC_LINK_SHARE_DATETIME,
-        this.$client.publicFiles.PUBLIC_LINK_SHARE_OWNER
-      ]
-      )
-      this.$client.publicFiles.list(this.publicLinkToken, this.publicLinkPassword, properties, '0').then(files => {
+      this.$client.publicFiles.listExtended(this.publicLinkToken, this.publicLinkPassword, '0').then(files => {
```
DeepDiver1975: I disagree to this - @labkode we have been chatting about this - the client (phoenix) needs a way to define the properties which shall be listed.

This applies to all other ocurances where davProperties has been removed from the code base.

package.json
```diff
@@ -105,7 +105,11 @@
     "webpack-version-file-plugin": "^0.4.0",
     "whatwg-fetch": "^3.0.0",
     "wicked-good-xpath": "^1.3.0",
-    "xml-js": "^1.6.11"
+    "xml-js": "^1.6.11",
+    "@grpc/proto-loader": "^0.3.0",
```
DeepDiver1975: Do we really need these deps in here? This feels wrong

apps/files/src/components/PublicLinks/FilesDrop.vue
diocas: If the client defines this, then we cannot support GRPC... The interfaces need to be defined a priori so that both webdav and grpc know how to honour them. 
And I don't see any reason why Phoenix needs to define this. If they need some extra value, then it should be implemented in the sdk and not in the client.

diocas: Hugo discussed this with me. I was unaware that you might want to filter information returned. We could think of a way to filter, but I would still remove all the webdav stuff from Phoenix. We can define internal filters in sdk.

But... Currently in GRPC we don't have filtering so the information will be returned to the client either way. We could change it.

package.json
diocas: Without the GRPC deps we cannot have the client working.

DeepDiver1975: But the GRPC deps are being sucked in via the sdk dependency - at least this is what I'm expecting.
Works that way with any module. Let me know if there is anything to help out with.

apps/files/src/components/PublicLinks/FilesDrop.vue
labkode: @DeepDiver1975 I think the idea of @diocas was to have two methods: one simple and minimal that doesn't accept anything: `list()` and another where you can filter out attributes `listExtended(attrs...)` to have a better developer experience if you don't want advanced use-cases.

DeepDiver1975: All I can say is that I started off with a simple list() and immediately did hit the limitation that one property was simply not given (afaik the favorities property).
So passing in the props one needs is the logical step.

I agree that these properties/attributes might not need to be named like webdav props - but then we need a mapping somewhere. Complexity grows ....

package.json
diocas: Ah sorry.. This was a previous attempt just for testing and I forgot to remove it. Again, this is just to kick start the discussion and is definitely not for merging. I have things half baked in many places.

DeepDiver1975: No problem at all - I'm just having a close eye :eyes: ;-)

apps/files/src/components/PublicLinks/FilesDrop.vue
diocas: I kept all the properties that were defined. I just moved them from phoenix to the sdk. And I think there is the correct place. 
From the current Phoenix code I only saw a place where there were more properties being requested than the default ones. So, as Hugo mentioned, I created a `listExtended` that adds these extra stuff internally. This was to be easier, as I didn't see in detail what this was returning, and we should give a more meaningful name to the method.

My point is that we shouldn't give the freedom to phoenix to ask whatever it wants. We should see what it needs and define the corresponding methods/interfaces in the sdk.
In GRPC we cannot request whatever to the same endpoint like in webdav, we need to construct the query and define which endpoint(s) to call first. For that, we need to have things properly defined.

In conclusion, my suggestion is that we should define what Phoenix needs first. If new - not though off - things appear, we add them to the specification and implement them in (both) sdk's. They will be just a method call away.

DeepDiver1975: > My point is that we shouldn't give the freedom to phoenix to ask whatever it wants. We should see what it needs and define the corresponding methods/interfaces in the sdk.
> In GRPC we cannot request whatever to the same endpoint like in webdav, we need to construct the query and define which endpoint(s) to call first. For that, we need to have things properly defined.

Well - afaik there is a map structure in the protobuf definition for any kind of meta data. This is similar to the props in webdav - as far as I understood this.

To me this is highly required to have this flexibility otherwise we cannot easily extend the frontend and display/manipulate additional metadata without rebuilding the full stack.

Looking at the extensibility of the ownCloud 10 frontend it is even possible that at runtime an app/extension is requiesting additional properties which will be queried right away when calling for the file list.

I hope this was part of the overall CS3 API design otherwise we have diverging requirements from my understanding.

diocas: Ok, got your point! Yes, we have the metadata field and we can get the extra stuff from there. We cannot forget that the webdav server is actually a converter GRPC-Webdav, so there's no information you get that I cannot retrieve from the web.
The difference is that, currently at least, GRPC doesn't filter. It returns all metadata in the file. But we could add a filtering mechanism.

DeepDiver1975: So a client would always get all meta data? This will only work easily if the metadata is directly stored with one file. But there are scenarios where such meta data can be computed or derived from higher elements in the tree or what so ever - this is at least how we implemented many features in ownCloud as of today.

Filtering is essential to not trigger computation for information which is not needed.

Nevertheless - we need to agree on property names and - as said above - define some mapping if necessary.

PR #2652, opened 2019-12-09
Hide quota on external storage

LukasHirt:
## Description
Hide quota in mounted folders.

## How Has This Been Tested?

- test environment: Manually
1. Navigate to a mounted folder
2. See no quota

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -47,7 +47,7 @@
                 &dash; {{ filesTotalSize | fileSize }}
               </template>
             </div>
-            <div v-if="!publicPage()" class="uk-visible@s uk-width-2-3 uk-width-1-2@xl uk-text-meta uk-flex" :class="{ 'uk-visible@xl' : _sidebarOpen  }">
+            <div v-if="quotaVisible" class="uk-visible@s uk-width-2-3 uk-width-1-2@xl uk-text-meta uk-flex" :class="{ 'uk-visible@xl' : _sidebarOpen  }">
               <span class="uk-margin-small-right">
                 <translate>Used space:</translate> {{ quota.used | fileSize }}
                 <template v-if="quota.definition !== 'default' && quota.definition !== 'none'">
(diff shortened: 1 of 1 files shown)


apps/files/src/components/AllFilesList.vue
```diff
@@ -158,11 +158,23 @@ export default {
...
     item () {
       return this.$route.params.item
+    },
+
+    quotaVisible () {
+      if (this.publicPage()) {
+        return false
+      }
+
+      if (this.configuration.hideQuotaOnExternalStorage) {
```
PVince81: I think we should **always** hide it on external storages as it never makes sense to show it then.

When talking about this, I thought we already had this logic in place before but it got broken ? Did you find out where the old logic was and what broke it ? Not understanding this makes us run the risk of having duplicate code and potentially dead code elsewhere

LukasHirt: I'd say it shouldn't be hidden always since it is not information tied to the current folder. Even though a user is in a mounted folder doesn't mean he shouldn't be able to see his quota. So unless users can operate only in mounted folders where it really doesn't make sense, I wouldn't suggest to hide it.

The bigger issue currently is the location of the quota. It seems too much related to the current list which could cause confusion. A better location would be e. g. in the bottom of the left menu if it would be visible all the time.

LukasHirt: I got confused with where we had the logic. It wasn't implemented for the display of quota but in the create/upload logic - https://github.com/owncloud/phoenix/blob/078a28ff9fb70b216b835e0c7210e8c31268d938/apps/files/src/components/FilesAppBar.vue#L253

PVince81: You're right that the location is currently confusing.

I'd still suggest to hide it always for mount points and we can sort out its location later on.
I don't think it's that useful to introduce yet another config switch that we might need to remove later.

If the left sidebar was always visible we could display it there for example.

LukasHirt: Ok - adjusted.

PR #2753, opened 2019-12-19
 allow to set the selenium host by env. variable when running on saucelabs

individual-it:
## Description
obey the `SELENIUM_HOST` env. variable even when running on saucelabs

## Motivation and Context
when trying to connect from a local development machine to saucelabs (e.g. through sc-proxy) we need to be able to set the `SELENIUM_HOST` manually

## How Has This Been Tested?
run IE tests on PR

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] revert commit that runs IE tests on PR

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -76,6 +76,9 @@ config = {
 			'browsers': [
 				'IE11',
 			],
+			'extraEnvironment': {
+				'SELENIUM_HOST': 'saucelabs'
+			},
 			'cronOnly': True,
 			'filterTags': '@smokeTest and not @skip and not @skipOnIE',
 		},
--- a/nightwatch.conf.js
+++ b/nightwatch.conf.js
@@ -70,7 +70,6 @@ module.exports = {
       }
     },
     saucelabs: {
-      selenium_host: 'saucelabs',
       desiredCapabilities: {
         tunnelIdentifier: SAUCELABS_TUNNEL_NAME,
         screenResolution: '1280x1024',
--- a/.drone.yml
+++ b/.drone.yml
@@ -4149,6 +4149,7 @@ steps:
       from_secret: sauce_access_key
     SAUCE_USERNAME:
       from_secret: sauce_username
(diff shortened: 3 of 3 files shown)


nightwatch.conf.js
```diff
@@ -70,7 +70,7 @@ module.exports = {
       }
     },
     saucelabs: {
-      selenium_host: 'saucelabs',
+      selenium_host: SELENIUM_HOST,
```
skshetry: This is already in the default, so, it's no longer needed here.
```suggestion
```

.drone.starlark
```diff
@@ -76,6 +76,9 @@ config = {
 			'browsers': [
 				'IE11',
 			],
+            'extraEnvironment': {
+                'SELENIUM_HOST': 'saucelabs'
+            },
```
skshetry: looks like there are spaces here. Please convert indentation to tabs.

.drone.starlark
```diff
@@ -76,7 +76,9 @@ config = {
 			'browsers': [
 				'IE11',
 			],
-			'cronOnly': True,
```
skshetry: Looks like `.drone.yml` is not generated for this.

individual-it:
@skshetry fixed the issues, please review again

nightwatch.conf.js
skshetry: I don't see any change for this file now in the diff. Isn't this PR all about this change?
To clarify, I simply suggested to remove the specific line, and after that, setting SELENIUM_HOST should work as it's on the default.

individual-it: :man_facepalming:

skshetry: Sorry, I could not be clearer. What i said is you no longer need that line. You can remove line 73, and it should work as it's already present in the default.

PR #2756, opened 2019-12-19
Show browser log

skshetry:
This PR adds a browser console logging in case the scenario fails. Currently, it does not fail the tests in case of console.error.

### DIFF
--- a/tests/acceptance/helpers/browserConsole.js
+++ b/tests/acceptance/helpers/browserConsole.js
@@ -1,0 +1,32 @@
+import { client } from 'nightwatch-api'
+
+function cleanupLogMessage (message) {
+  return message
+    .replace(/\\u003C/gi, '')
+    .replace(/\\n/g, '\n') // revive newlines
+}
+
+function formatLog (log) {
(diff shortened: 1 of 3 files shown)


PVince81:
seems the setup itself didn't work
image

@skshetry

tests/acceptance/helpers/browserConsole.js
```diff
@@ -0,0 +1,32 @@
...
+
+async function getAllLogs () {
+  let logs = []
+  await client.getLog('browser', entries => {
+    logs = entries
+  })
+  return logs
+}
+
+export async function getAllLogsWithDateTime (level = null) {
```
individual-it:
```suggestion
export async function getLogsOfLevel (level = null) {
```

skshetry: Here, `level` is just a filter to narrow the results down. If you do not provide any level, it will return unfiltered results. So, I don't prefer the above suggestion.

How about simple `getAllLogs()`? Or, any better suggestions?

PR #2821, opened 2020-01-07
[NOMERGE] Debug IE11 random test failures on CI

PVince81:
Add log entries to check if the Promise polyfill is always correctly loaded

For https://github.com/owncloud/phoenix/issues/2754

Based on https://github.com/owncloud/phoenix/pull/2804/

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -14,60 +14,73 @@ config = {
 	'yarnlint': True,
 
 	'acceptance': {
-		'webUI': {
+		'webUI-IE11': {
 			'suites': {
-				'webUIFavorites': 'Favorites',
-				'webUIFiles': 'Files',
-				'webUILogin': 'Login',
-				'webUINotifications': 'Notifications',
-				'webUIPrivateLinks': 'PrivateLinks',
-				'webUIRenameFiles': 'RenameFiles',
-				'webUIRenameFolders': 'RenameFolders',
-				'webUIRestrictSharing': 'RestrictSharing',
-				'webUISharingAutocompletion': 'SharingAutocompletion',
-				'webUISharingInternalGroups': 'SharingInternalGroups',
-				'webUISharingInternalUsers': 'SharingInternalUsers',
-				'webUISharingPermissionsUsers': 'SharingPermissionsUsers',
-				'webUISharingFilePermissionsGroups': 'SharingFilePermissionsGroups',
-				'webUISharingFolderPermissionsGroups': 'SharingFolderPermissionsGroups',
-				'webUIResharing': 'Resharing',
-				'webUISharingPublic': 'SharingPublic',
-				'webUITrashbin': 'Trashbin',
-				'webUIUpload': 'Upload',
-				'webUISharingAcceptShares': 'SharingAcceptShares',
-				'webUISharingPermissionMultipleUsers': 'SharingPermissionMultipleUsers',
-				'webUISharingNotifications': 'SharingNotifications',
-				'webUIAccount': 'Account'
+				'webUI[R,T,F]*': 'Rename-Trash-Files',
+				'webUISharing*': 'Sharing',
 			},
-		},
-		'webUIFederation': {
-			'suites': {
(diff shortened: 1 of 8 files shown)


PVince81:
@individual-it it seems that the log array is null:
image

maybe it's not possible to read log entries from IE11 ? :-/

PVince81:
seems that logs in IE 11 are not available at all: https://github.com/SeleniumHQ/selenium/issues/4802

so will need to use a different trick... possibly something that becomes visible in the screenshots, and already requires reenabling screenshots...

PVince81:
or: create a dummy element with the contents and retrieve the element's text with Selenium's usual selectors

PVince81:
I've used the latter trick: put the log text into a hidden element and then retrieve it with `elementIdText`. let's see...

PVince81:
Seems Selenium is unable to access that element, it's getting a 500 HTTP status:
```

√ Element <button[type=submit]> was not present after 753 milliseconds.
--

```
from https://drone.owncloud.com/owncloud/phoenix/7573/3/14

@individual-it any idea what's going on ?

PVince81:
context: since we cannot access IE11 logs directly, I've embedded the messages I wanted into an HTML DIV element with id "debugArray" and here I'm trying to access it after the login process to log its contents.

individual-it:
elementIdText needs the webElement (a magic number) not the CSS-id https://nightwatchjs.org/api/elementIdText.html
I think we are better off trying to use getValue() or getText()

PVince81:
hmm, the value is always null

PVince81:
Had to switch to `getAttribute` with `innerHTML` and at least with Chrome it shows something.
Let's see how it looks like in IE11...

PVince81:
Drone seems stuck, didn't start.

Rebased...

PVince81:
finally, some results !

In all tests the polyfill appears initialized properly.

However it seems that the test failure has morphed to something different than back when we started debugging:
```
Failed [strictEqual]: (displayed file name not as expected) - expected "testimage (2).jpg" but got: "[object Object]"
```

the problem used to be about `.finally()` not being called.

Maybe the above was introduced by recent changes and need fixing first.

PVince81:
I was not able to reproduce this locally. The "filename" attribute looks fine.

Might need some remote VM debugging on the Saucelab VM.
I don't understand why the attribute would contain an object here.

@individual-it can you help with this when you have time ?

individual-it:
The result object there wit IE11 is 
```
{ status: -1,
  value:
   { error:
      'An error occurred while running .getAttribute() command on <//span[span/text()=\'testimage\' and span/text()=".jpg"]/../../../../div[@data-is-visible="true"]//span[contains(@class, "file-row-name")]>:',
     message:
      'An error occurred while running .getAttribute() command on <//span[span/text()=\'testimage\' and span/text()=".jpg"]/../../../../div[@data-is-visible="true"]//span[contains(@class, "file-row-name")]>:',
     stack:
      'NoSuchElementError: An error occurred while running .getAttribute() command on <//span[span/text()=\'testimage\' and span/text()=".jpg"]/../../../../div[@data-is-visible="true"]//span[contains(@class, "file-row-name")]>:\n    at GetAttribute.noSuchElementError (/home/artur/www/owncloud-phoenix/node_modules/nightwatch/lib/element/command.js:370:12)\n    at Object.errorHandler (/home/artur/www/owncloud-phoenix/node_modules/nightwatch/lib/api/element-commands/_baseElementCommand.js:42:30)\n    at PeriodicPromise.runAction (/home/artur/www/owncloud-phoenix/node_modules/nightwatch/lib/util/periodic-promise.js:60:29)\n    at process._tickCallback (internal/process/next_tick.js:68:7)' } }

individual-it:
it seems to be a problem with the scrolling, when I try a file that is in view e.g. `data.tar.gz` it works fine :thinking:

LukasHirt:
I pushed a fix for the selector. For now, I'll keep it as a separate commit.

PVince81:
obsolete and unsupported

PR #2875, opened 2020-01-14
Update dependencies 20200114

DeepDiver1975:


.drone.star
```diff
@@ -603,7 +603,7 @@ def installFederatedServer(version, db, dbSuffix = '-federated'):
 def installNPM():
 	return [{
 		'name': 'npm-install',
-		'image': 'owncloudci/php:7.1',
+		'image': 'owncloudci/nodejs:10',
```
LukasHirt: Why 10? Newer ones are breaking some builds?

DeepDiver1975: Up to now we did execute our code with node 8 - node 10 is the next logical step ....

PR #2877, opened 2020-01-14
Share indicator for direct and indirect shares in file list

PVince81:
## Steps to view:

1. Setup a new OC 10.4 server
1. Run the script from https://github.com/owncloud/product/pull/40
1. Open Phoenix
1. Login as "bob" with password "test"
1. Browse the folders and enjoy the share indicators

image

Ref: https://github.com/owncloud/owncloud-design-system/pull/607
For story: https://github.com/owncloud/enterprise/issues/3701

Fixes https://github.com/owncloud/phoenix/issues/2060
Fixes https://github.com/owncloud/phoenix/issues/2894

@pmaier1 FYI

### DIFF
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -9,6 +9,7 @@
               <span class="oc-visually-hidden" v-text="favoritesHeaderText" />
               <oc-star id="files-table-header-star" aria-hidden="true" class="uk-display-block uk-disabled" />
             </div>
+            <div></div>
             <div class="uk-text-truncate uk-text-meta uk-width-expand" v-translate>Name</div>
             <div :class="{ 'uk-visible@s' : !_sidebarOpen, 'uk-hidden'  : _sidebarOpen }" class="uk-text-meta uk-width-small" v-translate>Size</div>
             <div type="head" :class="{ 'uk-visible@s' : !_sidebarOpen, 'uk-hidden'  : _sidebarOpen }" class="uk-text-nowrap uk-text-meta uk-width-small" v-translate>Updated</div>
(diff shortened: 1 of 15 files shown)


PVince81:
- [ ] TODO: also add fakes for link panel:
    - parent resource owner
    - inherited OLC link shares

PVince81:
- [ ] TODO: extend test samples to have subdir for "alice2bob2charlie+link-sub-bob2dave+link (IUC, OLC, OUC, OUP, OLP)"
   - this would illustrate having both "charlie" and "dave" in the inherited list, with different via entries

PVince81:
I've already added crude via links, here's a good example with both inherited resource owner and inherited share from parent resource:
image

PVince81:
- [ ] remove separation of users and groups

PVince81:
I'd like to reuse this PR and fill it with actual data once available.

PVince81:
Added share indicators for indirect **outgoing** shares.

TODOs:

- [ ] python script issue: IUP+OLP is missing link share!
- [ ] need to query webdav props on parents as well for oc:share-types for inheritent incoming shares
- [ ] share indicator event after creating/deleting link for direct share ? might need to reload just the row ?
- [ ] enable caching: only request if not in sharesTree yet

PVince81:
In OC 10 we don't load incoming shares for all parents, but only the Webdav share-types. While this might be enough to display indicators, it won't be enough for showing the inherited resource/share owners. So here in Phoenix let's go with an OCS call for incoming shares right away.

PVince81:
Added requests for incoming shares and now the indirect indicators work for both inheritance scenarios.
Added caching for the shares tree with pruning algorithm when switching folders.

Next up is to clean up stuff unrelated to share indicators to bring this into mergable state.

PVince81:
Squashed into a single commit.

The "dummy" parts for later tasks were moved to the branch "share-indirect-info" for continuing later.

PVince81:
- [ ] BUG: indicator for incoming user share on parent (IUP) should be grey. currently is absent

PVince81:
- [ ] verify that share indicators and its logic does NOT run in public link page

PVince81:
- [ ] share indicators into a separate vue component ?

PVince81:
Adjusted all tests, I expect all to pass except maybe the federation ones which I couldn't run locally.

PVince81:
added a few more fixes for federated tests (squashed)

PVince81:
Fixed federation tests where a copy-paste issue made it create the share recipient users on the wrong instance.

PVince81:
This doesn't make sense... two failures with login:
```
  ✖ And user "user2" has logged in using the webUI # tests/acceptance/stepDefinitions/loginContext.js:60
```

one thing I observed earlier is that if you attempt to log in as an already logged in user, the same error happens. However, in this one case the currently logged in user should be "user1".

PVince81:
weird, I looked at VNC and see that it's stuck on the authorize page showing "User One" while trying to login as user2. I thought we had logic in place to automatically switch the user there ?

PVince81:
I managed to fix the tests by using the "relogin" step to force clearing the session.
And also fixed a random failure due to missing delay in the share autocomplete input.

I'm expecting everything to pass now.

PVince81:
All passed, @kulmann please review

PR #2890, opened 2020-01-16
make Mediaviewer work in OCIS+Reva

madsi1m:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/media-viewer/src/Mediaviewer.vue
+++ b/apps/media-viewer/src/Mediaviewer.vue
@@ -207,11 +207,11 @@ export default {
         a: 1
       })
 
+      this.activeMediaFile.path = this.activeMediaFile.path.replace(/^\//, '');
       if (this.publicPage()) {
         const path = [
           '..',
-          'dav',
-          'public-files',
+          'webdav',
           this.activeMediaFile.path
         ].join('/')
 
@@ -219,8 +219,7 @@ export default {
       }
       const path = [
         '..',
-        'dav',
-        'files',
+        'webdav',
         this.$store.getters.user.id,
         this.activeMediaFile.path
       ].join('/')


apps/media-viewer/src/Mediaviewer.vue
```diff
@@ -207,20 +207,19 @@ export default {
         a: 1
       })
 
+      this.activeMediaFile.path = this.activeMediaFile.path.replace(/^\//, '');
       if (this.publicPage()) {
         const path = [
           '..',
-          'dav',
-          'public-files',
```
DeepDiver1975: This will break on public shares

apps/media-viewer/src/Mediaviewer.vue
```diff
@@ -207,20 +207,19 @@ export default {
...
+          'webdav',
           this.activeMediaFile.path
         ].join('/')
 
         return this.$client.files.getFileUrl(path) + '?' + query
       }
       const path = [
         '..',
-        'dav',
```
DeepDiver1975: Is the dav endpoint not available on reva/ocis?

apps/media-viewer/src/Mediaviewer.vue
madsi1m: a closer look implies that for some reason `this.publicPage()` is true for non public files all the time in reva/ocis

apps/media-viewer/src/Mediaviewer.vue
madsi1m: not yet (soon?)

apps/media-viewer/src/Mediaviewer.vue
madsi1m: Is something missing in ocs @butonic ?

PVince81:
I seem to remember that the media viewer is now working, now that the ocis-thumbnailer was developed/fixed. @C0rby can you confirm ?

in any case we need to fix the backend (if not already) to also generate thumbnails on those dav endpoints

C0rby:
> I seem to remember that the media viewer is now working, now that the ocis-thumbnailer was developed/fixed. @C0rby can you confirm ?

Yes, but we need to update the `ocis-thumbnails` dependency in `ocis` because there was a bug which prevented the generation of thumbnails.

Also because the thumbnail service now uses the predefined sizes, the mediaviewer only gets images up until the max predefined size. So we may have to add some more like "1920x1080" for FHD images.

C0rby:
The mediaviewer is working for me locally.
Is this pr still relevant?

butonic:
fixed using pre signed urls

PR #2912, opened 2020-01-23
Remove tests for Internet Explorer

### LINKED ISSUE #2908: delete IE11 magic from test-code
as there is no official IE11 support and tests are not run in internet explorer, we should remove all the IE related hacks from the test-code, deleted code is debugged code!

jasson99:
## Description
Remove all the tests and functions which were written for running the tests in internet explorer

## Related Issue

- Fixes #2908 

## Motivation and Context
Since there is no official IE11 support and tests are not run in internet explorer, we should remove all the IE related hacks from the test-code.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


tests/acceptance/customCommands/setValueBySingleKeys.js
```diff
@@ -1,9 +1,4 @@
 /**
```
dpakach: This file can be deleted as there is no need to set value by single keys now.

jasson99: it is needed for running federation tests even in chrome browser

individual-it: then please write that in the comment

jasson99: fixed

dpakach: In that case please fix this function while you're at it. For the federated sharing problem we don't need to send value by single keys. We just need to wait a short time while entering last few characters.
change line 15 to
``` js
 const charPromise = this.setValue(selector, chars)
```
change line 21 to
``` js
 return Promise.all([charPromise, ...charEndPromise])
```
That should fix the problem IMO but may need some tweaking

PR #2928, opened 2020-01-28
Move sharing indicators into own component and add extension point

### LINKED ISSUE #2895: Extension point sharing indicators for complex theming
Based on https://github.com/owncloud/phoenix/issues/2060.

- [ ] add extension point for populating the share indicator section in a file row
- [ ] add flag to be able to disable the default indicators

LukasHirt:
## Description
Move sharing indicators into smaller components, enable disabling of default indicators and add extension point for new indicators.

## Related Issue

- Fixes #2895

## How Has This Been Tested?

- test environment: Manually
1. Disable default indicators inside of theme
2. Create new indicators (example https://github.com/LukasHirt/phoenix-status-indicators) or add a custom indicators into already existing extension

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] Add changelog item
- [x] Implement click handlers for default indicators

### DIFF
--- a/apps/files/docs/extensionpoints.adoc
+++ b/apps/files/docs/extensionpoints.adoc
@@ -1,4 +1,4 @@
-## Extensionpoints
+## Extension points
 
 #### file-details-panel
 
@@ -17,3 +17,76 @@
 
 OC.$extend.request('files', 'file-details-panel', payload).then( response => {});
 ----
+
+### Files list status indicators
+
+An extension can extend status indicators in files list with custom component, properties and methods.
+Extending this list can be done by creating a new object inside of appInfo called `filesListIndicators`. This object can contain keys `name` and `component`.
+
+#### Props
+
+Please see known issues before using props in custom indicators.
+
+- `item`: resource to which are indicators assigned
+- `parentPath`: parent folder path of the item
+
+#### Example:
+
+##### App entry point
+[source,js]
+----
+import ExampleIndicators from './components/ExampleIndicators.vue'
+
+const appInfo = {
+  name: 'ExampleIndicators',
+  id: 'ExampleIndicators',
+  icon: 'info',
+  isFileEditor: false,
+  extensions: [],
+  filesListIndicators: [{
+    name: 'CustomIndicators',
+    component: ExampleIndicators
+  }]
+}
+----
+
+##### ExampleIndicators component
+[source,vue]
+----
+<template>
+  <oc-button
+    v-if="isFolder"
+    class="file-row-share-indicator uk-text-middle"
+    aria-label="Custom indicator"
+    @click="triggerAlert"
+    variation="raw"
+  >
+    <oc-icon
+      name="folder"
+      class="uk-text-middle"
+      size="small"
+      variation="active"
+    />
+  </oc-button>
+</template>
+
+<script>
+  export default {
+    name: 'ExampleIndicators',
+    computed: {
+      isFolder () {
+        return this.$parent.item.type === 'folder'
+      }
+    },
+    methods: {
+      triggerAlert () {
+        alert('Hello world')
+      }
+    }
+  }
+</script>
+----
+
+#### Known issues:
+
+Passing resource as a prop into the dynamic component causes Vuex mutation error. To avoid this, access resource direcly via parent component - `this.$parent.item`
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -29,14 +29,7 @@
                 class="uk-margin-small-left"
               />
             </div>
-            <div>
-              <oc-button v-if="$_isUserShare(item)" class="file-row-share-indicator uk-text-middle" :aria-label="$_shareUserIconLabel(item)" @click="$_openSideBar(item, 'files-sharing')" variation="raw">
-                <oc-icon name="group" class="uk-text-middle" size="small" :variation="$_shareUserIconVariation(item)"/>
-              </oc-button>
-              <oc-button v-if="$_isLinkShare(item)" class="file-row-share-indicator uk-text-middle" :aria-label="$_shareLinkIconLabel(item)" @click="$_openSideBar(item, 'file-link')" variation="raw">
-                <oc-icon name="link" class="uk-text-middle" size="small" :variation="$_shareLinkIconVariation(item)"/>
-              </oc-button>
-            </div>
+            <StatusIndicators :item="item" :parentPath="currentFolder.path" @click="$_openSideBar" />
             <div class="uk-text-meta uk-text-nowrap uk-width-small" :class="{ 'uk-visible@s' : !_sidebarOpen, 'uk-hidden'  : _sidebarOpen }">
               {{ item.size | fileSize }}
             </div>
@@ -75,18 +68,16 @@
 <script>
 import FileList from './FileList.vue'
 import { mapGetters, mapActions, mapState } from 'vuex'
-import { shareTypes } from '../helpers/shareTypes'
-import { getParentPaths } from '../helpers/path'
 
 import Mixins from '../mixins'
 import FileActions from '../fileactions'
-import intersection from 'lodash/intersection'
 
-const userShareTypes = [shareTypes.user, shareTypes.group, shareTypes.guest, shareTypes.remote]
+const StatusIndicators = () => import('./FilesLists/StatusIndicators/StatusIndicators.vue')
 
 export default {
   components: {
-    FileList
+    FileList,
+    StatusIndicators
   },
   mixins: [
     Mixins,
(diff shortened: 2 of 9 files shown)


apps/files/src/components/FilesLists/StatusIndicators/StatusIndicators.vue
```diff
@@ -0,0 +1,75 @@
...
+
+  created () {
+    this.loadCustomIndicators()
+  },
+
+  methods: {
+    loadCustomIndicators () {
+      const path = `/themes/${this.configuration.currentTheme}/status-indicators.js`
+
+      requirejs([path], this.injectCustomIndicators, this.onRequireError)
```
LukasHirt: Thinking now about the 404 that this will cause in case the file does not exist (which will be most of the time IMHO). I don't really want to do extra call which would first check for the file and then load it in case it exists. So what about another key in theme json which would define if the load should happen or not? If it would be true we can then expect that the file is there or it then makes sense to throw a 404. But it's a small extra effort for administrators in the end. Seems to me like a way to go though. What do you think @PVince81 @kulmann @DeepDiver1975 ?

LukasHirt: In the beginning, I was going to approach it as "just another extension" but then it wasn't really feeling right separating the theme-related stuff into two different places. Talking about simple or complex I think it should live at some united place not to make it complicated to look around where it actually lives. And looking at the complex theming requirement, most of it will live in the json theme file only as a value anyway.

Otherwise, I would simply call it an extension and wouldn't mix it into the theme. But then I'd say this could be done overall and we could drop simple/complex theming and just have theming and then extensions.

LukasHirt: Result of discussion with @PVince81 - we won't consider this a theming but it will be an extension/customisation. We'll reuse logic for loading extensions. (ref https://github.com/owncloud/phoenix/issues/2404#issuecomment-579673241)

LukasHirt:
Tests are still failing, will investigate more and there is one error in console regarding vuex mutations which I'm digging deeper in now. I think you can do a "first round" of the review though @PVince81

LukasHirt:
Seems that the error is coming from within the extension when the item prop is used. Not yet sure why. Continuing 🔎

apps/files/src/components/AllFilesList.vue
```diff
@@ -29,14 +29,7 @@
...
             </div>
-            <div>
-              <oc-button v-if="$_isUserShare(item)" class="file-row-share-indicator uk-text-middle" :aria-label="$_shareUserIconLabel(item)" @click="$_openSideBar(item, 'files-sharing')" variation="raw">
-                <oc-icon name="group" class="uk-text-middle" size="small" :variation="$_shareUserIconVariation(item)"/>
-              </oc-button>
-              <oc-button v-if="$_isLinkShare(item)" class="file-row-share-indicator uk-text-middle" :aria-label="$_shareLinkIconLabel(item)" @click="$_openSideBar(item, 'file-link')" variation="raw">
-                <oc-icon name="link" class="uk-text-middle" size="small" :variation="$_shareLinkIconVariation(item)"/>
-              </oc-button>
-            </div>
+            <StatusIndicators :item="item" :currentFolderPath="currentFolder.path" @click="$_openSideBar" />
```
PVince81: Shouldn't it rather be `parentPath` since it's the parent of the item ?
For context, think about flat lists where each item might have a different parent folder (like favorites), so the name "current folder" is misleading

Should the property be called `resource` as this is either a file or folder ?

apps/files/src/components/FilesLists/StatusIndicators/DefaultIndicators.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <div>
+    <oc-button
+      v-for="(indicator, index) in indicators"
+      :key="index"
+      class="file-row-share-indicator uk-text-middle"
+      :class="{ 'uk-margin-xsmall-left' : index > 0 }"
```
PVince81: did you change / adjust styling here ?

I didn't know that it's possible to have an attribute twice this way

apps/files/src/components/FilesLists/StatusIndicators/DefaultIndicators.vue
```diff
@@ -0,0 +1,142 @@
...
+          icon: 'link',
+          status: this.shareLinkIconVariation(this.item),
+          handler: this.indicatorHandler
+        })
+      }
+
+      return indicators
+    },
+
+    shareTypesIndirect () {
```
PVince81: why not keep all these methods private with `$_` prefix ?

apps/files/src/components/FilesLists/StatusIndicators/StatusIndicators.vue
```diff
@@ -0,0 +1,64 @@
...
+      type: String,
+      required: true
+    }
+  },
+
+  computed: {
+    ...mapGetters(['configuration', 'customFilesListIndicators']),
+
+    displayDefaultIndicators () {
+      return !this.configuration.theme.filesList.hideDefaultStatusIndicators
```
PVince81: beware in case "theme" and "theme.filesList" are not defined

maybe we need some utility method that we can query with a string ? or wasn't there an ES6 operator to accomodate for such situations ?

changelog/unreleased/2895
```diff
@@ -0,0 +1,8 @@
+Enhancement: Add status indicator extension point
```
PVince81: => "Add file list status indicator extension point" as this applies only to file list entries
(for those who are too lazy to read the description)

apps/files/src/components/AllFilesList.vue
LukasHirt: Right. Took it from the state originally and haven't realised the other use cases. Will change it. For the resource though I usually use `item` which seems to me as a good enough alternative. I can change that though to resource. Don't really care which one it will be.

apps/files/src/components/FilesLists/StatusIndicators/DefaultIndicators.vue
LukasHirt: Added just the margin class.

Should work as long as one of them is `v-bind`. Not sure though if this is only for the class. Haven't yet tried it with anything else.

apps/files/src/components/FilesLists/StatusIndicators/DefaultIndicators.vue
LukasHirt: Haven't seen a reason to keep them as private since it's encapsulated from anything else.

apps/files/src/components/FilesLists/StatusIndicators/StatusIndicators.vue
LukasHirt: This is coming from the store with a default value so it would have to be either manually deleted or it would be a bug. I haven't put there a check for it exactly because of the bug case so we would then see an error if something is wrong.

LukasHirt:
> Can you also extend "apps/files/docs/extensionpoints.adoc" and put a quick example code there to illustrate how a custom indicator implementation looks like ?

I completely missed that we have such a file until now 🙈

apps/files/docs/extensionpoints.adoc
```diff
@@ -17,3 +17,29 @@
...
+----
+const appInfo = {
+  name: 'StatusIndicators',
+  id: 'status-indicators',
+  icon: 'info',
+  isFileEditor: false,
+  extensions: [],
+  filesListIndicators: [{
+    name: 'indicators',
+    component: Indicators
```
PVince81: any hint about what properties the "Indicators" vue is supposed to take as input ? what events can/should it trigger if any ?

PVince81: can you also rename it to "ExampleIndicators" to make it clearer which part of this code is referencing actual Phoenix code and what part is custom ?

changelog/unreleased/2895-1
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add files list status indicator extension point
+
+We've added the ability for the extension to inject custom status indicator into files list.
```
PVince81: "indicator" -> "indicators"

apps/files/docs/extensionpoints.adoc
LukasHirt: Renamed and added props into docs. Haven't added events though since there are none for custom indicators. If we want to introduce some we need requirements for them. At the moment, every method needs to be implemented and taken care of inside of the custom indicators components.

LukasHirt:
@PVince81 Adjusted.

apps/files/docs/extensionpoints.adoc
```diff
@@ -17,3 +17,74 @@
...
+  extensions: [],
+  filesListIndicators: [{
+    name: 'CustomIndicators',
+    component: Indicators
+  }]
+}
+----
+
+##### Indicators component
+[source,vue]
```
PVince81: please add a mention that this is in fact the "ExampleIndicators.vue" file, for clarity

apps/files/docs/extensionpoints.adoc
```diff
@@ -17,3 +17,74 @@
...
+      class="uk-text-middle"
+      size="small"
+      variation="active"
+    />
+  </oc-button>
+</template>
+
+<script>
+  export default {
+    name: 'Indicators',
```
PVince81: isn't that the "ExampleIndicators" component ?

LukasHirt: Ah, you meant renaming the component? It still is called just Indicators. The name inside of appinfo is an arbitrary string and not tied to the name of the component.

PVince81: in that case please do rename it to something else to avoid the confusion with the same component(s) of Phoenix

my requests here are mostly about removing any possible confusion for someone reading the code the first time

LukasHirt: Renamed

apps/files/docs/extensionpoints.adoc
LukasHirt: Renamed.

LukasHirt:
@PVince81 Component and its references renamed.

PR #2952, opened 2020-01-30
Fix add/remove animations for collaborators and links

kulmann:
# Detailed Description
Collaborator and link forms (adding/editing) unnecessarily trigger mutations on loading states (`SHARES_LOADING` and `LINKS_LOADING`). This causes animations for adding and removing collaborators/links to fail in the respective panels. While fixing this, some old code has been cleaned up as well.
- [x] removed triggers for busy-state for collaborator and link saving/creating. using local variable in respective form components now. This fixes the animations in the sidebar panels when adding or removing items, as the loader is not triggered anymore. As saving can be represented within the respective form, there is no need to push this state into the store.
- [x] removed unused code in collaborator editing form
- [x] changed buttons in collaborator forms to represent a saving-state when still in saving mode
- [x] re-introduced a data change detection for collaborators.
- [x] switched promise handling when adding collaborators to using p-queue

## Related Issue

- Fixes https://github.com/owncloud/phoenix/issues/2937

## Motivation and Context
Fix animations and clean up old code.

## How Has This Ben Tested?

- local in Chrome and Firefox
- [ ] selenium

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/Collaborators/Collaborator.vue
+++ b/apps/files/src/components/Collaborators/Collaborator.vue
@@ -11,9 +11,10 @@
     </oc-table-row>
     <oc-table-row class="files-collaborators-collaborator-table-row-info">
       <oc-table-cell shrink>
-        <oc-button v-if="modifiable" :ariaLabel="$gettext('Delete share')" @click="$emit('onDelete', collaborator)" variation="raw" class="files-collaborators-collaborator-delete">
+        <oc-button v-if="$_deleteButtonVisible" :ariaLabel="$gettext('Delete share')" @click="$_removeShare" variation="raw" class="files-collaborators-collaborator-delete">
           <oc-icon name="close" />
         </oc-button>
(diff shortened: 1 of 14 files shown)


PVince81:
Code looks good.

However somehow now when deleting a collaborator or link there is a small 1-2 second delay where nothing happens until the animation begins. I think we do need some kind of spinner there, ideally also something that locks the row so one cannot click twice on delete.

kulmann:
@PVince81 The removal on a public link or collaborator now shows a loading spinner where the delete button was before, as soon as delete is clicked. The edit button is removed from the ui at the same time. Another thing I noticed is, that the first add and last removal didn't trigger the animation. Fixed that as well. Please re-review. :-)

PVince81:
As discussed, let's align the spinner sizes with the icon sizes in ODS

PVince81:
Just tested, looks fine as you said.

I've raised https://github.com/owncloud/owncloud-design-system/issues/631 to align spinner sizes with icon sizes.

PR #2961, opened 2020-02-04
We shall only call the sso end session endpoint in case we still have…

DeepDiver1975:
… the id_token. In some sceanrios the id_token is no longer known (expiry)

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/2961
+++ b/changelog/unreleased/2961
@@ -1,0 +1,8 @@
+Bugfix: Fix logout when no tokens are known anymore
+
+Single Log Out requires the id_token and in cases where this token is no
+longer known calling the SLO endpoint will result in an error.
+
+This has been fixed.
+
+https://github.com/owncloud/phoenix/pull/2961
--- a/src/services/auth.js
+++ b/src/services/auth.js
@@ -78,6 +78,17 @@ export function initVueAuthenticate (config) {
         }
         return null
       },
+      getStoredUserObject () {
+        const storageString = sessionStorage.getItem('oc_oAuth' + mgr._userStoreKey)
+        if (storageString) {
+          const user = User.fromStorageString(storageString)
+          if (user) {
+            mgr.events.load(user, false)
+            return user
+          }
+        }
+        return null
+      },
       isAuthenticated () {
         return this.getToken() !== null
       },
--- a/src/store/user.js
+++ b/src/store/user.js
@@ -32,12 +32,18 @@ const actions = {
       // force redirect to login page after logout
       router.push({ name: 'login' })
     }
-    vueAuthInstance.logout()
-      .then(logoutFinalizer)
-      .catch(error => {
-        console.error(error)
-        logoutFinalizer()
-      })
+    // TODO: only call logout if we still have the id token
+    const u = vueAuthInstance.getStoredUserObject()
+    if (u && u.id_token) {
+      vueAuthInstance.logout()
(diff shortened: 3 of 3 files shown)


DeepDiver1975:
I guess the test scenarios need to be adjusted ....

LukasHirt:
The problem with tests is that logout leads us to "Access denied" page. Not sure if this is happening also in a setup with oidc provider or only with oauth2.

LukasHirt:
The access denied page is appearing because of this line https://github.com/owncloud/phoenix/blob/f4eeac80a042b1d6ea4f72f004627e7631bc9476/src/store/user.js#L110 Would it be fine using here "Login" page instead or removing completely @DeepDiver1975 ? Can the user be unloaded because of some error? Or is it only when doing a logout? I guess also expiration? But even then I'd still expect to see login screen instead if I'd be the user.

kulmann:
> The access denied page is appearing because of this line
> 
> https://github.com/owncloud/phoenix/blob/f4eeac80a042b1d6ea4f72f004627e7631bc9476/src/store/user.js#L110
> 
> Would it be fine using here "Login" page instead or removing completely @DeepDiver1975 ? Can the user be unloaded because of some error? Or is it only when doing a logout? I guess also expiration? But even then I'd still expect to see login screen instead as a user.

As a bonus you would come back to the route where you were before the logout (also in case of expiration), right? As of now you come back to the access denied route when you re-login after expiration.

DeepDiver1975:
> Would it be fine using here "Login" page instead or removing completely @DeepDiver1975 ? Can the user be unloaded because of some error? Or is it only when doing a logout? I guess also expiration? But even then I'd still expect to see login screen instead as a user.

in the very first implementation we always redirected to the login screen - this was changed upon customer request.

I find it rather complicated to differentiate when to display which page .... :see_no_evil:

DeepDiver1975:
> Changelog pls

added

LukasHirt:
Hmm, looking at https://drone.owncloud.com/owncloud/phoenix/9041 the ocis tests seems to perform better than here. Maybe it is after all related somehow to this PR?

DeepDiver1975:
>  Maybe it is after all related somehow to this PR?

seems to be - who has enough understanding on having a look? I lost track with ocis these days :see_no_evil:

LukasHirt:
> who has enough understanding on having a look?

@dpakach Would you have time to take a look at the ocis tests here?

dpakach:
> @dpakach Would you have time to take a look at the ocis tests here?

@LukasHirt I will try to run them locally and see what's going on

dpakach:
All tests are failing on login so seems like something has changed in login flow

dpakach:
Sorry, l am unable to reproduce this locally. I just rebased just to see what happens.

LukasHirt:
> All tests are failing on login so seems like something has changed in login flow

The changes are actually in the opposite - logout.

> Sorry, l am unable to reproduce this locally. I just rebased just to see what happens.

It's green now, I'll try to run them one more time and if they'll still be green I'll merge it :wink: Thanks a lot for taking a look!

PR #2975, opened 2020-02-05
Add empty folder indicator

PVince81:
## Description
Add empty folder indicator

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/1910

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

image

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:
- [x] improve design, the colors don't match. Bigger size would be nice.
- [x] also process the other lists like "Favorites", "Shared with me", etc
- [x] changelog

### DIFF
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -37,6 +37,16 @@
               {{ formDateFromNow(item.mdate) }}
             </div>
           </template>
+          <template #noContentMessage>
+            <no-content-message v-if="!$_isFavoritesList" icon="folder">
+              <template #message><span v-translate>There are no resources in this folder.</span></template>
+              <template #callToAction><span v-translate>Drag files and folders here or use the "+ New" button to upload.</span></template>
+            </no-content-message>
+            <no-content-message v-else icon="star">
+              <template #message><span v-translate>There are no resources marked as favorite.</span></template>
+              <template #callToAction><span v-translate>You can mark some by clicking on the star icon in the file list.</span></template>
+            </no-content-message>
+          </template>
           <template #footer>
             <div v-if="activeFilesCount.folders > 0 || activeFilesCount.files > 0" class="uk-text-nowrap uk-text-meta">
               <span id="files-list-count-folders" v-text="activeFilesCount.folders" />
@@ -67,6 +77,7 @@
 </template>
 <script>
 import FileList from './FileList.vue'
+import NoContentMessage from './NoContentMessage.vue'
 import { mapGetters, mapActions, mapState } from 'vuex'
 
 import Mixins from '../mixins'
@@ -78,7 +89,8 @@ export default {
   name: 'AllFilesList',
   components: {
     FileList,
-    StatusIndicators
+    StatusIndicators,
+    NoContentMessage
   },
   mixins: [
     Mixins,
@@ -97,6 +109,10 @@ export default {
 
     item () {
       return this.$route.params.item
+    },
+
+    $_isFavoritesList () {
+      return (this.$route.name === 'files-favorites')
     },
 
     quotaVisible () {
@@ -174,7 +190,7 @@ export default {
       })
     },
     $_ocFileName (item) {
-      if (this.$route.name === 'files-favorites') {
+      if (this.$_isFavoritesList) {
         const pathSplit = item.path.substr(1).split('/')
         if (pathSplit.length === 2) return `${pathSplit[pathSplit.length - 2]}/${item.basename}`
         if (pathSplit.length > 2) return `…/${pathSplit[pathSplit.length - 2]}/${item.basename}`
--- a/apps/files/src/components/Collaborators/SharedFilesList.vue
+++ b/apps/files/src/components/Collaborators/SharedFilesList.vue
@@ -3,7 +3,7 @@
     :isActionEnabled="isActionEnabled">
     <template #headerColumns>
       <div class="uk-text-truncate uk-text-meta uk-width-expand" v-translate>Name</div>
-      <div class="uk-text-nowrap uk-text-meta uk-width-small" v-text="sharedCellTitle" />
+      <div class="uk-text-nowrap uk-text-meta uk-width-small" v-text="$_sharedCellTitle" />
       <div
         v-if="$route.name === 'files-shared-with-me'"
         shrink
@@ -45,6 +45,14 @@
       </div>
       <div class="uk-text-meta uk-text-nowrap uk-width-small" v-text="formDateFromNow(item.shareTime)" />
     </template>
+    <template #noContentMessage>
+      <no-content-message icon="share">
+        <template #message>
+          <span v-if="$_isSharedWithMe" v-translate>You are currently not collaborating on other people's resources.</span>
+          <span v-else v-translate>You are currently not collaborating on any of your resources with other people.</span>
+        </template>
+      </no-content-message>
+    </template>
   </file-list>
 </template>
 
@@ -53,11 +61,13 @@ import { mapGetters, mapActions } from 'vuex'
 import Mixins from '../../mixins'
 import FileActions from '../../fileactions'
 import FileList from '../FileList.vue'
+import NoContentMessage from '../NoContentMessage.vue'
 
 export default {
   name: 'SharedFilesList',
   components: {
-    FileList
+    FileList,
+    NoContentMessage
   },
   mixins: [
     Mixins,
@@ -76,12 +86,16 @@ export default {
   computed: {
     ...mapGetters('Files', ['loadingFolder']),
 
-    sharedCellTitle () {
-      if (this.$route.name === 'files-shared-with-me') {
-        return this.$gettext('Shared from')
+    $_isSharedWithMe () {
+      return (this.$route.name === 'files-shared-with-me')
+    },
+
+    $_sharedCellTitle () {
+      if (this.$_isSharedWithMe) {
+        return this.$gettext('Owner')
       }
 
-      return this.$gettext('Shared with')
+      return this.$gettext('Collaborators')
     }
   },
   watch: {
--- a/apps/files/src/components/FileList.vue
+++ b/apps/files/src/components/FileList.vue
@@ -2,7 +2,7 @@
   <!-- TODO: Take care of outside click overall and not just in files list -->
   <div :id="id" class="uk-height-1-1 uk-position-relative" @click="hideRowActionsDropdown">
     <div class="uk-flex uk-flex-column uk-height-1-1">
-      <oc-grid gutter="small" flex id="files-table-header" class="uk-padding-small">
+      <oc-grid gutter="small" flex id="files-table-header" class="uk-padding-small" v-if="fileData.length > 0" key="files-list-results-existence">
         <div>
           <oc-checkbox
             class="uk-margin-small-left"
@@ -65,6 +65,9 @@
             </oc-grid>
           </div>
         </RecycleScroller>
+        <div v-else class="uk-position-center files-list-no-content-message" key="files-list-results-absence">
+          <slot name="noContentMessage" />
+        </div>
       </div>
       <oc-grid gutter="large" class="uk-width-1-1 uk-padding-small" v-if="!loading">
         <slot name="footer" />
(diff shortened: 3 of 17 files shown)


PVince81:
I've added acceptance tests and also tweaked the code that retrieves the list to avoid another potentially waiting code when waiting is not needed => faster results

PVince81:
@individual-it I've pinged you as the tests part is less obvious. I've twisted them a bit to include the "empty message" detection inside the test step that checks if the resource list is empty.

The reason for that is that I wanted to avoid UI-specific language in the Gherkin step like "a message must appear" or something, so I thought having it implied but still checked through assertions would be the way to go. Let me know what you think.

LukasHirt:
> the colors don't match

If you leave out the variation in the icon it will get "system" (default) variation which has the same colour as meta text. Though I'd see this as a workaround only. But until we bring real oC colours into ODS I'd say it's good enough.

apps/files/src/components/AllFilesList.vue
```diff
@@ -37,6 +37,9 @@
               {{ formDateFromNow(item.mdate) }}
             </div>
           </template>
+          <template #emptyMessage>
+            <div class="uk-text-center"><oc-icon :name="$_emptyFolderIcon" type="div" size="large" variation="passive" /><div class="uk-text-muted" v-text="$_emptyFolderText" /></div>
```
LukasHirt: Separate this into own lines pls, it's too long.

apps/files/src/components/Collaborators/SharedFilesList.vue
```diff
@@ -45,6 +45,9 @@
       </div>
       <div class="uk-text-meta uk-text-nowrap uk-width-small" v-text="formDateFromNow(item.shareTime)" />
     </template>
+    <template #emptyMessage>
+      <div class="uk-text-center"><oc-icon name="share" type="div" size="large" variation="passive" /><div class="uk-text-muted" v-text="$_emptyFolderText" /></div>
```
LukasHirt: Separate this into own lines pls, it's too long.

apps/files/src/components/Trashbin.vue
```diff
@@ -30,6 +30,9 @@
             {{ formDateFromNow(item.deleteTimestamp) }}
           </div>
         </template>
+        <template #emptyMessage>
+          <div class="uk-text-center"><oc-icon name="delete" type="div" size="large" variation="passive" /><div class="uk-text-muted" v-translate>No deleted files</div></div>
```
LukasHirt: Separate this into own lines pls, it's too long.

apps/files/src/components/AllFilesList.vue
```diff
@@ -99,6 +102,26 @@ export default {
...
+      if (this.$_isFavoritesList) {
+        return 'star'
+      } else {
+        return 'folder'
+      }
+    },
+
+    $_emptyFolderText () {
+      if (this.$_isFavoritesList) {
+        return this.$gettext('No favorites defined')
```
LukasHirt: Maybe too much technical message? What about something like "No files/folders marked as favorite"?

apps/files/src/components/FilesAppBar.vue
```diff
@@ -50,7 +50,7 @@
             <oc-button v-if="selectedFiles.length > 0" icon="restore" @click="$_ocTrashbin_restoreFiles()">
               <translate>Restore selected</translate>
             </oc-button>
-            <oc-button id="delete-selected-btn" icon="delete" @click="selectedFiles.length < 1 ? $_ocTrashbin_empty() : $_ocTrashbin_deleteSelected()">
+            <oc-button id="delete-selected-btn" icon="delete" :disabled="files.length === 0" @click="selectedFiles.length < 1 ? $_ocTrashbin_empty() : $_ocTrashbin_deleteSelected()">
```
kulmann: Good catch to disabled the button! However, you can now remove the part from the click handler, that deals with an empty selection.

PVince81: empty selection doesn't mean no files
if you click the button with no files selected, it's the same as selecting all

PVince81:
I've addressed all comments, except for the design part.

- [ ] We need to update ODS to get the bigger icon sizes: https://github.com/owncloud/owncloud-design-system/pull/632
- [ ] Set size to xlarge and make sure the font size also matches

individual-it:
In the spirit of BDD and the discussions we had in the product repo about how to write steps I would suggest to explicitly have a step that says that there should be the message. I would argue that there is a new behaviour in the application, but that is not reflected in the feature files.
@HanaGemela @phil-davis how do you feel about that?

tests/acceptance/stepDefinitions/filesContext.js
```diff
@@ -320,10 +320,16 @@ When('the user unmarks the favorited file/folder {string} using the webUI', func
...
-  })
+Then('there should be no files/folders/resources listed on the webUI', async function () {
+  const emptyMessageVisible = await client
+    .page
+    .FilesPageElement.filesList()
+    .isEmptyMessageVisible()
+
+  const allRowsResult = await client.page.FilesPageElement.filesList().allFileRows()
+
+  assert.ok(emptyMessageVisible, 'Empty message must be visible')
```
individual-it: is the message empty, or is the message about the folder being empty? ;-)

PVince81: the message is about "folder is empty", see screen shot above.

I suspected that this wording will cause confusion, I'll adjust it

PVince81:
> In the spirit of BDD and the discussions we had in the product repo about how to write steps I would suggest to explicitly have a step that says that there should be the message. I would argue that there is a new behaviour in the application, but that is not reflected in the feature files.
> @HanaGemela @phil-davis how do you feel about that?

I was also a bit confused here and went for the idea that the "empty folder" message is an implementation/design detail (which could be considered obvious in any web app)) not related to the requirements. I guess if we say that it's an explicit product requirement to display a message, then we need it as a Gherkin step.

apps/files/src/components/AllFilesList.vue
```diff
@@ -118,9 +118,17 @@ export default {
...
-        return this.$gettext('Folder is empty')
+        return this.$gettext('There are no resources in this folder.')
+      }
+    },
+
+    $_emptyFolderSubText () {
+      if (this.$_isFavoritesList) {
+        return this.$gettext('You can mark some by clicking on the star icon.')
+      } else {
+        return this.$gettext('Drag files and folder here to upload, or browse files from your computer.') // TODO: link
```
kulmann: - `folder` missing an s (plural)
- maybe replace `computer` with `device`, as we're also targeting mobile

apps/files/src/components/AllFilesList.vue
```diff
@@ -118,9 +118,17 @@ export default {
...
+        return this.$gettext('There are no resources marked as favorite.')
       } else {
-        return this.$gettext('Folder is empty')
+        return this.$gettext('There are no resources in this folder.')
+      }
+    },
+
+    $_emptyFolderSubText () {
+      if (this.$_isFavoritesList) {
+        return this.$gettext('You can mark some by clicking on the star icon.')
```
kulmann: ... `star icon in the file list`? As a new user I might not know where to find the star icon.

tests/acceptance/stepDefinitions/filesContext.js
PVince81: renamed to "noContentMessage"

PVince81:
Fixed all "no content" messages.
Renamed EmptyMessage component to NoContentMessage to avoid confusion.
Separate styling for message and call to action.
Adjusted screenshot in original post.

@kulmann please recheck.

apps/files/src/components/Collaborators/SharedFilesList.vue
```diff
@@ -45,6 +45,14 @@
       </div>
       <div class="uk-text-meta uk-text-nowrap uk-width-small" v-text="formDateFromNow(item.shareTime)" />
     </template>
+    <template #noContentMessage>
+      <no-content-message icon="share">
+        <template #message>
+          <span v-if="$_isSharedWithMe"><translate>You are currently not collaborating on other people's resources.</translate></span>
+          <span v-if="!$_isSharedWithMe"><translate>You are currently not collaborating on any of your resources with other people.</translate></span>
```
kulmann: `v-else`?

PVince81:
For some reason I intuitively tend to avoid `v-else` as it makes the element depend on the previous one, which means I can't easily move around the element with `v-else`. in some cases, like the RecycleScroller, one needs to scroll a lot to find the original condition.

Anyway, I think we can agree to use `v-else` still. I've adjusted accordingly.

apps/files/src/components/AllFilesList.vue
```diff
@@ -37,6 +37,16 @@
               {{ formDateFromNow(item.mdate) }}
             </div>
           </template>
+          <template #noContentMessage>
+            <no-content-message v-if="!$_isFavoritesList" icon="folder">
+              <template #message><translate>There are no resources in this folder.</translate></template>
```
LukasHirt: Why not using the slot directly on `translate` tag?

apps/files/src/components/Collaborators/SharedFilesList.vue
```diff
@@ -45,6 +45,14 @@
       </div>
       <div class="uk-text-meta uk-text-nowrap uk-width-small" v-text="formDateFromNow(item.shareTime)" />
     </template>
+    <template #noContentMessage>
+      <no-content-message icon="share">
+        <template #message>
+          <span v-if="$_isSharedWithMe"><translate>You are currently not collaborating on other people's resources.</translate></span>
```
LukasHirt: `translate` tags default to span. Could we then either use `v-translate` directly on the span or dropping the span and moving the if statement to the translate tag? I'd prefer the former.

apps/files/src/components/AllFilesList.vue
PVince81: Do you mean putting v-translate on the template tag ? I wasn't sure whether this works.
I guess I could test it with a dummy temporary local translation

PVince81:
@LukasHirt I've addressed your comments and moved the `v-translate` on the previous tag.

also fixed the tests

apps/files/src/components/AllFilesList.vue
LukasHirt: Not sure if that works. I thought more about `<translate v-slot:message>`

LukasHirt: Ah, now I see this was deprecated in 2.6.0. My bad...

PVince81:
I did a few tests with translations:
- `<template #message v-translate>` the text stays in english
- `<translate v-slot:message>` kills the text completely
- `<span v-translate>` works correctly

PVince81:
@LukasHirt I went with `<span v-translate>` now to avoid span duplication.

LukasHirt:
@PVince81 Thanks for those tests! I saw that the `v-translate` is now in some parts on `<template>`. If I got your test right it's not translating, right? Maybe just a leftover from tests then?

PVince81:
@LukasHirt forgot pushing, please rechech

PVince81:
I messed up the test sentences while doing manual search and replace... Fixed now.

PVince81:
Yeah, the tests fail because they use an existing skeleton folder which contains the file `.gitkeep` :palm_tree: :face_with_head_bandage:

PVince81:
Seems the "no content" check is not compatible with the files drop page, so I've added an URL check and disabled the specific check for the message presence when on the files-drop page.

apps/files/src/components/FileList.vue
```diff
@@ -2,7 +2,7 @@
   <!-- TODO: Take care of outside click overall and not just in files list -->
   <div :id="id" class="uk-height-1-1 uk-position-relative" @click="hideRowActionsDropdown">
     <div class="uk-flex uk-flex-column uk-height-1-1">
-      <oc-grid gutter="small" flex id="files-table-header" class="uk-padding-small">
+      <oc-grid gutter="small" flex id="files-table-header" class="uk-padding-small" v-if="fileData.length > 0" >
```
LukasHirt: Add key pls to `v-if`/`v-else`

PR #2976, opened 2020-02-05
Test click authorize button directly

PVince81:
## Description
Don't wait for the element to be visible, as the click method will wait
on its own.

Hoping that this can fix the issues that sometimes clicking doesn't do
anything, maybe because the layout shifts during page load
after the element was detected.

## Related Issue
None raised.

## Motivation and Context
We've been learning more about how Nightwatch works over time, so this older code might need some rework to reduce unneeded waiting times and speed up tests.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/pageObjects/ownCloudAuthorizePage.js
+++ b/tests/acceptance/pageObjects/ownCloudAuthorizePage.js
@@ -12,7 +12,6 @@ module.exports = {
     {
       authorize: function () {
         return this
-          .waitForElementVisible('@authorizeButton')
           .click('@authorizeButton')
           .waitForElementNotPresent({
             selector: '@authorizeButton',


PVince81:
seems this did not solve the "WARNING: looks like I'm still on auth page. I will click the auth button again"

individual-it:
We could never find out why that click does not always work, I don't think `click()` waits for the element to be visible, the documentation mentions to scroll the element into view, but I guess it would fail if the element does not exist in the DOM. the example in the documentation also uses `wait` before `click` https://nightwatchjs.org/api/click.html

That might be our issue https://github.com/nightwatchjs/nightwatch/issues/1221 but if I recall correctly the angry click function did not solve it ether

PVince81:
that page states "click() will automatically wait for the element to be present (until the specified timeout)"

I suspect that the reason the click doesn't always go through is that maybe the CSS assets have not been loaded yet, so maybe the button has a different location on the screen.

From my understanding the click routine will first find the coordinates of the button with one HTTP request to the driver, then another request will make it click at the specified coordinates. It is possible that between the coordinate detection and the click, the CSS assets just got applied so the button location shifted.

I did have such similar occurrences in the past where Selenium was not able to click due to shifting layout after load. The solution was adding more waiting code. But here on the Authorize page there is no element that can represent "CSS assets have finished loading".

individual-it:
element to be present is not the same as element visible. present means to be in the DOM, but does not meant that its visible

PVince81:
okay, understood. thanks.
then this PR is likely not a good idea, closing.

PR #2983, opened 2020-02-07
Collaborator panel tweaks

kulmann:
## Description
This fixes the remaining UI/UX issues from the linked issue.
- Showing avatars for selected collaborators on a new share and fixed styling/layouting of said collaborators in the list.
- Added sensible margins on text about missing permissions for re-sharing in the sharing sidebar.
- Fixed alignment of displayed collaborator in editing view for collaborators.
- Removed separators from the forms that were cluttering the view.
- Moved role description on role selection (links and collaborators) into the form element. Not shown below the form element anymore.
- Fixed margins on spacers between sections in both sidebar panels.
- Removed animations for the initial loading of a links or shares list. Only the changes (adding or removing a share/link) have a transition now.
- Introduced sorting by name for public links
- Cleaned up code in the `FileLinkSidebar` (e.g. removed unnecessary filtering)

And one bugfix as well:
When typing text into the search box for new collaborators, selecting a user and a group with identical names was not possible.
This was due to the fact that when one (group or user) got selected, the other was excluded because of a matching name. Fixed by
including the share type (group or user) in matching.

## Related Issue

- Fixes https://github.com/owncloud/phoenix/issues/1186

## How Has This Been Tested?

- locally with Chrome and Firefox
- [ ] drone acceptance tests

Bildschirmfoto 2020-02-08 um 11 30 47

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
+++ b/apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
@@ -6,17 +6,14 @@
       :selectedRole="role"
       @roleSelected="selectRole"
     />
-    <div v-if="$_ocCollaborators_currentRoleDescription" class="uk-text-muted" v-text="$_ocCollaborators_currentRoleDescription"></div>
-    <hr v-if="$_ocCollaborators_hasAdditionalPermissions" class="divider" />
     <label class="oc-label" v-if="$_ocCollaborators_hasAdditionalPermissions">
-      <translate>Additional permissions</translate>
+      <translate>Additional permissions:</translate>
     </label>
     <additional-permissions
       :availablePermissions="role.additionalPermissions"
       :collaboratorsPermissions="collaboratorsPermissions"
       @permissionChecked="checkAdditionalPermissions"
     />
-    <hr v-if="$_ocCollaborators_expirationSupported" class="divider" />
     <oc-grid v-if="$_ocCollaborators_expirationSupported" gutter="small">
       <div class="uk-width-1-1">
           <label class="oc-label" for="files-collaborators-new-collaborator-expiration">
@@ -67,13 +64,6 @@ export default {
   computed: {
     $_ocCollaborators_expirationSupported () {
       return false
-    },
-
-    $_ocCollaborators_currentRoleDescription () {
-      if (this.selectedRole && this.selectedRole.name !== 'advancedRole') {
-        return this.selectedRole.description
-      }
-      return ''
     },
 
     $_ocCollaborators_hasAdditionalPermissions () {
--- a/apps/files/src/components/Collaborators/EditCollaborator.vue
+++ b/apps/files/src/components/Collaborators/EditCollaborator.vue
@@ -1,10 +1,7 @@
 <template>
   <div class="files-collaborators-collaborator-edit-dialog">
     <div v-if="user.id !== collaborator.info.uid_owner" class="uk-text-meta uk-flex uk-flex-middle uk-margin-small-bottom"><oc-icon name="repeat" class="uk-margin-small-right" /> {{ collaborator.info.displayname_owner }}</div>
-    <div class="uk-flex uk-flex-wrap uk-flex-middle">
-      <collaborator class="uk-width-expand" :collaborator="collaborator" />
-    </div>
-    <hr class="divider" />
+    <collaborator class="uk-width-expand" :collaborator="collaborator" :first-column="false" />
     <collaborators-edit-options
       :existingRole="$_originalRole"
       :collaboratorsPermissions="$_originalPermissions"
@@ -14,14 +11,16 @@
     <hr class="divider" />
     <oc-grid gutter="small" class="uk-margin-bottom">
       <div>
-        <oc-button class="files-collaborators-collaborator-cancel" :disabled="saving" @click="$_ocCollaborators_cancelChanges">
+        <oc-button :disabled="saving" @click="$_ocCollaborators_cancelChanges" class="files-collaborators-collaborator-cancel" key="edit-collaborator-cancel-button">
           <translate>Cancel</translate>
         </oc-button>
-        <oc-button v-if="saving" :disabled="true">
+        <oc-button :disabled="true" key="edit-collaborator-saving-button" v-if="saving">
           <oc-spinner :aria-label="$gettext('Saving Share')" class="uk-position-small uk-position-center-left" size="xsmall"/>
           <span :aria-hidden="true" class="uk-margin-small-left" v-translate>Saving Share</span>
         </oc-button>
-        <oc-button v-else variation="primary" :disabled="!$_hasChanges" :aria-label="$gettext('Save Share')" @click="$_ocCollaborators_saveChanges">
+        <oc-button :aria-label="$gettext('Save Share')" :disabled="!$_hasChanges" @click="$_ocCollaborators_saveChanges" id="files-collaborators-collaborator-save-share-button"
+                   key="edit-collaborator-saving-button" v-else
+                   variation="primary">
           <translate>Save Share</translate>
         </oc-button>
       </div>
--- a/apps/files/src/components/Collaborators/NewCollaborator.vue
+++ b/apps/files/src/components/Collaborators/NewCollaborator.vue
@@ -21,55 +21,44 @@
         </template>
       </oc-autocomplete>
     </oc-grid>
-    <oc-grid gutter="small">
-      <div
-        v-if="selectedCollaborators.length > 0"
-      >
+    <oc-grid gutter="small" v-if="selectedCollaborators.length > 0">
+      <div>
         <div>
-          <div>
-            <translate>Selected collaborators</translate>:
-          </div>
-          <div
-            v-for="(collaborator, index) in selectedCollaborators"
-            :key="collaborator.label"
-            class="uk-flex-inline uk-flex-row uk-flex-start uk-margin-small-bottom"
-            :class="{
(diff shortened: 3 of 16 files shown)


apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
```diff
@@ -6,17 +6,14 @@
       :selectedRole="role"
       @roleSelected="selectRole"
     />
-    <div v-if="$_ocCollaborators_currentRoleDescription" class="uk-text-muted" v-text="$_ocCollaborators_currentRoleDescription"></div>
-    <hr v-if="$_ocCollaborators_hasAdditionalPermissions" class="divider" />
     <label class="oc-label" v-if="$_ocCollaborators_hasAdditionalPermissions">
-      <translate>Additional permissions</translate>
+      <translate>Additional permissions</translate>:
```
PVince81: keep the colon inside the tag. in some translations the colon could be on the left side I guess (RTL mode)

I googled a bit and found this to confirm: http://docs.translatehouse.org/projects/localization-guide/en/latest/guide/translation/punctuation.html#colons

Also https://www.accelingo.com/software-localization-best-practices/ (search for Punctuation)

Please also adjust the other occurrences

apps/files/src/components/Collaborators/NewCollaborator.vue
```diff
@@ -165,15 +155,11 @@ export default {
...
-                return (
-                  collaborator.value.shareWith ===
-                  selectedCollaborator.value.shareWith
-                )
-              }
-            )
+          this.autocompleteResults = users.concat(groups, remotes).filter(collaborator => {
+            const selected = this.selectedCollaborators.find(selectedCollaborator => {
+              return collaborator.value.shareWith === selectedCollaborator.value.shareWith &&
+                parseInt(collaborator.value.shareType, 10) === parseInt(selectedCollaborator.value.shareType, 10)
```
PVince81: I thought we already parsed `shareType` to int in our last developments. But ok to keep it for safety

kulmann: Those are coming directly from the SDK, not from the store. So those are not parsed. There is only one occurrence a few lines down, where I was able to use an already parsed `shareType` (changed it now after you pointed it out, thanks!).

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
kulmann: That was an interesting read. `For example, there are spaces surrounding a colon in French.` from your last link. 😱 Changed every occurrence of punctuation and i18n that I could find.

PVince81: Yep. It's only recently that I learnt that both German and English do not put a space before a question mark while in French there's one.

PVince81:
There's some issue with an autocomplete test, I guess it might be relevant since you touched some of these bits ?

kulmann:
I will fix it tomorrow morning. There were already some other tests failing, related to changed xpath for selected collaborators. I changed to using classes and ids where possible, so that tests are already fixed. Probably missed one somewhere. ;-)

kulmann:
> There's some issue with an autocomplete test, I guess it might be relevant since you touched some of these bits ?

Looked into it just now. From my understanding the tests are wrong. Could you clarify please? Details:

Failing tests are
```
Scenario: Share file with a user and a group with same name (attempt 1, retried)
tests/acceptance/features/webUISharingInternalGroups/shareWithGroupsEdgeCases.feature:36
```
and
```
 Scenario: Share file with a group and a user with same name (attempt 1, retried)
tests/acceptance/features/webUISharingInternalGroups/shareWithGroupsEdgeCases.feature:57
```

Both tests are about sharing with a user and a group, where the user and the group have the same names. From my understanding this should be possible - and that's what I supposedly fixed in the code by adding this line:
https://github.com/owncloud/phoenix/blob/d9a7d00226e2a5d11e7bc3aaf175cb441fb1d963/apps/files/src/components/Collaborators/NewCollaborator.vue#L162

The fact that we check for shareTypes when looking into already existing shares here:
https://github.com/owncloud/phoenix/blob/d9a7d00226e2a5d11e7bc3aaf175cb441fb1d963/apps/files/src/components/Collaborators/NewCollaborator.vue#L168
supports my assumption.

Can you please clarify which behaviour would be correct? From my understanding, sharing with a user and group with identical names should be possible. If it shouldn't be, then I'll have to revert that line (162), fix/remove the other line (168) and remove the changelog item. If it's correct how I changed/fixed it I will just change the tests.

PS: I'll change the tests already, because I think my understanding is correct. 😁 and also because oc10 implementation is the same.

kulmann:
@PVince81 I made two more changes in this PR, please re-review.

1: visual improvement: the `hr` tags between lists caused different spacings for sections than within a section. I made style changes (through uikit margin classes) to ensure that all visual dividers produce the same distance a) from item before b) to divider c) to next item.

2: I don't remember where you wrote the comment, but you asked if we should have transitions on all list items, not only the items for direct shares/links. While I agree that all should be the same, I don't think that all items should have a transition BUT only changes should have a transition. Or in other words: when loading a list for the first time there should not be any animation. Only when you change something (add or remove), that change should have an animation, so that the user can easily see what changed. For the initial loading of a list, there is no use in an animation from a UX perspective (only design), so I made a commit which makes sure that only the changes are animated, not the initial loading of a list. Hope that you agree. :-)

apps/files/src/components/Collaborators/NewCollaborator.vue
```diff
@@ -21,39 +21,30 @@
...
-              <oc-icon
-                name="close"
-                variation="danger"
-              />
-            </oc-button>
-          </div>
+          <translate>Selected collaborators:</translate>
         </div>
+        <template v-for="collaborator in selectedCollaborators">
+          <oc-table middle :key="collaborator.value.shareWith + '-' + collaborator.value.shareType"
```
LukasHirt: Can we get selected collaborators into one table pls? It's better for accessibility. Also, why is it wrapped in the template? Is there some issue with using `v-for` directly in e. g. table tor?

apps/files/src/components/Collaborators/EditCollaborator.vue
```diff
@@ -21,7 +18,8 @@
           <oc-spinner :aria-label="$gettext('Saving Share')" class="uk-position-small uk-position-center-left" size="xsmall"/>
           <span :aria-hidden="true" class="uk-margin-small-left" v-translate>Saving Share</span>
         </oc-button>
-        <oc-button v-else variation="primary" :disabled="!$_hasChanges" :aria-label="$gettext('Save Share')" @click="$_ocCollaborators_saveChanges">
+        <oc-button v-else variation="primary" :disabled="!$_hasChanges" :aria-label="$gettext('Save Share')"
```
LukasHirt: I know this was missing there already before your PR but since you're touching this now, could you pls add also key to `v-if` and `v-else`?

apps/files/src/components/Collaborators/NewCollaborator.vue
kulmann: Erm... kind of ridiculous that I didn't use a single table in the first place. Thanks for pointing it out. :-) I am now using a single table and have the `v-for` on the `tr`s, like it should be. 😁

apps/files/src/components/Collaborators/EditCollaborator.vue
kulmann: Done. :-) Same for the buttons in `NewCollaborator.vue`. Will keep an open eye about this in the future.

apps/files/src/components/Collaborators/NewCollaborator.vue
LukasHirt: Thanks a lot! 🚀

apps/files/src/components/Collaborators/EditCollaborator.vue
LukasHirt: Nice! Thanks!

apps/files/src/components/FileSharingSidebar.vue
```diff
@@ -8,42 +8,44 @@
             <translate>Add Collaborators</translate>
           </oc-button>
         </div>
-        <translate
-          v-else
-          class="files-collaborators-no-reshare-permissions-message"
-          key="no-reshare-permissions-message"
-          v-text="noResharePermsMessage"
-        />
+        <p v-else>
```
LukasHirt: If the change is only because of having the `p` element then we can use either `v-translate` on that element or specify `tag="p"` on translate element. Either way, we can drop the extra element then.

kulmann: Now that you point it out, I am wondering why there was a translate tag in the first place. The `v-text` brings in text which is already translated. I merged the `p` and `translate` into a single `p` tag without any `v-translate`. ;-) Thanks for finding this 💪

LukasHirt: Thanks! 👍 It escaped me that it was already translated 🙈

LukasHirt:
Can you pls squash commits before merging? Thanks

kulmann:
> Can you pls squash commits before merging? Thanks

@LukasHirt Yes, of course! I know that it's a bit messy right now. But I wanted to give Vincent a chance to review only my latest commits instead of the full PR. ;-)

PR #2985, opened 2020-02-07
Improve design of versions tab a bit

PVince81:
## Description
Added file type icon.
Removed border for buttons.

## Related Issue
None raised

## Motivation and Context
We already have other buttons without border and they look nice, so why not here as well.

## How Has This Been Tested?
Manual test

### Before
image

### After
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] changelog


PVince81:
I'm unsure whether to include the file name also as it would be the same.
We don't preserve old file names in case of rename, so maybe it's best to not show it.

In the future those icons could become actual file previews.

PVince81:
@pmaier1 FYI see screenshots

PR #2987, opened 2020-02-07
Use `$ngettext` for pluralization in i18n

kulmann:
## Description
Use `$ngettext` instead of `$gettext` for messages with pluralization. This allows to make messages context sensitive, like `delete 1 file` and `delete 42 files` in one translation. Apparently, there are only 2 occurrences in the whole files app, where this is relevant.
Checked in easygettext that `$ngettext` is supposed to be exported to transifex as well, so I expect no problems in that matter.

## Motivation and Context
Correct pluralization of messages.

## How Has This Been Tested?

Checked the two messages manually in the browser.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FilesAppBar.vue
+++ b/apps/files/src/components/FilesAppBar.vue
@@ -525,14 +525,18 @@ export default {
     },
 
     $_ocTrashbin_deleteSelected () {
-      const translated = this.$gettext('%{numberOfFiles} items will be deleted immediately. You can’t undo this action.')
-      this.setTrashbinDeleteMessage(this.$gettextInterpolate(translated, { numberOfFiles: this.selectedFiles.length }, true))
+      const translated = this.$ngettext(
+        "%{numberOfFiles} item will be deleted immediately. You can't undo this action.",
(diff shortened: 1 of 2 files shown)


apps/files/src/components/FilesAppBar.vue
```diff
@@ -525,14 +525,18 @@ export default {
     },
 
     $_ocTrashbin_deleteSelected () {
-      const translated = this.$gettext('%{numberOfFiles} items will be deleted immediately. You can’t undo this action.')
-      this.setTrashbinDeleteMessage(this.$gettextInterpolate(translated, { numberOfFiles: this.selectedFiles.length }, true))
+      const translated = this.$ngettext(
+        '%{numberOfFiles} item will be deleted immediately. You can’t undo this action.',
```
PVince81: Mind also fixing that suspicious apostrophe that looks like a forward tick ?
You can also change it to "cannot"

PR #2996, opened 2020-02-10
Text tweaks for translation

PVince81:
Parametrize product name in welcome message.
Add space in "Log out" to be consistency with another location.

### DIFF
--- a/src/pages/login.vue
+++ b/src/pages/login.vue
@@ -6,7 +6,7 @@
             </h1>
             <div class="oc-login-card-body">
                 <h3 class="oc-login-card-title">
-                    <span v-translate>Welcome to</span> {{ configuration.theme.general.name }}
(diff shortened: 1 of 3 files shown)


src/pages/login.vue
```diff
@@ -6,7 +6,7 @@
             </h1>
             <div class="oc-login-card-body">
                 <h3 class="oc-login-card-title">
-                    <span v-translate>Welcome to</span> {{ configuration.theme.general.name }}
+                  <span v-text="$_welcomeMessage" />
```
kulmann: Want to try `<translate :translate-params="{productName: $_productName}">Welcome to %{productName}</translate>` here and only define `$_productName` as computed property? :-)

PVince81:
@kulmann Adjusted and works, please recheck

Issue #1952, opened 2019-09-11
Fixed table header


LukasHirt:
## Description
The table header needs to become fixed so the users still see easily which column is which.

PVince81:
- [ ] file list

PVince81:
@wuenschedesign for more feedback. See how it's done on mobile as well.

PR #2999, opened 2020-02-11
Fix filename column width in responsive mode

PVince81:
## Description
~~Removed uk-width-small and now using the icon width for column
alignment. This gives more space in responsive mode for the file name
column.
Removed the "More" column header label which felt redundant.
Adjusted file actions button id to not mention size any more, now that
we only have a single size.~~

- Removed uk-width-small and now using the icon width for column
alignment. This gives more space in responsive mode for the file name
column.
- Removed the "More" column header label which felt redundant.
- Adjusted file actions button id to not mention size any more, now that
we only have a single size.
- Share indicator column now always present even when indicators are not
rendered, for consistency with the header.
- On resize, copy the table body row width to the header row to make sure
that headers stay aligned when a scrollbar appears in the body as
the scrollbar does not cover the header part as the latter is
fixed/frozen on scroll.

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/2998

## How Has This Been Tested?
- manual test in responsive and regular mode to check file name column and also the column header alignment with column contents

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/FileList.vue
+++ b/apps/files/src/components/FileList.vue
@@ -2,7 +2,8 @@
   <!-- TODO: Take care of outside click overall and not just in files list -->
   <div :id="id" class="uk-height-1-1 uk-position-relative" @click="hideRowActionsDropdown">
     <div class="uk-flex uk-flex-column uk-height-1-1">
-      <oc-grid gutter="small" flex id="files-table-header" class="uk-padding-small" v-if="fileData.length > 0" key="files-list-results-existence">
+      <resize-observer @notify="$_resizeHeader" />
+      <oc-grid ref="headerRow" gutter="small" flex id="files-table-header" class="uk-padding-small" v-if="fileData.length > 0" key="files-list-results-existence">
         <div>
           <oc-checkbox
             class="uk-margin-small-left"
@@ -15,12 +16,7 @@
           />
         </div>
         <slot name="headerColumns"/>
-        <div
-          class="uk-text-meta uk-text-right uk-width-small uk-margin-small-right"
-          v-translate
-        >
-          More
-        </div>
+        <div class="uk-margin-small-right oc-icon" />
       </oc-grid>
       <div id="files-list-container" class="uk-flex-1 uk-overflow-auto" v-if="!loading">
         <RecycleScroller
@@ -35,7 +31,7 @@
             :data-is-visible="active"
             @click="selectRow(item, $event); hideRowActionsDropdown()"
           >
-            <oc-grid gutter="small" flex class="uk-padding-small oc-border-top" :class="_rowClasses(item)" :id="'file-row-' + item.id">
+            <oc-grid :ref="index === 0 ? 'firstRow' : null" gutter="small" flex class="uk-padding-small oc-border-top" :class="_rowClasses(item)" :id="'file-row-' + item.id">
               <div>
                 <oc-checkbox
                   class="uk-margin-small-left"
@@ -45,8 +41,8 @@
                   :hideLabel="true"
                 />
               </div>
-              <slot name="rowColumns" :item="item" :index="item.id" />
-              <div class="uk-width-small uk-text-right uk-margin-small-right">
+              <slot name="rowColumns" :item="item" :index="index" />
+              <div class="uk-text-right uk-margin-small-right">
                 <oc-button
                   :id="actionsDropdownButtonId(item.id, active)"
                   class="files-list-row-show-actions"
@@ -155,6 +151,12 @@ export default {
 
     item () {
       return this.$route.params.item
+    }
+  },
+  watch: {
+    compactMode (val) {
+      // sidebar opens, recalculate header sizes
+      this.$_resizeHeader()
     }
   },
   methods: {
@@ -302,10 +304,20 @@ export default {
 
     actionsDropdownButtonId (id, active) {
       if (active) {
-        return `files-file-list-action-button-small-resolution-${id}-active`
+        return `files-file-list-action-button-${id}-active`
       }
 
-      return `files-file-list-action-button-small-resolution-${id}`
+      return `files-file-list-action-button-${id}`
+    },
+
+    $_resizeHeader () {
+      this.$nextTick(() => {
+        const headerRow = this.$refs.headerRow
+        const firstRow = this.$refs.firstRow
+        if (headerRow && firstRow) {
+          headerRow.$el.style.width = getComputedStyle(firstRow.$el).width
+        }
+      })
     }
   }
 }
--- a/apps/files/src/components/FilesLists/RowActionsDropdown.vue
+++ b/apps/files/src/components/FilesLists/RowActionsDropdown.vue
@@ -1,9 +1,9 @@
 <template>
   <oc-drop
     v-if="displayed"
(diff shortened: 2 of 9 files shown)


apps/files/src/components/FilesLists/RowActionsDropdown.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <oc-drop
     v-if="displayed"
-    :boundary="`#files-file-list-action-button-small-resolution-${item.id}-active`"
+    :boundary="`#files-file-list-action-button-${item.id}-active`"
     :options="{ offset: 0 }"
```
kulmann: Since you are touching this code, mind putting `flip: false` inside the options json here? In responsive mode the dropdown can flip to the top and hide behind the top bar, see screenshot. flip: false prevents flipping to the top.

Bildschirmfoto 2020-02-11 um 17 51 52

PVince81:
@kulmann please recheck all, I've also added "flip: false" now to the new file menu after rebasing

apps/files/src/components/FileList.vue
```diff
@@ -302,11 +300,17 @@ export default {
...
       }
 
-      return `files-file-list-action-button-small-resolution-${id}`
+      return `files-file-list-action-button-${id}`
     }
   }
 }
 </script>
+<style scoped>
+  /* FIXME: move to ODS, it is useful to have a column as wide as just an icon */
```
kulmann: When you are raising a ticket for this, add icon size classes as an idea. So `oc-icon-width` with the default icon width, then `oc-icon-width-xsmall`, `oc-icon-width-medium`, etc.

PVince81:
Indeed I can just set "oc-icon" on the header div to get the width, done that now.

After rebasing on top of master and getting #1952, I have now a bigger alignment problem:
the header bar of the table has padding-small, but has no scroll bar. So we're missing some additional space on the right

So now I'd need to set the column width to 39px instead of just 24px to accommodate for that. The 24px only works fine when no scrollbar is here. I don't think I can detect the scrollbar width, and it would be hacky anyway.

With scrollbar:
image

Without scrollbar:
image

Any suggestions ? Would it be possible to have a fixed table header but still have the scrollbar pushing the layout ? @LukasHirt

PVince81:
Idea: use JS to measure the width of the name column in the body (it's the one that expands as far as possible) and then apply the same with to the name column in the header. This would help align the column headers and is likely easier than trying to guess the scrollbar size.

Otherwise I'd wish for a way with CSS to tell the browser to not count the scrollbar as being part of the layout, like Mac OS does.

PVince81:
It seems there used to be a value called `overflow: overlay` that would achieve the desired result, but it has been deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow

PVince81:
Seems we have an even bigger undiscovered problem: the share indicators column doesn't have a fixed width for neither the header nor the body... It seems we'd need the same calculation here as well.

kulmann:
> Seems we have an even bigger undiscovered problem: the share indicators column doesn't have a fixed width for neither the header nor the body... It seems we'd need the same calculation here as well.

@PVince81 what about calculating the width of `#files-list-container` and setting it as max-width on `#files-table-header`? So that you don't operate on column level, but instead just for the whole divs?

PVince81:
I have an ugly but working solution now: copy all column widths from the table rows to the header to have consistent sizes.

@kulmann processing column-wise is necessary because of the share indicators column width which is not predictable when using extensions (it's not always just two icons). Adjusting the parent's width wouldn't be enough to help with header label alignments.

I'm unsure yet whether I'll want to merge this solution but please do have a look and let me know what you think. @LukasHirt @kulmann

PVince81:
if this become a thing for more tables (at least the ones with RecycleScroller), we could consider wrapping all this ugliness into a separate vue component in the future...

PVince81:
After some thought I think I'll still try the other solution of setting the width of the parent, the share indicator column is likely not relevant there.

- [ ] try approach about setting parent width

apps/files/src/components/FileList.vue
```diff
@@ -301,6 +308,32 @@ export default {
...
       return `files-file-list-action-button-${id}`
+    },
+
+    $_resizeHeader () {
+      this.$nextTick(() => {
+        const headerRow = this.$refs.headerRow
+        const firstRow = this.$refs.firstRow
+        if (headerRow && firstRow) {
+          const headerColumns = Array.from(headerRow.$el.childNodes).filter(n => n.nodeType === Node.ELEMENT_NODE)
+          const rowColumns = Array.from(firstRow.$el.childNodes).filter(n => n.nodeType === Node.ELEMENT_NODE)
```
kulmann: You could rename `rowColumns` to `bodyColumns`, just to stay in the wording context of tables.

kulmann: Also in all variables and comments below.

apps/files/src/components/FileList.vue
```diff
@@ -301,6 +308,32 @@ export default {
...
+          for (let i = 0; i < headerColumns.length; i++) {
+            const headerColumn = headerColumns[i]
+            const rowColumn = rowColumns[i]
+            if (!rowColumn) {
```
kulmann: Do we want to start syncing sizes, if the column layout doesn't match between header and body? Are there valid use cases for that? If not we could check this once before the for loop (break on `if (headerColumns.length !== bodyColumns.length)`) instead of checking in every iteration.

apps/files/src/components/FileList.vue
```diff
@@ -301,6 +308,32 @@ export default {
...
+              // can happen in case of discrepancy between number of header and row columns
+              console.warn('no row column found for header ', headerColumn)
+              break
+            }
+            const rowComputedStyle = getComputedStyle(rowColumn)
+            headerColumn.classList.remove('uk-width-expand')
```
kulmann: Instead of removing ui lib specific classes, you could set both the width and max-width, to make sure that header columns can't expand to larger widths than the body columns.

PVince81:
@kulmann you were right, it seems to work fine! See last commit.

If all agree I'll squash.

There's a little UIKit bug I found while switching between multiple responsive mode: in some cases it seems UIKit is setting "uk-first-column" on the wrong column, here "Updated" which causes an additional top padding. I'd consider this unrelated but we could probably work it around later.

kulmann:
> @kulmann you were right, it seems to work fine! See last commit.
> 
> If all agree I'll squash.
> 
> There's a little UIKit bug I found while switching between multiple responsive mode: in some cases it seems UIKit is setting "uk-first-column" on the wrong column, here "Updated" which causes an additional top padding. I'd consider this unrelated but we could probably work it around later.

Yes, please squash. On mac this works fine. When you squashed I'll review again tomorrow, early in the morning.

PVince81:
As discussed:
- squashed
- removed "flip: false" from the changeset as it caused weird menu positioning

PVince81:
I'll add a changelog entry for this as a bugfix since it did not make it into 0.4.0

PR #3013, opened 2020-02-12
Separated navigation

### LINKED ISSUE #2746: Extension points app navigation vs app switcher
- [ ] properly manage the currently visible app
    - [ ] left side bar should only show routes from current app, not all of them
    - [ ] optional: actually display the app name in the top left
- [ ] extension point in app switcher to allow apps to inject more icons with URLs

LukasHirt:
## Description
In the sidebar is now only extension-specific navigation and into the apps switcher are automatically injected extensions which have at least one nav item. In the case that extension has only one nav item, the menu button is replaced by the name of the extension.

## Related Issue

- Fixes #2746

## How Has This Been Tested?

- test environment: Manually

image

image

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/Phoenix.vue
+++ b/src/Phoenix.vue
@@ -7,8 +7,20 @@
       </template>
       <template v-else>
         <message-bar :active-messages="activeMessages" @deleteMessage="$_deleteMessage" />
-        <top-bar :applicationsList="$_applicationsList" :activeNotifications="activeNotifications" :user-id="user.id" :user-display-name="user.displayname" :hasAppNavigation="!!appNavigationEntries.length" @toggleAppNavigation="$_toggleAppNavigation(!appNavigationVisible)"></top-bar>
-        <side-menu :visible="appNavigationVisible" :entries="appNavigationEntries" @closed="$_toggleAppNavigation(false)"></side-menu>
+        <top-bar
+          :applicationsList="$_applicationsList"
+          :activeNotifications="activeNotifications"
+          :user-id="user.id"
+          :user-display-name="user.displayname"
+          :hasAppNavigation="appNavigationEntries.length > 1"
+          @toggleAppNavigationVisibility="toggleAppNavigationVisibility"
+        />
+        <side-menu
+          v-if="appNavigationEntries.length > 1"
+          :visible="appNavigationVisible"
+          :entries="appNavigationEntries"
+          @closed="hideAppNavigation"
+        />
         <main id="main">
           <router-view id="oc-app-container" name="app" class="uk-height-1-1"></router-view>
         </main>
@@ -39,27 +51,47 @@ export default {
   },
   computed: {
     ...mapState(['route', 'user']),
-    ...mapGetters(['configuration', 'activeNotifications', 'activeMessages', 'capabilities']),
+    ...mapGetters(['configuration', 'activeNotifications', 'activeMessages', 'capabilities', 'apps']),
     $_applicationsList () {
-      return this.configuration.applications
+      const list = []
+
+      // Get extensions manually added into config
+      list.push(this.configuration.applications)
+
+      // Get extensions which have at least one nav item
+      const navItems = this.$root.navItems
+      for (const extensionId in navItems) {
+        list.push(this.apps[extensionId])
+      }
+
+      return list.flat()
     },
 
     appNavigationEntries () {
       if (this.publicPage()) {
         return []
       }
+
       // FIXME: use store or other ways, not $root
-      return this.$root.navItems.filter(item => {
-        // FIXME: filter to only show current app
+      const items = this.$root.navItems[this.currentExtension]
+
+      if (!items) {
+        return []
+      }
+
+      return items.filter(item => {
         if (item.enabled === undefined) {
           return true
         }
+
         if (this.capabilities === undefined) {
           return false
         }
+
         return item.enabled(this.capabilities)
       })
     },
+
     showHeader () {
       return this.$route.meta.hideHeadbar !== true
     },
@@ -109,15 +141,22 @@ export default {
   },
   methods: {
     ...mapActions(['initAuth', 'fetchNotifications', 'deleteMessage']),
-    $_toggleAppNavigation (state) {
-      this.appNavigationVisible = state
+
+    hideAppNavigation () {
+      this.appNavigationVisible = false
     },
+
+    toggleAppNavigationVisibility () {
+      this.appNavigationVisible = !this.appNavigationVisible
+    },
+
     $_updateNotifications () {
       this.fetchNotifications(this.$client).catch((error) => {
         console.error('Error while loading notifications: ', error)
         clearInterval(this.$_notificationsInterval)
       })
     },
+
     $_deleteMessage (item) {
       this.deleteMessage(item)
     }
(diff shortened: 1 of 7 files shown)


LukasHirt:
I have a small issue when displaying some title for the current extension. Currently, it is displayed only in case of extension having only one nav item when it replaces the menu button. But when we have the menu button there it doesn't look nice. I'm thinking if this might be solved with having the sidebar always visible as was once already discussed. Then this issue would be solved. I'm just a bit scared of how it would look like when switching between extension with menu and without. Suddenly appearing sidebar might not be so nice in the end as well.

Seems to me like a problem for a separate issue.

src/Phoenix.vue
```diff
@@ -39,17 +51,35 @@ export default {
...
-      return this.configuration.applications
+      const list = []
+
+      // Get extensions manually added into config
+      list.push(this.configuration.applications)
+
+      // Get extensions which have at least one nav item
+      const navItems = this.$root.navItems
+      for (const item in navItems) {
+        list.push(this.apps[item])
```
PVince81: confusing: does `navItems` contain app ids somewhere ?
from reading the code it seems `navItems` keys are the app ids and you're taking them from there.
if yes I suggest renaming `item` to `appId` to make this clearer

src/Phoenix.vue
```diff
@@ -39,17 +51,35 @@ export default {
...
+
+      const items = this.$root.navItems[this.currentExtension]
+
+      if (!items) {
+        return []
+      }
+
       // FIXME: use store or other ways, not $root
-      return this.$root.navItems.filter(item => {
+      return items.filter(item => {
```
PVince81: move the FIXME to above where this.$root is used ?

src/Phoenix.vue
```diff
@@ -39,17 +51,35 @@ export default {
...
+
+      return list.flat()
     },
 
     appNavigationEntries () {
       if (this.publicPage()) {
         return []
       }
+
+      const items = this.$root.navItems[this.currentExtension]
```
PVince81: aha, it took me a while to find where you do the actual filtering. Here it is :-)

PVince81:
I guess not easy to add acceptance tests ?

Regarding the trouble with the sidebar button: in what scenario would the button/sidebar flicker ? Is it when switching between an app that has a sidebar to one that doesn't ?

src/Phoenix.vue
LukasHirt: NavItems for each extension are stored under extension id. I changed the name. You're right it wasn't clear.

src/Phoenix.vue
LukasHirt: Right. Done :wink:

src/Phoenix.vue
LukasHirt: Originally wanted to move it somewhere else but since you're touching phoenix.js file and part of it is living in there I wanted to wait when that gets merged to align it together.

LukasHirt:
> I guess not easy to add acceptance tests ?

We'd need to load some other extension with routes to test the app switcher. And TBH I didn't want to implement acceptance tests for this part because I think this should be covered with unit tests. As soon as they're set up in Phoenix they can be implemented. I raised an issue for it here #3020

LukasHirt:
> Regarding the trouble with the sidebar button: in what scenario would the button/sidebar flicker ? Is it when switching between an app that has a sidebar to one that doesn't ?

Exactly. Basically scenario when an extension has only one nav item.

src/Phoenix.vue
```diff
@@ -7,8 +7,20 @@
...
         <message-bar :active-messages="activeMessages" @deleteMessage="$_deleteMessage" />
-        <top-bar :applicationsList="$_applicationsList" :activeNotifications="activeNotifications" :user-id="user.id" :user-display-name="user.displayname" :hasAppNavigation="!!appNavigationEntries.length" @toggleAppNavigation="$_toggleAppNavigation(!appNavigationVisible)"></top-bar>
-        <side-menu :visible="appNavigationVisible" :entries="appNavigationEntries" @closed="$_toggleAppNavigation(false)"></side-menu>
+        <top-bar
+          :applicationsList="$_applicationsList"
+          :activeNotifications="activeNotifications"
+          :user-id="user.id"
+          :user-display-name="user.displayname"
+          :hasAppNavigation="appNavigationEntries.length > 1"
+          @toggleAppNavigation="$_toggleAppNavigation(!appNavigationVisible)"
```
kulmann: Sorry for being nitpicking here 😁 in my opinion, a function with name "toggle" should not have a parameter. I would either name it `$_setAppNavigationVisible` and provide a new value, or name it `$_toggleAppNavigationVisibility` and don't provide a parameter. You are legally allowed to hit me for that when we see each other.

src/Phoenix.vue
```diff
@@ -7,8 +7,20 @@
...
+        />
+        <side-menu
+          v-if="appNavigationEntries.length > 1"
+          :visible="appNavigationVisible"
+          :entries="appNavigationEntries"
+          @closed="$_toggleAppNavigation(false)"
```
kulmann: Ok, since you set an explicit value here, I'd vote for `$_setAppNavigationVisible` as function name.

src/components/Top-Bar.vue
```diff
@@ -59,8 +70,14 @@ export default {
...
   },
   computed: {
+    ...mapGetters(['apps']),
+
     isPublicPage () {
       return !this.userId
+    },
+
+    currentExtensionName () {
+      return this.$gettext(this.apps[this.currentExtension].name)
```
kulmann: For apps provided in the `applications` part of config.json (see sample config), the extension doesn't live in `this.apps`. I've looked into the available data in store and hacked this (dirty):

```
if (this.apps[this.currentExtension]) {
        return this.$gettext(this.apps[this.currentExtension].name)
      } else {
        const applications = this.configuration.applications.filter(app => {
          if (app.path) {
            return app.path.includes(this.currentExtension)
          } else {
            return app.url.includes(this.currentExtension)
          }
        })
        if (applications) {
          // TODO: get title for correct lang or fallback lang from first application's title
          return 'App Title'
        } else {
          // TODO: I don't know
          return 'App Title'
        }
      }
``` 

You have to add `'configuration'` to `mapGetters` array as well.

kulmann: I hope that there is an easier way to select the correct application from `configuration.applications`, because to be honest, my filter also looks hacky as hell.

src/Phoenix.vue
LukasHirt: I went a bit different direction. Since this event is responsible only for closing I went with method called `hideAppNavigation` which doesn't take any argument. Hope that's okay :wink:

src/Phoenix.vue
LukasHirt: Implemented the "proper" toggle :wink:

src/Phoenix.vue
kulmann: perfect 😍

src/components/Top-Bar.vue
LukasHirt: As discussed, applications list in config is meant to include extensions which live in their own SPAs. This will split the responsibility for displaying the current extension. Let's see after we have such an extension as an example how we need to implement this.

LukasHirt:
@kulmann Changes implemented or addressed.

src/components/Top-Bar.vue
LukasHirt: I'll raise an issue for it so we don't forget it later

PR #3014, opened 2020-02-12
Fix l10n process

PVince81:
## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/3012

## How Has This Been Tested?
Run `make l10n-read` and see that the error is gone.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/store/actions.js
+++ b/apps/files/src/store/actions.js
@@ -809,7 +809,7 @@ export default {
   },
 
   // TODO: Think of a better name
-  pendingShare (context, { client, item, type, translate }) {
+  pendingShare (context, { client, item, type, $gettext }) {
     // TODO: Move request to owncloud-sdk
     client.requests.ocs({
       service: 'apps/files_sharing',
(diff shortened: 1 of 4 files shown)


apps/files/src/store/actions.js
```diff
@@ -819,12 +819,12 @@ export default {
       .then(_ => {
         context.dispatch('loadFolderSharedWithMe', {
           client: client,
-          $gettext: translate
+          $gettext: $gettext
```
kulmann: You can just write `$gettext` here, no key: value required.

DeepDiver1975:
It might be good to run make l10n-read in each and every PR to detect if there is something wrong with respect to translations .....

PVince81:
> It might be good to run make l10n-read in each and every PR to detect if there is something wrong with respect to translations .....

The likeliness of this happening is rather low I think, so won't add it.

However, unskipping errors for push/pull would be able to catch a broader scope of potential sync issues in the future.

DeepDiver1975:
> The likeliness of this happening is rather low I think, so won't add it.

Generally speaking: catching an error on PR level is better then days later on the translation sync job (which is monitored by whom? :man_shrugging: )

DeepDiver1975:
and compared to the time we spend on all the acceptance tests this is a rather small job ...

DeepDiver1975:
anyway - your project - your decision :muscle:

PR #3041, opened 2020-02-14
Replace join-path usages with the path library

PVince81:
#  Description
Since we now include the "path" library, we don't need the "join-path"
library which only provides a subset of its features.

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/2922

## How Has This Been Tested?
- Manual test of relevant code pieces

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/components/FilesAppBar.vue
+++ b/apps/files/src/components/FilesAppBar.vue
@@ -121,7 +121,7 @@ import FileDrop from './FileDrop.vue'
 import { mapActions, mapGetters, mapState } from 'vuex'
 import Mixins from '../mixins'
 import FileActions from '../fileactions'
-import join from 'join-path'
+import pathUtil from 'path'
 
 export default {
   components: {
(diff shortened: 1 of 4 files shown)


LukasHirt:
Looking at the tests error it seems it might be still somewhere imported?

PVince81:
oh, so it seems we used this lib in some tests as well. I'll recheck...

PVince81:
it looks like join-path doesn't behave exactly like `path.join` when dealing with URLs, so I can't replace it in the tests as a lot of test code has such assumption in place.

I'll keep the join-paths dependency for now for tests

PVince81:
pushed. let's see...

PVince81:
@LukasHirt it went through this time

LukasHirt:
I guess we don't need changelog item for this, right?

PR #3049, opened 2020-02-17
Improve collaborators and owner columns in shared file lists

PVince81:
## Description
       
When sharing with multiple targets, they must all be listed there instead of a single entry. This is fixed by replacing the code that make share entries unique into an aggregation that has been inspired by the one used in OC core.
Also includes group shares and link shares now.
Added avatars and icons for the displayed collaborators in the column.

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/2924

## Motivation and Context
Just doing a bugfix was too boring, but also the column contents didn't look very nice, so adding avatars there was a good move IMO.

## How Has This Been Tested?
Manual test
Acceptance tests

### Shared with others
image

### Shared with you
image

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:
- [x] test first!!!
- [x] add test for the group cases
- [x] implement the actual fix
- [x] changelog

### DIFF
--- a/apps/files/src/components/Collaborators/SharedFilesList.vue
+++ b/apps/files/src/components/Collaborators/SharedFilesList.vue
@@ -7,7 +7,9 @@
           <translate translate-context="Name column in files table">Name</translate>
         </sortable-column-header>
       </div>
-      <div class="uk-text-nowrap uk-text-meta uk-width-small" v-text="$_sharedCellTitle" />
+      <div><!-- indicators column --></div>
+      <div key="shared-with-header-cell" v-if="!$_isSharedWithMe" class="uk-text-nowrap uk-text-meta uk-width-medium" translate-context="Collaborators table column" v-text="$gettext('Collaborators')" />
+      <div key="shared-with-header-cell" v-else class="uk-text-nowrap uk-text-meta uk-width-small" translate-context="Owner table column" v-text="$gettext('Owner')" />
       <div
         v-if="$route.name === 'files-shared-with-me'"
         shrink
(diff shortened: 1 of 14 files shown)


PVince81:
- [ ] align with data structure from #3076 which changed a bit
- [ ] weird shifting in column headers

PVince81:
### Shared with others
image

### Shared with you
image

I could go further and try and add resharer/owner but it increases complexity. I think the current result is already a big step forward.

PVince81:
This is only a stepping stone as we will consider merging "shared with me" with "shared with others" view into a single one, so it will likely evolve further from here

PVince81:
Failed tests:
- renaming from share list: tests/acceptance/features/webUIRenameFiles/renameFiles.feature:160
- deleting from share list: tests/acceptance/features/webUIDeleteFilesFolders/deleteFilesFolders.feature:201
- resharing: tests/acceptance/features/webUIResharing/reshareUsers.feature:185
- some sharing: tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature:62
- accept/reject: tests/acceptance/features/webUISharingAcceptShares/acceptShares.feature:13

PVince81:
data structure aligned

PVince81:
- [ ] seems row height is bigger than before

PVince81:
row height issue unrelated to this PR => https://github.com/owncloud/phoenix/issues/3099

PVince81:
rebased and squashed.

next up will be adding/fixing tests

PVince81:
added and fixed tests, let's hope all pass now

PVince81:
raised https://github.com/owncloud/phoenix/issues/3101 which I discovered while trying to extend the test coverage

apps/files/src/components/Collaborators/SharedFilesList.vue
```diff
@@ -35,21 +37,30 @@
...
-            (<translate>group</translate>)
-          </span>
-        </div>
-        <div v-else key="shared-from-cell">
-          {{ item.shareOwnerDisplayname }}
-        </div>
+      <div><!-- indicators column --></div>
+      <div key="shared-with-cell" v-if="!$_isSharedWithMe" class="uk-text-meta uk-text-nowrap uk-text-truncate uk-width-medium uk-flex file-row-collaborators">
+        <span v-for="share in item.shares" :key="share.id" class="uk-margin-small-right uk-flex uk-flex-middle">
+          <avatar-image v-if="share.shareType === shareTypes.user && share.collaborator" class="uk-margin-xsmall-right" :width="24" :userid="share.collaborator.name" :userName="share.collaborator.displayName" />
```
LukasHirt: Add key pls

PVince81: keys added

PVince81:
I'll adjust the tests later in the morning (not saying tomorrow as it's past midnight...)

PVince81:
tests adjusted

background: the main reason for all these changes is that before this PR we were using `share.path` in the shared file list, which is always the original path. I changed it to use `share.file_target` which is the path in the point of view of the receiver. And that even when a share is declined, the latter stays. This is to align with the OC 10 logic. We could decide to change this later on if a different behavior makes sense.

PR #3078, opened 2020-02-21
File list columns are now sortable

PVince81:
## Description
Added ability to sort some columns in the file lists by clicking on
them. The sort direction appears with an indicator.

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/1854

## How Has This Been Tested?
- manual test
- acceptance tests

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] restore or keep sort order per file list (ex: sort by share time in "shared with others" then go back to "all files", no sort is set now)
- [x] add sortable columns in more lists (trashbin, shares)
- [x] add acceptance tests


PVince81:
Now thinking, it won't be easy to have a sort field per table because the sorting is happening in the global state. Would require moving the global state's `activeFiles` into the object, and possibly more of the mixins. For this task I'll try to limit the changes and just make sure it doesn't behave too badly when switching between views. Maybe it will simply fallback to the name column when switching for now.

PVince81:
- [ ] aria label for sort indicator
- [ ] own vue component for clickable sort column, there's already too much markup...

PVince81:
I've added clickable columns in more lists, and also the whole width is clickable.
Since there's a lot of repetition, let's move those into a separate Vue component.

Also will need a way to detect those indicators from the test to find out whether to click once or twice depending on the desired sort order.

PVince81:
- [ ] use raw oc-button instead of links

PVince81:
- [ ] use raw oc-button instead of links, requires ODS update to not look shitty

tests/acceptance/pageObjects/FilesPageElement/filesList.js
```diff
@@ -721,6 +721,29 @@ module.exports = {
       return indicators
     },
 
+    setSort: async function (column, isDesc = false) {
+      const columnSelector = util.format(this.elements.filesTableHeaderColumn.selector, xpathHelper.buildXpathLiteral(column))
+      await this.useXpath()
+        .waitForElementVisible(columnSelector)
```
LukasHirt: From nightwatch docs:

> click() will automatically wait for the element to be present (until the specified timeout).

Do we need to wait until the element is visible or is present enough?

PVince81: let's keep "visible" just to be on the safe side...

PR #3086, opened 2020-02-24
Added expiration date to collaborators

### LINKED ISSUE #2543: Expiration date field for regular shares
Same as https://github.com/owncloud/core/pull/36324 but for Phoenix

- [ ] If the capability `files_sharing.user.expireDate.enabled` is set, show expiration field
- [ ] Fix expiration field + date picker (currently hidden and broken)
- [ ] Consider enforcement fields/capabilities as well

LukasHirt:
## Related Issue

- Fixes #2543

## How Has This Been Tested?

- test environment: Manually and CI

image

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added


LukasHirt:
@PVince81 I implemented just two basic tests. There is a lot more to test but that would take 1 - 2 more days of work. If you're okay with it I'd say let's have just those two so this feature is finished in this sprint and then implement the additional ones in next sprint.

apps/files/src/components/Collaborators/EditCollaborator.vue
```diff
@@ -55,13 +59,28 @@ export default {
...
+      // Since expiration date can ba reset to null we can't depend on it when checking for changes
+      expirationDateResetted: false
     }
   },
   computed: {
     ...mapGetters('Files', ['highlightedFile']),
     ...mapGetters(['user']),
 
+    collaboratorType () {
+      if (this.collaborator.shareType === 0 || this.collaborator.shareType === 6) {
```
PVince81: please use the "shareTypes" file with the constants

apps/files/src/components/Collaborators/EditCollaborator.vue
```diff
@@ -82,8 +101,28 @@ export default {
         // if the role has changed, always return true. The user doesn't need to understand if two bitmasks of different roles are the same!
         return true
       }
+
+      // FIXME: Datepicker is not displaying correct timezone so for now we add it manually
+      const originalExpirationDate = moment(this.originalExpirationDate).add(moment().utcOffset(), 'm').toISOString()
```
PVince81: should we use a helper function for this hack ? this is the second time this appears

apps/files/src/components/Collaborators/EditCollaborator.vue
```diff
@@ -82,8 +101,28 @@ export default {
...
+
       const originalBitmask = roleToBitmask(this.$_originalRole, Object.keys(this.$_originalPermissions))
       return originalBitmask !== this.$_permissionsBitmask
+    },
+
+    originalExpirationDate () {
+      const expirationDate = this.collaborator.expires
+
+      if (expirationDate) {
+        const date = expirationDate
```
PVince81: return directly ?

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
```diff
@@ -51,19 +72,35 @@ export default {
...
+      enteredExpirationDate: null
     }
   },
+
   computed: {
-    $_ocCollaborators_expirationSupported () {
-      return false
+    ...mapGetters(['capabilities']),
+
+    editingUser () {
```
PVince81: would be nice to make all these private by prefixing with `$_`.

the other day we did a test and noticed that such public members are accessible from the outside when accessing the vue object instance

tests/acceptance/stepDefinitions/generalContext.js
```diff
@@ -206,3 +206,20 @@ Given('the app {string} has been disabled', function (app) {
       'app:disable', app
     ])
 })
+
+Given('default expiration date for users is set to {int} day/days', function (days) {
+  // FIXME: make array in runOcc functon work
+  occHelper.runOcc(
```
PVince81: you can also use the step `Given the setting "shareapi_default_expire_date_user_share" of app "core" has been set to "yes"`

apps/files/src/store/actions.js
```diff
@@ -283,7 +283,7 @@ function _buildCollaboratorShare (s, file) {
 
   // expiration:Object if unset, or string "2019-04-24 00:00:00"
   if (typeof s.expiration === 'string' || s.expiration instanceof String) {
-    share.expires = Date.parse(s.expiration)
+    share.expires = new Date(s.expiration)
```
PVince81: add `// FIXME: use moment`
parsing the date like this is likely to cause troubles with timezones

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
LukasHirt: Then something must be wrong if the methods were accessible from the Vue instance since this is not how Vue works. If you want to have methods like that they need to be directly assigned to the global instance. Otherwise, you need to do a few gymnastics to get to them (e.g. via `$parent$)

And if we make private everything then we are limiting ourselves and also introducing another step that outside developers would need to understand when contributing.

^ just my opinion

PVince81: I think the way to access is by using refs. You could then call methods on child vues.

I would only make stuff private if it's not intended to be used outside like in parent vues.

PVince81: here's an example: https://github.com/owncloud/phoenix/blob/master/apps/files/src/components/Collaborators/NewCollaborator.vue#L123

we deliberately exposed the "focus()" method on the autocomplete component from ODS.

apps/files/src/components/Collaborators/EditCollaborator.vue
LukasHirt: I'd prefer not to introduce workarounds as helpers since then it feels a bit like a justification of them. While this way it will be at least a nice reminder that we need to fix it. But it's more about the feeling here than real reasoning. I'd be okay in the end with creating it.

tests/acceptance/stepDefinitions/generalContext.js
LukasHirt: Yeah, wanted to keep the steps in scenario minimal. But will use that 👍

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
LukasHirt: Yeah, but that a bit different since this is coming from a design system component where indeed it does make sense for defining the private methods specifically. But in other cases, it is kind of an anti-pattern to access it like this since with that we are making the components dependent on each other. That's where the store should come into play.

PVince81: Another advantage of marking private members is that you directly see whether you're accessing a local private member vs one imported from the store when looking at the markup.

I'm fine solidifying either decision as long as we are consistent.

cc @kulmann as we discussed the public member thing before

LukasHirt: From Vue docs:

> **Use module scoping to keep private functions inaccessible from the outside**. If that’s not possible, always use the $_ prefix for custom private properties in a plugin, mixin, etc that should not be considered public API. Then to avoid conflicts with code by other authors, also include a named scope (e.g. $_yourPluginName_).

apps/files/src/store/actions.js
LukasHirt: With Date object, you get local timezone that is set in the OS of the user. This is an example of what date returns - `Sat Feb 29 2020 00:00:00 GMT+0100 (Central European Standard Time)`

I think it would be better not to use moment directly here because than we are enforcing it on all later use cases.

PVince81: Ah ok. So if "s.expiration" contains timezone info, then using `new Date()` is fine. I thought it was a milliseconds value in which case it would use the local timezone.

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
PVince81: ah, so those private guidelines are for plugins and mixins, not Vue components ?

I'm fine removing the prefixes then

LukasHirt:
@PVince81 Apart from open discussions all comments addressed.

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
LukasHirt: > ah, so those private guidelines are for plugins and mixins, not Vue components ?

Yes.

PVince81:
some failures, like tests/acceptance/features/webUISharingInternalUsers/shareWithUsers.feature:49

LukasHirt:
Yeah, looking into it now.

LukasHirt:
Hmm, need to change the way how the test for displaying an expiration date is done. Currently, it would the success of the test would depend on the time of a day it's run.

LukasHirt:
> Hmm, need to change the way how the test for displaying an expiration date is done. Currently, it would the success of the test would depend on the time of a day it's run.

Possible solution - Use full date instead of from now format. It might look like this:

image

LukasHirt:
The core issue is that momentjs is using "22 to 35 hours" as 1 day.

LukasHirt:
Ok, tests passed - @PVince81 pls re-review 🙏

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
```diff
@@ -98,8 +135,104 @@ export default {
...
+    },
+
+    canResetExpirationDate () {
+      return !this.expirationDateEnforced && this.enteredExpirationDate
     }
   },
+
+  mounted () {
+    // FIXME: Datepicker is not displaying correct timezone so for now we add it manually
+    // this.enteredExpirationDate = this.expirationDate ? moment(this.expirationDate).toISOString(true) : this.defaultExpirationDate
```
PVince81: is it worthy to raise a tech debt ticket for this ? or too minor ?

apps/files/src/components/Collaborators/EditCollaborator.vue
```diff
@@ -55,13 +60,34 @@ export default {
     return {
       selectedRole: null,
       additionalPermissions: null,
-      saving: false
+      saving: false,
+      expirationDate: null,
+      // Since expiration date can ba reset to null we can't depend on it when checking for changes
```
PVince81: s/ba/be/g

apps/files/src/components/Collaborators/EditCollaborator.vue
```diff
@@ -55,13 +60,34 @@ export default {
...
+      expirationDateResetted: false
```
PVince81: another pattern would be storing the old value and new value.
it does seem like the original value is available in `originalExpirationDate()`, why not use it ?

changelog/unreleased/2543
```diff
@@ -0,0 +1,7 @@
+Enhancement: Expiiration date for collaborators
```
PVince81: s/piir/pir/g

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
LukasHirt: IMHO the whole date picker component should use some different library than it does now because it is modal. So I'm not sure if it's worth it to raise issue for this. Maybe a general issue for changing the library?

LukasHirt:
@PVince81 Comments addressed.

PR #3136, opened 2020-03-05
Changed sort order for collaborators and public links

LukasHirt:
## Related issue:
- Fixes https://github.com/owncloud/enterprise/issues/3852

## How Has This Been Tested?

- test environment: Manually

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:
- [x] Sort links
- [x] Remove edit options from indirect shares

### DIFF
--- a/apps/files/src/components/FileSharingSidebar.vue
+++ b/apps/files/src/components/FileSharingSidebar.vue
@@ -22,7 +22,7 @@
               <collaborator :collaborator="$_currentUserAsCollaborator"/>
             </li>
           </ul>
-          <hr class="uk-margin-small-top uk-margin-small-bottom" v-if="$_directOutgoingShares.length > 0 || $_indirectOutgoingShares.length > 0" />
+          <hr class="uk-margin-small-top uk-margin-small-bottom" v-if="collaborators.length > 0" />
         </section>
         <section>
           <transition-group id="files-collaborators-list"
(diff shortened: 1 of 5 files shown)


LukasHirt:
Raised issue for unit tests - #3139

LukasHirt:
@PVince81 Ready for review :pray:

apps/files/src/components/FileSharingSidebar.vue
```diff
@@ -294,25 +285,30 @@ export default {
...
-      if (name1 === name2) {
-        const c1GroupShare = c1.shareType === shareTypes.group
-        const c2GroupShare = c2.shareType === shareTypes.group
-        if (c1GroupShare === c2GroupShare) {
-          return textUtils.naturalSortCompare(c1.id + '', c2.id + '')
-        } else {
-          return c1GroupShare ? -1 : 1
+      const c1UserShare = c1.shareType === shareTypes.user || c1.shareType === shareTypes.remote
+      const c2UserShare = c2.shareType === shareTypes.user || c1.shareType === shareTypes.remote
+      const c1DirectShare = !c1.outgoing
```
PVince81: confusion here: with "outgoing" I'd understand "outgoing share" but not direct vs indirect

I wonder if it was related to the hack that used that attribute simply to find out whether the share entry comes from the sharestree vs the regular one. Maybe add a comment there to clarify.

I think I'll want to clean this up soon as "the hack" is starting to spread everywhere.
Or if you want to tackle it quickly, maybe we can simply add an attribute "indirect" additionally to "incoming" and "outgoing" to reduce confusion. Not optimal but would do.

PVince81:
Please add a link to the original ticket, I can't find it by simply scrolling through this PR

apps/files/src/components/FileSharingSidebar.vue
LukasHirt: I originally added the indirect but than I saw there is the outgoing so I didn't want to introduce a new key for the same result. But I get there the confusion and didn't know that it was only a hack.

I'll adjust it.

PVince81: Thanks a lot and sorry for the old mess :-S

LukasHirt: Just went the quick way as you suggested and added there an `indirect` key. I'm not sure how you plan to tackle this in refactor so didn't want to touch it too much.

LukasHirt:
Issue link added and confusion addressed. Pls re-review @PVince81

PR #3167, opened 2020-03-10
Updated all deps

LukasHirt:


PVince81:
CI failure on the linter, maybe some of the rules got implicitly stricter ?

LukasHirt:
> CI failure on the linter, maybe some of the rules got implicitly stricter ?

Strange, this should have failed already in the previous CIs. I'll fix it

LukasHirt:
@PVince81 Removed the unused var

LukasHirt:
Tests are failing with timeout - will investigate

LukasHirt:
Locally doesn't pass the test for me even on master 🤢

LukasHirt:
Maybe found a reason - nightwatch upgrade slipped in. Reverted to prev version. Let's see if it will pass now.

Locally still not working but because of some issue with my setup. Will need to see what causes that.

PR #3187, opened 2020-03-16
Added file previews

PVince81:
## Related Issue
Part of https://github.com/owncloud/phoenix/issues/276

## How Has This Been Tested?
- [x] manual test, scroll wildly
- [x] preview in "all files list"
- [x] previews in public link files list
- [x] previews in shared file list
- acceptance tests

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [x] update ODS https://github.com/owncloud/owncloud-design-system/pull/678
- [x] changelog entry
- [x] acceptance tests

### DIFF
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -48,9 +48,13 @@
         <oc-star class="uk-display-block" @click.native.stop="toggleFileFavorite(item)" :shining="item.starred" />
       </div>
       <div class="uk-text-truncate uk-width-expand" :ref="index === 0 ? 'firstRowNameColumn' : null">
-        <oc-file @click.native.stop="item.type === 'folder' ? navigateTo(item.path.substr(1)) : openFileActionBar(item)"
-          :name="$_ocFileName(item)" :extension="item.extension" class="file-row-name" :icon="fileTypeIcon(item)"
-          :filename="item.name" :key="item.id"/>
+        <file-item
+          @click.native.stop="item.type === 'folder' ? navigateTo(item.path.substr(1)) : openFileActionBar(item)"
+          :item="item"
+          :dav-url="davUrl"
+          :show-path="$_isFavoritesList"
+          class="file-row-name"
+          :key="item.viewId"/>
         <oc-spinner
           v-if="actionInProgress(item)"
           size="small"
@@ -112,6 +116,7 @@
 <script>
 import FileList from './FileList.vue'
 import Indicators from './FilesLists/Indicators.vue'
+import FileItem from './FileItem.vue'
 import NoContentMessage from './NoContentMessage.vue'
 import { mapGetters, mapActions, mapState } from 'vuex'
 
@@ -127,6 +132,7 @@ export default {
   components: {
     FileList,
     Indicators,
+    FileItem,
     NoContentMessage,
     SortableColumnHeader
   },
@@ -186,6 +192,26 @@ export default {
       })
 
       return Object.keys(shareTypes).map(shareType => parseInt(shareType, 10))
+    },
+
+    davUrl () {
+      let davUrl
+      // FIXME: use SDK once it switches to DAV v2
+      if (this.publicPage()) {
+        davUrl = [
+          '..',
+          'dav',
+          'public-files'
+        ].join('/')
+      } else {
+        davUrl = [
+          '..',
+          'dav',
+          'files',
+          this.$store.getters.user.id
+        ].join('/')
+      }
+      return this.$client.files.getFileUrl(davUrl)
     },
 
     $_isFavoritesList () {
@@ -268,14 +294,6 @@ export default {
         client: this.$client,
         file: item
       })
-    },
-    $_ocFileName (item) {
-      if (this.$_isFavoritesList) {
-        const pathSplit = item.path.substr(1).split('/')
-        if (pathSplit.length === 2) return `${pathSplit[pathSplit.length - 2]}/${item.basename}`
-        if (pathSplit.length > 2) return `…/${pathSplit[pathSplit.length - 2]}/${item.basename}`
-      }
-      return item.basename
     },
 
     isActionEnabled (item, action) {
--- a/apps/files/src/components/FileItem.vue
+++ b/apps/files/src/components/FileItem.vue
@@ -1,0 +1,102 @@
+<template>
+  <oc-file
+    :name="fileName"
+    :extension="item.extension"
+    :icon="previewIcon"
+    :iconUrl="previewUrl"
+    :filename="item.name"
+    />
+</template>
+<script>
+import queryString from 'query-string'
+import Mixins from '../mixins'
+
+export default {
+  name: 'FileItem',
+  mixins: [
+    Mixins
+  ],
+  props: {
+    item: {
+      type: Object
+    },
+    // override for file name
+    name: {
+      type: String
+    },
+    davUrl: {
+      type: String
+    },
+    showPath: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data: function () {
+    return {
+      previewUrl: this.item.previewUrl
+    }
+  },
+  computed: {
+    fileName () {
+      if (this.name) {
+        return this.name
+      }
+      if (this.showPath) {
+        const pathSplit = this.item.path.substr(1).split('/')
+        if (pathSplit.length === 2) return `${pathSplit[pathSplit.length - 2]}/${this.item.basename}`
+        if (pathSplit.length > 2) return `…/${pathSplit[pathSplit.length - 2]}/${this.item.basename}`
+      }
+      return this.item.basename
+    },
+    previewIcon () {
+      return this.fileTypeIcon(this.item)
+    }
+  },
+  mounted () {
+    this.loadPreview()
+  },
+  methods: {
+    loadPreview () {
+      if (this.item.previewUrl) {
+        this.previewUrl = this.item.previewUrl
+        this.previewLoaded = true
+        return
+      }
+
+      // TODO: check if previews are globally enabled (requires capability entry)
+      // don't load previews for pending or rejected shares (status)
+      if (!this.davUrl || this.item.type === 'folder' || (typeof this.item.status !== 'undefined' && this.item.status !== 0)) {
+        this.previewLoaded = true
+        return
+      }
+
+      const query = {
+        x: this.thumbDimensions,
+        y: this.thumbDimensions,
+        scalingup: 0,
+        preview: 1,
+        a: 1
+      }
+
+      if (this.item.etag) {
+        // add etag for URL based caching
+        // strip double quotes from etag
+        query.c = this.item.etag.substr(2, this.item.etag.length - 2)
+      }
+
+      let itemPath = this.item.path
+      if (itemPath.charAt(0) === '/') {
+        itemPath = itemPath.substr(1)
+      }
+
+      const previewUrl = this.davUrl + '/' + this.encodePath(itemPath) + '?' + queryString.stringify(query)
+
+      this.mediaSource(previewUrl, 'url', this.requestHeaders).then(dataUrl => {
+        // cache inside item
+        this.previewUrl = this.item.previewUrl = dataUrl
+      })
+    }
+  }
+}
+</script>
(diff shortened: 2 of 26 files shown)


PVince81:
seems blob URL doesn't work for the icon component:
image

PVince81:
Previews do work when I scroll over the same element at least once.
I suspect that `mounted()` from oc-icon is not being called as "iconUrl" is undefined. In error cases it would be an empty string instead of undefined.
Likely related to the way how recycle scroller works.

PVince81:
- [ ] requires ODS update to fix preview loading issue: https://github.com/owncloud/owncloud-design-system/pull/681

PVince81:
added tests but discovered some new issues:

- [ ] preview not loading directly after upload ?
- [ ] scrolling down then back up quicky sometimes removes the already loaded previews: maybe the elements are re-created or something and we need to requery the URL ? (svg appears there for icon type instead of the expected img from the previous load, makes acceptance tests fail)

PVince81:
and now the lorem file is locked.
the failures are totally random... I'll try rebasing just in case...

PVince81:
rebased

PVince81:
Seems master is all green https://github.com/owncloud/phoenix/pull/3202

So there is something I changed here that is causing new side effects :cry:

PVince81:
seems there's a locking issue.

waiting for previews to always load is not a good idea as we need to use a delay, it will slow down the tests.

idea: add a hidden loading indicator, maybe "data-loading" on the preview while the media source is running, so that the tests know that they need to wait longer.

PVince81:
or use `waitForOutstandingAjaxCalls`

PVince81:
Added `waitForOutstandingAjaxCalls` to attempt to wait for previews to be loaded by `mediaSource()` before continuing. This is to prevent running into situations where the file itself might be locked.

apps/files/src/components/FileItem.vue
```diff
@@ -0,0 +1,88 @@
...
+</template>
+<script>
+import queryString from 'query-string'
+import fileTypeIconMappings from '../fileTypeIconMappings.json'
+
+export default {
+  name: 'FileItem',
+  props: {
+    /**
+         * The html element used for the avatar container.
```
LukasHirt: Some strange indentation here?

apps/files/src/components/FileItem.vue
```diff
@@ -0,0 +1,88 @@
...
+        x: this.thumbDimensions,
+        y: this.thumbDimensions,
+        c: this.item.etag,
+        scalingup: 0,
+        preview: 1,
+        a: 1
+      })
+
+      const previewUrl = this.davUrl + '/' + this.item.path + '?' + query
+      this.mediaSource(previewUrl, 'url', this.requestHeaders).then(dataUrl => {
```
LukasHirt: Aren't you missing the `requestHeaders` mixin in this component?

LukasHirt:
Pls post a screenshot into description

PVince81:
adjusted and rebased

PVince81:
while trying to debug the tests I found more issues which are fixed now:

- path names with special characters work now (try loading preview for "strängé folder name (duplicate #2 &)/strängé filename (duplicate #2 &).txt")
- added limit to concurrent downloads of previews

However there are still these issues:
- [ ] for some strange reason, the preview loading doesn't kick in when entering folder in CI environment
- [ ] due to concurrency, I suspect that the "wait for outstanding ajax calls" would think that the last ajax call was done when maybe next tick the next ajax call will start (in a p-queue env), so need to add a delay there.
- [ ] preview detection still feels unreliable, maybe need to use a data attribute instead with a loading state there...

apps/files/src/components/FileItem.vue
```diff
@@ -0,0 +1,87 @@
...
+        preview: 1,
+        a: 1
+      })
+
+      let itemPath = this.item.path
+      if (itemPath.charAt(0) === '/') {
+        itemPath = itemPath.substr(1)
+      }
+
+      const previewUrl = this.davUrl + '/' + this.encodePath(itemPath) + '?' + query
```
C0rby: I think the '/' is not necessary. It would generate an url with a double slash.

PVince81: see the code above that removes the slash on the left if there's one

PVince81:
Locking issues are still there https://drone.owncloud.com/owncloud/phoenix/8983/24/14

It is likely that waitForOutstandingAjaxCalls is not waiting properly as described, considering that some ajax calls might fire up one tick after we detected the termination of the previous one.

apps/files/src/components/FileItem.vue
C0rby: Oh right, I have overlooked that.

PVince81:
I've added an extra wait but it still doesn't work. I wonder if I should also add some "waiting for previews" before all possible file operations :-/

PVince81:
Plan:

- [ ] make preview loading detection more reliable: add data attribute with loading state and adjust the waiting code to wait for that attribute (on file scroll, etc)
- [ ] if not enough, add said waiting code in more locations

PVince81:
It turns out I didn't wrap fetch properly, and this has caused other issues. Fixed now.

I've now added an attribute "data-preview-loaded=true" whenever the preview has finished loading, even in fallback situations. This way the test code knows that it can move forward with the testing once the preview loading attempt has finished.

I've kept the ajax waiting in various places as it might be useful anyway.

Let's see if the tests pass now...

PVince81:
forgot to mention: I also tweaked mediasource to properly detect error codes. Whenever a status code 404 comes due to missing preview or unsupported preview, it doesn't go into the "catch()" promise block but goes to "then()" instead, where I needed to check `response.ok`.

PVince81:
@LukasHirt please recheck the frontend code.

@dpakach would be good if you could review the test code as I had to change many core stuff there

PVince81:
getting closer... now fixed an issue where we were waiting for the thumbnails of folders, but there is none. skipping that case now.

apps/files/src/components/AllFilesList.vue
```diff
@@ -48,9 +48,13 @@
...
-        <oc-file @click.native.stop="item.type === 'folder' ? navigateTo(item.path.substr(1)) : openFileActionBar(item)"
-          :name="$_ocFileName(item)" :extension="item.extension" class="file-row-name" :icon="fileTypeIcon(item)"
-          :filename="item.name" :key="item.id"/>
+        <file-item
+          @click.native.stop="item.type === 'folder' ? navigateTo(item.path.substr(1)) : openFileActionBar(item)"
+          :item="item"
+          :dav-url="davUrl"
+          :show-path="$_isFavoritesList"
+          class="file-row-name"
+          :key="item.id"/>
```
LukasHirt: Would it make sense to use here the `viewId` from https://github.com/owncloud/phoenix/pull/3248 ? Just for consistency.

PVince81:
another fix: set `previewLoaded` to true even for folders, and also when using the cached preview URL (happens when scrolling up and down repeatedly after having already loaded the preview)

apps/files/src/components/FileItem.vue
```diff
@@ -0,0 +1,98 @@
...
+      // TODO: check if previews are globally enabled (requires capability entry)
+      // don't load previews for pending or rejected shares (status)
+      if (!this.davUrl || this.item.type === 'folder' || (typeof this.item.status !== 'undefined' && this.item.status !== 0)) {
+        this.previewLoaded = true
+        return
+      }
+      const query = queryString.stringify({
+        x: this.thumbDimensions,
+        y: this.thumbDimensions,
+        c: this.item.etag,
```
C0rby: I think after `stringify` the etag will have `"` around it. They should be remove if we want to be equal to the OC10 url

PVince81: Nice one!

I just first looked into stripping the double quote at the source, but realized that it would affect other code paths that would require adding back the double quotes, so decided against a global solution.

I'll then push a fix that removes the double quotes locally, also applied to the media viewer code that has a similar logic.

apps/files/src/components/AllFilesList.vue
PVince81: done

PVince81:
I've added more checks, now a generic "waitForAllThumbnails" call for after the list is loaded.

Still, it feels that it's not enough for a proper coverage, because often times the scenarios only tell that the user logged in, but not explicitly says that the user navigated to the files page, so the logic for waiting for the progress bar and thumbnails is likely not always kicking in. I'd leave this for another PR as it would require touching a lot of scenarios. Thoughts ? @individual-it @dpakach

PVince81:
Seems that `waitForAllThumbnailsLoaded` is not working:
image

The code is in the right place: https://github.com/owncloud/phoenix/blob/filelist-previews/tests/acceptance/stepDefinitions/filesContext.js#L114

I've now added an extra 1 second delay after waiting, just in case the file locks need more time to release themselves...

PVince81:
now debugging waitForAllThumbnailsLoaded, seems the selector is not working because the attribute is missing when "previewLoaded" is false ?!

PVince81:
okay, seems Vue.js doesn't render HTML attributes if their value is boolean false. Had to set it to the string `"false"`.
also had to fix the selector for `waitForAllThumbnailsLoaded` to use xpath.

let's see...

PVince81:
I had an explicit check for a file entry in the file list with "upload overwrite" but doesn't cover all cases. We anyway don't need it any more due to the global waiting. So I've added "user browses to files page" before all the overwrite cases.

Squashed

ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUIOCIS</strong> failed. Please find the screenshots inside ...</summary>

https://drone.owncloud.com/owncloud/phoenix/9034/

<p>





</p></details>

PVince81:
added a few more fixes.
the trashbin name was not rendered properly as trashbin must display the original path as prefix

ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUIOCIS</strong> failed. Please find the screenshots inside ...</summary>

https://drone.owncloud.com/owncloud/phoenix/9036/

<p>





</p></details>

PVince81:
almost all green, I've pushed one last fix.

the OCIS tests fail, especially the ones related to thumbnail. Not sure if the thumbnailer service is already in place ? if not, we should skip them. do you mind having a look @dpakach ?

ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUIOCIS</strong> failed. Please find the screenshots inside ...</summary>

https://drone.owncloud.com/owncloud/phoenix/9037/

<p>





</p></details>

dpakach:
@PVince81, I think we can skip the tests on ocis until we properly implement thumbnail service in CI.

tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature
```diff
@@ -69,6 +69,7 @@ Feature: Sharing files and folders with internal groups
     When the user renames file "lorem.txt" to "new-lorem.txt" using the webUI
     And the user shares file "new-lorem.txt" with group "grp1" as "Editor" using the webUI
     And the user re-logs in as "user1" using the webUI
+    And the user browses to the files page
```
dpakach: can we call `waitTillThumbnailsLoaded` function in the `user has logged in` or `user has re-logged in` step. 
That way we can avoid having `user has browsed to the files page` everywhere after logging in.

dpakach:
The CI has passed but I restarted just to make sure it is stable

apps/files/src/components/FileItem.vue
```diff
@@ -0,0 +1,111 @@
...
+        y: this.thumbDimensions,
+        scalingup: 0,
+        preview: 1,
+        a: 1
+      }
+
+      if (this.item.etag) {
+        // add etag for URL based caching
+        // strip double quotes from etag
+        query.c = this.item.etag.substr(2, this.item.etag.length - 2)
```
C0rby: @LukasHirt, oh no here is a bug..
the `substr` should be `substr(1, this.item.etag.length - 2)` or else it will cut of the first character. :/

LukasHirt: Thanks for catching that! I'll open a PR with fix and ping you for review :wink:

PR #3202, opened 2020-03-19
[NOMERGE] Test if CI also fails randomly from master

PVince81:
CI fails randomly on https://github.com/owncloud/phoenix/pull/3187

This PR here is just to test if it also fails randomly on master.


PVince81:
hmmm, all green here... so there must be something in my PR...

PR #3239, opened 2020-03-22
Setup prettier in eslint

kulmann:
we might want to reduce the number of open PRs and then rebase before merging, as this affects all js and vue files.

## Description
This PR sets up the `prettier` plugin with `recommended` ruleset for eslint. As we already have lint fix running as commit hook, this automatically gets included now for files that were changed in the commit.

Applied `lint --fix` once so that we have everything with recommended code styles.

Would suggest to rebase this PR when the current sprint is done (so that the number of open PRs is smaller) and re-run the lint fixing again before merging.

## Related Issue
- Fixes https://github.com/owncloud/phoenix/issues/2992

## Motivation and Context
Prettier provides an out of the box awesome code style.

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,23 +1,25 @@
 {
-    "env": {
-        "browser": true,
-        "es6": true,
-        "amd": true
-    },
-    "extends": [
-        "standard",
-        "plugin:vue/essential"
-        ],
-    "parserOptions": {
-        "parser": "babel-eslint",
-        "sourceType": "module"
-    },
-    "rules": {
-      "require-await": "warn",
-      "vue/order-in-components": "error"
-    },
-    "globals": {
-      "require": false,
-      "requirejs": false
-    }
+  "env": {
+    "browser": true,
+    "es6": true,
+    "amd": true
+  },
+  "extends": [
+    "standard",
+    "plugin:vue/recommended",
+    "plugin:prettier/recommended",
+    "prettier/standard",
+    "prettier/vue"
+  ],
+  "parserOptions": {
+    "parser": "babel-eslint",
+    "sourceType": "module"
+  },
+  "rules": {
+    "require-await": "warn"
+  },
+  "globals": {
+    "require": false,
+    "requirejs": false
+  }
 }
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+  "printWidth": 100,
+  "semi": false,
+  "singleQuote": true
+}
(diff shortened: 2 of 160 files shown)


kulmann:
Open questions @LukasHirt @PVince81 
- do we want to stick with the `no semicolons` rule? Then I can add it to the prettier options
- do we want to have the number of columns as set by the default (80) or longer lines? I personally would go with something around 100-120, but have no strong preference.

LukasHirt:
> do we want to stick with the no semicolons rule? Then I can add it to the prettier options

I'd vote for still having none

> do we want to have the number of columns as set by the default (80) or longer lines?

100 LGTM

I'd also vote for one additional rule - `vueIndentScriptAndStyle: true`

.eslintrc.json
```diff
@@ -6,7 +6,8 @@
     },
     "extends": [
         "standard",
-        "plugin:vue/essential"
+        "plugin:vue/essential",
```
LukasHirt: We might also want to consider switching here to recommended

LukasHirt:
Any reason for switching to double quotes? If not, I'd prefer to keep single ones.

kulmann:
@LukasHirt last commit is - again - applying the rules to the current code base. I'd say that we rebase to current master on friday, drop the last commit and re-run the `lint --fix` step, so that the impact is minimal.

Important: the rule `vueIndentScriptAndStyle: true` doesn't work in eslint and I didn't find a reason why. Do you have an idea? Thought it could be an issue with the indentation rules in `.editorconfig`, but removing them didn't help. 🤔

LukasHirt:
> I'd say that we rebase to current master on friday, drop the last commit and re-run the lint --fix step, so that the impact is minimal.

Sounds good 👍 

> the rule vueIndentScriptAndStyle: true doesn't work in eslint and I didn't find a reason why

You're right. I tested it and had the same result. Strange is that most issues mentioning this are tied to IDEs and their extensions which I think wouldn't however explain why running `yarn lint` doesn't detect the missing indentation. Let's not bother then with this rule. I don't think it's worth the effort to try and debug it.

ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUIOCIS</strong> failed. Please find the screenshots inside ...</summary>

https://drone.owncloud.com/owncloud/phoenix/9387/

<p>























































































































</p></details>

PR #3242, opened 2020-03-23
[Tests-only] Use ocis-phoenix for running tests in drone

dpakach:
## Description

In CI right now we use the ocis binary to run phoenix service for acceptance tests. This PR changes that to ocis-phoenix binary so that we can catch error even earlier.

## Related Issue

- Fixes <issue_link>

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -79,7 +79,7 @@ config = {
 				'all': 'webUIOCIS'
 			},
 			'extraEnvironment': {
-				'SERVER_HOST': 'http://ocis:9100',
+				'SERVER_HOST': 'http://phoenix:9100',
 				'BACKEND_HOST': 'http://reva:9140',
 				'RUN_ON_OCIS': 'true',
 				'OCIS_SKELETON_DIR': '/var/www/owncloud/server/apps/testing/data/webUISkeleton',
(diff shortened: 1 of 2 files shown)


.drone.star
```diff
@@ -1046,8 +1047,7 @@ def ocisServices():
 		},
 		'commands': [
 			'cd /var/www/owncloud',
-			'./ocis devldap &',
```
kiranparajuli589: should that be deleted too?

.drone.star
```diff
@@ -89,6 +89,7 @@ config = {
 			},
 			'runningOnOCIS': True,
 			'filterTags': 'not @skip and not @skipOnOCIS',
+			'screenShots': True,
```
individual-it: debug or do we need the screenshots for longterm?

.drone.star
dpakach: yes, that should not have been there in the first place.

.drone.star
dpakach: I think we can leave it for now, will be helpful as soon as tests fail in ocis.

PR #3248, opened 2020-03-23
Use alternative temporary ids for uniqueness in views

PVince81:
## Description
When using OCIS as backend, the ids of resources is a string instead of
integer. So we cannot embed those into DOM node ids and need to use
another alternative. This fix introduces a unique listId which is only
there to provide uniqueness across the current list and should not be
used for any data related operation.

This fixes the file actions menu when using OCIS as backend.

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/3214
Fixes https://github.com/owncloud/ocis-phoenix/issues/51

## How Has This Been Tested?
Manual test with OCIS backend (see https://github.com/owncloud/phoenix/pull/3212 for setup)
Manual test with OC 10 backend just in case, also to test the "stime" sorting change which can currently only be tested with OC 10.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -240,7 +240,7 @@ export default {
           this.$nextTick(() => {
             const file = this.activeFiles.find(item => item.name === scrollTo)
             this.setHighlightedFile(file)
-            this.$scrollTo(`#file-row-${file.id}`, 500, {
+            this.$scrollTo(`#file-row-${file.viewId}`, 500, {
               container: '#files-list-container'
             })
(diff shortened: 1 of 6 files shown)


dpakach:
@PVince81 I just rebased this PR since the OCIS tests are failing and those have been skipped for now.

PVince81:
@kulmann you're right, I was also not comfortable with the name. I've renamed all listId to viewId now.

PR #3274, opened 2020-03-27
Do not remove first char in etag

LukasHirt:
## Description
Follow up on #3187. The first char of etag was being removed when stripping away double quotes.

### DIFF
--- a/apps/files/src/components/FileItem.vue
+++ b/apps/files/src/components/FileItem.vue
@@ -87,7 +87,7 @@ export default {
       if (this.item.etag) {
         // add etag for URL based caching
         // strip double quotes from etag
-        query.c = this.item.etag.substr(2, this.item.etag.length - 2)
+        query.c = this.item.etag.substr(1, this.item.etag.length - 2)
       }
 
       let itemPath = this.item.path
--- a/changelog/unreleased/3274
+++ b/changelog/unreleased/3274
@@ -1,0 +1,6 @@
+Bugfix: Do not remove first character of etag
+
+When stripping away double quotes in etag of the file thumbnails, we accidentally
+removed first character as well. We've stopped removing that character.
+
+https://github.com/owncloud/phoenix/pull/3274


C0rby:
> Why stripping anything at all? the double quotes are an integral part of the etag.
> They should never be removed!

The etags in the oc10 requests do not contain quotes. This request should be compatible to the oc10 request.

DeepDiver1975:
> The etags in the oc10 requests do not contain quotes. This request should be compatible to the oc10 request.

They do for sure - unless anybody did break anything in oc10 the past months :see_no_evil:

C0rby:
> > The etags in the oc10 requests do not contain quotes. This request should be compatible to the oc10 request.
> 
> They do for sure - unless anybody did break anything in oc10 the past months

Here is a preview request from oc10.
http://localhost/owncloud/remote.php/dav/files/admin/Photos/Paris.jpg?c=00356a07fe238448791f9c2526d546a1&x=32&y=32&forceIcon=0&preview=1

DeepDiver1975:
> Here is a preview request from oc10.
> http://localhost/owncloud/remote.php/dav/files/admin/Photos/Paris.jpg?c=00356a07fe238448791f9c2526d546a1&x=32&y=32&forceIcon=0&preview=1

Wow - this is so wrong .... anyway since you are rebuilding oc10 apis :+1:

DeepDiver1975:
FYI: on the PROPFIND response there are double quotes ...

LukasHirt:
> the double quotes are an integral part of the etag.

Why is it like that? @DeepDiver1975

LukasHirt:
@C0rby @DeepDiver1975 pls let me know about decision if keep/remove. I'd like to have this in today's release so we don't release previews with a bug 🙂

DeepDiver1975:
> Why is it like that? @DeepDiver1975

the rfc says so :man_shrugging: 

anyway - this is 'only' on the preview url parameter so I don't care much

PR #3282, opened 2020-04-01
[Tests-Only] Test for fix share expiration date changes

Talank:
## Description
test has been modified to cope with the changes made in share expiration date changes

## Motivation and Context
Changes in the code was causing the CI to fail since previously noticed error message do not appear anymore. So, test scenario has been modified accordingly.

## How Has This Been Tested?
locally

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/features/webUISharingPublic/shareByPublicLink.feature
+++ b/tests/acceptance/features/webUISharingPublic/shareByPublicLink.feature
@@ -412,14 +412,20 @@ Feature: Share by public link
       | lorem.txt       |
       | simple-folder   |
 
-  Scenario: user cannot set an expiry date when creating a public link to a date that is past the enforced max expiry date
+  Scenario: expiry date is set to enforced max expiry date when creating a public link to a date that is past the enforced max expiry date
     Given the setting "shareapi_default_expire_date" of app "core" has been set to "yes"
     And the setting "shareapi_enforce_expire_date" of app "core" has been set to "yes"
     And user "user1" has logged in using the webUI
     When the user tries to create a new public link for resource "simple-folder" using the webUI with
-      | expireDate | +8 |
-    Then the user should see an error message on the public link share dialog saying "Cannot set expiration date more than 7 days in the future"
-    And user "user1" should not have created any shares
+      | expireDate | +15 |
+    Then user "user1" should have a share with these details:
+      | field       | value          |
+      | share_type  | public_link    |
+      | uid_owner   | user1          |
+      | permissions | read           |
+      | path        | /simple-folder |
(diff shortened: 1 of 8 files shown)


tests/acceptance/features/webUISharingPublic/shareByPublicLink.feature
```diff
@@ -412,14 +412,20 @@ Feature: Share by public link
...
-    And user "user1" should not have created any shares
+      | expireDate | +15 |
+    Then user "user1" should have a share with these details:
+      | field       | value          |
+      | share_type  | public_link    |
+      | uid_owner   | user1          |
+      | permissions | read           |
+      | path        | /simple-folder |
+      | name        | Public link    |
+      | expiration  | +7             |
```
individual-it, line 428: that does not sound logical. I'm suggesting rather do something like:
```
    When the user tries to create a new public link for resource "simple-folder" using the webUI with
      | expireDate | +8 |
    Then the expiration date shown on the webUI should be "+7" days
    And user "user1" should not have created any shares
```
similar like in https://github.com/owncloud/phoenix/blob/c723f814afe9d4ffde7bf4d7045c948c90816664/tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature#L390
the problem is, that there is no difference currently between `When the user tries to create a new public link` and `When the user creates a new public link`, but there should be.
Better to implement it like in https://github.com/owncloud/phoenix/blob/d65297a63b8255fd7dccdd44840d184e1abe671b/tests/acceptance/stepDefinitions/sharingContext.js#L1024
where the underlining function does not save the share if the date could not have been selected see:
https://github.com/owncloud/phoenix/blob/6a5800edbd17883c83c07d3f9f42724abbb86415/tests/acceptance/pageObjects/FilesPageElement/sharingDialog.js#L151

Talank:
```
Then the expiration date shown on the webUI should be "+7" days
And user "user1" should not have created any shares
```
Here the first line checks if the expected date `+7` matches the date in the `input field for date` that appears while user makes new share or edit existing share. I find it to be useful while creating share which is not saved yet. But, in our case, the share can be created (so 2nd line fails as well) and we have to check the expiry date of the saved share.

individual-it: the `tries` function should not try to save when the date could not be set, see the function for group shares

PR #3288, opened 2020-04-02
Fix media-viewer on private pages

dschmidt:
## Description
I've made the mediaviewer app depend only on this.isAuthenticated instead of this.publicPage().

publicPage() is always true because (at least I assume so) it's rendered in its own route that is always public. In my tests I could use isAuthenticated to determine whether it was loaded from a private page or from a public link e.g.

It feels hackish, happy to hear if there's a better API to use.

## Related Issue

- Fixes https://github.com/owncloud/phoenix/issues/3287

## How Has This Been Tested?

- test environment:
- test case 1: I looked at the picture from regular files view
- test case 2: I looked at the file from public link view
- ...

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #3294, opened 2020-04-03
mediaviewer: add state to urls

dschmidt:
Hey everyone,

I'm sorry this PR addresses a bunch of separate issues/topics at once.
I wanted to figure out what can be done and what is possible in which scope. I would like to discuss the overall approach here, if desired I can send more fine grained PRs for individual aspects.

Commits can obviously still be cleaned up a bit :) 
Loader is also obviously a stupid name for the mixin, any better ideas?
I specifically tried to encapsulate all folder loading logic there so it can be reused in other apps, what do you think about that? Where could we put that? 

This includes the commit from #3288.

## Description
This PR adds state to app routes (i.e. the context from which they were opened and the file they open). They are accessible as `contextRouteName` and `filePath`.

Moreover it adds a local history to the mediaviewer app, so forward and back buttons work and we can smootly animate between different pictures without transitioning between routes (and to keep the changes somewhat limited in this regard).

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Apps opened from the files view did not contain state in the url, so F5 is broken. We should either not route to apps or make the urls useful (which I think is a pretty good idea for webapps anyway).

## How Has This Been Tested?
I tested local files, favorites and public links.
Navigating to a folder, opening pictures, clicking prev and next, using the forward and back buttons in the browser works as expected, I could not break it or find any behavior that I did not expect... 

shares don't allow me to open apps, so not tested.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] Loading the mediaviewer directly for a public link didn't work, even without a password needed ... I don't know why, I couldn't make loadFolder work.

### DIFF
--- a/apps/media-viewer/src/Mediaviewer.vue
+++ b/apps/media-viewer/src/Mediaviewer.vue
@@ -5,7 +5,7 @@
         <img v-show="!loading && activeMediaFileCached" :src="image.url" :alt="image.name" :data-id="image.id" style="max-width:90vw;max-height:90vh" class="uk-box-shadow-medium">
       </transition>
     </div>
-    <oc-spinner class="uk-position-center" v-if="loading" size="large" />
+    <oc-spinner :ariaLabel="this.$gettext('Loading media')" class="uk-position-center" v-if="loading" size="large" />
     <oc-icon v-if="failed" name="review" variation="danger" size="large" class="uk-position-center uk-z-index" />
 
     <div class="uk-position-medium uk-position-bottom-center">
@@ -14,7 +14,7 @@
         <div class="uk-width-large uk-flex uk-flex-middle uk-flex-center uk-flex-around" style="user-select:none;">
           <oc-icon role="button" class="oc-cursor-pointer" size="medium" @click="prev" name="chevron_left" />
           <!-- @TODO: Bring back working uk-light -->
-          <span class="uk-text-small" style="color:#fff"> {{ activeIndex + 1 }} <span v-translate>of</span> {{ mediaFiles.length }} </span>
+          <span v-if="!$_loader_folderLoading" class="uk-text-small" style="color:#fff"> {{ activeIndex + 1 }} <span v-translate>of</span> {{ mediaFiles.length }} </span>
           <oc-icon role="button" class="oc-cursor-pointer" size="medium" @click="next" name="chevron_right" />
           <oc-icon role="button" class="oc-cursor-pointer" @click="downloadImage" name="file_download"  />
           <oc-icon role="button" class="oc-cursor-pointer" @click="closeApp" name="close"/>
@@ -25,10 +25,13 @@
 </template>
 <script>
 import { mapGetters } from 'vuex'
-import queryString from 'query-string'
+import Loader from './mixins/loader.js'
 
 export default {
   name: 'Mediaviewer',
+  mixins: [
+    Loader
+  ],
   data () {
     return {
       loading: true,
@@ -46,7 +49,7 @@ export default {
   },
 
   computed: {
-    ...mapGetters('Files', ['activeFiles', 'publicLinkPassword']),
+    ...mapGetters('Files', ['activeFiles']),
     ...mapGetters(['getToken']),
 
     mediaFiles () {
@@ -80,22 +83,8 @@ export default {
         default: return 3840
       }
     },
-    headers () {
-      if (!this.isPublicContext()) {
-        return null
-      }
-
-      const headers = new Headers()
-      headers.append('X-Requested-With', 'XMLHttpRequest')
-
-      const password = this.publicLinkPassword
-      if (password) {
-        headers.append('Authorization', 'Basic ' + Buffer.from('public:' + password).toString('base64'))
-      }
-      return headers
-    },
     thumbPath () {
-      const query = queryString.stringify({
+      const query = {
         x: this.thumbDimensions,
         y: this.thumbDimensions,
         // strip double quotes from etag
@@ -103,27 +92,9 @@ export default {
         scalingup: 0,
         preview: 1,
         a: 1
-      })
-
-      if (this.isPublicContext()) {
-        const path = [
-          '..',
-          'dav',
-          'public-files',
-          this.activeMediaFile.path
-        ].join('/')
-
-        return this.$client.files.getFileUrl(path) + '?' + query
       }
-      const path = [
-        '..',
-        'dav',
-        'files',
-        this.$store.getters.user.id,
-        this.activeMediaFile.path
-      ].join('/')
 
-      return this.$client.files.getFileUrl(path) + '?' + query
+      return this.$_loader_getDavFilePath(this.activeMediaFile.path, query)
     }
   },
 
@@ -135,28 +106,21 @@ export default {
     }
   },
 
-  mounted () {
+  async mounted () {
     document.addEventListener('keyup', this.handleKeyPress)
 
-    // Return if no Image is selected
-    if (this.$store.getters.activeFile.path === '') {
-      this.$router.push({
-        path: '/files'
-      })
-      return
-    }
+    // keep a local history for this component
+    window.addEventListener('popstate', this.handleLocalHistoryEvent)
 
-    // Set initial file
-    for (let i = 0; i < this.mediaFiles.length; i++) {
-      if (this.mediaFiles[i].path === this.$store.getters.activeFile.path) {
-        this.activeIndex = i
-        break
-      }
-    }
+    const filePath = this.$route.params.filePath
+    await this.$_loader_loadFolder(this.$route.params.contextRouteName, filePath)
+    this.setCurrentFile(filePath)
   },
 
   beforeDestroy () {
     document.removeEventListener('keyup', this.handleKeyPress)
+
+    window.removeEventListener('popstate', this.handleLocalHistoryEvent)
 
     this.images.forEach(image => {
       window.URL.revokeObjectURL(image.url)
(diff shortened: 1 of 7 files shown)


dschmidt:
it would also be nice to scroll to the last opened file when closing, this comment hints their might be code to achieve that already: https://github.com/owncloud/phoenix/issues/3184#issuecomment-599453811

Can anyone point me to it?

apps/media-viewer/src/Mediaviewer.vue
```diff
@@ -24,11 +24,143 @@
...
+      headers.append('X-Requested-With', 'XMLHttpRequest')
+
+      const password = this.publicLinkPassword
+      if (password) {
+        headers.append('Authorization', 'Basic ' + Buffer.from('public:' + password).toString('base64'))
+      }
+      return headers
+    },
+
+    $_loader_navigateToContextRoute (contextRouteName, filePath) {
```
kulmann, line 146: For scrolling to the current file when going back to the files app you can do this:
`import { basename, dirname } from 'path'` and then
```
this.$router.push({
  name: contextRouteName,
  params: {
    item: dirname(filePath) || '/'
  },
  query: {
    scrollTo: basename(filePath)
  }
})
```
as the method body of `$_loader_navigateToContextRoute` (remove yours)

LukasHirt: Can you pls check if the scrolling works when the item is deep down in the list? There might be an issue with this related to https://github.com/owncloud/phoenix/issues/3265

dschmidt: Yes I can. But fixing any of that is way out of scope for this PR. And I'd hope the API would stay stable no matter if theres virtual scrolling or pagination or whatever in place.

LukasHirt: Sure, just wanted to be safe and know how it behaves here. Absolutely no need for a fix here. And I agree that the API should stay stable. :+1:

LukasHirt:
> I specifically tried to encapsulate all folder loading logic there so it can be reused in other apps, what do you think about that? Where could we put that?

Since we can't reach mixins from different apps, one option could be to bring this to store. That is however not such a nice solution. If you feel like this would be something really used often we could move it into the [owncloud-sdk](https://github.com/owncloud/owncloud-sdk/) or as a core plugin.

apps/media-viewer/src/Mediaviewer.vue
```diff
@@ -14,7 +14,7 @@
         <div class="uk-width-large uk-flex uk-flex-middle uk-flex-center uk-flex-around" style="user-select:none;">
           <oc-icon role="button" class="oc-cursor-pointer" size="medium" @click="prev" name="chevron_left" />
           <!-- @TODO: Bring back working uk-light -->
-          <span class="uk-text-small" style="color:#fff"> {{ activeIndex + 1 }} <span v-translate>of</span> {{ mediaFiles.length }} </span>
+          <span v-if="!folderLoading" class="uk-text-small" style="color:#fff"> {{ activeIndex + 1 }} <span v-translate>of</span> {{ mediaFiles.length }} </span>
```
LukasHirt, line 17: You can add the class `uk-light` which will take care of the text colour.

apps/media-viewer/src/Mediaviewer.vue
```diff
@@ -24,11 +24,143 @@
...
+  data () {
+    return {
+      folderLoading: true
+    }
+  },
+
+  methods: {
+    ...mapActions('Files', ['loadFolder']),
+
+    $_loader_publicContext () {
```
LukasHirt, line 44: I think this might be better of as a computed prop.

apps/media-viewer/src/Mediaviewer.vue
dschmidt: Alright, I tested it - I could not make it scroll anywhere :joy: 
What works tho: the file is selected, which is nice for starters.

dschmidt:
> > I specifically tried to encapsulate all folder loading logic there so it can be reused in other apps, what do you think about that? Where could we put that?
> 
> Since we can't reach mixins from different apps, one option could be to bring this to store. That is however not such a nice solution. If you feel like this would be something really used often we could move it into the [owncloud-sdk](https://github.com/owncloud/owncloud-sdk/) or as a core plugin.

I'm not so familiar with how you are doing things in phoenix, but yes, I think this is something that can and probably will be reused over and over again.

Basically anything that deals with actual files and folders will need at least parts of it. MediaViewer, PDF Viewer or just look at AllFilesList.vue. davUrl, $_ocFilesFolder_getFolder. It could use my mixin instead by passing in this.$route.name as contextRouteName.
I think the code should be shared between regular folder views and apps. No idea if this kind of mixin is the best way, if it should live in core or the sdk or if the sdk should otherwise be fixed to be usable more easily, but this PR paves the way for refactoring it further by at least separating data loading from display logic.

apps/media-viewer/src/Mediaviewer.vue
dschmidt: The line above says: 
```html
<!-- @TODO: Bring back working uk-light -->
```
and indeed I don't see light text and it's harder readable than without the manual style. So I'll leave it as it is for now.

dschmidt:
Hope I addressed all of the raised suggestions :)

apps/media-viewer/src/Mediaviewer.vue
```diff
@@ -14,7 +14,7 @@
         <div class="uk-width-large uk-flex uk-flex-middle uk-flex-center uk-flex-around" style="user-select:none;">
           <oc-icon role="button" class="oc-cursor-pointer" size="medium" @click="prev" name="chevron_left" />
           <!-- @TODO: Bring back working uk-light -->
-          <span class="uk-text-small" style="color:#fff"> {{ activeIndex + 1 }} <span v-translate>of</span> {{ mediaFiles.length }} </span>
+          <span v-if="!$data.$_loader_folderLoading" class="uk-text-small" style="color:#fff"> {{ activeIndex + 1 }} <span v-translate>of</span> {{ mediaFiles.length }} </span>
```
LukasHirt, line 17: This should be accessible without `$data`

dschmidt: It's not because it's prefixed with $ (would be the same for only _). Gave me a vue error that I needed to use $data.

LukasHirt:
LGTM. If @kulmann is also happy with changes I'd say let's merge it :wink:

apps/media-viewer/src/mixins/loader.js
```diff
@@ -0,0 +1,138 @@
...
+      $_loader_folderLoading: true
+    }
+  },
+
+  methods: {
+    ...mapActions('Files', ['loadFolder']),
+
+    // This methods ensures the folder is loaded if we don't have a folder loaded currently
+    async $_loader_loadFolder (contextRouteName, filePath) {
+      console.log('$_loader_loadFolder', contextRouteName, filePath)
```
kulmann, line 33: please remove ;-)

apps/media-viewer/src/mixins/loader.js
```diff
@@ -0,0 +1,138 @@
...
+
+      // TODO: support public-files as well
+      if (this.$store.getters.activeFile.path === '' && contextRouteName === 'public-files') {
+        console.warn('TODO: handle public-files with passwords and everything, until then we redirect to the main public link page')
```
kulmann, line 37: please remove `console.warn` and write a `FIXME` note instead

apps/media-viewer/src/mixins/loader.js
```diff
@@ -0,0 +1,138 @@
...
+        const path = this.$route.params.filePath.substring(1)
+        const token = path.substr(0, path.indexOf('/'))
+        this.$nextTick(() => {
+          console.log('wat')
```
kulmann, line 41: please remove ;-)

apps/media-viewer/src/mixins/loader.js
```diff
@@ -0,0 +1,138 @@
...
+      // Currently it works. We cannot use publicPage() because that will still return
+      // true when opening the mediaviewer from authenticated routes
+      return !this.isAuthenticated
+    }
+
+  },
+
+  data () {
+    return {
+      $_loader_folderLoading: true
```
kulmann, line 24: I don't really like that we have data in a mixin. But since there's a ticket from @LukasHirt now that reminds us to solve this in a better way, I'm ok with merging this anyway.

kulmann: ** using the data from the mixin directly.

dschmidt: It's especially bad if the consumer needs to access this with `$data`. I assume it'd be a breaking api change to make it a computed property then. Better to make it a computed property from the started. Added an additional `_internal_` infix to the data member to make it obvious it's not meant to be used from outside the mixin itself.

dschmidt: Hehehe, I wrote my comment without reloading - so I didn't see your second comment. 
Yep, full ack.

kulmann: Awesome, thanks! 👍

kulmann:
Are you fine with squashing this PR into one commit before the merge?

dschmidt:
If that's what you do in this project :shrug:

... usually I like having a proper history, but as there's quite a bit of back and forth here, I can squash, sure.

kulmann:
> If that's what you do in this project :shrug:
> 
> ... usually I like having a proper history, but as there's quite a bit of back and forth here, I can squash, sure.

Yes, I'm all for a clean history as well, but I don't think that it's worth the effort to do this for your 26 commits. ;-) And having 1 is nicer than having the 26. 😁

dschmidt:
Yeah, well ... squashed :)

PR #3420, opened 2020-04-30
Fix share indicators click to open the correct panel

PVince81:
## Description
When clicking on a share indicator inside a file list row, the correct share panel will now be displayed.

## Related Issue
Fixes https://github.com/owncloud/phoenix/issues/3324

## How Has This Been Tested?
Manual test

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/components/AllFilesList.vue
+++ b/apps/files/src/components/AllFilesList.vue
@@ -91,11 +91,7 @@
         />
       </div>
       <div v-if="!$_isFavoritesList" class="uk-flex uk-flex-middle uk-flex-right">
-        <Indicators
-          :default-indicators="indicatorArray(rowItem)"
-          :item="rowItem"
-          @click="$_openSideBar"
-        />
+        <Indicators :default-indicators="indicatorArray(rowItem)" :item="rowItem" />
       </div>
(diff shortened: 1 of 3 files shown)


LukasHirt:
@PVince81 Would it make sense to cover this with tests? To prevent the same error again in the future?

PVince81:
@LukasHirt I started with it and was wondering how to formulate the test.
In general we don't want to talk about clicks, but more about "the user opens the panel" or even not mention any UI elements. Since this is a shortcut, not sure.

@individual-it any thoughts ?

LukasHirt:
> I started with it and was wondering how to formulate the test.
In general we don't want to talk about clicks, but more about "the user opens the panel" or even not mention any UI elements. Since this is a shortcut, not sure.

I guess we might try something like "the user opens the collaborators panel via its indicator". But yeah, I know what you mean. If you want we can also merge this and discuss/implement the tests separately. Not to block this bugfix. Your call 😉

PR #3436, opened 2020-05-05
Add TUS creation-with-upload support

PVince81:
## Description
Whenever the server advertised the capability for creation-with-upload,
use it automatically for chunked uploads.

It is possible to override this and disable it by setting
uploadChunkDirect=false in config.json

## Related Issue
None

## How Has This Been Tested?
Manual test with latest reva

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] debug ocis/reva issue
- [ ] add changelog

### DIFF
--- a/apps/files/src/mixins.js
+++ b/apps/files/src/mixins.js
@@ -446,7 +446,16 @@ export default {
         relativePath = pathUtil.join(basePath, relativePath)
         // FIXME: this might break if relativePath is not the currentFolder
         // and is a mount point that has no chunk support
-        if (this.browserSupportsChunked && this.currentFolder.isChunkedUploadSupported) {
+        if (this.browserSupportsChunked && this.currentFolder.chunkUploadSupport) {
+          const extraOptions = {}
+          // enable direct upload if enabled and supported
+          // instead of create + upload in two requests
+          if (
+            this.configuration.uploadChunkDirect &&
+            this.currentFolder.chunkUploadSupport.extension.includes('creation-with-upload')
+          ) {
+            extraOptions.uploadDataDuringCreation = true
+          }
           let chunkSize = this.configuration.uploadChunkSize
           if (this.capabilities.files.tus_support.max_chunk_size > 0) {
             if (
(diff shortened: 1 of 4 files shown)


PVince81:
@jfd this is not working and returns a 400 error: https://github.com/cs3org/reva/issues/710

Also I noticed that the tus-js-client is not sending the header `Expect: 100-continue` to receive early feedback. Not sure if we should hack it in by passing that header ourselves or expect the lib to do so in the future.

- [ ] investigate possibility for expect header

PVince81:
Upload works with this Reva PR: https://github.com/cs3org/reva/pull/725

Now I found out that [tus-js-client](https://github.com/tus/tus-js-client#installation) is not using a fingerprint for uploads that use "creation-with-upload", which means those will not be resumable if the initial POST failed. It will not be resumable because the client has no way of keeping track of this upload so it will not know where to send a HEAD request if the server already had data.

In the light of this, I think we should not merge this PR until the library is fixed to start tracking uploads earlier. The doc does say that this feature is currently experimental.

Now thinking, the library would not be able to track the download before it gets the "Location" header from the response. This means that the first POST MUST go through. I'm not sure if the "100-Continue", once implemented, would be able to return that header earlier before the payload is sent. A workaround for this would be to purposefully send a smaller chunk for the initial upload to increase the chance of it going through and quickly get access to the "Location" header.

@butonic thoughts ?

PVince81:
As far as I could read, we cannot send additional headers from the server when sending the "100-continue" status code.

So the only way would be to decide to send a smaller chunk first to avoid wasting a request and then continuing with the rest in a subsequent request.

kulmann:
Version 2 of the `tus-js-client` has a lot of changes regarding the upload tracking / fingerprints. Worth to give this another try after a lib update.

kulmann:
Thanks for the ideas. We'll have a look at the new tus-js-client 2.3.0 soonish. Added this PR for reference in [this ticket](https://github.com/owncloud/web/issues/5371), so that we can give it another try after the upgrade.

PR #3459, opened 2020-05-13
[Test-only] Create selenium command with port forwarding for mac

LukasHirt:
## Motivation and Context
`--network=host` option does not work in Mac https://docs.docker.com/network/host/


individual-it:
could you also add some lines to the docs https://github.com/owncloud/phoenix/blob/master/docs/testing.md

LukasHirt:
> could you also add some lines to the docs

Done @individual-it

PR #3460, opened 2020-05-13
Search bar in top bar

LukasHirt:
## Description
Moved search bar from app bar into top bar

host docker internal_8300_index html(iPad Pro)

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -1,0 +1,109 @@
+<template>
+  <div>
+    <oc-search-bar
+      ref="globalSearchBar"
+      class="uk-width-large uk-visible@m"
+      :label="searchLabel"
+      :loading="isLoadingSearchResults"
+      :button-hidden="true"
+      @search="search"
+      @clear="resetSearch"
+    />
+    <oc-button
+      id="files-open-search-btn"
+      key="mobile-search-button"
+      class="uk-hidden@m"
+      icon="search"
+      :aria-label="displaySearchButtonLabel"
+      @click="focusMobileSearchInput"
+    />
+    <oc-drop
+      drop-id="oc-topbar-search-mobile"
+      toggle="#files-open-search-btn"
+      mode="click"
+      class="uk-margin-remove"
+      :options="{ delayHide: '0' }"
+    >
+      <oc-search-bar
+        ref="mobileSearch"
+        :label="searchLabel"
+        :loading="isLoadingSearchResults"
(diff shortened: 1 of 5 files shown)


LukasHirt:
For now, it is built on top of the [sidebar PR](https://github.com/owncloud/phoenix/pull/3442#issuecomment-627920838). After it'll get merged, I'll drop that commit in here.

LukasHirt:
There is one issue with search on mobile. Since we hide it in the drop, this results in the search being still visible after the search has been finished. However, if we use the `closeOnClick` then it works if we execute the search by clicking on the search button but it also hides it if we click only into the input. And of course, if the search is executed by via keyboard then the result is the same as in the beginning.

It's times like these I really wish for having drop with a simple `v-if` instead of the UIkit one...

LukasHirt:
Fixed the selector in tests for loading spinner in the search bar. Tests are now passing locally. Also, I've hidden the top bar in public links on wider screens because they had no content there

LukasHirt:
Ok, one more change - hidden top bar and sidebar in public pages. Top bar because there is no element which could be used (opened a ticket to make search work in public links #3521) and sidebar because there are no nav items (opened a ticket for branding in such case #3520)

LukasHirt:
Tests are green. Pls review @PVince81 @kulmann

src/components/SearchBar.vue
```diff
@@ -0,0 +1,103 @@
...
+      :button-hidden="true"
+      @search="search"
+      @clear="resetSearch"
+    />
+    <oc-button
+      id="files-open-search-btn"
+      key="mobile-search-button"
+      class="uk-hidden@m"
+      icon="search"
+      aria-label="search"
```
kulmann, line 17: aria-label should have translated content

src/components/SearchBar.vue
```diff
@@ -0,0 +1,103 @@
...
+      this.clearSearchBarValue()
+    }
+  },
+
+  methods: {
+    ...mapActions('Files', ['searchForFile', 'resetSearch']),
+
+    search(searchTerm) {
+      if (!searchTerm) {
+        return
```
kulmann, line 65: Since typing `enter` emits `search`, it would make sense to call `resetSearch` here just above the return.

LukasHirt:
@kulmann Label translated to "Open search bar" and search reset when hitting enter added.

PR #3473, opened 2020-05-18
[Tests-only] Fix folder existence check in tests

PVince81:
Instead of doing an illegal GET request use PROPFIND instead.

Extracted from https://github.com/owncloud/phoenix/pull/3457


PVince81:
Fixes https://github.com/owncloud/ocis-reva/issues/106

PR #3484, opened 2020-05-20
Fetch settings values from sdk and use language code

kulmann:
This PR depends on https://github.com/owncloud/owncloud-sdk/pull/475, so this has to stay WIP until the SDK-PR is merged and released.

## Description
We have a settings service, that is supposed to be integrated in Phoenix like described [in the docs](https://owncloud.github.io/extensions/ocis_settings/)

This PR uses settings values, coming from the settings service, to adapt the Phoenix UI for the authenticated user. At the moment, only the `language` setting is used: Changing the language in the settings frontend (has to be configured as a Phoenix app) results in having another language in Phoenix. In the future we can make use of this for a variety of customization:
- chosen theme
- notification preferences
- preferences regarding the shown elements in the files list (e.g. different quick actions)
- ...

## Related Issue
https://github.com/owncloud/product/issues/17

## Motivation and Context
We want to provide a customized user experience for users.

## How Has This Been Tested?
So far only in Chrome and Firefox. No tests implemented so far.

#### English UI
Screenshot 2020-05-20 at 10 48 12

#### Changed language to French
Screenshot 2020-05-20 at 10 49 16

#### French UI (at least everything that's translated already)
Screenshot 2020-05-20 at 10 48 55

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] Depends on release of https://github.com/owncloud/owncloud-sdk/pull/475
- [x] Discuss tests

### DIFF
--- a/changelog/unreleased/3484-1
+++ b/changelog/unreleased/3484-1
@@ -1,0 +1,7 @@
+Change: Make settings available in phoenix
+
+We upgraded to a new owncloud-sdk version which provides loading settings from the
+settings service, if available. The settings values are available throughout phoenix
+and all extensions.
+
+https://github.com/owncloud/phoenix/pull/3484
--- a/src/Phoenix.vue
+++ b/src/Phoenix.vue
@@ -64,7 +64,8 @@ export default {
       'activeNotifications',
       'activeMessages',
       'capabilities',
-      'apps'
+      'apps',
+      'getSettingsValueByIdentifier'
     ]),
     $_applicationsList() {
       const list = []
@@ -154,6 +155,14 @@ export default {
       }
 
       return 'fade'
+    },
+
+    selectedLanguage() {
+      return this.getSettingsValueByIdentifier({
+        extension: 'ocis-accounts',
+        bundleKey: 'profile',
+        settingKey: 'language'
+      })
     }
   },
   watch: {
@@ -175,6 +184,16 @@ export default {
           this.$_updateNotifications()
         }, 30000)
       }
+    },
+    selectedLanguage: {
+      handler(language) {
+        if (language !== null) {
+          this.$language.current = language.listValue.values[0].stringValue
+        } else if (this.$language.defaultLanguage) {
+          this.$language.current = this.$language.defaultLanguage
+        }
+      },
+      immediate: true
     }
   },
 
--- a/src/Phoenix.vue
+++ b/src/Phoenix.vue
@@ -187,10 +187,12 @@ export default {
     },
     selectedLanguage: {
       handler(language) {
-        if (language !== null) {
-          this.$language.current = language.listValue.values[0].stringValue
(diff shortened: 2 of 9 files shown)


PVince81:
please update owncloud-sdk and add some error handling

kulmann:
Screenshot 2020-05-22 at 21 20 08

changelog/unreleased/3484-1
```diff
@@ -0,0 +1,7 @@
+Change: Make settings available in phoenix
```
LukasHirt, line 1: "Enhancement" might be better in this case?

src/Phoenix.vue
```diff
@@ -175,6 +184,16 @@ export default {
           this.$_updateNotifications()
         }, 30000)
       }
+    },
+    selectedLanguage: {
+      handler(language) {
+        if (language !== null) {
+          this.$language.current = language.listValue.values[0].stringValue
+        } else if (this.$language.defaultLanguage) {
```
LukasHirt, line 192: Is `defaultLanguage` always defined?

src/store/user.js
```diff
@@ -29,6 +29,7 @@ const actions = {
   logout(context) {
     const logoutFinalizer = () => {
       context.dispatch('cleanUpLoginState')
+      context.dispatch('loadSettingsValues')
```
LukasHirt, line 32: Why load the settings during logout?

changelog/unreleased/3484-1
kulmann: Nope, it's a change. ;-) Introduces settings to phoenix => entirely new feature.

src/Phoenix.vue
kulmann: Default language is configured as `navigator.language.substring(0, 2)`, so it should always be available. Still, I changed the code in a new commit, because it didn't allow to unset the setting. Please re-review. :-)

src/store/user.js
kulmann: Because there will be settings that exist without user context. For example, when we provide multiple themes in the future, an admin could set a default theme in settings. That has to be available even if you don't have a user context. So as a result, all changes to the authenticated user (changing to another one or signing out) need to trigger a settings reload.

LukasHirt: I see. I am just a bit afraid of relying on authentication in this case because of the public link's context.

src/Phoenix.vue
LukasHirt: I just wondered why was there the `else if`. But the current approach looks good 👍

changelog/unreleased/3484-1
LukasHirt: Exactly, a new feature which doesn't change any existing functionality

PR #3548, opened 2020-05-28
[tests-only] Simplify test run

### LINKED ISSUE #3456: make a script that helps to setup and start tests easier
something that setups the env. with default values and runs the tests, maybe as yarn command.
The goal is to reduce the pain and sources of mistakes when running the tests as described in the docs, specially for ocis backend https://owncloud.github.io/clients/web/testing/#with-ocis-backend

individual-it:
## Description
1. a couple of commands to setup and run tests with ocis
2. more default settings, so that less need to be configured in a default environment

## Related Issue
- Fixes #3456

## Motivation and Context
make our live easier

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


docs/testing.md
```diff
@@ -61,50 +61,25 @@ The feature files are located in the "tests/acceptance/features" subdirectories.
...
-      "markdown-editor",
-      "media-viewer"
-    ]
-  }
-  ```
-
-- Run the OCIS server with necessary configurations
+    yarn run redis-server
+    ```
+3. create a new phoenix config.json file.
```
LukasHirt, line 80: If I understand it correctly, oCIS generates it's own config if the path to an existing one is not specified. What about writing this step as optional?

individual-it: I wasn't aware of that auto-generation I will try it

individual-it: I've deleted the step all-together, also in the auto run

docs/testing.md
```diff
@@ -61,50 +61,25 @@ The feature files are located in the "tests/acceptance/features" subdirectories.
...
+   As starting point and example that should work when running every service on localhost use `tests/acceptance/ocis-config-local.json`
```
LukasHirt, line 81: This now points to non-existing config file.

LukasHirt:
Apart from that reference to a config which doesn't exists looks all good to me. Just tried to start the tests from fresh Phoenix clone and it's so awesome that we need just one yarn command and have all the required part of ocis and ocis itself! 🚀 Just one point. I need to point instead of `localhost` to `host.docker.internal` as a Mac user so I need to use the `PHOENIX_WEB_CONFIG` when running the `test-requirements:ocis:mac` command. Would it make sense to include it somehow into the docs?

docs/testing.md
individual-it: :facepalm: I've removed the step now for real

individual-it:
In my imagination you should not need `host.docker.internal` because all the docker containers use port-forwarding and everything should be accessible through `localhost`, but I might be wrong for Mac

LukasHirt:
The config.json is now fine but there is a different issue. It still tries to point to `localhost` when trying to log in because of OpenID config.

From https://host.docker.internal:9200/.well-known/openid-configuration
```
"issuer": "https://localhost:9200",
"authorization_endpoint": "https://localhost:9200/signin/v1/identifier/_/authorize",
"token_endpoint": "https://localhost:9200/konnect/v1/token",
"userinfo_endpoint": "https://localhost:9200/konnect/v1/userinfo", "end_session_endpoint": "https://localhost:9200/signin/v1/identifier/_/endsession",
"check_session_iframe": "https://localhost:9200/konnect/v1/session/check-session.html",
"jwks_uri": "https://localhost:9200/konnect/v1/jwks.json"
```

individual-it:
could you try to create a new identifier-registration.yml  similar to https://github.com/owncloud/phoenix/blob/master/tests/drone/identifier-registration.yml but with with `https://host.docker.internal:9200/` everywhere
and then in `ocis:mac` add these env. variables
`REVA_OIDC_ISSUER='https://host.docker.internal:9200/'`
`KONNECTD_IDENTIFIER_REGISTRATION_CONF='/your/folder/identifier-registration.yml'`
`KONNECTD_ISS='https://host.docker.internal:9200'`
`KONNECTD_TLS='true'`

PR #3568, opened 2020-06-02
Use chunk size and method override from caps

PVince81:
Use the chunk size from the capabilities if it is higher than the
locally configured one.

Use HTTP method override mode if the TUS capabilities says so.

For https://github.com/owncloud/ocis-reva/issues/177.
Requires https://github.com/owncloud/ocis-reva/pull/228 for testing the capabilities
Requires https://github.com/cs3org/reva/pull/792 for testing the method override.

I've tested with the above PRs and the upload is using the correct chunk value and is also setting `overridePatchMethod` whenever the matching capability is set.

Note: merging this should be safe due to default values.

### DIFF
--- a/apps/files/src/mixins.js
+++ b/apps/files/src/mixins.js
@@ -447,9 +447,20 @@ export default {
         // FIXME: this might break if relativePath is not the currentFolder
         // and is a mount point that has no chunk support
         if (this.browserSupportsChunked && this.currentFolder.isChunkedUploadSupported) {
+          let chunkSize = this.configuration.uploadChunkSize
+          if (this.capabilities.files.tus_support.max_chunk_size > 0) {
+            if (
+              chunkSize === null ||
+              chunkSize === 0 ||
+              chunkSize > this.capabilities.files.tus_support.max_chunk_size
+            ) {
+              chunkSize = this.capabilities.files.tus_support.max_chunk_size
+            }
+          }
           promise = this.uploadQueue.add(() => {
             return this.uploadChunkedFile(file, pathUtil.dirname(relativePath), {
-              chunkSize: this.configuration.uploadChunkSize,
+              chunkSize: chunkSize,
+              overridePatchMethod: !!this.capabilities.files.tus_support.http_method_override,
               emitProgress: progress => {
                 this.$_ocUpload_onProgress(progress, file)
               }
--- a/changelog/unreleased/3568
+++ b/changelog/unreleased/3568
@@ -1,0 +1,11 @@
+Enhancement: Use TUS settings from capabilities
+
+The TUS settings advertise the maximum chunk size, so we now use the smallest chunk size from
+the one configured in config.json and the one from the capabilities.
+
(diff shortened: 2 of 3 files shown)


apps/files/src/mixins.js
```diff
@@ -447,9 +447,23 @@ export default {
         // FIXME: this might break if relativePath is not the currentFolder
         // and is a mount point that has no chunk support
         if (this.browserSupportsChunked && this.currentFolder.isChunkedUploadSupported) {
+          let chunkSize = this.configuration.uploadChunkSize
+          console.log('Configured chunk size:', chunkSize)
```
LukasHirt, line 451: Are those console logs left here on purpose?

apps/files/src/mixins.js
```diff
@@ -447,9 +447,23 @@ export default {
...
+          if (this.capabilities.files.tus_support.max_chunk_size > 0) {
+            if (
+              chunkSize === null ||
+              chunkSize === 0 ||
+              chunkSize > this.capabilities.files.tus_support.max_chunk_size
+            ) {
+              chunkSize = this.capabilities.files.tus_support.max_chunk_size
```
LukasHirt, line 458: Could this be shortened with `Math.max()`?

LukasHirt: Ah, it would return the wrong one in case of 0 or null. Nevermind 🤦

PVince81: please recheck, I managed to find a way: converting null to 0.

0 means "unlimited" so the logic can still be summarized with a Math.max

apps/files/src/mixins.js
PVince81: removed now

apps/files/src/mixins.js
PVince81: I mixed up max and min... we do want the admin to override and specify a smaller chunk size than the advertised one

PVince81: I've reverted now to the logic we had in the first iteration

LukasHirt: Yeah, I mixed them as well when I wrote the first comment (realised after it and edited it)

apps/files/src/mixins.js
```diff
@@ -447,9 +447,22 @@ export default {
...
+            }
+          }
+          console.log('tus_support: ', this.capabilities.files.tus_support)
```
LukasHirt, line 460: Forgotten as well?

apps/files/src/mixins.js
```diff
@@ -447,9 +447,22 @@ export default {
...
+          console.log('Using chunk size: ', chunkSize)
```
LukasHirt, line 461: Forgotten as well?

PVince81: yeah... used the undo feature of my editor and didn't see that it brought those back as well :facepalm:

PVince81:
I've pushed again, hopefully all should be fine now...

PR #3570, opened 2020-06-04
Move navItems into a store module

kulmann:
## Description
NavItems now live in a `navigation` store module in phoenix core. They are internally separated into static and dynamic navItems. Static is for the navItems loaded from an extension config. Dynamic is for dynamically loaded navItems (most likely from loaded data). A known limitation of this implementation is, that only statically loaded navItems (the ones coming from appInfo) lead to the extension appearing in the app switcher. As the app switcher is likely to change in the near future, we can touch this separately.

## Related Issue
- Fixes https://github.com/owncloud/phoenix/issues/3497 

## Motivation and Context
We need dynamically loaded navItems that are based on loaded data, not on configuration.

## How Has This Been Tested?
For now with Chrome and Firefox locally, with code in settings ui depending on this change (injecting navItems through the store). Statically loaded navItems are now loaded differently (into the store) but the functionality stays the same. Acceptance tests are covering that.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/dynamic-nav-items
+++ b/changelog/unreleased/dynamic-nav-items
@@ -2,7 +2,7 @@ Enhancement: Dynamically loaded nav items
 
 We have moved the navItems from application configuration into a store module. We extended it's functionality
 by introducing statically and dynamically loaded navItems. This way navItems can be loaded based on extension
-data, as soon as the extension becomes active. Please note that at having least one static navItem (coming
+data, as soon as the extension becomes active. Please note that having at least one static navItem (coming
 from the appInfo object of an extension) is still a requirement for the extension appearing in the app switcher.
 
 https://github.com/owncloud/phoenix/issues/3497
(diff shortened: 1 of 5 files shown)


changelog/unreleased/dynamic-nav-items
```diff
@@ -0,0 +1,9 @@
+Enhancement: Dynamically loaded nav items
+
+We have moved the navItems from application configuration into a store module. We extended it's functionality
+by introducing statically and dynamically loaded navItems. This way navItems can be loaded based on extension
+data, as soon as the extension becomes active. Please note that at having least one static navItem (coming
```
LukasHirt, line 5: `at` should be after `having`?

kulmann: Yes, thanks! :-)

PR #3584, opened 2020-06-09
[Tests-Only]Run multiple times on drone for fix intermittently failing test

Talank:
## Description
Run the scenario `Create a folder with default name` 20 times on drone


Talank:
All tests have passed.

PR #3585, opened 2020-06-09
Add collaborators quick action

LukasHirt:
## Description
Added a new quick action which opens the new collaborators tab in the files list sidebar.

## Related Issue

- Refs https://github.com/owncloud/product/issues/80

## How Has This Been Tested?

- test environment: Acceptance tests

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Open tasks:

- [x] Rebase after merging quick actions
- [ ] Fix clicking on quick action when sidebar is already opened

### DIFF
--- a/apps/files/src/quickActions.js
+++ b/apps/files/src/quickActions.js
@@ -39,11 +39,37 @@ function createPublicLink(ctx) {
   })
 }
 
+function openNewCollaboratorsPanel(ctx) {
+  ctx.store.dispatch('Files/setHighlightedFile', ctx.item)
+
+  // Workaround for displaying the new collaborators panel even when one is already opened
+  // Creating timeout takes care of the event loop
+  // TODO: Get rid of this after we use overlay instead of sidebar
+  setTimeout(() => {
+    ctx.store.commit('Files/SET_CURRENT_SIDEBAR_TAB', {
+      tab: 'files-sharing',
+      options: { collaboratorsCurrentPanel: 'newCollaborator' }
+    })
+  })
+}
+
+function canShare(item, store) {
+  return store.state.user.capabilities.files_sharing.api_enabled && item.canShare()
+}
+
 export default {
+  collaborators: {
+    id: 'collaborators',
+    label: $gettext('Add new collaborators'),
+    icon: 'group',
+    handler: openNewCollaboratorsPanel,
+    displayed: canShare
+  },
   publicLink: {
     id: 'public-link',
     label: $gettext('Create and copy public link'),
     icon: 'link',
-    handler: createPublicLink
+    handler: createPublicLink,
+    displayed: canShare
   }
 }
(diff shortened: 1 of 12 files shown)


LukasHirt:
This PR is built on top of #3573 After it will get merged I will rebase this PR.

LukasHirt:
There is an issue that when we're switching highlighted file, the current tab and current panel resets. So if we click on the quick action while already having sidebar opened, we land in the default tab

One solution could be to set first the highlighted file to null which would do the reset before we set the new tab. This leads to a "jumpy" sidebar which quickly closes and opened => very ugly.

I'll try to see if there'd be some nicer solution. But I don't want to invest much time into debugging this since we'll change it later to overlay which will hide the quick actions anyway :shrug:

apps/files/src/store/mutations.js
```diff
@@ -244,5 +244,13 @@ export default {
       }
     })
     state.files = []
+  },
+
+  SET_CURRENT_SIDEBAR_TAB(state, tab) {
+    state.currentSidebarTab = tab
+  },
+
+  SET_CURRENT_COLLABORATORS_TAB_PANEL(state, panel) {
```
kulmann, line 253: IMO this is too specific. What about an options object as an optional second argument for the current sidebar tab? Then we would be able to pass in context specific data without changing the store everytime we add something to the sidebar.

Could be called like this: `SET_CURRENT_SIDEBAR_TAB(state, { tab, options: {'collaborators-tab-sub-panel': 'newCollaborator'} }`

LukasHirt: I still introduced mutation though which sets only options and not the whole tab object. This is because in collaborators we often change only the panel without changing the tab and it would feel strange to always have to define the collaborators' tab all over again.

LukasHirt:
@kulmann I've brought back the commented out part of tests and introduced the options object. Pls, re-review.

apps/files/src/quickActions.js
```diff
@@ -53,8 +53,8 @@ function openNewCollaboratorsPanel(ctx) {
   })
 }
 
-function canShare(item) {
-  return item.canShare()
+function canShare(ctx) {
```
kulmann, line 56: If you want to have `store` and `item` as named variables, you can unwrap the ctx like this:
`function canShare({ item, store }) {`. Of course also fine like it is. 👍

LukasHirt:
> There is one bug: when I open the sidebar and then click on the link indicator of another row, the sidebar switches to the collaborators tab instead of the links tab... happens reliably in Chrome. Can you confirm that?

Yes, happens to me as well. I'll jump to debugging first thing tomorrow morning

LukasHirt:
Fixed the bug. Had to use `setTimeout` again because the `beforeDestroy` which resets the tab happened after the links tab has been set.

apps/files/src/quickActions.js
LukasHirt: Right, 🤦 I did it already automatically after doing it in the handlers of quickActions. I got rid of the object completely and just pass it in.

PR #3593, opened 2020-06-11
removed vue-clipboard2 from package.json

### LINKED ISSUE #3588: Remove vue-clipboard2
We now have a new package `copy-to-clipboard` which can be used not only in Vue templates but also in plain js files. The `vue-clipboard2` should be replaced with this not to have two clipboard deps.

Julian1998:
## Description

Removed vue-clipboard2 from the package.json file dependencies

## Related Issue

- Fixes #3588 

## Motivation and Context

We already got copy-to-clipboard so vue-clipboard would be redundant

## How Has This Been Tested?

- tested in chrome

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/apps/files/src/components/PublicLinksSidebar/PublicLinkListItem.vue
+++ b/apps/files/src/components/PublicLinksSidebar/PublicLinkListItem.vue
@@ -63,8 +63,7 @@
         >
           <oc-icon
             v-if="!linksCopied[link.url]"
-            v-clipboard:copy="link.url"
-            v-clipboard:success="$_clipboardSuccessHandler"
+            @click="$_copyPublicLinkToClipboard()"
             name="copy_to_clipboard"
             size="small"
           />
@@ -101,6 +100,7 @@
 <script>
 import { basename, dirname } from 'path'
 import mixins from '../../mixins'
+import copyToClipboard from 'copy-to-clipboard'
 
 export default {
   name: 'PublicLinkListItem',
@@ -174,12 +174,13 @@ export default {
     }
   },
   methods: {
-    $_clipboardSuccessHandler(event) {
-      this.$emit('onCopy', event)
-    },
     $_removeLink() {
       this.removalInProgress = true
       this.$emit('onDelete', this.link)
+    },
+    $_copyPublicLinkToClipboard() {
+      copyToClipboard(this.link.url)
+      this.$emit('onCopy', { action: 'copy', text: this.link.url })
     }
   }
 }
--- a/apps/files/src/components/FileLinkSidebar.vue
+++ b/apps/files/src/components/FileLinkSidebar.vue
@@ -18,8 +18,7 @@
               <oc-icon
                 v-if="!linksCopied[$_privateLinkOfHighlightedFile]"
                 id="files-sidebar-private-link-label"
-                v-clipboard:copy="$_privateLinkOfHighlightedFile"
-                v-clipboard:success="$_clipboardSuccessHandler"
+                @click="$_copyPrivateLinkToClipboard()"
                 name="copy_to_clipboard"
                 size="small"
               />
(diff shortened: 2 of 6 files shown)


Julian1998:
@LukasHirt I absolutely don't like my part where I create a custom event for not crashing the event handler method. Do you think we can delete the whole event handler as it just seems to prevent fast double clicks on the copy to clipboard icon/button.

LukasHirt:
> I absolutely don't like my part where I create a custom event for not crashing the event handler method. Do you think we can delete the whole event handler as it just seems to prevent fast double clicks on the copy to clipboard icon/button.

I think part of that is also responsible for giving feedback to the user that the link has been copied. At first glance, it might be enough to use local data object in the component. But there might be something that I have overlooked so you can try it and see if it'll work still fine 😉

Julian1998:
> > I absolutely don't like my part where I create a custom event for not crashing the event handler method. Do you think we can delete the whole event handler as it just seems to prevent fast double clicks on the copy to clipboard icon/button.
> 
> I think part of that is also responsible for giving feedback to the user that the link has been copied. At first glance, it might be enough to use local data object in the component. But there might be something that I have overlooked so you can try it and see if it'll work still fine

Yeah it is responsible for feedback... BUT we never implemented the visual feedback in the sidebar

LukasHirt:
> BUT we never implemented the visual feedback in the sidebar

We did.

image

Julian1998:
Ah I was talking about something like a notification message

Julian1998:
@LukasHirt well I slightly adapted my code but we can't get rid of these events. We need to emit an event so the success handler method gets triggered. I guess there are several ways to deal with that, but emitting an event seems to be the best in my opinion.

What do you think?

Julian1998:
@LukasHirt please re-review :)

apps/files/src/components/PublicLinksSidebar/PublicLinkListItem.vue
```diff
@@ -63,8 +63,7 @@
         >
           <oc-icon
             v-if="!linksCopied[link.url]"
-            v-clipboard:copy="link.url"
-            v-clipboard:success="$_clipboardSuccessHandler"
+            @click="$_copyPublicLinkToClipboard()"
```
LukasHirt, line 66: Move the click pls to the wrapping button.

LukasHirt:
> I guess there are several ways to deal with that, but emitting an event seems to be the best in my opinion.

Is there something connected to the event outside the link item component? As far as I understood (haven't looked that close though so I might be missing something) we emit the event, then set the copied link globally and then read it again in the link item component. Is that correct? What about not emitting the event but instead using local data object in the component?

Julian1998:
Well.... the linkListItem triggers the onCopy event which gets processed in the fileLinkSidebar component. This is done so we can reuse the successHandler in the filesidebar for the linkListItem as well as some code for the clipboard in the sidebar itself

LukasHirt:
@Julian1998 Address pls this comment https://github.com/owncloud/phoenix/pull/3593#discussion_r453519970

Julian1998:
> @Julian1998 Address pls this comment [#3593 (comment)](https://github.com/owncloud/phoenix/pull/3593#discussion_r453519970)

My requested re-review was not on intended. I am sorry

LukasHirt:
> My requested re-review was not on intended. I am sorry. Unluckily there was no way to undo this.

No worries 😉

kulmann:
@Julian1998 same here, do you know when you will be able to schedule time for this? or should someone else take over?

pascalwengerter:
@julian1998 just stumbled across this when checking open issues, shall we finish your work or would you like to continue working on it? The project structure has changed a bit, happy to offer some guidance if applicable!

Julian1998:
@pascalwengerter I forgot to unassign. We could keep this PR open and someone else could build a new solution based on this branch. But I probably won't finish it. Best luck

pascalwengerter:
Superseded by #5147

PR #3605, opened 2020-06-15
Rework account dropdown

LukasHirt:
## Description
We've removed user avatar, user email and version from the account dropdown.
The log out button has been changed into a link.
All links in account dropdown are now inside of a list.

## Related Issue

- Fixes https://github.com/owncloud/product/issues/82

## Motivation and Context
Keep focus only on the actionable items inside of the dropdown.

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/rework-account-dropdown
+++ b/changelog/unreleased/rework-account-dropdown
@@ -1,0 +1,8 @@
+Change: Rework account dropdown
+
+We've removed user avatar, user email and version from the account dropdown.
+The log out button has been changed into a link.
+All links in account dropdown are now inside of a list.
+
+https://github.com/owncloud/product/issues/82
+https://github.com/owncloud/phoenix/pull/3605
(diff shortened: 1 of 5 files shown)


LukasHirt:
The design is still a bit off from mockups in Figma but I think to achieve that we should first adjust the oc-drop component itself.

LukasHirt:
Pushed a new commit regarding the tests. Locally it's passing fine. Let's see here.

LukasHirt:
Ok, hopefully, the last one. Was missing id in the logout item.

kulmann:
What about the size of the drop? The default size of 300px looks way too big for this menu. Do we have a way to shrink it to the width of the list items? 🤔

LukasHirt:
> What about the size of the drop? The default size of 300px looks way too big for this menu. Do we have a way to shrink it to the width of the list items? 🤔

I originally didn't want to touch the UI of the dropdown before we do something with the drop component itself. But yeah, this is a quick win. Width decreased to the content width :wink:

kulmann:
For the record:
Screenshot 2020-06-16 at 20 15 29

PR #3617, opened 2020-06-16
Move indicators under resource name

LukasHirt:
## Motivation and Context
Better separation from quick actions -> probably only temporarily solution (will be tested with real users).

image

## Types of changes

- [ ] Technical debt
- [x] Tests

### DIFF
--- a/apps/files/src/components/FileItem.vue
+++ b/apps/files/src/components/FileItem.vue
@@ -1,20 +1,62 @@
 <template>
-  <oc-file
-    :name="fileName"
-    :extension="item.extension"
-    :icon="previewIcon"
-    :icon-url="previewUrl"
-    :filename="item.name"
-    :data-preview-loaded="previewLoaded"
-  />
+  <div class="oc-file uk-flex uk-flex-middle" :data-preview-loaded="previewLoaded">
+    <oc-img
+      v-if="previewUrl"
+      key="file-preview"
+      class="files-list-file-preview uk-margin-small-right"
+      :class="{ 'files-list-file-preview-small': !hasTwoRows }"
+      :src="previewUrl"
+      alt=""
+    />
+    <oc-icon
+      v-else
+      key="file-icon"
+      :name="previewIcon"
+      :size="hasTwoRows ? 'medium' : 'small'"
+      aria-hidden="true"
+      class="uk-margin-small-right"
+    />
+    <div class="uk-width-expand">
+      <div class="file-row-name uk-text-truncate" :filename="item.name">
+        <span
+          class="uk-text-bold oc-cursor-pointer oc-file-name uk-padding-remove-left"
+          role="button"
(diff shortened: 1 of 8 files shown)


apps/files/src/components/FileItem.vue
```diff
@@ -1,20 +1,57 @@
...
+          role="button"
+          v-text="fileName"
+        /><span
+          v-if="item.extension"
+          class="uk-text-meta oc-file-extension"
+          v-text="'.' + item.extension"
+        />
+      </div>
+      <div class="uk-flex uk-flex-middle">
+        <translate class="uk-margin-small-right">State:</translate>
```
kulmann, line 28: IMO `State` together with indicator icons or with the `-` is not self explanatory. What do you think about `Shares:` or `Shared:` instead?

changelog/unreleased/move-indicators-under-resource
```diff
@@ -0,0 +1,5 @@
+Change: Move status indicators under the resource name
+
+We've moved the sharing status indicators from own column in files list under the resource name.
```
kulmann, line 3: Proposing `We've moved the sharing status indicators from an own column in the files list to a second row under the resource name.`

apps/files/src/components/FileItem.vue
LukasHirt: The problem is that `Shared:` is not covering our use case since the indicators can come from an extension and can indicate whatever it wants. We tried to come up together with @micbar with some good way of labelling the indicators and `State:` seems like the easiest and best option for now. @micbar did quick research also on the web on the meaning of words `Status` and `State` and in this case, `State` is the one to go with because `Status` is meant for singular indicator (e.g. `Accepted`)

kulmann: Sounds a lot like we need labels per indicator instead of one generic label.

LukasHirt:
@kulmann I've removed the underline effect when hovering over the indicators and display it only when hovering the name directly. It's done directly via styles in the component same as the sizes of the preview. I'd like to adjust the ODS component to handle two rows and then move the styles in there as well.

I've also hidden the second row in all the lists but all files and favourites as you correctly mentioned that it doesn't make much sense in there with the current implementation.

The changelog is also adjusted with your proposal :wink:

I haven't unfortunately figured out what to do with the tests. I cannot reproduce the error locally either when running the tests or manually. I guess it cannot find some resources, but that doesn't really make sense to me, that it doesn't do in more scenarios and why does it pass locally.

dpakach:
There was a bug in the test code that was scrolling through the file list and checking if the file was visible. Because of that, some elements would not be seen as visible no matter how much it scrolled. 
this change (https://github.com/owncloud/phoenix/pull/3617/commits/d106fb1aca418e8b68f2a3609ea60f3785945533) fixes that

PR #3621, opened 2020-06-17
Update ODS

LukasHirt:
Updated ODS to bring the new branding

### DIFF
--- a/apps/files/src/components/Roles/RolesSelect.vue
+++ b/apps/files/src/components/Roles/RolesSelect.vue
@@ -3,8 +3,11 @@
     <label class="oc-label">
       <translate>Role:</translate>
     </label>
-    <oc-button :id="`files-${mode}-role-button`" :class="`uk-width-1-1 files-${mode}-role-button`">
-      <role-item :role="selectedRole" class="uk-margin-small-bottom" />
+    <oc-button
+      :id="`files-${mode}-role-button`"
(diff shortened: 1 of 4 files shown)


apps/files/src/components/Roles/RolesSelect.vue
```diff
@@ -3,8 +3,11 @@
     <label class="oc-label">
       <translate>Role:</translate>
     </label>
-    <oc-button :id="`files-${mode}-role-button`" :class="`uk-width-1-1 files-${mode}-role-button`">
-      <role-item :role="selectedRole" class="uk-margin-small-bottom" />
+    <oc-button
+      :id="`files-${mode}-role-button`"
+      :class="`uk-width-1-1 files-${mode}-role-button uk-flex-left`"
+    >
+      <role-item :role="selectedRole" class="uk-margin-xsmall-top uk-margin-xsmall-bottom" />
```
LukasHirt, line 10: Padding classes would be much better here but we don't have an xsmall ones 🙁

kulmann: Looks nice, let's keep it like this for now 👍

LukasHirt:
Made the "+" in create new button an icon instead of string. The button itself is quite bigger now but at least it matches the height of delete selected button and also fixes the a11y.

image

PR #3622, opened 2020-06-17
Move create new button

LukasHirt:
Adjusted also a bit the spacing around table header.

image

### DIFF
--- a/apps/files/src/components/FileList.vue
+++ b/apps/files/src/components/FileList.vue
@@ -10,7 +10,7 @@
         key="files-list-results-existence"
         gutter="small"
         flex
-        class="uk-padding-small"
+        class="uk-padding-small uk-padding-remove-top uk-padding-remove-bottom uk-margin-xsmall-bottom"
       >
         <div>
           <oc-checkbox
--- a/apps/files/src/components/FilesAppBar.vue
+++ b/apps/files/src/components/FilesAppBar.vue
@@ -10,7 +10,7 @@
       @progress="onFileProgress"
     />
     <oc-grid flex gutter="small">
-      <div class="uk-flex-1">
+      <div>
         <div class="uk-flex">
           <oc-breadcrumb
             v-if="showBreadcrumb"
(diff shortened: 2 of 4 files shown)


LukasHirt:
Hmm, not sure why tests are failing in CI. Locally are passing fine and there is no other change then different width of breadcrumbs 🤷

LukasHirt:
I tried rebasing the PR. Let's see.

LukasHirt:
I haven't figured out what is happening with the tests. Locally cannot reproduce the errors because it all passes just fine. And I don't get that only changing width of an element can fail two separate tests. 🤷

individual-it:
the tests don't really click on the "Upload File" button but click the "+ New" button, check that the "Upload File" button exists and then upload the file with "setValue" it happens that the dropdown overlays the modal dialog
image

in https://github.com/owncloud/phoenix/blob/533e8cd9e1451a4d27778f1ef51ba7aa271d2dc4/tests/acceptance/pageObjects/filesPage.js#L261 the test click on the modal dialog to close the dropdown, but because of the overlay it clicks "New mark-down file..." and that app seems to be broken and redirects to an empty window

So we could:
1. click not on 0,0 but e.g. 100,100 - but how to make sure that a similar issue hits us again when the dropdown gets longer?
2. close the dropdown directly in `selectFileForUpload()` again
3. don't open the dropdown at all

kulmann:
IMO we should close the dropdown by doing another click on the `+ New` button, as it toggles the dropdown.

LukasHirt:
I've moved the click to the actual dropdown. Since the dropdown has padding, we do not click on any link inside and are also safe in the future in case it would be growing in size. Pls, check if you're fine with this solution @individual-it @kulmann 

> IMO we should close the dropdown by doing another click on the + New button, as it toggles the dropdown.

I've tried this solution. Using `click()` directly on that button doesn't work at all. And I am a bit afraid of clicking at the location of the button because I am not 100 % sure it wouldn't get covered in case the dropdown would be way expanded too much.

kulmann:
Yes, awesome!

Issue #3603, opened 2020-06-15
phoenix translation sync failing


phil-davis:
See issue https://github.com/owncloud/translation-sync/issues/35

Needs someone to investigate.

phil-davis:
@micbar who looks after this sort of thing?

micbar:
any engineer capable of doing it :-)
@lhirt @kulmann ?

kulmann:
It looks like easygettext doesn't support `??` (nullish coalescing operator), which is used in `src/store/modal.js`.

There are two ways of fixing this:
1: fall back to `||`. Need @LukasHirt to clarify, if that is valid or if it breaks his idea of the implementation.
2: add support for parsing `??` upstream, can probably be done similar to this: https://github.com/laget-se/react-gettext-parser/commit/cc9d71d6a21c65f401525d17fd74ab1cbd22da04

LukasHirt:
> 1: fall back to ||. Need @LukasHirt to clarify, if that is valid or if it breaks his idea of the implementation.
> 2: add support for parsing ?? upstream, can probably be done similar to this: laget-se/react-gettext-parser@cc9d71d

There shouldn't be a real issue with using OR operator. The idea was to use the nullish coalescing one since in some modal values it is valid to pass `false`. So it doesn't break anything but simply doesn't look nice when it uses `false` backup value instead of the one that was passed.

LukasHirt:
> add support for parsing ?? upstream, can probably be done similar to this: laget-se/react-gettext-parser@cc9d71d

Easygettext seems to be using acorn for parsing the JS code which should support nullish coalescing operator https://github.com/acornjs/acorn/pull/890 upstream is using some outdated version?

phil-davis:
Still failing - e.g. https://drone.owncloud.com/owncloud/translation-sync/1583/33/5

So no new translations are coming.

LukasHirt:
> Nobody has been assigned to this issue.

I'll take care.

phil-davis:
Translations arrived last night - https://github.com/owncloud/phoenix/commit/cce62d12b94db22537741b512f8a0fe338607480
Works

phil-davis:
Some translation came, but last night: https://drone.owncloud.com/owncloud/translation-sync/1588/33/5
```
tx INFO: Pushing source file (template.pot)
tx ERROR: Could not upload source file. You can use --skip to ignore this error and continue the execution.
[...]

PR #3733, opened 2020-07-06
Stop using nullish coalescing operator and optional chaning

LukasHirt:
## Description
Using Nullish coalescing operator and optional changing resulted in syntax errors in translations sync.

## Related Issue

- Refs #3603

## Motivation and Context
Passing translations sync

## How Has This Been Tested?

Run `make l10n-read`

### DIFF
--- a/apps/files/src/components/FileDetails.vue
+++ b/apps/files/src/components/FileDetails.vue
@@ -85,7 +85,11 @@ export default {
     },
 
     currentTab() {
-      return this.currentSidebarTab?.tab || this.defaultTab
+      if (this.currentSidebarTab && this.currentSidebarTab.tab) {
+        return this.currentSidebarTab.tab
+      }
+
+      return this.defaultTab
     },
 
     activeTabComponent() {
--- a/apps/files/src/components/FileSharingSidebar.vue
+++ b/apps/files/src/components/FileSharingSidebar.vue
@@ -143,9 +143,13 @@ export default {
     ...mapState(['user']),
 
     currentPanel() {
-      const panel = this.currentSidebarTab.options?.collaboratorsCurrentPanel
+      const tabOptions = this.currentSidebarTab.options
 
-      return panel || PANEL_SHOW
+      if (tabOptions && tabOptions.collaboratorsCurrentPanel) {
+        return tabOptions.collaboratorsCurrentPanel
+      }
+
+      return PANEL_SHOW
     },
 
     $_transitionGroupEnter() {
--- a/apps/files/src/mixins.js
+++ b/apps/files/src/mixins.js
@@ -193,7 +193,8 @@ export default {
 
     displayOverwriteDialog() {
       const translated = this.$gettext('File %{file} already exists')
-      const isVersioningEnabled = !this.publicPage() && this.capabilities.files?.versioning
+      const isVersioningEnabled =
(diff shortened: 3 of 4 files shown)


LukasHirt:
@phil-davis This PR only fixes errors happening when reading and extracting. Not sure if it will also fix the error with push since I cannot run that locally.

apps/files/src/components/FileDetails.vue
```diff
@@ -85,7 +85,11 @@ export default {
     },
 
     currentTab() {
-      return this.currentSidebarTab?.tab || this.defaultTab
+      if (this.currentTab) {
```
kulmann, line 88: Here and in the next line it has to be `this.currentSidebarTab` instead of `this.currentTab`

apps/files/src/components/FileDetails.vue
```diff
@@ -85,7 +85,11 @@ export default {
...
+        return this.currentTab.tab
```
kulmann, line 89: Is the `tab` field guaranteed to have a value? I guess so... but if not, then you would have to check that as well to have the same logic as before.

apps/files/src/components/FileSharingSidebar.vue
```diff
@@ -143,9 +143,13 @@ export default {
     ...mapState(['user']),
 
     currentPanel() {
-      const panel = this.currentSidebarTab.options?.collaboratorsCurrentPanel
+      const tabOptions = this.currentSidebarTab.options
 
-      return panel || PANEL_SHOW
+      if (tabOptions) {
+        return tabOptions.collaboratorsCurrentPanel
```
kulmann, line 149: Same as above, is `tabOptions.collaboratorsCurrentPanel` guaranteed to have a value? The logic we had before falls back to `PANEL_SHOW` if the options are not there or if the `collaboratorsCurrentPanel` is empty. New code only performs the first of the two checks.

apps/files/src/components/FileDetails.vue
LukasHirt: It shouldn't happen that it would be undefined or null but anything can go wrong so I added it back. Better safe than sorry

apps/files/src/components/FileDetails.vue
```diff
@@ -85,7 +85,11 @@ export default {
     },
 
     currentTab() {
-      return this.currentSidebarTab?.tab || this.defaultTab
+      if (this.currentSidebarTab) {
```
kulmann, line 88: What about `if(this.currentSidebarTab && this.currentSidebarTab.tab) {` here? Feels a bit weird to have to fallbacks to the same variable in place. ;-)

apps/files/src/components/FileSharingSidebar.vue
```diff
@@ -143,9 +143,13 @@ export default {
     ...mapState(['user']),
 
     currentPanel() {
-      const panel = this.currentSidebarTab.options?.collaboratorsCurrentPanel
+      const tabOptions = this.currentSidebarTab.options
 
-      return panel || PANEL_SHOW
+      if (tabOptions) {
```
kulmann, line 148: Same here, you could go with `if (tabOptions && tabOptions.collaboratorsCurrentPanel) {` to avoid falling back to the same value in two different places.

PR #3739, opened 2020-07-07
Feature/move

LukasHirt:
## Description
We've added move action to the files list. It is possible now to move folders and files.
The move is executed via a new page called location picker.

## Related Issue

- Closes https://github.com/owncloud/product/issues/101
- Refs https://github.com/owncloud/product/issues/108

## How Has This Been Tested?

- Acceptance tests

image

## Types of changes

- [ ] Technical debt
- [x] Tests


apps/files/src/components/FilesAppBar.vue
```diff
@@ -583,6 +606,22 @@ export default {
       }
       this.resetFileSelection()
       this.setHighlightedFile(null)
+    },
+
+    moveResources() {
+      const resources = JSON.parse(JSON.stringify(this.selectedFiles))
```
PVince81, line 612: strange gymnastics ? is that a trick to do a deep copy ?
would it make sense to only gather the file ids ?

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+    >
+      <template #headerColumns>
+        <div ref="headerNameColumn" class="uk-text-truncate uk-text-meta uk-width-expand">
+          <sortable-column-header
+            :aria-label="$gettext('Sort files by name')"
+            :is-active="fileSortField === 'name'"
+            :is-desc="fileSortDirectionDesc"
+            @click="toggleSort('name')"
+          >
+            <translate translate-context="Name column in files table">Name</translate>
```
PVince81, line 40: should we replace "files table" with "location picker" ?

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+
+      return resources
+    },
+
+    resourcesCount() {
+      return this.resources.length
+    },
+
+    target() {
+      return JSON.parse(JSON.stringify(this.$route.query.target))
```
PVince81, line 166: another weird copy

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+    },
+
+    basePath() {
+      return this.$route.path
+    },
+
+    resourcesQuery() {
+      return (
+        '&resource=' +
+        this.resources.map(resource => encodeURIComponent(resource)).join('&resource=')
```
PVince81, line 176: didn't we have a library already somewhere for encoding URL params ?

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+    },
+
+    breadcrumbs() {
+      const breadcrumbs = []
+
+      if (!this.isPublicPage) {
+        breadcrumbs.push({
+          index: 0,
+          text: this.$gettext('Home'),
+          to: this.basePath + `?action=${this.currentAction}&target=` + this.resourcesQuery
```
PVince81, line 191: encode action or use query builder library

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+
+        return this.$gettextInterpolate(title, { count: count })
+      }
+
+      return null
+    },
+
+    confirmBtnText() {
+      if (this.currentAction === 'move') {
+        return this.$gettext('Move here')
```
PVince81, line 236: add translation context, where is this text appearing ?

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        await promise
+          .move(resource, (target += '/' + resourceName), this.publicLinkPassword)
+          .catch(error => errors.push(error))
+      }
+
+      if (errors.length === 1) {
+        this.showMessage({
+          title: this.$gettext('Move has failed'),
```
PVince81, line 335: "An error occurred while moving ${file}" 

same below

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+          status: 'danger'
+        })
+
+        return
+      }
+
+      if (errors.length > 1) {
+        this.showMessage({
+          title: this.$gettext('Move has failed'),
+          desc: this.$gettext('Multiple resources could not be moved'),
```
PVince81, line 346: "Several" instead of "Multiple" ? should we say how many ?

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,376 @@
...
+          status: 'danger'
+        })
+        console.error('Move failed:', errors)
+
+        return
+      }
+
+      if (this.isPublicPage) {
```
PVince81, line 354: duplicate logic from "leaving picker" ?

apps/files/src/components/LocationPicker/MoveSidebarMainContent.vue
```diff
@@ -0,0 +1,22 @@
+<template functional>
+  <oc-grid gutter="small" child-width="1-1">
+    <translate translate-comment="Location picker guide on how to execute move."
+      >Navigate into the desired folder and move selected resources into it.</translate
+    >
+    <translate translate-comment="Location picker guide on how to navigate into a folder."
+      >You can navigate into a folder by clicking on its name.</translate
+    >
+    <translate translate-comment="Location picker guide on how to navigate to previous folders."
+      >To navigate back, you can use the breadcrumbs.</translate
```
PVince81, line 10: use => click on the breadcrumbs

apps/files/src/helpers/path.js
```diff
@@ -34,3 +34,16 @@ export function getParentPaths(path, includeCurrent = false) {
 
   return paths
 }
+
+/**
+ * Removes tha resource name and returns only the path
+ * @param {String} path Full resource path containing the name of the resource
+ * @return {String} path without the resource name
+ */
+export function getResourcePath(path) {
```
PVince81, line 43: isn't that the same as `path.dirname` (from the path or pathutil lib) ?

apps/files/src/helpers/resourceInfo.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Removes the path of the resource and returns only its name
+ * @param {String} path Full path of the resource containing the resource name
+ * @return {String} Resource name
+ */
+export function getResourceName(path) {
```
PVince81, line 6: isn't that the same as `pathutil.basename` ?

apps/files/src/store/actions.js
```diff
@@ -88,6 +88,9 @@ function _buildFile(file) {
     canShare: function() {
       return this.permissions.indexOf('R') >= 0
     },
+    canCreate: function() {
+      return this.permissions.indexOf('C') > -1
```
PVince81, line 92: consistency :eyes:

changelog/unreleased/move
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add move action
```
PVince81, line 1: "Ability to move files and folders into a different location"

apps/files/src/components/LocationPicker/LocationPicker.vue
DeepDiver1975: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

apps/files/src/components/LocationPicker/LocationPicker.vue
DeepDiver1975: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,369 @@
...
+      this.$router.push({ name: 'files-list', params: { item: target } })
+    },
+
+    async moveResources() {
+      const errors = []
+      const promise = this.isPublicPage ? this.$client.publicFiles : this.$client.files
+
+      for (const resource of this.resources) {
+        let target = this.target || '/'
+        const resourceName = pathUtil.basename(resource)
```
kulmann, line 314: I know that you just changed to using the `pathUtil.basename`. But one more remark on this: in `PublicLinkListItem.vue` and `Collaborator.vue` we were using the `basename` function like this:
`import { basename, dirname } from 'path'` and then just `basename(...)`. Would be nice to have consistent usages.

apps/files/src/components/LocationPicker/LocationPicker.vue
```diff
@@ -0,0 +1,369 @@
...
+        return
+      }
+
+      if (errors.length > 1) {
+        this.showMessage({
+          title: this.$gettext('Move has failed'),
+          desc: this.$gettext('Multiple resources could not be moved'),
+          status: 'danger'
+        })
+        console.error('Move failed:', errors)
```
kulmann, line 348: Debugging leftover?

apps/files/src/components/LocationPicker/MoveSidebarMainContent.vue
```diff
@@ -0,0 +1,22 @@
+<template functional>
```
kulmann, line 1: Is this visible, ever? When trying the move operation I had no sidebar and no topbar, thus no help content visible. Are there plans to change that? Otherwise I would consider this obsolete.

apps/files/src/components/Trashbin.vue
```diff
@@ -40,7 +40,12 @@
         </template>
         <template #rowColumns="{ item }">
           <div class="uk-width-expand">
-            <file-item :key="item.viewId" :item="item" :name="$_ocTrashbin_fileName(item)" />
+            <file-item
+              :key="item.viewId"
+              :item="item"
+              :name="$_ocTrashbin_fileName(item)"
+              :display-preview="false"
```
kulmann, line 47: Why hide the thumbnails within the trash bin? 🤔

apps/files/src/fileactions.js
```diff
@@ -30,14 +56,17 @@ export default {
...
+            const parent = pathUtil.dirname(resource.path)
+            this.$router.push({
+              name: 'location-picker',
+              query: { action: 'move', target: parent, resource: resource.path }
+            })
           },
-          isEnabled: function(item) {
-            return item.canDownload()
-          }
+          ariaLabel: () => this.$gettext('Move'),
```
kulmann, line 68: As Vincent pointed out, translation context would be great here :-)

apps/files/src/components/LocationPicker/LocationPicker.vue
LukasHirt: Is there an issue with using the native `encodeURIComponent`? If possible I'd prefer to keep it this way because using the `URLSearchParams` as @DeepDiver1975 linked seems to create a bit more code so if it wouldn't be necessary I'd like to avoid it.

apps/files/src/components/FilesAppBar.vue
LukasHirt: This copies only the values without the watchers, etc. related to the vuex state. By doing this we are safe that we don't get any errors later on in case of modifying the values + we do not modify the original state. It looks a bit odd but I haven't yet found in any vue issue/forum different way to do this. Well, I saw one with lodash but I'd prefer to avoid that 🤷

apps/files/src/components/LocationPicker/LocationPicker.vue
LukasHirt: See comment above

PVince81: should we maybe create our own utility function then ? the name of the function will already be explicit enough so we know what and why this is used

apps/files/src/components/LocationPicker/MoveSidebarMainContent.vue
LukasHirt: Was a bug. The sidebar is supposed to be visible. Will be fixed in the next commit.

apps/files/src/components/Trashbin.vue
LukasHirt: The thumbnails are not yet being displayed in trashbin at all and I do not think implementing that should be part of this PR so disabling them until it's done is good enough here IMO.

apps/files/src/components/LocationPicker/LocationPicker.vue
LukasHirt: It's left on purpose. Until we have added the logic to notifications to "stack" and then display all of the resources that failed, this is a "workaround" to be able to still see all the errors separately.

apps/files/src/components/LocationPicker/LocationPicker.vue
LukasHirt: Done

LukasHirt:
@PVince81 @kulmann All comments addressed. Pls re-review

LukasHirt:
Seems that the breadcrumbs step is failing on mobile resolution. I'll investigate

LukasHirt:
Found the issue with breadcrumbs. The tests found them twice on mobile resolution which resulted in error. Until we [fix the breadcrumbs in ODS](https://github.com/owncloud/owncloud-design-system/issues/809), I adjusted it to look for the mobile ones only if the "normal" ones are not found.

apps/files/src/components/Trashbin.vue
kulmann: 👍

apps/files/src/components/LocationPicker/LocationPicker.vue
kulmann: 👍

LukasHirt:
@kulmann I opened a ticket for the breadcrumbs bug #3753

PR #3740, opened 2020-07-07
Do not fallback to appId if the file action route is not defined

LukasHirt:
## Description
When opening a file in a editor or a viewer the path was falling back to an appId.
This made it impossible to navigate via the file actions into an app which doesn't have duplicate appId in the route.
We've stopped falling back to this value and in case that the route of the file action is not defined, we use the root path of the app.

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/356
- Refs https://github.com/owncloud/product/issues/69

## How Has This Been Tested?

- open pdf file
- open md file

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/apps/pdf-viewer/src/PdfViewer.vue
+++ b/apps/pdf-viewer/src/PdfViewer.vue
@@ -1,14 +1,15 @@
 <template lang="html">
-  <div id="pdf-viewer">
+  <div id="pdf-viewer" class="uk-flex uk-flex-column">
     <pdf-viewer-app-bar />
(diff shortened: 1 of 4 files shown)


LukasHirt:
> We should aim for having proper URLs on opened resources in file viewers

Absolutely agree 👍

kulmann:
> > We should aim for having proper URLs on opened resources in file viewers
> 
> Absolutely agree 👍

fyi, raised it in product board for next weeks product board https://jira.owncloud.com/browse/OC-2038

PR #3797, opened 2020-07-20
Use pre-signed URLs to download files

LukasHirt:
## Description
Get URL of the file, use sdk to pre-sign it and then download the file via it.

## Motivation and Context
No need to fetch the files first.

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/src/plugins/phoenix.js
+++ b/src/plugins/phoenix.js
@@ -24,20 +24,17 @@ export default {
         downloadFile(file, publicContext = null) {
           if (publicContext === null) {
             const url = this.$client.helpers._webdavUrl + file.path
-
-            return this.$client.signUrl(url).then(signedUrl => {
-              fetch(signedUrl, {
-                method: 'HEAD',
-                headers: { Authorization: 'Bearer ' + this.getToken }
-              }).then(res => {
-                if (res.status === 200) {
+            return fetch(url, {
+              method: 'HEAD',
+              headers: { Authorization: 'Bearer ' + this.getToken }
+            }).then(res => {
+              if (res.status === 200) {
+                this.$client.signUrl(url).then(signedUrl => {
                   const a = document.createElement('a')
                   a.style.display = 'none'
                   document.body.appendChild(a)
-
                   a.href = signedUrl
 
-                  console.log(file)
                   // Use download attribute to set desired file name
                   a.setAttribute('download', file.name)
 
@@ -46,14 +43,14 @@ export default {
 
                   // Cleanup
                   document.body.removeChild(a)
-                } else if (res.status === 404) {
-                  this.showMessage({
-                    title: this.$gettext('Download failed'),
-                    desc: this.$gettext('File could not be located'),
-                    status: 'danger'
-                  })
-                }
-              })
+                })
+              } else if (res.status === 404) {
+                this.showMessage({
+                  title: this.$gettext('Download failed'),
+                  desc: this.$gettext('File could not be located'),
+                  status: 'danger'
+                })
+              }
             })
           }
 
--- a/src/plugins/phoenix.js
+++ b/src/plugins/phoenix.js
@@ -11,7 +11,7 @@ export default {
           return this.$route.path.split('/')[1]
         },
 
-        isDownloadUrlSigningEnabled() {
+        isUrlSigningEnabled() {
           return this.capabilities.core && this.capabilities.core['support-url-signing']
         }
       },
@@ -45,7 +45,7 @@ export default {
           }
 
           // download with signing enabled
-          if (this.isDownloadUrlSigningEnabled) {
+          if (this.isUrlSigningEnabled) {
             try {
               const response = await fetch(url, {
                 method: 'HEAD',
@@ -77,12 +77,9 @@ export default {
 
           // download without url signing enabled
           this.addActionToProgress(file)
-          if (publicPage) {
-            return this.downloadFileFromUrlAsBlob(url, headers, file)
-          }
-          return this.downloadFileFromUrlAsBlob(url, headers, file)
+          return this.$_downloadFileAsBlob(url, headers, file)
         },
-        downloadFileFromUrlAsBlob(url, headers, file) {
+        $_downloadFileAsBlob(url, headers, file) {
           const request = new XMLHttpRequest()
           request.open('GET', url)
           request.responseType = 'blob'
--- a/src/plugins/phoenix.js
+++ b/src/plugins/phoenix.js
@@ -22,6 +22,14 @@ export default {
         },
         // FIXME: optional publicContext parameter is a mess
         downloadFile(file, publicContext = null) {
+          if (publicContext === null) {
+            const url = this.$client.helpers._webdavUrl + file.path
+
+            return this.$client.signUrl(url).then(signedUrl => {
+              window.location = signedUrl
+            })
+          }
+
           this.addActionToProgress(file)
           const publicPage = publicContext !== null ? publicContext : this.publicPage()
           let headers = {}
--- a/src/plugins/phoenix.js
+++ b/src/plugins/phoenix.js
@@ -26,7 +26,20 @@ export default {
             const url = this.$client.helpers._webdavUrl + file.path
 
             return this.$client.signUrl(url).then(signedUrl => {
-              window.location = signedUrl
+              fetch(signedUrl, {
+                method: 'HEAD',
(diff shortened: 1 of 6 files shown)


LukasHirt:
The download seems to work fine now apart from one issue - if the file was deleted before the download, it lands on a "Not found" page. This is breaking the behaviour we've had before where "only" a fail message was shown.

image

LukasHirt:
Trying to download files with oCIS results in `Error: Secret key is required`

LukasHirt:
I included a fetch with method `HEAD` to check if the file exists before download. This is not that nice so if anyone has a better suggestion I am listening 🙂

individual-it:
added QA team label to look into the tests

src/plugins/phoenix.js
```diff
@@ -22,6 +22,41 @@ export default {
         },
         // FIXME: optional publicContext parameter is a mess
         downloadFile(file, publicContext = null) {
+          if (publicContext === null) {
+            const url = this.$client.helpers._webdavUrl + file.path
+
+            return this.$client.signUrl(url).then(signedUrl => {
+              fetch(signedUrl, {
+                method: 'HEAD',
```
butonic, line 30: on the server side the signed url will still be checked, but since you are using HEAD it will never match because the signed URL includes the method (GET). Why even chech at all? if it is a 404 the browser will tell you that the file does not exist ... why do you want to double check that?

dpakach:
Ocis tests are failing with this logs in server
```
2020-07-22T16:22:10+05:45 ERR invalid transfer token error="datagateway: error validating transfer token: error parsing token: token contains an invalid number of segments" pkg=rhttp service=reva token=testimage.eps traceid=ab8e7c5fdfc78687a4f66717bc71c824
2020-07-22T16:22:10+05:45 WRN http end="22/Jul/2020:16:22:10 +0545" host=192.168.100.210 method=GET pkg=rhttp proto=HTTP/1.1 service=reva size=0 start="22/Jul/2020:16:22:10 +0545" status=403 time_ns=156803 traceid=ab8e7c5fdfc78687a4f66717bc71c824 uri=/data/user1/testimage.eps url=/data/user1/testimage.eps
2020-07-22T16:22:10+05:45 ERR http end="22/Jul/2020:16:22:10 +0545" host=127.0.0.1 method=GET pkg=rhttp proto=HTTP/1.1 service=reva size=0 start="22/Jul/2020:16:22:10 +0545" status=500 time_ns=40051953 traceid=ab8e7c5fdfc78687a4f66717bc71c824 uri=/remote.php/dav/files/user1/testimage.eps?a=1&c=6a41de85d8ebdd746d0ebf693d5f99c8&preview=1&scalingup=0 url=/remote.php/dav/files/user1/testimage.eps?a=1&c=6a41de85d8ebdd746d0ebf693d5f99c8&preview=1&scalingup=0

```
Something with the number of segments.

src/plugins/phoenix.js
```diff
@@ -22,6 +22,41 @@ export default {
         },
         // FIXME: optional publicContext parameter is a mess
         downloadFile(file, publicContext = null) {
+          if (publicContext === null) {
+            const url = this.$client.helpers._webdavUrl + file.path
+
+            return this.$client.signUrl(url).then(signedUrl => {
+              fetch(signedUrl, {
```
butonic, line 29: in any case this should be  `fetch(url, {`

dpakach:
@LukasHirt does this feature work with oc10 too? I was unable to reproduce the failure in CI, but when I try to download the file from the UI it sends the request to `http://172.17.0.1/core/ocs/v1.php/cloud/user/signing-key` URL and fails. This may be causing CI failure.

kulmann:
Adapted this 

> @LukasHirt does this feature work with oc10 too? I was unable to reproduce the failure in CI, but when I try to download the file from the UI it sends the request to `http://172.17.0.1/core/ocs/v1.php/cloud/user/signing-key` URL and fails. This may be causing CI failure.

Not supported in oc10, but the code was not taking the signing capability into consideration. That is fixed now, please retry. In oc10 the download should again fall back to the BLOB based download (which is limited to 2GB). In ocis, the signing capability is enabled, so ocis-web will trigger a download with presigned url.

kulmann:
@butonic @IljaN @LukasHirt please test and review, thanks!

src/plugins/phoenix.js
kulmann: please check the most recent solution

dpakach:
@kulmann  Looks like signing url works in oc10 too https://github.com/owncloud/core/pull/37634.
But now downloading does not work properly. If you download files like `.zip` then it works as expected but when downloading files like `.txt` it just opens the file content in the browser. I thank that is breaking the tests in CI right now.
When I try to download a file I get this
Screenshot from 2020-07-24 08-39-39

src/plugins/phoenix.js
kulmann: implemented it that way. We need to keep the HEAD request for now, to show errors in the web ui instead of seeing them in the download progress list of the browser.

LukasHirt:
@kulmann would maybe this endpoint help with the issue of files being opened? `/index.php/apps/files/ajax/download.php?${filePath}`

kulmann:
Wow.... @dpakach you're right, since owncloud core 10.5, downloads don't have a content disposition header anymore. Because of that, the respective file gets opened directly in the browser. It only worked in phoenix so far, because without presigned urls we were using blobs for download, without minding the content disposition header.

DeepDiver1975:
> @kulmann would maybe this endpoint help with the issue of files being opened? `/index.php/apps/files/ajax/download.php?${filePath}` Not sure about oCIS though shrug

Please only use this endpoint for zipped download - any other download must go via the webdav endpoints

kulmann:
> > @kulmann would maybe this endpoint help with the issue of files being opened? `/index.php/apps/files/ajax/download.php?${filePath}` Not sure about oCIS though shrug
> 
> Please only use this endpoint for zipped download - any other download must go via the webdav endpoints

Ok, then I won't bother trying, thanks!

@DeepDiver1975 do you happen to have an idea why the `Content-Disposition` header is missing when running with oc10.5? 🤔  Tests against oc10.5 are failing because the respective file directly opens in the browser, and thus all the tests don't have a chance to continue after triggering a file download.

kulmann:
> > @DeepDiver1975 do you happen to have an idea why the `Content-Disposition` header is missing when running with oc10.5? thinking Tests against oc10.5 are failing because the respective file directly opens in the browser, and thus all the tests don't have a chance to continue after triggering a file download.
> 
> is debug set to true in the config.php?

Wow 😱 Thank you @DeepDiver1975, that was it! 🤯

kulmann:
Probably CI running with debug enabled as well...

butonic:
for some more digging to understand what is causing the change in behavior follow https://github.com/owncloud/core/pull/25254#issuecomment-228880204

kulmann:
@individual-it could you clarify why we need the `debug` config variable set to `true`? It's set [here](https://github.com/owncloud/phoenix/blob/1f34eda92fa1bd95ac9e885cc2c49630550ac510/.drone.star#L1016) and a test run without it being set has basically all tests in a failing state, see here: https://drone.owncloud.com/owncloud/phoenix/10511/10/19

However, if it is set to true the download tests against oc10.5 are failing because of what [owncloud/core#25254 (comment)](https://github.com/owncloud/core/pull/25254#issuecomment-228880204).

Also, we should probably introduce a test suite for download tests with oc10.4, as those behave different than downloads on oc10.5:
- oc10.4: download as blob, because it doesn't have the capability to sign urls and downloads can't have authentication
- oc10.5: download with signed url

At the moment we're blocked, CI needs to be changed somehow...

individual-it:
the debug mode is needed for https://github.com/owncloud/openidconnect/blob/bbaaa9cb6e5bd8d9993701db5d7df03312a34087/lib/Client.php#L69-L71
otherwise the openidconnect app will not accept self-signed URLs 
```

**Comment by butonic** [comment]
created https://github.com/owncloud/openidconnect/pull/89 to separate the certificate checking configuration from debug mode

**Comment by butonic** [comment]
@jasson99 @individual-it with https://github.com/owncloud/openidconnect/pull/89it should be possible to switch from
```
'php occ config:system:set debug --value=true --type=bool',
```
to 
```
'php occ config:system:set openid-connect insecure --value=true --type=bool',
```

PR #3800, opened 2020-07-21
[Tests-only] Adjust browse to files page method to reduce waiting time

### LINKED ISSUE #3660: improve speed of tests
after login the tests are waiting for `#files-list-progress` to appear, looks like we are often too late the progressbar is already gone.
Make that smarter or at least decrease the timeout

dpakach:
## Description

Wait for the new file button to be loaded when browsing to the files page instead of the progress bar which disappears immediately and sometimes not found.

## Related Issue

- Fixes #3660

## Motivation and Context

if the progress bar is not found, the test just waits for 10 seconds until timeout which reduces the test execution time very greatly.

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/helpers/navigationHelper.js
+++ b/tests/acceptance/helpers/navigationHelper.js
@@ -18,5 +18,22 @@ module.exports = {
     return client
       .waitForElementPresent({ ...locator, abortOnFailure: false }) // don't fail if we are too late
       .waitForElementNotPresent(locator)
+  },
+
+  /**
+   * @param {string} url
+   * @param {object} ElementSelector - object that can have properties {selector [,locateStrategy]}
+   */
+  navigateAndWaitTillElementPresent: function(url, ElementSelector) {
+    client.url(url)
+    client.refresh() // IE11 needs an extra refresh because of the hash in the url
+    const locator = {}
+    if (typeof ElementSelector === 'object') {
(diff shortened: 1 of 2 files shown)


individual-it:
I'm not sure if that will be stable enough. What if the button is loaded but the AJAX call to load the file-list takes time or the creation of the DOM is slow.
What about simply having a shorter timeout in https://github.com/owncloud/phoenix/pull/3800/files#diff-c3b794a8b474e92cfa78c0cfb7dd3955R19 ?
and also get rid of the IE11 reload

dpakach:
> I'm not sure if that will be stable enough. What if the button is loaded but the AJAX call to load the file-list takes time or the creation of the DOM is slow.
> What about simply having a shorter timeout in https://github.com/owncloud/phoenix/pull/3800/files#diff-c3b794a8b474e92cfa78c0cfb7dd3955R19 ?
> and also get rid of the IE11 reload

@individual-it, I think this will be pretty stable. From what I see the new file button will be loaded only after the filelist is loaded and the fileprogress bar is gone. So it will wait at least until the file progress bar is gone (Which we used to do before this).
output

PR #3806, opened 2020-07-22
Fix settings

kulmann:
The data model in ocis-settings has changed. Because of that we need to align our usage of it.

- [x] requires owncloud-sdk release with [this pull request](https://github.com/owncloud/owncloud-sdk/pull/520) first
until that is merged, you can run `yarn build:system` in owncloud sdk on your local machine and then `yarn link` it in phoenix.


LukasHirt (approved):
LGTM

refs (approved):

PR #3808, opened 2020-07-22
Remove "Shared with:" from file list rows

pmaier1:
;)

Please check whether that works ^_^

Closes https://github.com/owncloud/product/issues/123
Closes https://github.com/owncloud/ocis/issues/408

### DIFF
--- a/apps/files/src/components/FileItem.vue
+++ b/apps/files/src/components/FileItem.vue
@@ -31,16 +31,14 @@
           v-text="'.' + item.extension"
         />
       </div>
-      <div v-if="hasTwoRows" class="uk-flex uk-flex-middle">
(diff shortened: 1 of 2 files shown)


kulmann:
Margins are looking good.
Screenshot 2020-07-22 at 10 03 34

However, I'm just thinking if we should default back to the "one line" row layout if there are no indicators, instead of showing the `-`. This is how it looks now:
Screenshot 2020-07-22 at 10 02 59

This is what I mean by "back to one line row layout":
Screenshot 2020-07-22 at 10 07 25

kulmann:
Screenshot 2020-07-22 at 10 12 49

@pmaier1 ^ this is what it would look like with one row for files without indicators.

kulmann:
@refs could you review this? then we can merge and include it in the release...

PR #3809, opened 2020-07-22
Remove head request

kulmann:
Same as https://github.com/owncloud/phoenix/pull/3797, but without the HEAD request. If a download fails, it just fails. No need to determine beforehand if it could fail.

### DIFF
--- a/src/plugins/phoenix.js
+++ b/src/plugins/phoenix.js
@@ -26,34 +26,28 @@ export default {
             const url = this.$client.helpers._webdavUrl + file.path
 
             return this.$client.signUrl(url).then(signedUrl => {
-              fetch(signedUrl, {
-                method: 'HEAD',
-                headers: { Authorization: 'Bearer ' + this.getToken }
-              }).then(res => {
-                if (res.status === 200) {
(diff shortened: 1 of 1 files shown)


kulmann:
@IljaN check out this PR and run `yarn build` in phoenix. Restarting ocis should not be required. If the download works in the browser, everything is fine.

IljaN:
Works for me:
signed_download

kulmann:
@IljaN the actual download is not visible in the network tab - only as file in your download list. did that appear?

IljaN:
Yeah sorry I see know that the image is cropped wrongly... Nothing more visible  in the Network tab. But browser Download-File picker appeared and I could download file.

kulmann:
Screenshot 2020-07-22 at 19 01 06

successfully downloaded a 6GB random data file (on localhost) with this PR.

kulmann:
failure handling needs to be fixed before we can merge this.

kulmann:
Closing this, since we decided to keep (and fix) the HEAD request for now. Otherwise we'd need to change all download related tests.

PR #3828, opened 2020-07-27
Removed docs how to work with hugo

LukasHirt:
## Motivation and Context
Moving into oCIS docs as it is the top level of our dev docs https://github.com/owncloud/ocis/pull/420


LukasHirt:
I tried rebasing this PR to see if the tests fails will be gone.

PR #3849, opened 2020-07-28
Display quota in sidebar and move information about current folder below files list

LukasHirt:
## Related Issue

- Fixes https://github.com/owncloud/product/issues/120
- Fixes https://github.com/owncloud/product/issues/121

## How Has This Been Tested?

- Set quota to user
- Fill quota above 80%
- Fill full files list with resources

image
image

### DIFF
--- a/apps/files/src/store/actions.js
+++ b/apps/files/src/store/actions.js
@@ -4,6 +4,7 @@ import { getParentPaths } from '../helpers/path'
 import { bitmaskToRole, permissionsBitmask } from '../helpers/collaborators'
 import { shareTypes } from '../helpers/shareTypes'
 import path from 'path'
+import SidebarQuota from '../components/SidebarQuota.vue'
 const { default: PQueue } = require('p-queue')
 
 function _extName(fileName) {
@@ -399,7 +400,20 @@ export default {
         .finally(() => {
           context.commit('UPDATE_FOLDER_LOADING', false)
           client.users.getUser(context.rootGetters.user.id).then(res => {
-            context.commit('CHECK_QUOTA', res.quota)
+            const quota = res.quota
+
+            context.commit('CHECK_QUOTA', quota)
+
+            // Display quota in the sidebar
+            if (
+              !isPublicPage &&
+              !context.getters.currentFolder.isMounted() &&
+              quota.definition !== 'default' &&
+              quota.definition !== 'none'
+            ) {
+              context.commit('SET_SIDEBAR_FOOTER_CONTENT_COMPONENT', SidebarQuota, { root: true })
+            }
+
             resolve()
           })
         })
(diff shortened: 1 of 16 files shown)


pmaier1:
Great work, love it :+1: I'd remove the "occupied" text as I think that's quite clear and would make it cleaner.

pmaier1:
> Please indicate warning not only with color change. Maybe add additional warning sign…

Even if you can't distinguish the colors you will notice that the bar is quite full, won't you? In addition the text below will tell you.
I really don't think we need another hint for that. Please, keep it simple and clean.

michaelstingl:
@pmaier1 then not use color change at all to treat everyone equal?

With the color change, you want to get extra-attention, so users can have a closer look about the problem. Why exclude people there?

pmaier1:
> @pmaier1 then not use color change at all to treat everyone equal?

If you think that it is more important to "treat everyone equally", then I'd remove the color indication instead of adding another indicator :man_shrugging:

pmaier1:
Ok, so let's not use the color for now to keep things simple without excluding anybody. Maybe later on another cool solution will come up.

LukasHirt:
@pmaier1 @michaelstingl New version without the variation and `occupied` text:

image

pmaier1:
Excellent!

src/Phoenix.vue
```diff
@@ -21,6 +21,23 @@
             <template v-if="sidebar.mainContentComponent" v-slot:mainContent>
               <component :is="sidebar.mainContentComponent" />
             </template>
+            <template v-if="isQuotaVisible" v-slot:footer>
+              <div class="uk-text-center">
+                <oc-progress
+                  :value="parseInt(quota.relative)"
```
LukasHirt, line 27: Adjust in state

LukasHirt:
- [ ] Discuss if quota should be visible from different extensions
- [ ] Test switching between extensions if the quota state is really available and how the load of it works

LukasHirt:
@PVince81 I added the extension point to display content in the footer of the sidebar. The quota is now included via that extension point and is only displayed in files app.

apps/files/src/store/actions.js
```diff
@@ -399,7 +400,20 @@ export default {
...
+            context.commit('CHECK_QUOTA', quota)
+
+            // Display quota in the sidebar
+            if (
+              !isPublicPage &&
+              !context.getters.currentFolder.isMounted() &&
+              quota.definition !== 'default' &&
+              quota.definition !== 'none'
+            ) {
+              context.commit('SET_SIDEBAR_FOOTER_CONTENT_COMPONENT', SidebarQuota, { root: true })
```
PVince81, line 414: is there a code path where maybe the quota might have changed and we need to set an empty Vue into that extension point ? I could imagine this happening if the user is switching between views/apps while the admin might have changed the quota from a true value to infinite. it's quite a corner case though

LukasHirt: Right - I added an `else` which will reset the component in there.

PR #3850, opened 2020-07-28
Add info about number of selected items and their size

LukasHirt:
## Description
We've added information about the number of selected items and their size above the files list next to batch actions.

## Related Issue

- Fixes https://github.com/owncloud/product/issues/122

## How Has This Been Tested?

- Select multiple resources in different files lists.

image
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/tests/acceptance/pageObjects/FilesPageElement/filesList.js
+++ b/tests/acceptance/pageObjects/FilesPageElement/filesList.js
@@ -457,7 +457,7 @@ module.exports = {
       await this.initAjaxCounters()
       await this.waitForElementVisible('@virtualScrollWrapper')
       await this.api.executeAsync(
-        function({ itemName, scrollWrapperSelector, listHeaderSelector }, done) {
+        function({ itemName, scrollWrapperSelector, listHeaderSelector, listItemSelector }, done) {
           const virtualScrollWrapper = document.querySelector(scrollWrapperSelector)
           const tableHeaderPosition = document
             .querySelector(listHeaderSelector)
@@ -484,7 +484,9 @@ module.exports = {
               return
             }
 
-            scrollDistance += virtualScrollWrapper.clientHeight
+            const listItemHeight = document.querySelector(listItemSelector).clientHeight
+
+            scrollDistance += listItemHeight * 5
             virtualScrollWrapper.scrollTop = scrollDistance
             setTimeout(function() {
               scrollUntilElementVisible()
@@ -497,7 +499,8 @@ module.exports = {
           {
             itemName: itemName,
             scrollWrapperSelector: this.elements.virtualScrollWrapper.selector,
-            listHeaderSelector: this.elements.filesTableHeader.selector
+            listHeaderSelector: this.elements.filesTableHeader.selector,
+            listItemSelector: this.elements.filesListItem.selector
           }
         ]
       )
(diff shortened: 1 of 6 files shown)


LukasHirt:
The interaction to clear selection doesn't look different at all currently. Is taken care of in https://github.com/owncloud/owncloud-design-system/pull/418

LukasHirt:
- [ ] remove selected text from buttons

LukasHirt:
> When you shrink the windows size so that the left sidebar disappears, the info about selected items and the clear selection button disappear entirely. Might be intended, but this way shrinking the windows causes the ui to lose information and one quite handy action (clearing the selection).

This is on purpose until we figure out the app bar layout on smaller screens. Adding there more elements makes it almost unusable already.

> selected suffix from the action buttons still needs to be removed ;-)

It was already removed in the previous commit. Just forgot to remove it now from tests - will do next

LukasHirt:
> Clear selection raw button has a height of 22px, while the text left of it has a height of 17px. Because of this, the Clear selection text overflows to the bottom, which looks off.

I've aligned it with flexbox.

> The pipe | between text and clear selection button is quite short. I would expect it to have the same height as all the other text.

With the flexbox in place, it should be now aligned as well. But since it's using font size it's possible that some screens might slightly differ.

Here's a screenshot with the new alignment:

image

image

tests/acceptance/pageObjects/FilesPageElement/filesList.js
```diff
@@ -484,7 +484,9 @@ module.exports = {
               return
             }
 
-            scrollDistance += virtualScrollWrapper.clientHeight
+            const listItemHeight = document.querySelector(listItemSelector).clientHeight
+
+            scrollDistance += listItemHeight * 5
```
LukasHirt, line 489: This will make the function slightly slower but it will make it much more stable. Some rows in the files list were skipped when doing a scroll with the visible height of the table. This should ensure that the row is always found (if it exists of course 🙂 )

kulmann: Nice! However, the `scroll item to the top of the list` didn't work. I removed it, since I don't see a reason why we need to scroll it back to the top of the list. If there is a reason, please elaborate. :-)

LukasHirt: Sometimes can happen that the actions dropdown is cut off (either top or bottom part) and selectors are then failing. Happening in small resolution tests. If it would be scrolled to the top we'd have then consistent behaviour on all resolutions and no menu would be cut off.

kulmann: Ah ok, thanks for the explanation. Might be best then to revert my commit and fix the `scroll item to top`, which was not working.

kulmann: iOS tests are still failing, so this is probably what you wanted to achieve...

LukasHirt: > iOS tests are still failing, so this is probably what you wanted to achieve...

Exactly. I'll try to debug why the scroll was failing.

LukasHirt: Reverted all new scrolling magic and skipped the tests on iPhone instead until we have a fix for the actions dropdown.

PR #3852, opened 2020-07-29
[Tests-Only]Fixed failing test on public link share

Talank:
## Description
The test `Scenario: user edits the password of an already existing public link and tries to access with old password` was failing because two shares were created fast with the same `stime`.  So, a new function is created to wait for at least one second between the creation of shares.

## Related Issue
- Fixes https://github.com/owncloud/phoenix/issues/3830

## How Has This Been Tested?
- :robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/stepDefinitions/sharingContext.js
+++ b/tests/acceptance/stepDefinitions/sharingContext.js
@@ -13,6 +13,7 @@ const _ = require('lodash')
 const path = require('../helpers/path')
 const util = require('util')
 const { COLLABORATOR_PERMISSION_ARRAY, calculateDate } = require('../helpers/sharingHelper')
+let timeOfLastShareOperation = Date.now()
 
 /**
  *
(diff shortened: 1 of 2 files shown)


tests/acceptance/stepDefinitions/sharingContext.js
```diff
@@ -107,6 +108,17 @@ Given(
...
 
+/**
+ * makes sure share operations are carried out maximum once a second to avoid same stime
+ */
+const waitBetweenShareOperations = async function() {
+  const timeSinceLastShare = Date.now() - timeOfLastShareOperation
+  if (timeSinceLastShare <= 1001) {
+    await client.pause(1001 - timeSinceLastShare)
+  }
+  timeOfLastShareOperation = Date.now()
```
dpakach, line 119: I think this should be updated after sharing as sharing might take some time

tests/acceptance/stepDefinitions/sharingContext.js
```diff
@@ -510,6 +529,7 @@ Given(
       null,
       { password: password }
     )
+    await updateTimeOfLastShareOperation()
```
dpakach, line 532: add this line in `shareFilefolder()` function as there are other steps too that do share.

PR #3938, opened 2020-08-11
Differentiate between username and user-id

LukasHirt:
## Description
With oCIS user-id and username are not the same as is the case in ownCloud 10.
We've started differentiating between them to correctly display all information in the accounts page.
If the username is not available (oC10), we fall back to using user-id as the username.

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/440

## How Has This Been Tested?

- test environment: oC10 and oCIS

image

### DIFF
--- a/src/store/user.js
+++ b/src/store/user.js
@@ -77,7 +77,8 @@ const actions = {
                 context.commit('SET_CAPABILITIES', cap)
                 context.commit('SET_USER', {
                   id: res.id,
-                  displayname: res['display-name'],
+                  username: res.username,
+                  displayname: res.displayname || res['display-name'],
                   email: !Object.keys(res.email).length ? '' : res.email,
                   token,
                   isAuthenticated: true,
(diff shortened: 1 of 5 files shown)


src/store/user.js
```diff
@@ -77,7 +77,8 @@ const actions = {
                 context.commit('SET_CAPABILITIES', cap)
                 context.commit('SET_USER', {
                   id: res.id,
-                  displayname: res['display-name'],
+                  username: res.username,
+                  displayname: res.displayname || res['display-name'],
```
PVince81, line 81: seems you also fixed display name here but didn't mention it in changelog

LukasHirt: A forgot that one - added an item now :wink:

PVince81:
some test failure

kulmann:
CI failure was a random unrelated failure. Tracking it in another ticket.

PR #4005, opened 2020-08-31
[tests-only] Add ocis-accounts default admin user to tests

kulmann:
This PR adds the default admin user of ocis-accounts to the `userSettings.js` helper.


dpakach (approved):
LGTM :+1:

PR #4083, opened 2020-09-18
Remember public link password on refresh

PVince81:
## Description
Stores the public link password in sessionStorage so that it survives
page refresh. Closing the browser tab or to another tab doesn't preserve
it.

## Related Issue
Part 8b of https://github.com/owncloud/product/issues/231

## How Has This Been Tested?
After entering password on public link page:

- [x] TEST: refresh the page doesn't ask for password again
- [x] TEST: closing page and opening the URL again asks for the URL again (so password is forgotten correctly)
- [x] TEST: change public link password to create a discrepancy with saved one: password is asked again

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/src/components/PublicLinks/PublicLink.vue
+++ b/apps/files/src/components/PublicLinks/PublicLink.vue
@@ -55,6 +55,7 @@ export default {
       firstTime: true,
       passwordRequired: false,
       password: null,
+      savedPassword: null,
       inputErrorMessage: null
     }
   },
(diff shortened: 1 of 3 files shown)


PVince81:
One little issue though is that the page blinks:

1. When opening the public link, you first see the file list loading.
2. Then the PROPFIND fails, it goes to the "public link password" vue.
3. In that vue, it gets the saved password and shows "loading public link" spinner
4. Then back to file list

Ideally we should already read the stored password earlier to avoid going to that "loading public link" page.
Maybe something for later ?

LukasHirt (approved):
> Ideally we should already read the stored password earlier to avoid going to that "loading public link" page.
Maybe something for later ?

Do you think it would be simple to fix? If it would be something quick I'd vote for taking care of it already in this PR. Depending on how much disturbing is the jump between list and loading page. Do not want to block this PR though so taking care of it separately would be also fine for me

PVince81:
I have ideas for alternative approaches, I'll give them a try

PVince81:
alternative PR: https://github.com/owncloud/phoenix/pull/4084

PR #4086, opened 2020-09-18
Update ods 1.10.1

kulmann:


LukasHirt (approved):
Checked the UI and didn't find any more buttons which would be broken or not adjusted to the updated component. Awesome job! 🚀

PR #4112, opened 2020-09-28
[tests-only] Fix mysql in drone

kulmann:
Based on https://bugzilla.redhat.com/show_bug.cgi?id=1281733


kulmann:
looks like this is not needed

PR #4125, opened 2020-10-01
[Tests-Only] fix xpath of newfile button in filespage in ui tests

dpakach:
## Description

Fix the xpath of new button in files page for UI tests.

## Related Issue

- Fixes test performance by some margin

## Motivation and Context

Since the element didn't exist the tests would wait for 10 seconds every time
```
Given the user has browsed to the files page
Timed out while waiting for element <//button[@id="new-file-menu-btn" and contains(@class, "uk-button-primary")]> to be present for 10000 milliseconds. - expected "found" but got: "not found" (10005ms)
```

## How Has This Been Tested?

:robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/pageObjects/filesPage.js
+++ b/tests/acceptance/pageObjects/filesPage.js
@@ -405,7 +405,7 @@ module.exports = {
       selector: '#files-breadcrumb li:nth-of-type(2)'
     },
     newFileButtonLoaded: {
-      selector: '//button[@id="new-file-menu-btn" and contains(@class, "uk-button-primary")]',
+      selector: '//button[@id="new-file-menu-btn" and contains(@class, "oc-button-primary")]',
       locateStrategy: 'xpath'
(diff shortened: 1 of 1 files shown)


HariBhandari07 (commented):
LGTM

phil-davis:
The "new file" button should always be found after browsing to the top-level files page.
If the button is not found within 10 seconds, then I expected that the step should fail.

If there are scenarios where the files page gets displayed without the "new file" button, then for those cases we should not even look/wait for the button. (Maybe this case never happens)

phil-davis (approved):
This is good to fix test performance.
Maybe we should also fail if the button has not appeared within the timeout?

PR #4137, opened 2020-10-07
[Tests-only] Bump ocis commit id to get a fix about autoprovisioned users in

kulmann:
Bump ocis commit id to get a fix about autoprovisioned users in

See https://github.com/owncloud/product/issues/219 for details

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -134,7 +134,7 @@ config = {
 	'defaults': {
 		'acceptance': {
 			'ocisBranch': 'master',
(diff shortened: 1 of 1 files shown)


phil-davis:
CI fails to run with OCIS because the "new OCIS" is the mono-repo. IMO it needs issue https://github.com/owncloud/product/issues/246 "Phoenix CI: use ocis from new mono repo" to also be sorted out here.

@individual-it is looking at it.

kulmann:
Will be solved in https://github.com/owncloud/phoenix/pull/4139 I guess 👍

PR #4138, opened 2020-10-07
Enable autoredirect to IdP

LukasHirt:
We've added a key into the theme to enable auto-redirect to the IdP when entering ocis-web instead of displaying the login page first. The default value is set to true.

### DIFF
--- a/changelog/unreleased/login-autoredirect
+++ b/changelog/unreleased/login-autoredirect
@@ -1,0 +1,6 @@
+Change: Enable autoredirect to the IdP
+
+We've added a key into the theme to enable autoredirect to the IdP when entering ocis-web instead of displaying the login page first.
+The default value is set to true.
+
+https://github.com/owncloud/phoenix/pull/4138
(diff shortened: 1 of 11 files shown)


LukasHirt:
Originally I tried to do this in the router but that required to handle this as well in the logout logic and had to take care that no components are displayed before the redirect. Handling this at the login page makes it much easier.

LukasHirt:
- [ ] Adjust acceptance tests

kulmann (commented):
One thing I noticed when trying this is, that you get redirected to the IdP login page directly after logout. I guess that's want we want - but still, wanted to point it out. @pmaier1 @LukasHirt

LukasHirt:
Adjusted tests - let's see if it needs some more love or if that is enough. I'd love to add also a test for displaying the login page when it's set but I guess this won't be possible until we have some endpoint for theming where it could be changed dynamically?

kulmann (approved):
Awesome 🚀

pmaier1:
Nice, THX! Great to see this coming!

> One thing I noticed when trying this is, that you get redirected to the IdP login page directly after logout. I guess that's want we want - but still, wanted to point it out. @pmaier1 @LukasHirt

Yes, that's the expected behavior. Thanks for pointing out, though :+1:

pmaier1:
Fixes https://github.com/owncloud/product/issues/170
Fixes https://github.com/owncloud/ocis/issues/432

PR #4162, opened 2020-10-14
Move folder loading switch outside of finally method

LukasHirt:
## Description
It seems that the `finally` method was being called too early. Moving the switch of loading into `then` and `catch` solved the flashy empty folder messages.

## Related Issue

- Fixes https://github.com/owncloud/product/issues/263

### DIFF
--- a/changelog/unreleased/flashy-messages-when-loading-folders
+++ b/changelog/unreleased/flashy-messages-when-loading-folders
@@ -1,0 +1,6 @@
+Bugfix: Do not display "empty folder" message when there is any content
+
+We've fixed that when some of the file/share lists were being loaded, the "empty folder" message sometimes briefly appeared even though the list wasn't empty.
+
+https://github.com/owncloud/product/issues/263
+https://github.com/owncloud/phoenix/pull/4162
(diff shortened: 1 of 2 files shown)


changelog/unreleased/flashy-messages-when-loading-folders
```diff
@@ -0,0 +1,6 @@
+Bugfix: Do not display "empty folder" message when there is any content
+
+We've fixed that when some of the list were being loaded, sometimes briefly appeared the "empty folder" message even though the list wasn't empty.
```
kulmann, line 3: Proposing: `We've fixed that when some of the file/share lists were being loaded, the "empty folder" message sometimes briefly appeared even though the list wasn't empty.`

kulmann:
CI failed. Don't know if it's a random failure.. I restarted it, let's see what happens.

changelog/unreleased/flashy-messages-when-loading-folders
LukasHirt: Done, thanks! :wink:

kulmann (approved):
🚀

PR #4204, opened 2020-10-19
[tests-only] make sure userId is used as string

individual-it:
## Description
`userId` is a single value array e.g. `[user1]`, to transform it into a string before sending it to the next function

## Motivation and Context
`join()` function fails with `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array`

## How Has This Been Tested?
:robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/stepDefinitions/provisioningContext.js
+++ b/tests/acceptance/stepDefinitions/provisioningContext.js
@@ -178,7 +178,7 @@ Given('these users have been created with default attributes but not initialized
 ) {
   return Promise.all(
     dataTable.rows().map(userId => {
-      return deleteUser(userId).then(() => createDefaultUser(userId))
+      return deleteUser(userId.toString()).then(() => createDefaultUser(userId.toString()))
     })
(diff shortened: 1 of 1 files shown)


kulmann (approved):
👍

tests/acceptance/stepDefinitions/provisioningContext.js
```diff
@@ -178,7 +178,7 @@ Given('these users have been created with default attributes but not initialized
 ) {
   return Promise.all(
     dataTable.rows().map(userId => {
```
kulmann, line 180: In order to keep the scope of `map` minimal, you could also introduce another `map` which just does `.map(userIds => userIds.toString())` before the map that deletes the user.

PR #4211, opened 2020-10-20
Fix logout with OIDC

LukasHirt:
## Description
Open signout URL directly.

## Related Issue

- Fixes https://github.com/owncloud/product/issues/266


kulmann (approved):
Looks good 👍  Just for the record: with this change we log out the user from the IdP. If it's a company wide IdP and we are just one application, this might be annoying. A better solution might be to unload the user and the user to click some "Login" button (without needing to enter credentials) when the user is still logged in at the IdP. CC @pmaier1

PR #4216, opened 2020-10-20
Attach share permission to roles

LukasHirt:
## Description
We've attached the share permission of collaborators to roles. There is no longer a share additional permission. In oCIS, the share permission is hidden as it is not yet possible to enable re-sharing.

### DIFF
--- a/apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
+++ b/apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
@@ -6,14 +6,16 @@
       :selected-role="role"
       @roleSelected="selectRole"
     />
-    <label v-if="$_ocCollaborators_hasAdditionalPermissions" class="oc-label">
-      <translate>Additional permissions:</translate>
-    </label>
-    <additional-permissions
-      :available-permissions="role.additionalPermissions"
-      :collaborators-permissions="collaboratorsPermissions"
-      @permissionChecked="checkAdditionalPermissions"
-    />
+    <template v-if="$_ocCollaborators_hasAdditionalPermissions">
+      <label v-if="selectedRole.name !== 'advnacedRole'" class="oc-label">
+        <translate>Additional permissions:</translate>
+      </label>
+      <additional-permissions
+        :available-permissions="role.additionalPermissions"
+        :collaborators-permissions="collaboratorsPermissions"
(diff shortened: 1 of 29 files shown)


apps/files/src/helpers/collaborators.js
```diff
@@ -42,10 +42,10 @@ export function roleToBitmask(role, additionalPermissions = []) {
  * @param {boolean} isFolder Defines if the item is folder
  * @returns {object} Role mapped to the bitmask
  */
-export function bitmaskToRole(bitmask, isFolder) {
+export function bitmaskToRole(bitmask, isFolder, allowSharePerm) {
```
kulmann, line 45: Add new param to docblock

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
```diff
@@ -106,7 +108,7 @@ export default {
     },
 
     $_ocCollaborators_hasAdditionalPermissions() {
-      if (this.selectedRole && this.selectedRole.name !== 'advancedRole') {
+      if (this.selectedRole && this.selectedRole.additionalPermissions) {
```
kulmann, line 111: The body of this new if statement is always `true`. So you could simplify the whole function body to `return this.selectedRole && this.selectedRole.additionalPermissions`.

apps/files/src/components/Collaborators/CollaboratorsEditOptions.vue
```diff
@@ -6,14 +6,16 @@
...
-    <label v-if="$_ocCollaborators_hasAdditionalPermissions" class="oc-label">
-      <translate>Additional permissions:</translate>
-    </label>
-    <additional-permissions
-      :available-permissions="role.additionalPermissions"
-      :collaborators-permissions="collaboratorsPermissions"
-      @permissionChecked="checkAdditionalPermissions"
-    />
+    <template v-if="$_ocCollaborators_hasAdditionalPermissions">
+      <label v-if="selectedRole.name !== 'advnacedRole'" class="oc-label">
```
kulmann, line 10: Typo in the string `advnacedRole`, needs to be `advancedRole`.

kulmann (changes requested):
Found some minor issues

kulmann (approved):

Issue #4102, opened 2020-09-24
Share received inside `share_folder` cannot be unshared


dpakach:
**Tested with OC10(latest git master)**
By default the shares are received in the root folder. But we can set config `share_folder` to set default share receive folder.
### Steps to reproduce
1. Set share folder for receiving share.
2. Login as a user and share a file to another user.
3. Try to login as the receiver and delete the file.

### Expected behaviour
The delete option should be available

### Actual behaviour
The delete option is not available
![Screenshot from 2020-09-24 11-52-04](https://user-images.githubusercontent.com/19869344/94110933-17a30b00-fe63-11ea-8733-6ae9d50a8b2d.png)

LukasHirt:
Will be fixed in #4255

PR #4255, opened 2020-10-30
Move actions to sidebar

LukasHirt:
## Description
Moving resource actions from file list into the sidebar.

## How Has This Been Tested?

- test environment: Drone and manually

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Open tasks:

~~- [ ] Improve styles of buttons~~
~~- [ ] Adjust tests~~
- [x] Add changelog item
- [x] add `folder-open` icon from fontawesome for the `navigate` action
- [x] find an icon for the `actions` accordion item
- [x] unskip test that failed due to missing scrolling in actions dropdown


LukasHirt:
Fixes #4102

LukasHirt:
Skipping the failing scenario because it's actually an issue in master as well #4310 tested both locally and on ocis.owncloud.works. Not sure at all how the test could pass in other PRs then 🤷

kulmann (changes requested):
The code itself looks good and the consistency of having everything in the right sidebar is super nice now.
Unfortunately I found some beef on this PR while trying it out again locally.

"shared with me" page:
- There is a "download" action for folders
- "mark as favorite" doesn't work
- if status of share = pending -> we shouldn't allow anything! maybe we can just hide the accordion entirely and display a "please accept or decline this share" message?
- "rename" action should only be available if the sharing role permits it

"shared with others" page:
- file details (top row above the accordion) looks different: missing favorites-star, missing file details
- folders have a "Download" action. Should not be available.

"deleted files" page:
- actions accordion item is not collapsible. Stays open on click.

Also, we wanted to re-activate a test scenario that was disabled in a previous PR (issue with scrolling in actions dropdown). Added a checkbox to the PR description for this.

kulmann:
Screenshot 2020-11-13 at 15 13 13

I added a commit that does two tiny things (see screenshot). Please revert the commit if you don't agree. But we should at least find an icon for the actions accordion item, so that all the accordion items have one.
a) Set an icon for the actions accordion item. Please check if you agree with the icon. A "play button" makes a reasonable impression for me.
b) Added a slight horizontal padding for the contents of the accordion items. Without it I had a pretty hard time to find a visible separation between the respective accordion items.

LukasHirt:
Pushed a new commit yesterday which tackles those bugs @kulmann Some notes though:

> "mark as favorite" doesn't work

Removed this action from shared lists since we do not have `starred` value when building resources there. I would see this out of the scope of this PR to tackle that.

> "rename" action should only be available if the sharing role permits it

Rename should be always available. Renaming shared resource shouldn't rename the original one but only the local one for the user.

> actions accordion item is not collapsible. Stays open on click.

Was happening everywhere because sometimes the opened accordion was set to `null` and `open-first` took care of opening the accordion which resulted on trying to change `opened-accordion-id` from `null` to `null`. Fixed as well.

kulmann (changes requested):
Awesome, thanks for adjusting!

More small things that I found:
1: "shared with others" and "shared with me" pages
- when you select multiple files, there should only be the `Delete` batch action. Please remove the `Copy` and `Move` batch actions there.

2: "shared with others" page
- I can't reproduce it, but I had an issue navigating into a folder from the ride sidebar of this page.

kulmann (approved):
Yeeeees, awesome! 🚀 👍 🥇

Issue #4266, opened 2020-11-04
Remove skipOnOCIS tags from acceptance tests


phil-davis:
Currently there are Phoenix UI tests with `skipOnOCIS` tag. Maybe it would be possible to run these in OCIS and have an expected-failures file in `owncloud-ocis` like we do for core API acceptance tests.

That will require some way to process the nightwatch UI test result output, get a list of scenarios that failed and compare it to the expected failures.

- [x] see if nightwatch can produce a nice summary output of failed scenarios - see #4268 
- [x] think if there is a practical way to process the failed scenarios and compare with a list of expected-failures

phil-davis:
See PR #4268 for a demo of the failing test output that cucumber-js can produce.
Maybe we can parse that with some script and compare it to an `expected-failures.txt` file?

phil-davis:
PR #4512 implements the `run.sh` script with currently-empty expected-failures files.

phil-davis:
PR #4534 removes most `skipOnOCIS` tags and adds to the expected-failures file. When that PR is finished then this issue can be closed.

phil-davis:
PR #4534 was merged.

PR #4268, opened 2020-11-04
[tests-only] Try different test output formats

phil-davis:
investigation for issue #4266 

https://drone.owncloud.com/owncloud/phoenix/11654/12/20
```
The acceptance test run exited with error status 1
```

`failed-scenarios.txt` has:
```
tests/acceptance/features/webUIFiles/copy.feature:10:10:89:89
tests/acceptance/features/webUIFiles/versions.feature:14:14
```

It seems to give 1 line for each feature that has a fail. That line contains the feature path and a list of the scenario lines that failed. Each scenario that failed is listed twice, I think that is because these UI  scenarios are being run twice, and so they fail twice

That information can be quite easily parsed into an array with strings like:
```
webUIFiles/copy.feature:10
webUIFiles/copy.feature:89
webUIFiles/versions.feature:14
```

which can then be compared with entries for the `webUIFiles` suite in an `expected-failures.txt` file, and differences reported.


phil-davis:
Issue #4266 is done.

PR #4316, opened 2020-11-17
[tests-only] Bump OCIS commit id

phil-davis:
To the current https://github.com/owncloud/ocis/commits/master

This will show that the current Phoenix and OCIS master branches work together.

Then we can look at why https://github.com/owncloud/ocis/pull/868 is having trouble.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -298,7 +298,7 @@ config = {
 	'defaults': {
 		'acceptance': {
 			'ocisBranch': 'master',
-			'ocisCommit': '7f1eef36d887b5955a178d9a1c82b8a9b0113675',
+			'ocisCommit': 'd1543f76a29c06ab2640c613a5429e401f23ce2a',
 		},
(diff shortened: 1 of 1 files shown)


individual-it:
are the failing tests just random or real issues?

phil-davis:
First example
https://drone.owncloud.com/owncloud/phoenix/11776/71/11 OCISTrashbinFilesFolders passed but after restart
https://drone.owncloud.com/owncloud/phoenix/11778/71/11 OCISTrashbinFilesFolders failed.

and other suites failed also which would need looking at.

I restarted again.

The commit id needs to be "constantly" updated so that we know quickly if something in OCIS makes the tests less reliable.

individual-it:
I've rebased (force-pushed) and updated the commit id of ocis again to the current tip of the branch

individual-it (approved):

Issue #4335, opened 2020-11-18
[Bug] Messages not shown in fullscreen apps


VirtualEvan:
### Steps to reproduce
1. Define an app as full screen
    ```js
    {
      name: 'draw-io-edit',
      path: '/edit/:filePath',
      components: {
        fullscreen: DrawIoEditor
      },
      meta: { hideHeadbar: true }
    }
    ```
2. Use the `showMessage` function

The message is shown if the `fullscreen` property is changed to `app` and `hideHeadbar` is removed
```js
{
  name: 'draw-io-edit',
  path: '/edit/:filePath',
  components: {
    app: DrawIoEditor
  }
}
```

### Expected behaviour
The speified message should be shown

### Actual behaviour
The message is not displayed

pascalwengerter:
Hey @VirtualEvan, is this issue still open after your PR has been merged? :)

VirtualEvan:
Yes, this issue is something I found while working in the draw.io app, my PR did not addressed it

kulmann:
Apps are now always forced into an application layout which contains the message-bar component and handles the active messages. The fullscreen mode has been removed entirely in https://github.com/owncloud/web/pull/6401 - closing this issue here as solved.

PR #4337, opened 2020-11-18
[draw.io] Deployment config + .vsdx support

### LINKED ISSUE #4327: [Feature] draw.io vsdx support
Right now the only way to import a .vsdx (MS Visio files) is to create a new diagram and import the vsdx file from the local storage.
Draw.io should be able to open .vsdx directly from OC by using post messages. [Embed mode docs](https://desk.draw.io/support/solutions/articles/16000042544-embed-mode)

### LINKED ISSUE #4328: [Feature] Add custom draw.io settings
Due to data privacy reasons, our org is hosting a private draw.io server.
It would be nice to have an easy way to point OC to custom servers rather than the official draw.io one.
Additionally to this, support for specific settings could be added, i.e. enable autosave or select theme

VirtualEvan:
## Description

1. This PR allows to import .vsdx (Microsoft Visio files) directly from OC.
2. Once the file is open, if saved, a new `<filename>_<timestamp>.drawio` file is created with the modifications. (This avoids overwriting the original file with XML content). A message informing this should be shown ( #4335 )
3. By default, the draw-io app now points to `https://embed.diagrams.net` (The official embed-specific draw.io server)
4. If specified, the configuration in `config.json` will be used for the connections to draw.io (server url, autosave and theme)
    ```json
    {
      ...
      "applications": [
        {
          "title": {
            "en": "draw-io"
          },
          "url": "https://custom.draw.io.url",
          "autosave": true,
          "theme": "atlas"
        }
      ]
    }
    ```

## Related Issue

- Closes #4327 
- Closes #4328 

## Motivation and Context

- Privacy reasons
- Flexibility
- Support for more formats

## How Has This Been Tested?

- test environment:
Docker OC 10 + local Phoenix: following [this docs](https://owncloud.github.io/clients/web/getting-started)
- test case 1:
Use custom draw.io server, enable autosave, use the 'atlas' theme
- test case 2:
Open a .vsdx file by clicking on it

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] Import .vsdx files directly from OwnCloud
- [x] Custom draw.io server support
- [x] Autosave support 
- [x] Different themes support 
- [x] Specify draw.io settings in the global configuration
- [x] Use by default the embed-specific diagrams.net server
- [x] View mode for read-only diagrams

### DIFF
--- a/apps/draw-io/src/DrawIoEditor.vue
+++ b/apps/draw-io/src/DrawIoEditor.vue
@@ -26,7 +26,7 @@ export default {
         ui: 'minimal'
       })
 
-      return 'https://www.draw.io?' + query
+      return `https://embed.diagrams.net?${query}`
     }
   },
   created() {
--- a/apps/draw-io/src/DrawIoEditor.vue
+++ b/apps/draw-io/src/DrawIoEditor.vue
@@ -3,9 +3,9 @@
 </template>
 <script>
 import { mapGetters, mapActions } from 'vuex'
+import { basename } from 'path'
 import queryString from 'query-string'
 import moment from 'moment'
-import { basename } from 'path'
 
 export default {
   name: 'DrawIoEditor',
@@ -15,27 +15,32 @@ export default {
     currentETag: null
   }),
   computed: {
-    ...mapGetters(['getToken']),
+    ...mapGetters(['getToken', 'configuration']),
     loading() {
       return this.content === ''
+    },
+    config() {
+      const { url = 'https://embed.diagrams.net', theme = 'minimal', autosave = false } =
+        this.configuration.applications.find(app => app.title.en === 'draw-io') || {}
+      return { url, theme, autosave: autosave ? 1 : 0 }
     },
     iframeSource() {
       const query = queryString.stringify({
         embed: 1,
+        chrome: 1,
         picker: 0,
         stealth: 1,
         spin: 1,
         proto: 'json',
-        ui: 'minimal'
+        ui: this.config.theme
       })
 
-      return `https://embed.diagrams.net?${query}`
+      return `${this.config.url}?${query}`
     }
   },
   created() {
     this.filePath = this.$route.params.filePath
     this.fileExtension = this.filePath.split('.').pop()
-
     window.addEventListener('message', event => {
       if (event.data.length > 0) {
         var payload = JSON.parse(event.data)
@@ -75,7 +80,7 @@ export default {
             JSON.stringify({
               action: 'load',
               xml: resp.body,
-              autosave: 1
+              autosave: this.config.autosave
             }),
             '*'
           )
@@ -120,7 +125,7 @@ export default {
               JSON.stringify({
                 action: 'load',
                 xml: reader.result,
-                autosave: 1
+                autosave: this.config.autosave
               }),
               '*'
             )
--- a/apps/draw-io/src/DrawIoEditor.vue
+++ b/apps/draw-io/src/DrawIoEditor.vue
@@ -9,6 +9,7 @@ export default {
   name: 'DrawIoEditor',
   data: () => ({
     filePath: '',
+    fileExtension: '',
     currentETag: null
   }),
   computed: {
@@ -31,14 +32,14 @@ export default {
   },
   created() {
     this.filePath = this.$route.params.filePath
-    this.extension = this.filePath.split('.').pop()
+    this.fileExtension = this.filePath.split('.').pop()
 
     window.addEventListener('message', event => {
       if (event.data.length > 0) {
         var payload = JSON.parse(event.data)
         switch (payload.event) {
           case 'init':
-            this.extension === 'vsdx' ? this.importVisio() : this.load()
+            this.fileExtension === 'vsdx' ? this.importVisio() : this.load()
             break
           case 'autosave':
           case 'save':
@@ -63,7 +64,6 @@ export default {
         }
       })
     },
-
     load() {
       this.$client.files
         .getFileContents(this.filePath, { resolveWithResponseObject: true })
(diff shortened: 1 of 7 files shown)


VirtualEvan:
I have a couple of questions regarding the implementation of these functionalities.

1. I could not find a way to get the file permissions from inside an app, can that be achieved somehow? I am interest in knowing if the current user has read-only permissions on the open diagram.

2. The only place I found situable to put the draw.io configuration settings was the "applications" array in `config.json` as there are similar configurations for other apps. However, as it is an array, it is necessary to filter by the `title` and `en` (english name), in order to get the right app.
Is there some other place where this information could be set in a more optimal and reliable way? Or maybe change it from being an array to be a named object? Not sure how this change would affect other functionalities of Phoenix.

LukasHirt:
> I could not find a way to get the file permissions from inside an app, can that be achieved somehow? I am interest in knowing if the current user has read-only permissions on the open diagram.

You could do a call via our sdk - `this.$client.files.fileInfo(filePath, ['{http://owncloud.org/ns}permissions'])`

> The only place I found situable to put the draw.io configuration settings was the "applications" array in config.json...

This will be possible after merging #4380. Applications array is only for displaying the app in the app switcher. The config wouldn't be respected when opening the app. Otherwise, there's also settings service which enables you to create a settings bundle for extension https://owncloud.github.io/extensions/settings/bundles For this basic config it might be too much effort though.

apps/draw-io/src/DrawIoEditor.vue
```diff
@@ -76,20 +89,77 @@ export default {
...
+      // Change the working file after the import so the original file is not overwritten
+      this.filePath += `_${this.getTimestamp()}.drawio`
+      this.showMessage({
+        title: this.$gettext('Diagram imported'),
+        desc: getDescription(),
+        autoClose: {
+          enabled: true
+        }
+      })
+      fetch(url, { headers })
```
LukasHirt, line 113: Wouldn't this methods work here? https://github.com/owncloud/owncloud-sdk/blob/c4b96e0b50bf989e792899913fe3a70fd1698b43/src/fileManagement.js#L76

VirtualEvan: I tried with `getFileContent`, but it looks like we cannot specify the responsetype. Drawio is able to handle the vsdx file as base64, so we need the file itself insted of its content. Also tried to re-create the file from its binary content, but I could not find a way to do so, the resultant file was always different.

VirtualEvan:
Sorry for the delay, I've been quite busy and wanted to test as much as possible before comming back to you

> You could do a call via our sdk - this.$client.files.fileInfo(filePath, ['{http://owncloud.org/ns}permissions'])

I am trying to show diagrams in chromeless mode when they are read-only.  
The SDK allowed me to know if the file is read-only, but as it is asychronous I had to delay the render until the response is ready.

> This will be possible after merging #4380. Applications array is only for displaying the app in the app switcher. The config shouldn't be used for settings outside of that...

By this I understand that the `applications` array should not be used to store configuration specific to applications, but the `external_apps` can be used for this, is it right? I added it under `external_apps` and was able to get it directly from the store.

LukasHirt:
> I am trying to show diagrams in chromeless mode when they are read-only.
The SDK allowed me to know if the file is read-only, but as it is asychronous I had to delay the render until the response is ready.

Is the delay okay or does it cause some issues? AFAICT the delay shouldn't take too long so from users POV it shouldn't be a big problem.

> By this I understand that the applications array should not be used to store configuration specific to applications, but the external_apps can be used for this, is it right? I added it under external_apps and was able to get it directly from the store.

Correct. Nice!

LukasHirt:
Can you try to rebase the PR? It might solve the failing tests.

VirtualEvan:
Rebasing fixed the issue with the tests, thanks!

> Is the delay okay or does it cause some issues? AFAICT the delay shouldn't take too long so from users POV it shouldn't be a big problem.
It is not a problem in my local instance, but I was not sure how long could it take for other people with different/more complex setups. If you think it is OK then is perfect! :)

LukasHirt (approved):
Nice 🚀

LukasHirt:
@VirtualEvan Let me pls know if there's something left you would like to commit here or if I can hit the merge button :wink:

VirtualEvan:
No, I think that's all. Please merge :)

Do you want me to add some documentation somewhere mentioning the options that can be configured for the app?

```json
"external_apps": [
  {
    "id": "draw-io",
    "path": "this_should_exist_and_cannot_be_empty",
    "config": {
      "url": "https://drawio.web.cern.ch",
      "autosave": true,
      "theme": "atlas"
    }
  }
]
```

By the way, I noticed that for external apps, `path` is mandatory.  
If it does not exist or is empty, then everything fails. Just putting a dummy value makes it work.  
Not sure if it's a bug or if I missed something.

LukasHirt:
> Do you want me to add some documentation somewhere mentioning the options that can be configured for the app?

That would be awesome to have it documented. Maybe you could put it in the sample config.json files?

> By the way, I noticed that for external apps, path is mandatory.

Yep, both `id` and `path` are mandatory. Without the path, the frontend doesn't know where to locate the app. I know that it's now creating confusion with the `apps` array where drawio is loaded. If you check the console, there should be actually a 404 error. That could be fixed by specifying the correct path and removing drawio from `apps` array to prevent having the app loaded twice.

VirtualEvan:
OK, I think I got it.  
Added default configuration to the `oc10` and `ocis` sample config.json files

Let me know if there is something else I can add that could be helpful

LukasHirt:
Thank you! I think that this is now in a perfect shape for merging! I'll hit the button when CI is green :wink:

PR #4341, opened 2020-11-18
Fix docker image build step

LukasHirt:


LukasHirt:
@micbar Is this enough to make the docker image build step run again?

micbar:
under which condition should it run?

LukasHirt:
> under which condition should it run?

After it gets merged into master.

micbar:
@LukasHirt then ok.

micbar (approved):

PR #4373, opened 2020-11-23
Theme redirect page

LukasHirt:
We've adjusted the theme on OIDC redirect and access denied pages to use correct logo and background.
We've also added those two values into the theming capabilities.

host docker internal_8300_index html (5)
host docker internal_8300_index html (4)

## Open tasks:
- [x] Release and update ODS
- [x] Changelog item
- [x] Add screenshot to PR desc


LukasHirt:
Something to think about - when missing config page is triggered we do not have the current theme available. Need to think about a solution for that page.

kulmann (approved):

PR #4380, opened 2020-11-24
Load extension config

LukasHirt:
Extensions can now define own config object in the `external_apps` in config.json


kulmann (approved):
🚀

PR #4403, opened 2020-12-01
Bootstrap Vue

LukasHirt:
Previously the Vue class wouldn't be assigned to the window object. To do so, we have to move Vue import into own file and assign the class there.


LukasHirt:
TBH I am not sure if we need changelog for this. Even though it is kinda a bugfix, I don't think it's necessary in this case. What do you think @kulmann ?

kulmann (approved):
Fine without changelog IMO

PR #4410, opened 2020-12-02
Display collaborator info in labels

LukasHirt:
## Description
- Use `oc-tag` component to display share info
- Display public and private links
- Refactor public links sidebar components

Requires release and update of ODS with new icons and fixed `router-link` variation of `oc-tag` https://github.com/owncloud/owncloud-design-system/pull/975.

## Screenshots
image

image

## Test env
- Drone

### DIFF
--- a/changelog/unreleased/share-info-labels
+++ b/changelog/unreleased/share-info-labels
@@ -1,0 +1,5 @@
+Change: Use labels to display share info
+
+We've change the way of displaying share information for public links and people. Every information is now displayed in own label.
+
+https://github.com/owncloud/phoenix/pull/4410
--- a/changelog/unreleased/visible-links
+++ b/changelog/unreleased/visible-links
@@ -1,0 +1,5 @@
+Enhancement: Display full public and private links
+
+We've added under names of public and private links a full link so that user can copy it without using the copy to clipboard button.
+
+https://github.com/owncloud/phoenix/pull/4410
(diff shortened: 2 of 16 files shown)


LukasHirt:
Commit history is rather chaotic for this one but still keeping it for now in case it would be at least a bit helpful for the review. After the PR is approved I'll squash it.

changelog/unreleased/visible-links
```diff
@@ -0,0 +1,5 @@
+Enhancement: Display full public and private links
+
+We've added under names of public and private links a full link so that user can copy it without using the copy to clipboard button.
```
kulmann, line 3:
```suggestion
Below the names of public and private links we've added the respective full URL so that users can copy it without the copy to clipboard button.
```

changelog/unreleased/share-info-labels
```diff
@@ -0,0 +1,5 @@
+Change: Use labels to display share info
+
+We've change the way of displaying share information for public links and people. Every information is now displayed in own label.
```
kulmann, line 3:
```suggestion
We've changed the way of displaying share information for public links and people. Every information is now displayed in its own label.
```

LukasHirt:
With updated ODS node version `11` was not compatible. As we do not have version `12` in owncloud-ci, I switched here to the one from webhippie /cc @micbar

kulmann (approved):
❤️

PR #4422, opened 2020-12-04
check content length before trying to parse the content

C0rby:
Every time phoenix calls the notifications endpoint it tries to do a json parse of the response content.
But there is no response content and therefore the parsing fails and log a message to the console.
This pr tries to prevent that so that the console doesn't get spammed with unnecessary logs.

- [ ] Add changelog

### DIFF
--- a/src/store/app.js
+++ b/src/store/app.js
@@ -34,6 +34,9 @@ const actions = {
         action: 'api/v1/notifications'
       })
       .then(response => {
+        if (response.headers.get('Content-Length') === '0') {
+          return
+        }
         response.json().then(json => {
           if (response.ok) {
             context.commit('UPDATE_NOTIFICATIONS', json.ocs.data)


src/store/app.js
```diff
@@ -34,6 +34,9 @@ const actions = {
         action: 'api/v1/notifications'
       })
       .then(response => {
+        if (response.headers.get('Content-Length') === '0') {
```
kulmann, line 37:
```suggestion
        if (parseInt(response.headers.get('Content-Length')) === 0) {
```

C0rby: Why is that better?
The result would be the same but cost more because of the parsing.

kulmann: Yeah sorry... always a string anyway 🤦

kulmann (approved):
Yes, good thing. Failed notification requests are not critical.

I would be fine to merge this without a changelog entry.

PR #4428, opened 2020-12-07
emit files in dev mode

fschade:
right now no changes in main app get written to disk. This is a problem if you start to develop ocis with custom asset path (PHOENIX_ASSET_PATH). This PR fixes this

### DIFF
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -12,6 +12,7 @@ module.exports = merge(common, {
     symlinks: false
   },
   devServer: {
+    writeToDisk: true,
     contentBase: path.resolve(__dirname),
     historyApiFallback: true,
     publicPath: '/',


LukasHirt:
Could we make this optional? Sometimes we're developing against oC10 backend where this is not needed. I'd still love to see the `watch-all-ocis` command to spawn a dev server on ocis-phoenix port so we can develop without that service running. That's something for a separate PR though 🙂

LukasHirt:
Rebasing PR should btw help with the node error. We are using now different nodejs images with version 12.

fschade:
let's discuss the watch-all-ocis idea today. I think i have time to do it today

fschade:
I close this in fav to the experimental rollup switch

PR #4430, opened 2020-12-07
Copy nav item state

fschade:
fix: https://github.com/owncloud/ocis/issues/1031


kulmann (changes requested):
Cool! 🚀 

Please remove the `writeToDisk` config option. Should not be in scope of this PR. And I don't want to have it without setting up a ramdisk first.

Also, please add a changelog. The fix is worth being mentioned. ;-)

kulmann (approved):
Awesome! ❤️  You need to sign the CLA, then we can merge this PR.

PR #4434, opened 2020-12-08
add the option to retroactively decline an accepted share

C0rby:
Declined shares could be accepted retroactively but accepted shares could not be declined.
Implements https://github.com/owncloud/ocis/issues/985


kulmann (approved):
Cool 👍

PR #4435, opened 2020-12-08
[tests-only] Fixup test suites in ci

phil-davis:
1) sort the UI test suites in `.drone.star` so that it is easier to notice when something is missing or wrong.
2) fixup duplications of `webUICreateFilesFolders` so that it actually runs `webUIDeleteFilesFolders` suite
3) fixup mentions of `webUITrashbin` (which is not a suite name any more) so that it actually runs `webUITrashbinDelete` (which was missing from `.drone.star`)
4) there are some UI test suites that currently have all scenarios skipped because the features are not yet implemented. Add those into the pipelines so that when some scenarios do get enabled, they will automagically get run. Otherwise people will forget to add these into `.drone.star`
5) skip a failing trashbin-delete scenario for now. See issue #4437


phil-davis:
https://drone.owncloud.com/owncloud/phoenix/12090/42/20
`webUITrashbinDelete` 6 scenarios (6 passed) - good

individual-it (approved):

PR #4444, opened 2020-12-09
fix ocis-1043

fschade:
partially fixes: https://github.com/owncloud/ocis/issues/1043
fixes: https://github.com/owncloud/ocis/issues/884

### DIFF
--- a/src/Phoenix.vue
+++ b/src/Phoenix.vue
@@ -2,7 +2,17 @@
   <div class="uk-height-1-1">
     <skip-to target="main">Skip to main</skip-to>
     <div id="Phoenix" class="uk-height-1-1">
-      <template v-if="!showHeader">
+      <div
+        v-if="user.isAuthenticated && !user.userReady"
+        class="loading-overlay"
+        :style="{
+          backgroundImage: 'url(' + configuration.theme.loginPage.backgroundImg + ')'
+        }"
+        uk-height-viewport
+      >
+        <oc-spinner size="xxxlarge" :aria-label="$gettext('Loading')" class="uk-position-center" />
+      </div>
+      <template v-else-if="!showHeader">
         <router-view name="fullscreen" />
       </template>
       <div v-else key="core-content" class="uk-height-1-1 uk-flex uk-flex-row uk-flex-row">
@@ -124,7 +134,6 @@ export default {
 
       return list
     },
-
     showHeader() {
       return this.$route.meta.hideHeadbar !== true
     },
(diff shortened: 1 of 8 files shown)


LukasHirt (approved):
I guess cloning the whole `items` instead of one by one wouldn't make a real difference here with performance, right? Since I guess there won't ever be hundreds of nav items :shrug:

fschade:
> I guess cloning the whole `items` instead of one by one wouldn't make a real difference here with performance, right? Since I guess there won't ever be hundreds of nav items 🤷

yes its more costy to do it one by one (but not much it's a low level api). The problem if we do

```
const o = [1, 2, 3]
const n = [...o]
n[0] = 99
```

the result would be [99,2,3]

but if we do 
```
const o = [{foo: 1}]
const n = [...o]
n[0].foo = 99
```

we still mutate the original object in o

objects in arrays get refferenced and not deep copied.

src/Phoenix.vue
```diff
@@ -2,7 +2,17 @@
...
-      <template v-if="!showHeader">
+      <div
+        v-if="user.isAuthenticated && !user.userReady"
+        class="loading-overlay"
+        :style="{
+          backgroundImage: 'url(' + configuration.theme.loginPage.backgroundImg + ')'
+        }"
+        uk-height-viewport
+      >
+        <oc-spinner size="xxxlarge" :aria-label="$gettext('Loading')" class="uk-position-center" />
```
kulmann, line 13: From my personal feeling the spinner is too big. I would choose `xlarge` size and border sizes in the css class `10px`.

src/store/user.js
```diff
@@ -142,7 +147,7 @@ const actions = {
     }
     const token = vueAuthInstance.getToken()
     if (token) {
-      init(this._vm.$client, token)
+      await init(this._vm.$client, token)
```
kulmann, line 150: The `await` here should not be needed, as it's the last statement within the function.

src/store/navigation.js
```diff
@@ -32,7 +32,9 @@ const mutations = {
   ADD_NAV_ITEM(state, { extension, navItem }) {
     const dynamicNavItems = { ...state.dynamicNavItems }
     dynamicNavItems[extension] = dynamicNavItems[extension] || []
-    const index = dynamicNavItems[extension].findIndex(item => item.name === navItem.name)
+    const index = dynamicNavItems[extension].findIndex(
+      item => item.route.path === navItem.route.path
```
kulmann, line 36: Thanks for spotting this ❤️

kulmann (changes requested):

kulmann (approved):
👍

PR #4446, opened 2020-12-10
[tests-only] Bump OCIS commit id for tests

phil-davis:
The updated commit-id has nothing interesting/new, but the CI against core uses `daily-master-qa` and I want to make sure that the tests pass with that. So this is an easy way to get a full drone CI result.


phil-davis:
This passed when the repo was called `owncloud/phoenix`, so that is good to know.

Closing - this is not actually important right now and is easily done after the `owncloud/web` CI is all set up again and passing.

PR #4484, opened 2020-12-13
[tests-only] add legacy dist target

micbar:
# Description

To keep the web app compatible with our default build and signing process
https://gitea.owncloud.services/devops/releasescripts
BTW: @jnweiger you need to update that and recreate your build container. ownCloud web needs node 12

- We need to add a `dist` target on the default makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,21 @@ core: dist/core/core.bundle.js
 
 .PHONY: clean-core
 clean-core:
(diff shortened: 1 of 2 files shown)


LukasHirt (approved):
:rocket:

LukasHirt:
Tests should start passing again after #4485 gets merged.

PR #4485, opened 2020-12-14
[tests-only] Change more phoenix to web now that owncloud/ocis has web extension

phil-davis:
## Description
PR https://github.com/owncloud/ocis/pull/1041 switched `owncloud/ocis` to having a `web` extension, rather than `ocis-phoenix`

Adjust CI here so that it understands the changes in `owncloud/ocis`

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -330,7 +330,7 @@ config = {
 	'defaults': {
 		'acceptance': {
 			'ocisBranch': 'master',
-			'ocisCommit': '76fefa325594c3d58560ca06e0e5311298104832',
+			'ocisCommit': 'ee7b3213b39ec8e76b4a176ed85cbf2588009554',
 		}
 	},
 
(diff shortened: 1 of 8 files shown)


.drone.star
```diff
@@ -1430,9 +1430,9 @@ def buildOcisWeb():
 		'pull': 'always',
 		'commands': [
 			'cd $GOPATH/src/github.com/owncloud/ocis',
-			'cd ocis-phoenix',
+			'cd web',
 			'make build',
-			'cp bin/ocis-phoenix /var/www/owncloud'
+			'cp bin/web /var/www/owncloud/ocis-web'
```
phil-davis, line 1435: Actually I  wonder why this step exists at all - from `owncloud/web` we test with a version of `web` that is built from the PR branch in `owncloud/web`. We should no need the version that is currently in `owncloud/ocis` `web` folder.

But I can try removing this step completely in a followup PR. Because this looks like it was not needed even when it was called `ocis-phoenix`

phil-davis: Note: there is already the "real" `/var/www/owncloud/web` that has been built from the `owncloud/web` branch, which is what we use for testing. So this copy has to go somewhere different, otherwise it gives an error message if we try to copy it on top of the existing `/var/www/owncloud/web`

.drone.star
```diff
@@ -1451,13 +1451,13 @@ def ocisWebService():
...
-			'PHOENIX_ASSET_PATH': '/var/www/owncloud/web/dist',
-			'PHOENIX_OIDC_CLIENT_ID': 'phoenix'
+			'WEB_UI_CONFIG': '/srv/config/drone/config.json',
+			'WEB_ASSET_PATH': '/var/www/owncloud/web/dist',
+			'WEB_OIDC_CLIENT_ID': 'phoenix'
 		},
 		'commands': [
 			'cd /var/www/owncloud',
-			'./ocis-phoenix --log-level debug server',
+			'./ocis-web --log-level debug server',
```
phil-davis, line 1460: Hmmm - I  think this should be running `./web`. `./ocis-web` will be the version of `web` that has been "built-in" to `owncloud/ocis`

phil-davis: This has been changed in a recent commit.

LukasHirt (approved):
:rocket:

phil-davis:
I broke it in the last 3 commits - I need to either sort out why, or remove those commits for now.

phil-davis:
> I need to either sort out why, or remove those commits for now.

I removed those commits and squashed a bit. This should pass now.

In a separate PR I will try to sort out what the acceptance tests against an oC10 server are really testing. I think that is an issue that has been there since before the phoenix->web change.

PR #4491, opened 2020-12-14
typo in oc10-app.md

jnweiger:
The package is called web-app.tar.gz, and github project and top folder are just web. Is this correct?


LukasHirt:
In #4492 I completely dropped the section about tarball in favour of marketplace download. Let me pls know if you think it still makes sense to keep this in.

PR #4492, opened 2020-12-14
Improve app docs

LukasHirt:
Show example oauth2 entry, mention the difference between `apps` and `apps-custom`, add examples of app switcher config and point to the marketplace in the deployment section.

/cc @jnweiger @davitol


docs/deployments/oc10-app.md
```diff
@@ -9,40 +9,36 @@ geekdocFilePath: oc10-app.md
 
 {{< toc >}}
 
-One of the build artifacts of the ownCloud Web is a ownCloud 10 app bundle.
+The ownCloud Web is being deployed as an app to [ownCloud marketplace](https://market.owncloud.com) to enable easy early integration into existing ownCloud 10 instances.
+After completing this setup, ownCloud Web will be available on `https://<your-owncloud-server>/apps/web`.
+
+{{< hint info >}}
+Depending on your setup, it might be required to use `apps-custom` instead of `apps` in the URL. Keep this in mind when setting all mandatory URL addresses.
```
phil-davis, line 16: In new oC10 installs, in the default `config.php`, `apps-paths` has `apps-external` which is where "extra" apps are installed. Will installations like this have to put `apps-external` in the URL?

If the URL is based on the actual folder name which the app is installed in, then should we use `apps-external` here?

Or is there some way that we can make the URL always be `/apps/web` ?

LukasHirt: I don't have much knowledge about how and where marketplace apps are installed. But I guess it will always land in the external folder - always have to be `external/web`?

We saw `custom` folder in the setup of @micbar Maybe that's some custom change though? I guess we can adjust this to mention `apps-external` and warn that it could have even different name?

micbar: yes. use `apps-external`

LukasHirt: Done :wink:

docs/deployments/oc10-app.md
```diff
@@ -9,42 +9,38 @@ geekdocFilePath: oc10-app.md
 
 {{< toc >}}
 
-One of the build artifacts of the ownCloud Web is a ownCloud 10 app bundle.
+The ownCloud Web is being deployed as an app to [ownCloud marketplace](https://market.owncloud.com) to enable easy early integration into existing ownCloud 10 instances.
+After completing this setup, ownCloud Web will be available on `https://<your-owncloud-server>/apps-external/web`.
+
+{{< hint info >}}
+Depending on your setup, the name of `apps-external` folder can vary. It is important to use the corrent name in all of mentioned URLs.
```
phil-davis, line 16:
```suggestion
Depending on your setup, the name of `apps-external` folder can vary. It is important to use the correct name in all of the mentioned URLs.
```

And it is important to use the correct spelling of "corrent" ;)

phil-davis (approved):

PR #4493, opened 2020-12-14
Add media viewer and draw-io to default apps

LukasHirt:
When release tarballs are created, we are copying the config.json.dist into them as a default config. This file had so far only "files" app enabled. This adds also "media viewer" and "draw-io" into apps enabled by default.

### DIFF
--- a/changelog/unreleased/default-apps
+++ b/changelog/unreleased/default-apps
@@ -1,0 +1,6 @@
+Enhancement: Extend default apps
+
+When release tarballs are created, we are copying the config.json.dist into them as a default config.
+In that file were so far only "files" app enabled. This adds also "media viewer" and "draw-io" into apps enabled by default.
+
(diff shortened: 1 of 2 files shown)


phil-davis:
Failed 1 scenario: https://drone.owncloud.com/owncloud/web/12265/6/20
`1) Scenario: Delete all files at once (attempt 2) # tests/acceptance/features/webUIDeleteFilesFolders/deleteFilesFolders.feature:66`

IMO that test might be subject to timing errors sometimes???

Restarted drone.

LukasHirt:
> IMO that test might be subject to timing errors sometimes???

Delete ops are limited to 4 requests at the time via [p-queue](https://www.npmjs.com/package/p-queue) so I guess that's possible that if the CI performance would slow down a bit that it could not delete all files in time.

phil-davis (approved):

PR #4498, opened 2020-12-14
[DO NOT MERGE] Fix vue router paths

fschade:
fixes https://github.com/owncloud/ocis/issues/988

### DIFF
--- a/apps/files/src/default.js
+++ b/apps/files/src/default.js
@@ -114,14 +114,14 @@ const routes = [
     }
   },
   {
-    path: '/list/:item?',
+    path: '/list/:item(.*)',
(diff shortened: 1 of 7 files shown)


kulmann (changes requested):
Navigating into random folders, I see two issues:
1: `[vue-router] missing param for named route "files-list": Expected "item" to be defined`
2: empty folders list their own folder name as subfolder, although it doesn't exist (create a folder `test`, navigate into it. You will see a folder `test` in the list although you just created the parent and it is empty).

butonic:
outdated. will be fixed by moving to the new vue router.

PR #4500, opened 2020-12-15
Fix opening editors in new tab

LukasHirt:
When opening the editors view in a new tab, we were using the origin of location. This would break in case we have Web deployed to a different path than root e.g. `http://owncloud/apps/web`.

### DIFF
--- a/apps/files/src/mixins/fileActions.js
+++ b/apps/files/src/mixins/fileActions.js
@@ -84,9 +84,8 @@ export default {
           name: editor.routeName,
           params: { filePath: filePath }
         }).href
-        const url = window.location.origin + '/' + path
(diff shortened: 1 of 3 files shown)


fschade (approved):
💯

phil-davis:
https://drone.owncloud.com/owncloud/web/12278/52/11
@LukasHirt various test scenarios that open stuff in media viewer are failing.
Without having tried it, my first guess is that the files get opened in a new tab, and so the test code does not see them.

LukasHirt:
@phil-davis This was actually because of an incorrect route name. We dropped the automatic prefixing with app id in the file actions because some route names are not doing that but I forgot to fix the route in media viewer. Already pushed fixed.

kulmann (approved):

Issue #4426, opened 2020-12-05
Phoenix UI tests for public links


phil-davis:
PR #4425 fixed some issues related to public links. But it required no changes to tests.

Investigate what UI tests currently exist for public links.

- enable any that do exist
- add a "reasonable" set of tests for public links

phil-davis:
PR #4504 fixed a permissions field for public links that was accidentally removed. Somehow no tests failed when the field was accidentally removed.

Look into why.

PR #4504, opened 2020-12-15
Bring back the permissions field in public links editing

kulmann:
## Description
We accidentally removed the `permissions` field in the public links editing form. This PR brings it back.

## How Has This Been Tested?

- drone

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


refs (approved):

kulmann:
@phil-davis @individual-it might be worth to look into this. Since we have a lot of tests regarding different share permissions, I don't get why this bug was not blowing up in CI. 😅

kulmann:
@LukasHirt fyi I placed the changelog item directly in the respective versioned folder so that we can tag a beta4 after this PR is merged, without needing another release preparation PR. Needs fixing in ocis as soon as possible IMO.

phil-davis:
> @phil-davis @individual-it might be worth to look into this. Since we have a lot of tests regarding different share permissions, I don't get why this bug was not blowing up in CI.

Agree - see issue #4426 - my guess is that there is actually not much test coverage of the public links UI. That issue needs looking at.

micbar (approved):

phil-davis:
Some public link test scenarios failed: https://drone.owncloud.com/owncloud/web/12298/39/20

There are test scenarios that try to delete the password on an existing public link share. They expect to get a message "Passwords are enforced for link shares" But the message is not found.

Maybe the DOM elements have changed now that the `permissions` filed is back, and the tests need to be adjusted to find the expected error text. Or maybe there is a real problem. Needs investigating.

kulmann:
The permissions were not correctly loaded upon creation of the vue component. Fixed now. Good that the test suite caught that 😎

butonic (approved):

PR #4470, opened 2020-12-11
Random typos

phil-davis:
This is on top of #4443 - actually I just wanted to edit something to see if CI is OK. My IDE found these for me.

PR #4508, opened 2020-12-16
[tests-only] Fix random typos

phil-davis:
These are in tests and comments only, so no changelog needed.
I found them a few days ago in #4470 while I just wanted to run a test of CI. May as well commit them.


kulmann (approved):
Thanks for taking care of those! 👍

PR #4538, opened 2020-12-20
[Tests Only] Add acceptance tests to restore a trashbin file whose parent folder is renamed or is not restored

jasson99:
## Description
This PR adds acceptance tests for:
- restoring a trashbin file whose parent folder is renamed
- restoring a trashbin file whose parent folder is not restored

Also, the method to assert the error/success message is changed in this or. Previously, when the expected error/success message and the actual displayed error/success message in the webUI were different, the error thrown was not clear about what was actually found. Can be seen in this: https://drone.owncloud.com/owncloud/web/12372/44/20 . With the recent change, even when the expected and actual text do not match, we can at least find out what was been expected and what was actually found, as in this: https://drone.owncloud.com/owncloud/web/12407/44/20

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/1057

## How Has This Been Tested?
- CI

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/stepDefinitions/generalContext.js
+++ b/tests/acceptance/stepDefinitions/generalContext.js
@@ -95,6 +95,22 @@ Then('the success/error message with header {string} should be displayed on the 
     .text.to.equal(message)
 })
 
+Then(
+  'the success/error message with header {string} and subheader {string} should be displayed on the webUI',
+  async function(headerMessage, subheaderMessage) {
+    await client.page
+      .webPage()
+      .waitForElementVisible('@message')
+      .expect.element('@message')
+      .text.to.equal(headerMessage)
+    return client.page
+      .webPage()
+      .waitForElementVisible('@messageSubHeader')
+      .expect.element('@messageSubHeader')
+      .text.to.equal(subheaderMessage)
+  }
+)
+
 Then('the following success/error message should be displayed on the webUI', function(message) {
   return client.page
     .webPage()
--- a/tests/acceptance/stepDefinitions/generalContext.js
+++ b/tests/acceptance/stepDefinitions/generalContext.js
@@ -3,7 +3,6 @@ const { After, Before, Given, Then, When } = require('cucumber')
 const webdavHelper = require('../helpers/webdavHelper')
 const httpHelper = require('../helpers/httpHelper')
 const backendHelper = require('../helpers/backendHelper')
-const assert = require('assert')
 const fs = require('fs')
 const occHelper = require('../helpers/occHelper')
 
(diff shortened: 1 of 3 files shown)


Talank (commented):
LGTM

tests/acceptance/stepDefinitions/generalContext.js
```diff
@@ -95,12 +96,11 @@ Then('the success/error message with header {string} should be displayed on the
...
-  return client.page
-    .webPage()
-    .waitForElementVisible('@messages')
-    .expect.element('@messages')
-    .text.to.equal(message)
+Then('the following success/error message should be displayed on the webUI', async function(
+  message
+) {
+  const displayedMessage = await client.page.webPage().getDisplayedMessage()
+  assert.strictEqual(displayedMessage, message)
```
HariBhandari07, line 103: You can also add user friendly `expected message` for `strictEqual`

dpakach:
@jasson99, I get both the expected and actual values from the error message in the current implementation, 
```
   Then the following error message should be displayed on the webUI
      """
      File upload failed…
      tus: failed to upload chunk at offset
      """
√ Element <//*[contains(@class, "uk-notification-message")]> was visible after 22 milliseconds.
Expected element <//*[contains(@class, "uk-notification-message")]> text to equal: "File upload failed…
tus: failed to upload chunk at offset" - expected "equal 'File upload failed…
tus: failed to upload chunk at offset'" but got: "File upload failed…
tus: failed to upload chunk at offset 3063808, caused by [object ProgressEvent], originated from request (response code: 0, response text: )" (10007ms)
undefined    ✖ failed
      Error while running "element" command: Expected element <//*[contains(@class, "uk-notification-message")]> text to equal: "File upload failed…
      tus: failed to upload chunk at offset" - expected "equal 'File upload failed…
      tus: failed to upload chunk at offset'" but got: "File upload failed…
      tus: failed to upload chunk at offset 3063808, caused by [object ProgressEvent], originated from request (response code: 0, response text: )" (10007ms)
          at World.<anonymous> (/home/dipakacharya/www/phoenix/tests/acceptance/stepDefinitions/generalContext.js:117:13)
    And file "0" should not be listed on the webUI

```

The reason you are not getting proper message in your case is because the element itself was not there so it is giving `stale element reference`
Maybe there is a intermittent problem with this scenario itself.

jasson99:
Yes, for example in the case of the failed message when trashbin restoration is tried, the message stays there only for a certain period of time, for example 2-4 seconds. And then the message disappears. And then the error message is not accurate. Therefore I tried to make the error message clear whenever the expected message is not equal to the displayed one, and then the message disappears if a few seconds.

individual-it (approved):

PR #4539, opened 2020-12-21
Remove markdown editor

LukasHirt:
Markdown editor is living in its [own repository](https://github.com/owncloud/ocis-markdown-editor). Since we are not even delivering it as an extension enabled by default - there's no reason to keep this in this repo anymore because of the license.

PR #4563, opened 2020-12-28
Start markdown editor service in CI

LukasHirt:
In order to remove markdown editor from this repo (#4539), we need to start a service with this extension to enable us running file creation tests.


LukasHirt:
New decision is to keep the editor in the Web repo

PR #4567, opened 2020-12-29
Create files journey

micbar:
## Related Issue

- Fixes <issue_link>

## Motivation and Context
- Add a basic user journey

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...


package.json
```diff
@@ -23,8 +23,8 @@
...
-    "acceptance-tests-ocis": "NODE_TLS_REJECT_UNAUTHORIZED=0 RUN_ON_OCIS=true cucumber-js --require-module @babel/register --require-module @babel/polyfill --require tests/acceptance/setup.js --require tests/acceptance/stepDefinitions --format node_modules/cucumber-pretty -t \"${TEST_TAGS:-not @skip and not @skipOnOCIS and not @notToImplementOnOCIS}\"",
-    "acceptance-tests-with-ldap": "RUN_ON_OCIS=true RUN_WITH_LDAP=true cucumber-js --require-module @babel/register --require-module @babel/polyfill --require tests/acceptance/setup.js --require tests/acceptance/stepDefinitions --format node_modules/cucumber-pretty -t \"${TEST_TAGS:-not @skip and not @skipOnOCIS and not @notToImplementOnOCIS}\"",
+    "acceptance-tests-ocis": "WEB_UI_CONFIG='./tests/acceptance/ocis-mac-config.json' SERVER_HOST='https://host.docker.internal:9200' BACKEND_HOST='https://host.docker.internal:9200' NODE_TLS_REJECT_UNAUTHORIZED=0 RUN_ON_OCIS=true cucumber-js --require-module @babel/register --require-module @babel/polyfill --require tests/acceptance/setup.js --require tests/acceptance/stepDefinitions --format node_modules/cucumber-pretty -t \"${TEST_TAGS:-not @skip and not @skipOnOCIS}\"",
```
phil-davis, line 26: I don't see any `tests/acceptance/ocis-mac-config.json` in this repo, and it is not added in this PR.
I guess this is for running locally on a Mac? 
What is the intention here? Or is this accidentally committed?

package.json
```diff
@@ -34,9 +34,9 @@
     "ocis-kill": "killall ocis",
     "docker-kill": "docker kill web-tests-selenium web-tests-redis",
     "testing-app": "if [ -d tests/testing-app ]; then (cd tests/testing-app && git pull && pwd); else git clone --depth 1 https://github.com/owncloud/testing.git tests/testing-app; fi;",
-    "build-ocis": "if [ -d tests/ocis ]; then (cd tests/ocis && git reset --hard master && git pull); else git clone --depth 1 https://github.com/owncloud/ocis.git tests/ocis; fi; cd tests/ocis/ocis; make clean; make",
+    "build-ocis": "if [ -d tests/ocis ]; then (cd tests/ocis && git reset --hard master && git pull); else git clone --depth 1 https://github.com/owncloud/ocis.git tests/ocis; fi; cd tests/ocis/ocis; make clean; make generate build",
```
phil-davis, line 37: Adding `make generate build` is probably "a good thing". It seems like it is not related to this PR, but might be good to do anyway.

phil-davis (commented):
I commented on `package.json` changes.

All things that are done in this user journey are already in various feature files.

The "user journey" scenario tests a sequence of many things, so it "breaks the rules" of "a scenario should test one thing". But I guess the point of this is to have some overall examples of sequences that a user would typically do. And to have them in one place so that we do not have to look in lots of feature files to find all the pieces.

package.json
micbar: make alone does nothing. Problem was that the package.json items were no working at all.

micbar:
> I commented on `package.json` changes.
> 
> All things that are done in this user journey are already in various feature files.
> 
> The "user journey" scenario tests a sequence of many things, so it "breaks the rules" of "a scenario should test one thing". But I guess the point of this is to have some overall examples of sequences that a user would typically do. And to have them in one place so that we do not have to look in lots of feature files to find all the pieces.

Yes. This is the point.

It is a more general thing. IMO our acceptance tests are behaving a bit like "unit"  tests which test too many variations of the same scenario. IMO acceptance tests should ensure that the software works in general. The variations should be tested on a lower level.

phil-davis:
> IMO acceptance tests should ensure that the software works in general. The variations should be tested on a lower level.

Yes, this is true. Something in between low-level unit tests, and full end-to-end acceptance tests. And in a single-project test suite that is what [should|could] have been done.

Now we are making 2 products, oC10 and OCIS, which need to "work the same" to a large degree. But they are written in completely different languages and architecture. We do need some common set of test cases that demonstrate that the behaviour is consistent for the 2 implementations. So that rules out using phpunit or some GRPC test framework for the "common" tests. We are pretty much forced to have end-to-end tests for this purpose, which could be implemented in any test framework. They happen to be in PHP-Behat just because of history.

But yes, the user journeys are useful to have.

package.json
micbar: it is there. I didnt add it 🤷

phil-davis: I must be going crazy. It is there! I was probably looking in some other `tests/acceptance` directory in some other app.

phil-davis (approved):

PR #4583, opened 2021-01-04
[Tests-Only] Add jest and write example unit test

### LINKED ISSUE #2026: Introduce JS unit test framework for utility functions
Currently we already have a lot of acceptance tests.

These are useful to test user interaction in the UI.

Once we start going into some details like checking with every possible file name, like testing if natural sort works, using acceptance tests through the UI are overkill. They will take a long time to run for not much gain.

So here the idea is to introduce JS unit tests. The JS unit tests should never test any UI components but focus on making sure that the **utility functions** (like in the mixins) are working properly for any set of inputs, including corner cases.

We could look at https://vuejs.org/v2/guide/unit-testing.html

Here's a list of things we could cover with unit tests:
- utility functions for mapping permissions to roles (for https://github.com/owncloud/phoenix/issues/1897)
- utility functions for natural sort (for https://github.com/owncloud/phoenix/pull/1972)

@DeepDiver1975 @individual-it @LukasHirt

LukasHirt:
This PR brings jest tests framework and vue-test-utils to enable us writing unit tests for code and components. As an example, there are already written 3 super basic tests.

- Closes #2026

### Open tasks:
- [x] Run unit tests in drone


LukasHirt:
Strange error in the CI. It shouldn't mount the `UserMenu` component at all so shouldn't complain about imports in there :shrug: Locally passes fine. I'll try to dig deeper.

tests/unit/specs/__snapshots__/TopBar.spec.js.snap
```diff
@@ -0,0 +1,37 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
individual-it, line 1: do we need this file? It seems to be regenerated on every test run

.drone.star
```diff
@@ -534,6 +534,38 @@ def changelog(ctx):
...
 
+def unitTests():
+	return [{
+		'kind': 'pipeline',
+		'type': 'docker',
+		'name': 'unit-tests',
+		'steps': [
+			{
+				'name': 'install-deps',
+				'image': 'owncloudci/nodejs:12',
```
individual-it, line 545: do we care about running the tests on different node version?

individual-it (commented):
I like it very much! Just small comments

Also in the SDK we are moving towards jest https://github.com/owncloud/owncloud-sdk/pull/637

.drone.star
```diff
@@ -534,6 +534,38 @@ def changelog(ctx):
...
+				'pull': 'always',
+				'commands': [
+					'yarn install',
```
individual-it, line 548: can we run the `installNPM()` instead of making an own step here?

individual-it:
when I start with a clean clone of web I get the same error like in drone

individual-it:
`yarn build` before `yarn run test:unit` seems to solve the problem locally. Do we need a build for running the unit tests?

tests/unit/specs/__snapshots__/TopBar.spec.js.snap
LukasHirt: Yes, this file is used for comparison with the generated one. It will fail the test in case the HTML of the component would change. If the change is intentional, tests need to be run with `yarn test:unit -u` and then commit and push the updated file.

.drone.star
LukasHirt: I was thinking about this and TBH I'm not sure. We're limiting node version 10 - 13. This means 4 potential different versions of Node. We do not have an image for all of them though.

.drone.star
LukasHirt: Done

LukasHirt:
> yarn build before yarn run test:unit seems to solve the problem locally. Do we need a build for running the unit tests?

It solved the error in CI as well. Thanks for spotting @individual-it! I thought that the vue-jest loader is enough to use the component in tests. Seems that it's not the case.

I did also one small additional change - the coverage output is now set only to lcov. This is to prevent the output in the console. After the PR will be approved I'll squash commits but for now leaving them for easier review :wink:

individual-it (approved):

PR #4584, opened 2021-01-04
Switch from webpack to rollup

fschade:
## Motivation of change:
before we used webPack to bundle the web repository, over time other ocis addons switched to rollup wich speeds up the whole development proccess and thanks to its tree-shaking makes whole bundle smaller. We also started to restructure the repository to make it easier to understand and get started.

## Changes:
- replace webpack with rollup
- remove unused dependencies
- introduce yarn workspaces (monorepo)
- cleanup scripts in package.json

## Open tasks:
- [x] review makefiles
- [x] adjust docs
- [x] adjust drone
- [x] cleanup rollup config
- [x] resolve conflicts
- [x] request review


kulmann (commented):
Some questions / things I noticed:
- when running `build:w`, a change in one workspace results in recompiling all workspaces. possible to limit this to the changed workspace only?
- for running with oc10 we need to have a dev server - ideally not persisting to disk but to RAM, to save SSD cell max write counts.

Things I noticed about the scripts in package.json:
- afaict scripts with `ldap` (`docker:ldap` and `test:acceptance:ldap`) are not needed. I added a commit that removes those.
- would make sense to keep the script for killing dockerized selenium and redis in docker. I added a commit that adds those.
- it should not be in scope of this repo to take care of running an ocis instance, so IMO it's good that you removed the related scripts.

Other questions:
- Why is everything a devDependency? I can't see real dependencies anywhere. Is it specific to the workspaces feature that all dependencies are devDependencies?

fschade:
> Some questions / things I noticed:
> 
> * when running `build:w`, a change in one workspace results in recompiling all workspaces. possible to limit this to the changed workspace only?
> * for running with oc10 we need to have a dev server - ideally not persisting to disk but to RAM, to save SSD cell max write counts.
> 
> Things I noticed about the scripts in package.json:
> 
> * afaict scripts with `ldap` (`docker:ldap` and `test:acceptance:ldap`) are not needed. I added a commit that removes those.
> * would make sense to keep the script for killing dockerized selenium and redis in docker. I added a commit that adds those.
> * it should not be in scope of this repo to take care of running an ocis instance, so IMO it's good that you removed the related scripts.
> 
> Other questions:
> 
> * Why is everything a devDependency? I can't see real dependencies anywhere. Is it specific to the workspaces feature that all dependencies are devDependencies?

Thanks for the input, true it only should rebuild the required parts.

This is because the vendor is required by all other apps and require js needs to import the updated vendor bundle in every app. I have a look if this can be fixed. 

I will bring back the dev server, thanks for the input :)

everything is a devDependency because we do not have a real runtime dependency like a index.js that is served by node. Instead we bundle everything to dist and in this stage there is no requirement to serve anything from node_modules. 

Npm mentions this here https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

> "dependencies": Packages required by your application in production.
"devDependencies": Packages that are only needed for local development and testing.

individual-it:
`it should not be in scope of this repo to take care of running an ocis instance, so IMO it's good that you removed the related scripts.`
those script have been after the request to make the test runs simpler so that the web developer just can fire up the tests without having to think about how to setup ocis, testing app, core etc. see https://owncloud.github.io/clients/web/testing/#with-ocis-backend 
so if we delete it, we will lose all that

LukasHirt:
Dev server is not working with my oauth2 setup. Trying on master works fine - trying to debug this.

LukasHirt:
> Dev server is not working with my oc10 oauth2 setup. Trying on master works fine - trying to debug this.

Found the issue... There was a newer version of oidc-client lib with which there was always a request to .well-known even though it was an oauth2 setup. For now, I'll pin the version to 1.10.1 where it still works. I'll open a ticket to update it to the latest version and see what changed there. Not scope of this PR.

LukasHirt:
Hmm, thinking now about ports... we can set an arbitrary port for Web when running against oC10 so we could actually always pick port 9100 and that way wouldn't need two scripts for devServer.

packages/web-runtime/src/components/UserMenu.vue
```diff
@@ -88,7 +88,7 @@
 </template>
 
 <script>
-import appVersionJson from '../../build/version.json'
+import { version } from '../../package.json'
```
kulmann, line 91: Nice, but version in that package.json is not up to date. Also, would be quite nice to have only one version defined in the top level package.json (has no version at all at the moment).

packages/web-runtime/src/services/clientRegistration.js
```diff
@@ -1,4 +1,4 @@
-import appVersionJson from '../../build/version.json'
+import { version } from '../../package.json'
```
kulmann, line 1: Same as before. Do we want to import the version from the runtime or top level?

packages/web-runtime/src/store/config.js
```diff
@@ -69,7 +69,7 @@ const actions = {
 
 const mutations = {
   LOAD_CONFIG(state, config) {
-    state.server = config.server.endsWith('/') ? config.server : config.server + '/'
```
kulmann, line 72: Why did you get rid of this? I wanted to enforce here, that we can rely on a) always having a trailing slash or b) never having a trailing slash. I don't care which one, but getting rid of this leaves potential for faulty configuration from the sysadmin.

kulmann (changes requested):
Really, really awesome job, I love it. There are some quirks that need fixing:

For both running on oc10 and ocis backend there is a JS error in the console:
`null:1 Failed to load resource: the server responded with a status of 404 (Not Found)`. Happens on app load and app switch.

When running on ocis, switching to the accounts UI, the list of accounts is empty.

`packages/web-runtime/l10n/Makefile` is broken -> let's move it to a top level l10n folder and make it generic for all apps + runtime from packages.

oc10 marketplace app build is broken (remove signing step from makefile, see for yourself).

LukasHirt:
> When running on ocis, switching to the accounts UI, the list of accounts is empty.

This is not a problem with this PR. We're using `oc-table` and haven't yet adjusted it after the ODS update.

kulmann:
> > When running on ocis, switching to the accounts UI, the list of accounts is empty.
> 
> This is not a problem with this PR. We're using `oc-table` in accounts and haven't yet adjusted it after the ODS update.

Ah yes, right :-)

fschade:
> > > When running on ocis, switching to the accounts UI, the list of accounts is empty.
> > 
> > 
> > This is not a problem with this PR. We're using `oc-table` in accounts and haven't yet adjusted it after the ODS update.
> 
> Ah yes, right :-)

OCIS PR is in progress

LukasHirt:
> For both running on oc10 and ocis backend there is a JS error in the console:
null:1 Failed to load resource: the server responded with a status of 404 (Not Found). Happens on app load and app switch.

Coming from `oc-icon`. For some strange reason it's trying to load the icon from URL even though only name is passed as a prop. Needs to be fixed in ODS most probably since I'm getting this error also on master and the demo instance.

@kulmann AFAICT we wanted to remove the URL part from the icon anyway, didn't we?

kulmann:
Now that the config.server is enforced to be without a trailing slash (which I like) we have more to fix. With my oc10 server URL `http://host.docker.internal/px1` the webdav propfind for my folder listing happens on a URL that is missing one slash: `http://host.docker.internal/px1remote.php/webdav/`. Since a trailing forward slash denotes a directory, I'd like to call veto on myself and vote for having the trailing slash enforced again. That would also fix this issue here. What do you think? @LukasHirt @fschade

LukasHirt:
@kulmann Fine by me :wink:

fschade:
> Now that the config.server is enforced to be without a trailing slash (which I like) we have more to fix. With my oc10 server URL `http://host.docker.internal/px1` the webdav propfind for my folder listing happens on a URL that is missing one slash: `http://host.docker.internal/px1remote.php/webdav/`. Since a trailing forward slash denotes a directory, I'd like to call veto on myself and vote for having the trailing slash enforced again. That would also fix this issue here. What do you think? @LukasHirt @fschade

done

kulmann (approved):
As discussed, this needs fixing for the i18n setup right after merge. Awesome job guys, this is a huge step! 🚀 ❤️ 🥳

PR #4590, opened 2021-01-07
Rename copy confirmation button

LukasHirt:
## Description
We've changed the label of the confirmation button in copy view. Instead of "Copy here", we used "Paste here".

## Motivation and Context
Be closer to desktop explorers


fschade (approved):
👍

kulmann (approved):

PR #4622, opened 2021-01-18
[docs-only] Bump version for oc10 app

kulmann:
## Description
Since we forgot to set the correct version in the oc10 appinfo we need a new release. This time with the correct version. Also updating the release process docs.


kulmann:
Please note that this is only for technical reasons. We don't have anything to release, hence the empty changelog.

micbar (approved):

PR #4669, opened 2021-01-29
[Tests-Only] add webUI tests for moving shared file

swoichha:
## Description
This PR adds webUI test for moving file within a received shared folder to new destination.

## Related Issue
- https://github.com/owncloud/ocis/issues/873

## How Has This Been Tested?
- locally

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...


tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
```diff
@@ -118,3 +118,17 @@ Feature: move files
...
       | lorem.txt   |
       | testapp.zip |
+
+  Scenario: sharee moves a file shared by sharer into another folder
+    Given the setting "shareapi_auto_accept_share" of app "core" has been set to "no"
+    And the administrator has set the default folder for received shares to "Shares"
+    And user "user2" has been created with default attributes
+    And user "user1" has uploaded file with content "test content" to "simple-empty-folder/testFile.txt"
+    And user "user1" has shared folder "/simple-empty-folder" with user "user2"
+    And user "user2" accepts the share "simple-empty-folder" offered by user "user1" using the sharing API
```
jasson99, line 128:
```suggestion
    And user "user2" has accepted the share "simple-empty-folder" offered by user "user1" using the sharing API
```

tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
```diff
@@ -118,3 +118,17 @@ Feature: move files
...
+    And user "user2" has logged in using the webUI
+    And user "user2" has created folder "/Shares/testFolder"
+    And the user has opened folder "Shares"
+    And the user has opened folder "simple-empty-folder"
+    When the user tries to move file "testFile.txt" into new folder "testFolder" using the webUI
```
jasson99, line 133: why tries to here?

jasson99 (dismissed):

tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
swoichha: @jasson99 Because the step ` When the user moves file "tesFile.txt" into folder "testFolder" using the webUI` moves the file using file action menu which will only show current directory (i.e Home/Shares/simple-empty-folder) as destination. But we want to move this file by selecting the checkbox and then clicking on `MOVE` button into new location within the `Shares` folder.

jasson99: But IMO, that tries to in the step looks like the expected then step would be that the move is not successful or something similar. It would be clearer if the `when the user tries to move` could be replaced with `when the user moves` even if having same implementations underneath.

jasson99: may be we can discuss about this?? @dpakach @HariBhandari07 @kiranparajuli589 @individual-it @swoichha @phil-davis @saw-jan

individual-it: I agree with @jasson99 if the difference in the step is **how** it does the operation the sentence should reflect that.
E.g. we could (if we don't have it yet) create a step to move multiple files (by selecting them using the check button) and if that step is used with only one file that would be what @swoichha wants to achieve
```
 When the user moves these files into the folder "testFolder" by selecting them all and using the MOVE button on the webUI 
  | item-names |
  |  testFile.txt  |
```

phil-davis: The existing step could also say better what it does, maybe:
```
When the user uses the file actions menu to move file "something.txt" into folder "testFolder" using the webUI
```
or
```
When the user moves file "something.txt" into folder "testFolder" using the file actions menu on the webUI
```
And then other way can be:
```
When the user uses the move button to move file "something.txt" into folder "testFolder" using the webUI
```
or
```
When the user moves file "something.txt" into folder "testFolder" using the MOVE button on the webUI
```

And steps can be made for selecting and moving multiple files/folders...

swoichha: After having some chat with @phil-davis @individual-it we have decide to use batch move for this test scenario and for future we have decide to make "tries to" in Gherkin more specific using some words that explain enough of what happens in the test. So, for that purpose I have opened an issue here https://github.com/owncloud/web/issues/4680

jasson99 (commented):
LGTM. Please restart CI

individual-it:
CI restarted

individual-it (approved):

PR #4682, opened 2021-02-02
[Tests-Only] add webUI tests for user session of a deleted user

swoichha:
## Description
This PR adds webUI test for user session of a deleted user should not be valid for newly created user of same name.

## Related Issue
- https://github.com/owncloud/ocis/issues/904

## How Has This Been Tested?
- CI

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/features/webUILogin/login.feature
+++ b/tests/acceptance/features/webUILogin/login.feature
@@ -70,3 +70,16 @@ Feature: login users
     Then the user should be redirected to the login error page
     When the user exits the login error page
     Then the user should be redirected to the login page
+
+  Scenario: the user session of a deleted user should not be valid for newly created user of same name
+    Given these users have been created with default attributes:
+      | username |
+      | user1    |
+    And user "user1" has logged in using the webUI
+    And the user has browsed to the files page
+    And user "user1" has been deleted
+    And the user has reloaded the current page of the webUI
+    And the user has been redirected to the login error page
+    And user "user1" has been created with default attributes
+    When the user exits the login error page
+    Then the user should be redirected to the login page
--- a/tests/acceptance/features/webUILogin/login.feature
+++ b/tests/acceptance/features/webUILogin/login.feature
@@ -76,7 +76,6 @@ Feature: login users
       | username |
       | user1    |
     And user "user1" has logged in using the webUI
-    And the user has browsed to the files page
     And user "user1" has been deleted
     And user "user1" has been created with default attributes
     When the user reloads the current page of the webUI
--- a/tests/acceptance/features/webUILogin/login.feature
+++ b/tests/acceptance/features/webUILogin/login.feature
@@ -78,8 +78,8 @@ Feature: login users
     And user "user1" has logged in using the webUI
     And the user has browsed to the files page
     And user "user1" has been deleted
-    And the user has reloaded the current page of the webUI
-    And the user has been redirected to the login error page
     And user "user1" has been created with default attributes
+    When the user reloads the current page of the webUI
+    Then the user should be redirected to the login error page
     When the user exits the login error page
     Then the user should be redirected to the login page
--- a/tests/acceptance/stepDefinitions/loginContext.js
+++ b/tests/acceptance/stepDefinitions/loginContext.js
@@ -110,6 +110,10 @@ Then('the user should be redirected to the IdP login page', function() {
   return client.page.ownCloudAuthorizePage().waitForPage()
 })
 
+Given('the user has been redirected to the login error page', function() {
+  return client.page.loginErrorPage().waitTillLoaded()
+})
+
 Then('the user should be redirected to the login error page', function() {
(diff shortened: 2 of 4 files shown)


dpakach (commented):
LGTM :+1:

tests/acceptance/features/webUILogin/login.feature
```diff
@@ -70,3 +70,16 @@ Feature: login users
     Then the user should be redirected to the login error page
     When the user exits the login error page
     Then the user should be redirected to the login page
+
+  Scenario: the user session of a deleted user should not be valid for newly created user of same name
+    Given these users have been created with default attributes:
+      | username |
+      | user1    |
+    And user "user1" has logged in using the webUI
+    And the user has browsed to the files page
```
jasson99, line 79: is this necessary?

tests/acceptance/features/webUILogin/login.feature
```diff
@@ -70,3 +70,16 @@ Feature: login users
...
+    And user "user1" has been deleted
+    And the user has reloaded the current page of the webUI
+    And the user has been redirected to the login error page
```
jasson99, line 82: I think I have some confusion in the given steps. for example in this case, `And the user has been redirected to the login error page` is in the given step, but it is actually the then step of the previous line i.e `the user has reloaded the current page of the WebUI`. What I think is, these two can be combined actually, something like `And the user has been redirected to the login error page after reloading the current page of the webui` Or we can have a new scenario where the user after being deleted reloads the current page in the `when` step and then its then step is `it should be redirected to the login error page` and in this scenario, we can use a single given step for that such as `And the deleted user has reloaded the current page of the WebUI` whose implementation can be the reload action and the assertion of being redirected to the login error page. But I am not sure if the present implementation is fine. Can you make me clear @phil-davis @individual-it @swoichha ?

phil-davis, line 82:
```suggestion
    When the user reloads the current page of the webUI
    Then the user should have been redirected to the login error page
```

Maybe this scenario can end here, and then have another scenario with similar given steps plus 
```
And user "user1" has been created with default attributes
```
which checks the case where the original user has been deleted, and the same user name has been created again, all while the original user stayed logged-in at the browser.

Because actually we probably want to cover both scenarios.

phil-davis: Hmmm - that looks just like the scenario at line 62. So we already have done that.

So for this new scenario we can just move `And user "user1" has been created with default attributes` up to be straight after `And user "user1" has been deleted` and then tthe user can reload the page and test what happens

individual-it: I think we have 2 tests here:
1. When the user is deleted And user reload the page Then the user should be redirected to new page
2. When the user is deleted And the user is recreated And user reload the page Then the user should be redirected to new page

The first test should pass, the second one should fail because of the bug

Will separating the test help to solve the confusion? And maybe we even don't need new steps for that

swoichha: @jasson99 You are right line 82 is the `Then` step of line 81. I wrote it in `Given` step because it is the precondition for this tests. I can combine these two into a single step as you have suggested but making a new scenario is not necessary for "where the user after being deleted reloads the current page" as there are already some test for that case.

swoichha: Or maybe I can just change them as @phil-davis suggested here https://github.com/owncloud/web/pull/4682#discussion_r569130670

tests/acceptance/features/webUILogin/login.feature
individual-it: I believe after the login the user will be on the files-page anyway

tests/acceptance/stepDefinitions/loginContext.js
```diff
@@ -110,6 +110,10 @@ Then('the user should be redirected to the IdP login page', function() {
   return client.page.ownCloudAuthorizePage().waitForPage()
 })
 
+Given('the user has been redirected to the login error page', function() {
```
individual-it, line 113: unused step AFAICS

individual-it (changes requested):

tests/acceptance/stepDefinitions/loginContext.js
swoichha: Removed

jasson99: It isn't removed @swoichha

tests/acceptance/features/webUILogin/login.feature
jasson99, line 79: @swoichha I think we don't need this step.

jasson99 (commented):
LGTM now :+1:

individual-it (approved):

PR #4689, opened 2021-02-03
download password protected public shared files using pre-signed url

C0rby:
## Description
Replaced the blob download with normal downloads using pre-signed urls provided by the backend.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

The blob download had some disadvantages like delayed download for big files and high memory consumption.

## How Has This Been Tested?

- locally using the OC10 backend

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/mixins/actions/fetch.js
+++ b/packages/web-app-files/src/mixins/actions/fetch.js
@@ -1,17 +1,20 @@
-import { checkRoute } from '../../helpers/route'
+import MixinRoutes from '../routes'
+import { mapGetters } from 'vuex'
 
 export default {
+  mixins: [MixinRoutes],
   computed: {
+    ...mapGetters(['configuration']),
     $_fetch_items() {
       return [
         {
           icon: 'remove_red_eye',
-          handler: file => this.$_fetch_trigger(file.path, 'application/pdf'),
+          handler: file => this.$_fetch_trigger(file, 'application/pdf', this.isPublicFilesRoute),
           ariaLabel: () => {
             return this.$gettext('Open in browser')
           },
           isEnabled: ({ resource }) => {
-            if (checkRoute(['files-trashbin'], this.$route.name)) {
+            if (this.isTrashbinRoute) {
               return false
             }
 
@@ -23,8 +26,14 @@ export default {
     }
   },
   methods: {
-    $_fetch_trigger(filePath, mimetype) {
-      const url = this.$client.helpers._webdavUrl + filePath
+    $_fetch_trigger(file, mimetype, isPublicFile) {
+      if (isPublicFile) {
+        const url = file.downloadURL
+        window.open(url, '_blank')
+        return
+      }
+
+      const url = this.$client.helpers._webdavUrl + file.path
       const headers = new Headers()
 
       headers.append('Authorization', 'Bearer ' + this.getToken)
--- a/packages/web-app-media-viewer/src/App.vue
+++ b/packages/web-app-media-viewer/src/App.vue
@@ -167,10 +167,10 @@ export default {
         a: 1
       }
 
-      return this.$_loader_getDavFilePath(this.activeMediaFile.path, query)
+      return this.$_loader_getDavFilePath(this.activeMediaFile, query)
     },
     rawMediaUrl() {
-      return this.$_loader_getDavFilePath(this.activeMediaFile.path)
+      return this.$_loader_getDavFilePath(this.activeMediaFile)
     },
 
     videoExtensions() {
@@ -263,7 +263,7 @@ export default {
       // workaround for now: Load file as blob for images, load as signed url (if supported) for everything else.
       let promise
       if (this.isActiveMediaFileTypeImage || !this.isUrlSigningEnabled) {
-        promise = this.mediaSource(url, 'url', null)
+        promise = this.mediaSource(decodeURIComponent(url), 'url', null)
       } else {
         promise = this.$client.signUrl(url, 86400) // Timeout of the signed URL = 24 hours
       }
--- a/packages/web-runtime/src/plugins/web.js
+++ b/packages/web-runtime/src/plugins/web.js
@@ -6,6 +6,7 @@ export default {
       computed: {
         ...mapGetters(['getToken', 'isAuthenticated', 'capabilities']),
         ...mapGetters('Files', ['publicLinkPassword']),
+        ...mapGetters(['configuration']),
 
         currentExtension() {
           return this.$route.path.split('/')[1]
@@ -32,7 +33,7 @@ export default {
           let url = null
           let headers = {}
           if (publicPage) {
-            url = this.$client.publicFiles.getFileUrl(file.path)
+            url = file.downloadURL
             const password = this.publicLinkPassword
             if (password) {
               headers = {
@@ -49,7 +50,7 @@ export default {
           }
 
           // download with signing enabled
-          if (this.isUrlSigningEnabled) {
+          if (!publicPage && this.isUrlSigningEnabled) {
             try {
               const response = await fetch(url, {
                 method: 'HEAD',
@@ -57,17 +58,7 @@ export default {
               })
               if (response.status === 200) {
                 const signedUrl = await this.$client.signUrl(url)
-                // create anchor dom element
-                const a = document.createElement('a')
-                a.style.display = 'none'
-                document.body.appendChild(a)
-                a.href = signedUrl
-                // use download attribute to set desired file name
-                a.setAttribute('download', file.name)
-                // trigger the download by simulating click
-                a.click()
-                // cleanup
-                document.body.removeChild(a)
+                this.triggerDownload(signedUrl, file.name)
                 return
               }
             } catch (e) {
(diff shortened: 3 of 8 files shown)


apps/files/src/store/state.js
```diff
@@ -19,7 +19,8 @@ export default {
     '{http://owncloud.org/ns}size',
     '{DAV:}getlastmodified',
     '{DAV:}getetag',
-    '{DAV:}resourcetype'
+    '{DAV:}resourcetype',
+    '{http://owncloud.org/ns}downloadURL'
```
LukasHirt, line 23: Is the download URL available only for public links or also for private files?

kulmann: Hehehehehe, I asked the same thing in the chat. It is not. I would love to have all download urls already by the server. Then we could get rid of the url signing in the sdk and would not need to make any difference between public and private context.

src/plugins/web.js
```diff
@@ -84,70 +85,16 @@ export default {
...
-          // request.addEventListener('progress', e => {
-          //   const progress = (e.loaded / e.total) * 100
-          //   this.updateFileProgress({
-          //     fileName: file.name,
-          //     progress
-          //   })
-          // })
-
-          request.send()
+          const a = document.createElement('a')
```
kulmann, line 88: I would like to get rid of this code duplication with the `a` element.

kulmann: Can be moved out into its own method.

apps/files/src/store/state.js
LukasHirt: Hehe. Exactly my thoughts!

src/plugins/web.js
C0rby: Should I tackle this here in this PR?

kulmann: Yes please :-)

C0rby:
The tests are failing because the backends which are used to test against do not contain the API changes yet.

C0rby:
Here is a PR which fixes some of the test failures: https://github.com/owncloud/ocis/pull/1785

C0rby:
I don't get it...these tests only fail in the CI. Locally they run successfully.
@kulmann @LukasHirt @phil-davis, any idea?

## webUIBasic-chrome
webUIPreview/imageMediaViewer.feature:15
webUIPreview/imageMediaViewer.feature:14
webUIPreview/imageMediaViewer.feature:65
webUIPreview/imageMediaViewer.feature:13
webUIPreview/imageMediaViewer.feature:145
webUIPreview/imageMediaViewer.feature:127
webUIPreview/imageMediaViewer.feature:54
webUIPreview/imageMediaViewer.feature:57
webUIPreview/imageMediaViewer.feature:74

## webUIOCISBasic-chrome
webUIPreview/imageMediaViewer.feature:14
webUIPreview/imageMediaViewer.feature:65
webUIPreview/imageMediaViewer.feature:13
webUIPreview/imageMediaViewer.feature:54
webUIPreview/imageMediaViewer.feature:57
webUIPreview/imageMediaViewer.feature:74

packages/web-app-files/src/mixins/actions/fetch.js
```diff
@@ -23,8 +28,14 @@ export default {
     }
   },
   methods: {
-    $_fetch_trigger(filePath, mimetype) {
-      const url = this.$client.helpers._webdavUrl + filePath
+    $_fetch_trigger(file, mimetype, isPublicFile) {
+      if (isPublicFile) {
+        const url = this.$client.helpers.instance + file.downloadURL.substring(1)
```
kulmann, line 33: Please use the server url from the store.
Add `import { mapGetters } from 'vuex'` as import and `...mapGetters(['configuration']),` within the computed properties. The you can use the server url like this: `this.configuration.server`. It is guaranteed to have a trailing slash.
The `$client` naturally has the same url, because it is initialized with it. But you should grab it from the source, not from an intermediate object.

kulmann: And yes, it is already super ugly to use the webdav url like this from the client. ;-) But that can be fixed later. For the server url it's a quick win.

packages/web-app-media-viewer/src/App.vue
```diff
@@ -167,7 +167,18 @@ export default {
         a: 1
       }
 
-      return this.$_loader_getDavFilePath(this.activeMediaFile.path, query)
+      if (this.$_loader_publicContext && this.activeMediaFile.downloadURL) {
```
kulmann, line 170: Could you move this entire block into the `$_loader_getDavFilePath` mixin and just provide the entire `activeMediaFile` instead of the path? Then you're also able to get rid of the path traversal with `../../../`. The URL format is well known to us, I'd prefer string replaces instead of traversals if possible.

packages/web-runtime/src/plugins/web.js
```diff
@@ -32,7 +32,8 @@ export default {
           let url = null
           let headers = {}
           if (publicPage) {
-            url = this.$client.publicFiles.getFileUrl(file.path)
+            // url = this.$client.publicFiles.getFileUrl(file.downloadURL)
+            url = this.$client.helpers.instance + file.downloadURL.replace(/\//, '')
```
kulmann, line 36: use the `configuration.server` url again instead.

kulmann: And get rid of the commented line please :-)

kulmann (changes requested):
Some small changes needed. Thanks for pushing this change! It's a really good step forward. 🚀

kulmann (changes requested):
Found something ugly. The backend returns the following downloadURL:
`/px1/remote.php/dav/public-files/hXzyzP24D3YC2hQ/Screenshot 2021-03-16 at 13.35.31.png?signature=7e922b4728b71f969c0024d1e826141591615a2a7543f45a67daed2237b1b043&expires=2021-03-16T14%3A35%3A01%2B00%3A00`
and it gets concatenated with the server url, which is `http://host.docker.internal/px1/`, resulting in a duplicated `px1` path. My ownCloud is in a subfolder, which is a valid setup.
Does this require a core change again? 👀

C0rby:
Ah damn it.. Possible... let me check

C0rby:
@kulmann, what do you think would make more sense?

Put the full url into `downloadURL` like: `http://localhost:8080/owncloud/remote.php/dav/public-files/hXzyzP24D3YC2hQ/Screenshot 2021-03-16 at 13.35.31.png?signature=7e922b4728b71f969c0024d1e826141591615a2a7543f45a67daed2237b1b043&expires=2021-03-16T14%3A35%3A01%2B00%3A00`
Or always only put `/remote.php/dav/public-files/hXzyzP24D3YC2hQ/Screenshot 2021-03-16 at 13.35.31.png?signature=7e922b4728b71f969c0024d1e826141591615a2a7543f45a67daed2237b1b043&expires=2021-03-16T14%3A35%3A01%2B00%3A00`

kulmann:
> @kulmann, what do you think would make more sense?
> 
> Put the full url into `downloadURL` like: `http://localhost:8080/owncloud/remote.php/dav/public-files/hXzyzP24D3YC2hQ/Screenshot 2021-03-16 at 13.35.31.png?signature=7e922b4728b71f969c0024d1e826141591615a2a7543f45a67daed2237b1b043&expires=2021-03-16T14%3A35%3A01%2B00%3A00`
> Or always only put `/remote.php/dav/public-files/hXzyzP24D3YC2hQ/Screenshot 2021-03-16 at 13.35.31.png?signature=7e922b4728b71f969c0024d1e826141591615a2a7543f45a67daed2237b1b043&expires=2021-03-16T14%3A35%3A01%2B00%3A00`

Would be super handy to the full download url already available. But is that again more error prone (regarding server side configuration)? If yes, that it's not an issue having to prefix the path with the server url from the web config.

C0rby:
Yeah, I think just sending the relative URL is more flexible for more deployment scenarios.

jnweiger:
> Yeah, I think just sending the relative URL is more flexible for different deployment scenarios.

Why do we go for flexibility? I'd argue that a full URL is more robust, as no errors can be made during URL-assembly.

kulmann (approved):
LGTM now 🚀

kulmann:
CI is failing for `webUIPreview/imageMediaViewer.feature:24` and `webUIPreview/imageMediaViewer.feature:33` because the core changes are not in master, yet. You'd need to switch to the release branch and respective commit id in the `.drone.env` file for now. I don't know if this is fine for @phil-davis and @individual-it though.

individual-it:
you can change the branch in `.drone.env` to prove that the change is working, but I would not merge it that way, better to get the change into core first or add the failing test to the expected failures

micbar:
@individual-it we are a bit in a cross-blocking situation.

The change in core is committed to the `release-10.7.0` branch.
It will be merged back to master after the release has been done.

Currently the server release is waiting for a new web release with this fix here.

We need to unblock here.

Raising to P1

PR #4745, opened 2021-02-15
Fix l10n paths

LukasHirt:
Use correct paths in l10n makefile after switching to rollup


LukasHirt:
@kulmann Is this enough to make translations happy or something else needed?

kulmann (approved):

kulmann:
I added a commit with some more fixes. Thanks for pushing this!

PR #4759, opened 2021-02-19
Supress redirect error

LukasHirt:
We've suppressed the error appearing in the console which warned about redirect happening after the oidc callback page. This error is being shown because after the oidc callback has successfully processed the authorization request we are redirecting to the `/` path which immediately does another redirect to the extension set as default one. In the context of Vue router, this is considered an error even though for us it is a valid use case. The error is only informative thus no issue is going to surface if we suppress it. This way we are getting closer to a clean console without errors.


kulmann (approved):
👍

PR #4783, opened 2021-03-02
Preserve symlinks

LukasHirt:
Preserve symlinks so we can use local links for development. Increase the limit for node version to < 15.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "@rollup/plugin-json": "^4.1.0",
     "@vue/test-utils": "^1.1.2",
     "archiver": "^3.0.0",
+    "autoprefixer": "^10.2.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "@rollup/plugin-json": "^4.1.0",
     "@vue/test-utils": "^1.1.2",
     "archiver": "^3.0.0",
+    "autoprefixer": "^10.2.4",
```
LukasHirt, line 43: We're not using this directly on the Web but it's needed when using the link for ODS

pascalwengerter:
Works like a charm for me! 👌🏽

pascalwengerter (approved):

Issue #4680, opened 2021-02-02
Adjust webUI test scenarios with "tries to" in its step


swoichha:
Currently we have tests with "tries to" in Gherkin. Which are used when the actions (like MOVE, COPY, RENAME) performed intended to give an error which are checked mostly in `Then` step. 

But, now we want to check if the action is actually performed or not. 

For example:
```
  Scenario: cancel moving a file
    Given user "user1" has logged in using the webUI
    And the user has browsed to the files page
    When the user tries to move file "data.zip" into folder "simple-empty-folder" using the webUI
    And the user cancels the attempt to move file into folder "simple-empty-folder" using the webUI
    Then file "data.zip" should be listed on the webUI
    But  file "data.zip" should not be listed in the folder "simple-empty-folder" on the webUI
```
 Here in this scenario we want to cancel moving of a file but this test doesn't actually move a file. Maybe we can do more precise while writing step definition or we can have more explicit description of each specific scenario.

phil-davis:
We need to think of words that explain enough of what is happening here. Instead of "tries to" we could write "start to" - or write more words explaining what point in the user workflow the user is getting to.

phil-davis:
```
When the user opens the move file dialog to move file "data.zip" into folder "simple-empty-folder" using the webUI
```
Maybe?

phil-davis:
For these sort of cases, when we want to test scenarios that "interrupt" the normal workflow of the UI, we will write much smaller `When` steps that take the scenario through more detail of the UI workflow. That will allow the scenario to stop at any point in the workflow and then take another path (for example, "the user cancels...").

For example, the UI workflow might have 3 steps like:
- the user opens the file actions menu
- the user selects the "Move" action
- the user selects "some-folder" as the place to move the file

For the "normal" workflow, we keep having a single step written in the scenario. The single describes what the user wants to achieve, not the detail of all the "clicks". That makes the "normal" scenarios more readable and less brittle.

phil-davis:
I will add some documentation. See issue https://github.com/owncloud/docs/issues/3128 for that.

pascalwengerter:
@SwikritiT is this resolved now? :)

SwikritiT:
@pascalwengerter the issue described is addressed by the PR above but we still need to do some refactoring in test scenarios with `tries to` in its step. 

For Example 
```
 Scenario: copy a file into another folder with no change permission
    Given user "Brian" has been created with default attributes
[...]

PR #4800, opened 2021-03-09
[tests-only] Adjusts webUI test scenarios with tries to in its step while canceling copy and move

SwikritiT:
## Description
Some of the scenarios that had `tries to` in its step and didn't follow the normal workflow of the UI , weren't described in a detail manner .This PR tries to explain those kind of scenarios in a more detailed manner of the UI workflow.

## Related Issue

- https://github.com/owncloud/web/issues/4680

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
+++ b/tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
@@ -88,9 +88,7 @@ Feature: move files
     Given user "Brian" has been created with default attributes
     And user "Brian" has shared folder "simple-folder" with user "Alice" with "read" permissions
     And user "Alice" has logged in using the webUI
-    When the user opens the file action menu of file "lorem.txt" using the webUI
-    And the user selects move action for file "lorem.txt" using the webUI
-    And the user selects the folder "simple-folder (2)" as a place to move the file using the webUI
+    When the user tries to move file "lorem.txt" into folder "simple-folder (2)" using the webUI
     Then it should not be possible to move into folder "simple-folder (2)" using the webUI
 
   Scenario: cancel moving a file
--- a/tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
+++ b/tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
@@ -88,13 +88,17 @@ Feature: move files
     Given user "Brian" has been created with default attributes
     And user "Brian" has shared folder "simple-folder" with user "Alice" with "read" permissions
     And user "Alice" has logged in using the webUI
-    When the user tries to move file "lorem.txt" into folder "simple-folder (2)" using the webUI
+    When the user opens the file action menu of file "lorem.txt" using the webUI
+    And the user selects move action for file "lorem.txt" using the webUI
+    And the user selects the folder "simple-folder (2)" as a place to move the file using the webUI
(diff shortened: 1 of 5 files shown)


tests/acceptance/stepDefinitions/filesContext.js
```diff
@@ -1199,6 +1207,10 @@ When(
   }
 )
 
+When('the user uses the move button to move files using the webUI', function() {
```
Talank, line 1210:
```suggestion
When('the user selects move to move files using the webUI', function() {
```
How about this? :thinking:

tests/acceptance/stepDefinitions/filesContext.js
```diff
@@ -1235,6 +1270,18 @@ When(
...
 
+When('the user selects following files/folders to copy/move using the webUI', async function(
+  resources
+) {
+  for (const item of resources.rows()) {
+    await client.page.FilesPageElement.filesList().toggleFileOrFolderCheckbox('enable', item[0])
+  }
+})
+
+When('the user uses the copy button to copy files using the webUI', function() {
```
dpakach, line 1281: I think something like `the user clicks the copy button...` sounds better here.

tests/acceptance/stepDefinitions/filesContext.js
```diff
@@ -1235,6 +1270,18 @@ When(
   }
 )
 
+When('the user selects following files/folders to copy/move using the webUI', async function(
```
dpakach, line 1273: We may already have a step for batch selection. If not maybe change this step to 
`the user selects following files/folders for batch action using the webUI`

tests/acceptance/stepDefinitions/filesContext.js
swoichha: IMO it would be better if we have something like this `When('the user selects move button to move files using the webUI', function() {`. For me the `selects move to move` seems a bit confusing.

swoichha (commented):
lgtm

tests/acceptance/stepDefinitions/filesContext.js
jasson99: But we don't use click normally, do we?

tests/acceptance/features/webUIMoveFilesFolders/moveFiles.feature
```diff
@@ -88,25 +88,31 @@ Feature: move files
...
   Scenario: cancel moving of multiple files at once
     Given user "Alice" has logged in using the webUI
     And the user has browsed to the files page
-    When the user tries to batch move these files into folder "simple-empty-folder" using the webUI
+    When the user marks these files for batch action using the webUI
       | file_name   |
       | data.zip    |
       | lorem.txt   |
       | testapp.zip |
+    And the user clicks the move button to move files using the webUI
```
jasson99, line 114: I would not prefer using click here, as we do not use it normally. May be we can use select here. You can decide about what would be better.

Talank (commented):
lgtm

jasson99 (commented):
LGTM

individual-it (approved):

PR #4837, opened 2021-03-22
Added only office extension opening into the existing window

rpocklin:
## Description
This is an adaptation of the Only Office external app for OCIS which is compatible with OC10 backend without requiring any OCIS server or runtime.  This simplifies the integration and avoids using a customer `handler` function to integrate Only Office.

This is partly raised to demonstrate another approach for integrating OO, and could be used as a stepping stone to a deeper integration down the track to actually use the JS APIs in OO directly (for things like automated background saves and triggering behaviour via post messages between the application and the extension.

I have raised another similar PR https://github.com/owncloud/web/pull/4836 which opens in a new window (as per existing design).  If this approach continued to implement the direct JS API, I believe opening in the same window is a better approach (in terms of UX) but would warrant being able to hide the left-hand sidebar (which is possible) and also adding a top 'Back' button or other actions at the top to facilitate navigating out of the integrated document editor.

* Could also specify `hideSideBar: true` in route to claim more screen real estate (will raise in a separate PR) but this will disappear if the user has a constrainted viewport by default anyway.

## Motivation and Context
It simplifies the integration point of Only Office in OCIS.  This implementation still uses the OC10 OO plugin, but could be updated to not use it at all (and use direct JS integration with the OO JS API).

## How Has This Been Tested?
Tested on the latest version of **owncloud/web** for a week, involving opening each type of document, making changes, going back into the parent app, seeing the changes then re-editing or downloading to confirm changes were applied.

Keep in mind you will have to add:
```
apps: [..., "onlyoffice"],
```
to your `config.json` to enable this app.

image
image

## Types of changes
- [x] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/only-office-integration-same-window
+++ b/changelog/unreleased/only-office-integration-same-window
@@ -1,0 +1,5 @@
+Enhancement: Added only office extension opening into the existing window
+
+This integrates the OC10 OnlyOffice extension into the OCIS web frontend rendering it in the same window.
+
+https://github.com/owncloud/web/pull/4837
--- a/packages/web-app-files/src/mixins/fileActions.js
+++ b/packages/web-app-files/src/mixins/fileActions.js
@@ -49,11 +49,11 @@ export default {
           icon: this.apps.meta[editor.app].icon,
           handler: item => this.$_fileActions_openEditor(editor, item.path, item.id),
           isEnabled: ({ resource }) => {
-            if (editor.routes && !checkRoute(editor.routes, this.$route.name)) {
-              return false
+            if (editor.handler || (editor.routes && checkRoute(editor.routes, this.$route.name))) {
+              return resource.extension === editor.extension
             }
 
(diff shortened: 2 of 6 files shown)


rpocklin:
Rebased and updated (now that `hideSideBar` feature has been merged into `master`)

rpocklin:
Closing due to lack of feedback.

kulmann:
Hey, sorry for not getting back to you on this. The onlyoffice team is currently working on an ownCloud Web extension for their oc10 connector app. Happening here: https://github.com/ONLYOFFICE/onlyoffice-owncloud-web/tree/develop
I bet they'd be more than happy if you contributed there! Our `web` repo is just not the right place since this is not supposed to be a monorepo for all extensions we could imagine. But you're working on a hot topic here. Much appreciated, thank you!

PR #4840, opened 2021-03-23
Added upload pending prompt on close/reload

### LINKED ISSUE #2590: Warn leaving users about running uploads
### Steps

1. Upload many files
1. During the upload, close the browser tab or navigate to another URL

### Expected result
A browser warning appears to tell that the upload will cancel if the user proceeds to leave.

### Actual result
No warning, upload is cancelled. Confused user.

This should be a metter of adding a `window.onbeforeunload` or similar to the right places and check for pending uploads (or any other operation, really)

@LukasHirt @hodyroff as discussed

anaswaratrajan:
## Description

Added an unload event listener that prompts user if there are pending uploads while user closes/ reloads/ navigates to another URL. 

## Related Issue

- Fixes #2590

## Motivation and Context
The pending uploads are cancelled on closing/reloading window. A prompt is required to warn users of upload cancellation. 

## How Has This Been Tested?

- tested with the latest changes from owncloud/web master branch
- tested by uploading many files and also with large files
- tested on tab close/ reload/ navigating to another URL
- tested on Chrome 88.0 and Firefox 86.0

owncloud-2590

## Types of changes

- [ ] Detect tab close/ reload/ navigation and prompt if any uploads are pending. 

## Checklist:

- [ ] Added unload event listener 
- [ ] Remove unload event listener before component destroy

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/UploadProgress.vue
+++ b/packages/web-app-files/src/components/UploadProgress.vue
@@ -108,6 +108,10 @@ export default {
       } else {
         return 100
       }
+    },
+
+    uploadPending() {
+      return this.totalUploadProgress < 100
     }
   },
   watch: {
@@ -122,12 +126,19 @@ export default {
     this.$root.$on('upload-start', () => {
       this.$nextTick(() => {
         this.delayForScreenreader(() => this.$refs.progressbar.$el.focus())
+        window.addEventListener('beforeunload', this.handlerClose);
       })
     })
   },
   methods: {
     $_toggleExpanded() {
       this.expanded = !this.expanded
+    },
+    handlerClose(event) {
+      if (this.uploadPending) {
+        event.preventDefault()    
+        event.returnValue = ""
+      }
     }
   }
 }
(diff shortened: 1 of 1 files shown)


pascalwengerter (changes requested):
Works like a charm locally, I'll do a quick research and see if we can add a test to our suite :)

@anaswaratrajan could you add a changelog item to this PR inside the changelog/unreleased folder? Name of the file should follow the schema of `enhancement-your-change` (since it's neither a bugfix nor a breaking change), and for the content of the file you can take inspiration from the existing changelog items!

pascalwengerter (approved):
LGTM 🚀

pascalwengerter:
@kulmann there seems to be a way to test this behaviour (see https://github.com/SeleniumHQ/selenium/issues/8638), but not sure if we want to have tests for this and if so on which level to add them..

anaswaratrajan:
@pascalwengerter @kulmann Should I look into adding a unit test using jest?

pascalwengerter:
> @pascalwengerter @kulmann Should I look into adding a unit test using jest?

That would be lovely, yeah! Not sure how to handle long-running uploads in there, but perhaps setting the `uploadPending` somewhere between 0 and 100 manually would be a first step?

kulmann (approved):
Thanks for your contribution! 🚀

PR #4863, opened 2021-03-25
Remove <main> tags from container & runtime and move them to extension level

### LINKED ISSUE #1652: Document has no main content area

pascalwengerter:
## Description
As explained in #1652 we only can have one `<main>` tag per HTML document. This PR removes the ones in `web-container` and `web-runtime` and adds one to each extension (files-list, mediaviewer, markdowneditor, drawio) since they can't be loaded at the same time.

## Related Issue
- Fixes #1652

## Motivation and Context
Important for accessibility-compliance (and to produce valid HTML)

## Types of changes


kulmann (approved):
LGTM 👍

PR #4822, opened 2021-03-17
Allow for custom theming

pascalwengerter:
## Description
Initialize the ownCloud Design System plugin by passing in the theme provided in `config.json`. Since we're breaking things here anyways I've refactored `theme.json` to being able to load more than one theme (while starting the application with a `"default"` one) so switching themes (e.g. into a darkmode one) won't introduce another breaking change.

## Related Issue
- Fixes #2362 partly

## Motivation and Context
As requested by customers and users, the new ownCloud `web` should be themable by providing custom images/logos, colors, fontsizes and spacing. This PR introduces the ability to change images/logos and colors and adds documentation on how theming can be achieved.

## How Has This Been Tested?
- We'll have unit tests in ODS, checking that CSS variables are initialized from defaults and overwritten if custom values are provided
- We'll add a visual regression test to the web CI later

## Types of changes
- [X] Change

## Checklist:
- [X] Code changes

## Open tasks:
- [x] Wait for ODS 4.0.0 to be merged into `web` (#4862)
- [x] Visual regression test added
- [ ] Documentation added
- [ ] Changelog item added

## Todos:
- [x] Change loading, make loading from URL and from within the bundle possible.
  - Option 1: specify base-url or folder name in the config.json (`"theme": "owncloud"` / `"theme": "https://owncloud.online/themes/"`). In this case we need a fixed name for the json file (`theme.json`) which cannot be changed.
  - Option 2: specify url or path to the theme json file, so that the name of the json file is not set in stone / can be decided by admin (e.g. `"theme": "owncloud/hello.json"` / `"theme": "https://owncloud.online/themes/hello.json"`).
  - Important in any case: fallback to bundled owncloud theme if theme loading fails
  - Think about asset paths in theme.json. Should be relative and as short as possible.
  - Bonus: Figure out if we can get rid of `WEB_UI_CONFIG_THEME` in oCIS or how to handle it when loading the theme
- [x] theming: login page (disable auto-redirect first)
- [x] theming: public link with uploader role
- [x] theming: dropzone in normal file listing

PR #4904, opened 2021-04-01
test

kulmann:
triggering a CI run to see if user creation in CI fails coming directly from master


kulmann:
turned out that https://github.com/owncloud/web/pull/4822 had an issue in the code leading to no pages being rendered. of course that failed all the pipelines. closing this here now.

PR #4917, opened 2021-04-06
Update appSwitcher titles & fix typo in group memberships

pascalwengerter:
## Description
Related to accessibility & translations
- Changed "Groups membership" => "Group memberships"
- Changed the way extension titles are rendered in the appSwitcher, so they hopefully don't show up as `{{ n.title }}` in Transifex anymore

### DIFF
--- a/packages/web-runtime/src/components/ApplicationsMenu.vue
+++ b/packages/web-runtime/src/components/ApplicationsMenu.vue
@@ -23,11 +23,11 @@
         <div v-for="(n, nid) in menuItems" :key="`apps-menu-${nid}`" class="uk-width-1-3">
           <a v-if="n.url" key="apps-menu-external-link" :target="n.target" :href="n.url">
             <oc-icon :name="n.iconMaterial" size="xlarge" />
-            <div v-translate>{{ n.title }}</div>
+            <span class="uk-display-block" v-text="$gettext(n.title)" />
           </a>
           <router-link v-else key="apps-menu-internal-link" :to="n.path">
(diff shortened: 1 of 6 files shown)


fschade (approved):
Nice, thanks LGMT 👍

packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -23,11 +23,11 @@
         <div v-for="(n, nid) in menuItems" :key="`apps-menu-${nid}`" class="uk-width-1-3">
           <a v-if="n.url" key="apps-menu-external-link" :target="n.target" :href="n.url">
             <oc-icon :name="n.iconMaterial" size="xlarge" />
-            <div v-translate>{{ n.title }}</div>
+            <div v-text="$gettext(n.title)" />
```
kulmann, line 26: `v-translate` should be supported by vue-gettext. Maybe it's unsupported to use it on `div`s? But this brings me to another issue: a `div` inside an `a` tag is invalid html markup. Could you try if this is visually still the same if you use a span? You could also use a `<translate>` tag, that will result in a span as well.

pascalwengerter: Uff, great eye! Will go with a `<span>` and `display: block;`.

Also, had a discussion with @fschade yesterday and we figured dropping `v-translate` and `<translate>` in favor of `$gettext()` would be benefical for statical code analysis down the road

kulmann (approved):
👍

PR #4934, opened 2021-04-08
Import ODS translations & merge with web translations

pascalwengerter:
## Description
Needs another ODS release (4.4.0) to work and a rebase after the ODS release is in web master, but should be good to go

Edit: Depends on #4940

### DIFF
--- a/packages/web-runtime/src/index.js
+++ b/packages/web-runtime/src/index.js
@@ -25,6 +25,7 @@ import Vue2TouchEvents from 'vue2-touch-events'
 // --- Gettext ----
 import GetTextPlugin from 'vue-gettext'
 import coreTranslations from '../l10n/translations.json'
+import odsTranslations from 'owncloud-design-system/dist/system/translations.json'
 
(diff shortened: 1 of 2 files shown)


packages/web-runtime/src/index.js
```diff
@@ -81,7 +82,8 @@ const supportedLanguages = {
   it: 'Italiano',
   gl: 'Galego'
 }
-const translations = coreTranslations
+
+const translations = Object.assign(coreTranslations, odsTranslations)
```
fschade, line 86: object.assign is not deep. For example

```javascript
Object.assign({de: {foo: 1}}, {de: {bar: 1}})
```

will result in 
```javascript
{de: {bar: 1}}
```

and not 
```javascript
{de: {foo: 1, bar: 1}}
```

it's saver to use lodash
```javascript
const foo = merge({}, coreTranslations, odsTranslations)
```

fschade (approved):
👍

PR #4942, opened 2021-04-09
Update web container

pascalwengerter:
## Description
- A `meta` element with an `http-equiv` attribute whose value is X-UA-Compatible must have a content attribute with the value `IE=edge`. (fixed in rollup)
- Element `h3` not allowed as child of element `span` in this context.
- The type attribute is unnecessary for JavaScript resources.


fschade (approved):
👍

kulmann (approved):

PR #4993, opened 2021-04-21
Focus-Management

### LINKED ISSUE #4992: Focus management
## Acceptance criteria
- never lose keyboard focus
- meaningful order of focussable elements
- always stay in context / resume old context

fschade:
## Description
to get a better background why this is needed please read #4992 beforehand.

## Related Issue

- Fixes #4992 

## Motivation and Context
This pr introduces the concept of user driven focus management.

This is helpful when a user navigates using the keyboard and the given context switches, for example the sidebar or and a modal opens which has a different tabindex and the user will never get there (more or less). 

Now it's possible to decide which item should be focused after a given action (modal open, sidebar show). 
The important part here is that those interactions need to be able to be rolled back.  (replayed in a reverse order).

- if a user closes a modal, the focus should jump back to the exact position where he started.
- if a sidebar opens the focus should jump into the sidebar or vice versa if it gets closed.

following worst case scenario is part of the idea behind it.
- user tabs through the page
- user opens the sidebar by pressing enter on the dedicated button
- the sidebar opens and the focus needs to jump to the close button of it
- the user tabs through the links in the sidebar and hits enter on "rename"
- the rename modal opens, the user tabs to update
- the modal closes and the focus must jump back to "rename" in sidebar
- the user tabs back to the sidebar close button and press enter which closes the sidebar
- the focus must jump back to the element where the user started (the three dots for example)

all this needs to be replayable in random order.

On top of this we also fixed some minor issues regarding tabindex in the top-bar. It also includes a new global lifecycle mixing which makes it possible to listen to events like mounted, replayable, ... without the requirement that those are defined by the consumed component.

## How Has This Been Tested?

- manual

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] tests
- [ ] discuss the scope of this
- [ ] changelog

### DIFF
--- a/changelog/unreleased/enhancement-focus-management
+++ b/changelog/unreleased/enhancement-focus-management
@@ -1,0 +1,7 @@
+Enhancement: Focus management
+
+We added a mixin that makes it able to manage, record and reverse-replay the focus for the current document.
+The first components that using it are modal and sidebar in the files app.
+
+https://github.com/owncloud/web/pull/4993
+https://github.com/owncloud/web/issues/4992
--- a/packages/web-app-files/src/App.vue
+++ b/packages/web-app-files/src/App.vue
@@ -14,8 +14,13 @@
     <sidebar
       v-if="_sidebarOpen"
       id="files-sidebar"
+      ref="filesSidebar"
+      tabindex="-1"
       class="uk-width-1-1 uk-width-1-2@m uk-width-1-3@xl"
       @reset="setHighlightedFile(null)"
+      @beforeDestroy="focusSideBar"
+      @mounted="focusSideBar"
+      @fileChanged="focusSideBar"
     />
   </main>
 </template>
@@ -89,6 +94,13 @@ export default {
     openSideBar(file, sideBarName) {
       this.setHighlightedFile(file)
       this.SET_APP_SIDEBAR_EXPANDED_ACCORDION(sideBarName)
+    },
+    focusSideBar(component, event) {
+      this.focus({
+        from: document.activeElement,
+        to: this.$refs.filesSidebar?.$el,
+        revert: event === 'beforeDestroy'
+      })
     },
 
     $_ocApp_dragOver() {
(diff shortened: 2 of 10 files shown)


packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -21,11 +21,17 @@
...
+            key="apps-menu-external-link"
+            :target="n.target"
+            :href="n.url"
+            tabindex="0"
+          >
             <oc-icon :name="n.iconMaterial" size="xlarge" />
             <span class="uk-display-block" v-text="$gettext(n.title)" />
           </a>
-          <router-link v-else key="apps-menu-internal-link" :to="n.path">
+          <router-link v-else key="apps-menu-internal-link" :to="n.path" tabindex="0">
```
pascalwengerter, line 34: `tabinde="0"` shouldn't strictly be necessary (tab navigation works in Chrome without it and doesn't work in Firefox/Safari either way)

pascalwengerter (commented):
Just a question regarding tabindex, rest looks good & works like a charm!

packages/web-runtime/src/components/ApplicationsMenu.vue
fschade: solved

kulmann:
@fschade The initial focus of the right sidebar is the `close` button now. That seems to be wrong. IMO it should be on the sidebar itself, not on the close button. AFAIK a screen reader would start reading with the close button now instead of with the resource name.

kulmann:
Clicking the file row instead of the three dots button results in resetting the focus to the top of the page after closing the right sidebar.

fschade:
> @fschade The initial focus of the right sidebar is the `close` button now. That seems to be wrong. IMO it should be on the sidebar itself, not on the close button. AFAIK a screen reader would start reading with the close button now instead of with the resource name.

good point, fixed. Thanks

kulmann (approved):
nice 😍

PR #4997, opened 2021-04-22
Enhance a11y on app-switcher

JanAckermann:
## Description
* Fix aria-expand on the button, which was set to false, while opening the oc-drop
* Wrap app switcher in nav element
* Use list element for items
* Replace the visible function with UI Kit functions because i did not work at all  

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/components/ApplicationsMenu.vue
+++ b/packages/web-runtime/src/components/ApplicationsMenu.vue
@@ -1,14 +1,13 @@
 <template>
-  <div>
+  <nav>
     <oc-button
       id="_appSwitcherButton"
       ref="menubutton"
       appearance="raw"
       class="oc-topbar-menu-burger"
       :aria-label="applicationSwitcherLabel"
-      :uk-tooltip="applicationSwitcherLabel"
     >
-      <oc-icon name="apps" class="uk-flex" />
+      <oc-icon :uk-tooltip="applicationSwitcherLabel" name="apps" class="uk-flex" />
     </oc-button>
     <oc-drop
       ref="menu"
@@ -19,8 +18,8 @@
       class="uk-width-large"
       close-on-click
     >
-      <div class="uk-grid-small uk-text-center" uk-grid>
-        <div v-for="(n, nid) in menuItems" :key="`apps-menu-${nid}`" class="uk-width-1-3">
+      <ul class="uk-grid-small uk-text-center" uk-grid>
+        <li v-for="(n, nid) in menuItems" :key="`apps-menu-${nid}`" class="uk-width-1-3">
           <a v-if="n.url" key="apps-menu-external-link" :target="n.target" :href="n.url">
             <oc-icon :name="n.iconMaterial" size="xlarge" />
             <span class="uk-display-block" v-text="$gettext(n.title)" />
@@ -29,23 +28,19 @@
             <oc-icon :name="n.iconMaterial" size="xlarge" />
             <span class="uk-display-block" v-text="$gettext(n.title)" />
           </router-link>
-        </div>
-      </div>
+        </li>
+      </ul>
     </oc-drop>
-  </div>
+  </nav>
 </template>
 
 <script>
 import NavigationMixin from '../mixins/navigationMixin'
+import UiKit from 'uikit'
 
 export default {
   mixins: [NavigationMixin],
   props: {
-    visible: {
-      type: Boolean,
-      required: false,
-      default: false
-    },
     applicationsList: {
       type: Array,
       required: false,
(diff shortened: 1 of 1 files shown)


JanAckermann:
@pascalwengerter @kulmann we are not quite sure with the UI Kit events here, I think this should be moved in to the oc-drop component where we can pass callbacks for onClose and onOpen

@pascalwengerter There is also an issue, when opening the app switcher the first item gets focused but the style won't be applied, any idea?

packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -1,14 +1,13 @@
...
     <oc-button
       id="_appSwitcherButton"
       ref="menubutton"
       appearance="raw"
       class="oc-topbar-menu-burger"
       :aria-label="applicationSwitcherLabel"
-      :uk-tooltip="applicationSwitcherLabel"
     >
-      <oc-icon name="apps" class="uk-flex" />
+      <oc-icon :uk-tooltip="applicationSwitcherLabel" name="apps" class="uk-flex" />
```
JanAckermann, line 10: Moved uk-tooltip to the icon because it overwrites the aria expand of the button

packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -60,14 +57,15 @@ export default {
...
     }
   },
-  watch: {
-    visible(val) {
-      if (val) {
-        this.focusFirstLink()
-      } else {
-        this.$emit('closed')
-      }
-    }
```
JanAckermann, line 70: this never worked

JammingBen:
> @pascalwengerter There is also an issue, when opening the app switcher the first item gets focused but the style won't be applied, any idea?

A bit more context to that: Focus seem to work a little different depending on the element. You can notice that when clicking a button -> it will have a focus border. In contrast to clicking a link -> it wont have a focus border. When navigating via keyboard, each element has a focus border.

Now the elements in the application switcher are links. I think, because of this, the border on the first element wont show when opening the application switcher. It works with those elements being buttons. It also works when opening the switcher via keyboard. Just not on click.

packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -20,32 +19,30 @@
...
-          <a v-if="n.url" key="apps-menu-external-link" :target="n.target" :href="n.url">
-            <oc-icon :name="n.iconMaterial" size="xlarge" />
-            <span class="uk-display-block" v-text="$gettext(n.title)" />
-          </a>
-          <router-link v-else key="apps-menu-internal-link" :to="n.path">
-            <oc-icon :name="n.iconMaterial" size="xlarge" />
-            <span class="uk-display-block" v-text="$gettext(n.title)" />
-          </router-link>
-        </div>
+        <ul v-for="(n, nid) in menuItems" :key="`apps-menu-${nid}`" class="uk-width-1-3 uk-list">
```
pascalwengerter, line 22: v-for needs to happen on the `<li>` level if I'm not mistaken? 🤔

pascalwengerter (changes requested):

pascalwengerter (commented):
The code looks fine but I'm having troubles noticing anything different when running locally and checking for anything in the browser. Is it purely for screenreaders, does it only work in specific browsers or am I missing something? :)

JanAckermann:
> The code looks fine but I'm having troubles noticing anything different when running locally and checking for anything in the browser. Is it purely for screenreaders, does it only work in specific browsers or am I missing something? :)

Can you give some context according to your question?

pascalwengerter:
> > The code looks fine but I'm having troubles noticing anything different when running locally and checking for anything in the browser. Is it purely for screenreaders, does it only work in specific browsers or am I missing something? :)
> 
> Can you give some context according to your question?

Sure, sorry my comment was a bit obscure in retrospect. What part of a11y is improved by this PR? I didn't notice different behaviour when checking out the changes locally and looking at the UI + navigating with the keyboard (which I kinda assumed would've been improved by your changes, but I may be mistaken 😉 )

pascalwengerter:
Aside from that, if you're done here you could remove the `[WIP]` in the PR title and we could merge it?

JanAckermann:
@pascalwengerter to make it short:
Opening the App switcher menu will focus on the first link now, but unfortunately, the focus style won't be applied if you click on the app switcher button. But if you hit the app switcher button with tab and press enter, the style will be applied.
I have no clue why this is the case, and not sure if we can merge it in that state

pascalwengerter (approved):
LGTM, focus management in FF still weird, but we'll update this with the new layout soon too I guess

PR #5008, opened 2021-04-23
Move location picker hint under breadcrumbs

LukasHirt:
## Description
We've moved the hint that is describing how to use the Location picker from sidebar under the breadcrumbs. There is navigation of the Files extension displayed in the sidebar now instead.

image

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-runtime/src/App.vue
```diff
@@ -158,7 +154,7 @@ export default {
     },
 
     sidebarNavItems() {
-      if (this.publicPage()) {
+      if (!this.user.token) {
```
pascalwengerter, line 157: What was the motivation behind this change here?

packages/web-runtime/src/App.vue
```diff
@@ -26,14 +26,10 @@
             :logo-img="logoImage"
             :product-name="productName"
             :nav-items="sidebarNavItems"
-            :hide-nav="sidebar.navigationHidden"
```
pascalwengerter, line 29: Are we removing the general capability of hidding the sidebar? 🤔

pascalwengerter (commented):
Looks great visually, yet I sometimes end up with a forever loading spinner, an `PROPFIND` error or an `Uncaught (in promise)` error

packages/web-runtime/src/App.vue
LukasHirt: `publicPage` is true for a page that can be used in both authorised/unauthorised context. So if the user would be in location picker with his own private resources he wouldn't see the navigation because it is a public page.

packages/web-runtime/src/App.vue
LukasHirt: The whole sidebar can be still hidden with `isSidebarVisible`. This is only about the navigation items. And from my understanding the sidebar should behave now only as navigation so I do not see much point in giving the option to hide it :shrug:

LukasHirt:
> I sometimes end up with a forever loading spinner, an PROPFIND error or an Uncaught (in promise) error

I guess that happens only in the location picker? I'll dig into it

PR #5011, opened 2021-04-23
Re-enable yarn link ods by disabling postcss-config in rollup

pascalwengerter:
## Description
After changing the way dependencies in ODS were declared some weeks ago, running `yarn link owncloud-design-system` failed due to a PostCSS config error. This change re-enables linking & testing locally between `web` and `owncloud-design-system`


LukasHirt (approved):
Nice! :rocket:

PR #5030, opened 2021-04-27
Translate quick action labels/tooltips

pascalwengerter:
## Description
See code/changelog item. Found another unsused CSS class (deleted by the evil PR ™️ ) that we could let go


kulmann (approved):
👍

PR #5067, opened 2021-05-04
Make both share buttons outlined & passive variation

pascalwengerter:


kulmann:
Already pushed directly to the a11y-swarming branch here https://github.com/owncloud/web/commit/3b29131b3516324ae685232766446346f5c2a1f6

PR #5076, opened 2021-05-07
[Tests-Only] removed redis information from web testing docs

kiranparajuli589:
## Description
After the use of `ocis` server with `ocis storage driver`, _redis_ usage is not required. With this PR that information is also removed from the testing documentation

## Related Issue

- https://github.com/owncloud/ocis/issues/2008


phil-davis (approved):
LGTM

PR #5122, opened 2021-05-18
Fix indirect via-share in collaborator sidebar section

### LINKED ISSUE #5082: The via button sends you to the parent of the share entry
The "Via" button send you to the parent of the share entry, however this folder may not be accessible to the user.
The "Via" button should send you the to share entry.

For example:

If I have received a share to folder Two, the user will see the following path:
`/eos/user/d/diogo/One/Two/Three`

If I am in Three and click "Via" button I get redirected to "One" where I should have been redirected to "Two".

pascalwengerter:
## Description
The "Via" button in the collaborators according section in the sidebar sends you to the parent of the share entry, while it actually should send you the to share entry itself.

## Related Issue
- Fixes #5082


LukasHirt (approved):
LGTM 🚀

PR #5149, opened 2021-05-25
[debatable] Hide left sidebar if no navitems present

pascalwengerter:
## Description
Felt surprisingly easy, I'd vote in favor of only hiding the navigation (and not the whole sidebar) to keep the iconic CI blue plus OC logo present for public links though...

### DIFF
--- a/packages/web-runtime/src/App.vue
+++ b/packages/web-runtime/src/App.vue
@@ -225,6 +225,9 @@ export default {
     },
 
     isSidebarVisible() {
+      if (this.sidebarNavItems.length === 0) {
+        return false
+      }
       return this.windowWidth >= 1200 || this.appNavigationVisible
     },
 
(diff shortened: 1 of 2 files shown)


packages/web-runtime/src/App.vue
```diff
@@ -220,6 +220,10 @@ export default {
       return 'uk-visible@l'
     },
 
+    isSidebarItems() {
+      return this.sidebarNavItems.length > 0
```
kulmann, line 224: Please add
```js
if (this.sidebarNavItems.length === 0) {
    return false
}
```
to `isSidebarVisible` instead.

kulmann (changes requested):

packages/web-runtime/src/App.vue
kulmann: Would like to move towards `early returns` in code style.

pascalwengerter: Code is updated to reflect this 🤓

kulmann (approved):
LGTM. I agree that it's not an ideal solution regarding usability. But an empty left sidebar feels even weirder, so I would do it this way until we have the changed layout with the two left sidebar.

PR #5154, opened 2021-05-26
Change main navigation aria-label

pascalwengerter:
## Description
Updated aria-label for main navigation `<nav>` element


kulmann (approved):
👍

PR #5155, opened 2021-05-26
Add tooltip/ariaLabel for New Button if user has permissions

pascalwengerter:
## Description
Add tooltip/ariaLabel for New Button if user has permissions to upload/create resources

### DIFF
--- a/packages/web-app-files/src/components/AppBar.vue
+++ b/packages/web-app-files/src/components/AppBar.vue
@@ -22,80 +22,78 @@
         v-if="showActions || selectedFiles.length > 0 || isTrashbinRoute"
         class="uk-flex uk-flex-middle oc-p-s"
       >
-        <template v-if="showActions">
-          <template v-if="areDefaultActionsVisible">
-            <oc-button
-              id="new-file-menu-btn"
-              key="new-file-menu-btn-enabled"
-              v-oc-tooltip="_cannotCreateDialogText"
-              :aria-label="_cannotCreateDialogText"
-              variation="primary"
-              appearance="filled"
-              :disabled="isNewBtnDisabled"
-            >
-              <oc-icon name="add" />
-              <translate>New</translate>
-            </oc-button>
-            <oc-drop
-              drop-id="new-file-menu-drop"
-              toggle="#new-file-menu-btn"
-              mode="click"
-              close-on-click
-              :options="{ delayHide: 0 }"
-            >
-              <ul class="uk-list">
-                <li>
-                  <file-upload
-                    :path="currentPath"
-                    :headers="headers"
-                    @success="onFileSuccess"
-                    @error="onFileError"
-                    @progress="onFileProgress"
-                  />
-                </li>
-                <li v-if="checkIfBrowserSupportsFolderUpload">
-                  <folder-upload
-                    v-if="!isIE11()"
-                    :root-path="currentPath"
-                    :path="currentPath"
-                    :headers="headers"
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/AppBar.vue
```diff
@@ -147,14 +145,14 @@ export default {
...
-    _cannotCreateDialogText() {
+    newButtonDescription() {
       if (!this.canUpload) {
         return this.$gettext('You have no permission to upload!')
       }
       if (!this.hasFreeSpace) {
         return this.$gettext('You have not enough space left to upload!')
       }
-      return null
+      return "Add files or folders"
```
kulmann, line 155: This now adds a tooltip for a non-error case as well. IMO not good. Looks like it's required to separate into two different computed props, one for the tooltip and one for the aria-label.

pascalwengerter: Please check the solution I went for and let me know if you like it!

packages/web-app-files/src/components/AppBar.vue
```diff
@@ -306,6 +300,19 @@ export default {
...
+      if (!this.canUpload) {
+        return this.$gettext('You have no permission to upload!')
+      }
+      if (!this.hasFreeSpace) {
+        return this.$gettext('You have not enough space left to upload!')
+      }
+      if (isToolTip === true) {
+        return null
+      }
+      return 'Add files or folders'
```
kulmann, line 313: This line lacks `this.$gettext`

packages/web-app-files/src/components/AppBar.vue
```diff
@@ -146,15 +144,11 @@ export default {
...
-    _cannotCreateDialogText() {
-      if (!this.canUpload) {
-        return this.$gettext('You have no permission to upload!')
-      }
-      if (!this.hasFreeSpace) {
-        return this.$gettext('You have not enough space left to upload!')
-      }
-      return null
+    newButtonTooltip() {
+      return this.newButtonDescription(true)
```
kulmann, line 148: `newButtonTooltip()` could do what we had before:
```js
if (!this.canUpload) {
         return this.$gettext('You have no permission to upload!')
       }
       if (!this.hasFreeSpace) {
         return this.$gettext('You have not enough space left to upload!')
       }
       return null
```

packages/web-app-files/src/components/AppBar.vue
```diff
@@ -146,15 +144,11 @@ export default {
...
+    },
+    newButtonAriaLabel() {
```
kulmann, line 150: ... and `newButtonAriaLabel()` could then do the following:
```js
const tooltip = this.newButtonTooltip
if (tooltip) {
    return tooltip
}
return this.$gettext('Add files or folders')
```

packages/web-app-files/src/components/AppBar.vue
```diff
@@ -306,6 +300,19 @@ export default {
       this.createModal(modal)
     },
 
+    newButtonDescription(isToolTip) {
```
kulmann, line 303: While this is totally fine technically, it is hard to read in the location where it's called. `this.newButtonDescription(true)` or `this.newButtonDescription(false)` doesn't give a hint what effect the boolean has on the output. I'd prefer a solution without the boolean, as pointed out in the two comments above.

kulmann (changes requested):

kulmann (approved):
LGTM 🤖

PR #5166, opened 2021-05-28
Focus breadcrumb on route change

pascalwengerter:
## Description
Focus current active breadcrumb on route change and announce contents of current folder to screenreader

### DIFF
--- a/packages/web-app-files/src/mixins/accessibleBreadcrumb.js
+++ b/packages/web-app-files/src/mixins/accessibleBreadcrumb.js
@@ -1,0 +1,35 @@
+import last from 'lodash-es/last'
+import { mapGetters } from 'vuex'
+
+export default {
+  computed: {
+    ...mapGetters('Files', ['activeFilesCount'])
+  },
+  methods: {
+    accessibleBreadcrumb_focusAndAnnounceBreadcrumb() {
+      const activeBreadcrumb = last(
+        document.getElementById('files-breadcrumb').children[0].children
+      )
+      const activeBreadcrumbItem = activeBreadcrumb.lastChild
+
+      const announcement = this.$gettextInterpolate(
+        'containing %{filesCount} files and %{folderCount} folders',
+        { filesCount: this.activeFilesCount.files, folderCount: this.activeFilesCount.folders }
+      )
+
+      const translatedHint =
+        this.activeFilesCount.folders > 0 || this.activeFilesCount.files > 0
+          ? announcement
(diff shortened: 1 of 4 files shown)


kulmann (changes requested):
Found one issue that I didn't think about before. This now steals the focus away from the nav directly after clicking in the left sidebar. As far as I understand this behaviour is only supposed to happen upon navigating within the file list or through the breadcrumbs.

pascalwengerter:
> Found one issue that I didn't think about before. This now steals the focus away from the nav directly after clicking in the left sidebar. As far as I understand this behaviour is only supposed to happen upon navigating within the file list or through the breadcrumbs.

Tried preventing it with what we discussed, checking if we're staying within the same route and only setting the focus on the last breadcrumb item if that's the case

packages/web-app-files/src/mixins/accessibleBreadcrumb.js
```diff
@@ -0,0 +1,37 @@
...
+
+      const announcement = this.$gettextInterpolate(
+        'containing %{filesCount} files and %{folderCount} folders',
+        { filesCount: this.activeFilesCount.files, folderCount: this.activeFilesCount.folders }
+      )
+
+      const translatedHint =
+        this.activeFilesCount.folders > 0 || this.activeFilesCount.files > 0
+          ? announcement
+          : this.$gettext('This folder has no content')
```
kulmann, line 23:
```suggestion
          : this.$gettext('This folder has no content.')
```

packages/web-app-files/src/mixins/accessibleBreadcrumb.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+  methods: {
+    accessibleBreadcrumb_focusAndAnnounceBreadcrumb(sameRoute) {
+      const activeBreadcrumb = last(
+        document.getElementById('files-breadcrumb').children[0].children
+      )
+      const activeBreadcrumbItem = activeBreadcrumb.lastChild
+
+      const announcement = this.$gettextInterpolate(
+        'containing %{filesCount} files and %{folderCount} folders',
```
kulmann, line 16: Don't make the distinction between files and folders. Just use `items` and with correct pluralization. `This folder contains %{ itemCount } item.` for singular form and `This folder contains %{ itemCount } items.` plural form.

kulmann (changes requested):

kulmann (approved):
LGTM 🚀

PR #5169, opened 2021-05-28
[tests-only] send apache log to /dev/null

individual-it:
## Description

reduce apache output log in tests

## Motivation and Context
these logs are usually not needed and with that change we will save a couple of hundreds lines of log output in every step of the pipeline

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1124,6 +1124,7 @@ def owncloudService():
         "pull": "always",
         "environment": {
             "APACHE_WEBROOT": "/var/www/owncloud/server/",
+            "APACHE_LOGGING_PATH": "/dev/null",
         },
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
LGTM.
note: I thought about adding a switch up in the config section to control this, to maybe make it easier for developers to turn it off. But IMO the logs will be very rarely needed, and it is very simple to just delete the `APACHE_LOGGING_PATH` lines in order to get the logs back.

Issue #4933, opened 2021-04-08
adjust webUI test scenarios with "tries to" in its step


kiranparajuli589:
Description: see issue description from here https://github.com/owncloud/web/issues/4680 

After issue https://github.com/owncloud/web/issues/4680, tries to steps with following are refactored:
- [x] tries to move
- [x] tries to copy

TODO:
- [x] tries to rename https://github.com/owncloud/web/pull/5134
- [x] tries to create a public link share https://github.com/owncloud/web/pull/5137
- [x] tries to create files/folder https://github.com/owncloud/web/pull/5157
- [x] tries to login https://github.com/owncloud/web/pull/5162
- [x] tries to navigate private liink https://github.com/owncloud/web/pull/5174

SwikritiT:
I think we still need some adjustments in webUI test scenarios with "tries to" in its step for renaming , so I'm un-checking the box and adding it to TODO 

for eg: 
https://github.com/owncloud/web/blob/4bf7899f57dbf8f10a2c0258e4c6e5bdb2022671/tests/acceptance/features/webUIRenameFiles/renameFiles.feature#L182 
The file should be renamed in this case and we shouldn't use `tries to rename`  we should use `rename` instead. 

Similar case here 
https://github.com/owncloud/web/blob/4bf7899f57dbf8f10a2c0258e4c6e5bdb2022671/tests/acceptance/features/webUIRenameFolders/renameFolders.feature#L113

https://github.com/owncloud/web/blob/4bf7899f57dbf8f10a2c0258e4c6e5bdb2022671/tests/acceptance/features/webUIRenameFolders/renameFolders.feature#L144

SwikritiT:
I couldn't find anything that can be refactored with `tries to` share so I'm removing it from todo

SwikritiT:
@individual-it @phil-davis can you take a look at this issue and make sure if we can close it?

PR #5174, opened 2021-05-31
[tests-only] Adjust steps that try to navigate private links

SwikritiT:
## Description
This PR adjusts the test step that tries to navigate private link, meaning that the navigation process is supposed to be unsuccessful.

The changes made by this PR and how they work

- When a user `tries to` navigate a private link which they don't have permission to , they shouldn't be able to do so and should be redirected to the login page. When a user with permission accesses the private link then `#webContainer` should be visible to them and they shouldn't be logged out . But for user trying to access it without permission  `#webContainer` shloudn't be present and they should be redirected to login page. This PR adds that extra check

## Related Issue

- part of : https://github.com/owncloud/web/issues/4933

## Motivation and Context
These changes were necessary because previously in steps dealing with `tries to` we used to check the errors in `then` and in `when `step we used to check the behavior of the webUI while trying to do certain operation like `navigating private link` in this case but we didn't check everything.
Now we want our `when` step to perform all the task that we perform while actually `navigating private link` and we want to make sure that the process is unsuccessful at when step, so that we know the cause if that step fails for some reason or starts behaving in undesired manner.

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/stepDefinitions/privateLinksContext.js
+++ b/tests/acceptance/stepDefinitions/privateLinksContext.js
@@ -36,6 +36,9 @@ When(
   'the user tries to navigate to the private link created by user {string} for file/folder {string}',
   async function(user, resource) {
     const item = await webdav.getProperties(resource, user, ['oc:privatelink'])
-    await client.url(item['oc:privatelink'])
+    await client
+      .url(item['oc:privatelink'])
+      .page.webPage()
+      .waitForElementNotPresent('@webContainer')
   }
 )


kiranparajuli589:
looks good!

kiranparajuli589 (commented):
looks good!

individual-it (approved):

PR #5232, opened 2021-06-10
Fix the way .feature files are handled in .gitattributes

pascalwengerter:
## Description
Added the wrong line of code 🥴 now it should work fine


kulmann (approved):
:trollface:

PR #5235, opened 2021-06-10
[CI-only] fix path in continuous deployed example

wkloucek:
## Description
Nightly deployment / rebuild of the deployment example fails because of a wrong path: https://drone.owncloud.com/owncloud/web/16102/66/3


pascalwengerter (approved):
LGTM 🍁

PR #5265, opened 2021-06-15
[full-ci] [tests-only] Do not run acceptance tests if only unit tests files are changed

kiranparajuli589:
## Description
With this PR, acceptance test suites are checked before and decided whether to run or skip w.r.t diff. If diff contains unit test files only, acceptance tests steps are skipped.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5231

## How Has This Been Tested?

- 🤖 
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #5329, opened 2021-06-22
[tests-only] [full-ci] Fix some typos in acceptance test code

phil-davis:
This is on top of PR #5265 to verify what CI runs.


phil-davis:
I will test after PR #5265 is merged. The PR has changes to `.drone.star` and other scripts etc, so it should always run all of CI.
We have to merge it first, then we can make other PRs that only change tests... and observe the real effect.

PR #5341, opened 2021-06-22
[tests-only] change something in a unit test

phil-davis:


phil-davis:
https://drone.owncloud.com/owncloud/web/16731

That works as expected - the acceptance tests skip through all the steps and do not actually run anything.

Note: we do not have a way to embed the calculate-diff step into the starlark drone pipeline generation process. If we could do that somehow, then we could not even generate the acceptance test pipelines. For that, we would need access to the cloned GitHub repo during the running of the starlark (which is on the drone server), so that a `git diff` could be analysed to see which files have changed.

PR #5279, opened 2021-06-16
[full-ci][tests-only]Changed some steps that were creating user with default attributes

Talank:
## Description
Some steps were missed while refactoring scenarios to avoid the use of skeleton as much as possible. Those missed steps has been fixed in this PR. Also, since the step `Given('user {string} has been created with default attributes', async function(userId)` is being no longer used, I have removed that step definition.

## Related Issue
- Part of https://github.com/owncloud/QA/issues/659
- 
## How Has This Been Tested?
- 🤖 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #5354, opened 2021-06-23
[tests-only] Add a scenario to expected-failures that does not fail

phil-davis:
Investigate why #5279 did not report an unexpected pass - there seems to be some problem that the CI pipeline is not finding unexpected passes.


phil-davis:
That's a problem - an unexpected pass should have been reported, and a pipeline should have failed.

phil-davis:
I raised issue #5356

phil-davis:
Fixed by PR #5403

Issue #5356, opened 2021-06-23
unexpected passes are not reported in CI


phil-davis:
See PR #5354 

If I add some extra scenarios to an expected-failures file, then a CI pipeline should report some unexpected passes (actually some expected failures that did not fail - they might be feature file line numbers that do not exist)

This problem caused a bit of difficulty working out what was going wrong in PR #5279 - in that case an existing feature file was edited and line numbers changed. It looked like a scenario was suddenly unexpectedly failing.

If the CI had also reported an "unexpected pass" then we would have straight away realized to just adjust the feature file line number.

We could also change the words "unexpected passes" to "expected failures that did not fail" - will that make more sense, or will it confuse everyone even more?

phil-davis:
See example in #5354 and #5361 that demonstrate that unexpected pass/success does not get reported.

PR #5361, opened 2021-06-24
[tests-only] Add a scenario to expected-failures that actually does pass

phil-davis:
To investigate issue #5356 

In PR #5354 I added a "scenario" to expected-failures that referenced a line number that is not actually a scenario. (That should have caused a fail, but did not)

In this PR, I add a "scenario" to expected-failures that references a line number that is an actual scenario that runs and passes.
That should cause an "unexpected pass" to be reported, and a pipeline fail.


ownclouders:
Results for <strong>oCISSharingBasic</strong> https://drone.owncloud.com/owncloud/web/16810/47/1

phil-davis:
https://drone.owncloud.com/owncloud/web/16810/45/10
`openidLogin.feature:24` actually passed. But at the end there are "no unexpected success".

So the code is not processing unexpected pass/success correctly.

phil-davis:
Fixed by PR  #5403

PR #5366, opened 2021-06-24
[CI] start acceptance tests only after unit and integration tests are complete

dpakach:
… complete

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -744,11 +744,15 @@ def beforePipelines(ctx):
     return yarnlint() + changelog(ctx) + website(ctx) + cacheOcisPipeline(ctx)
 
 def stagePipelines(ctx):
+    unitTestPipelines = unitTests(ctx)
     acceptancePipelines = acceptance(ctx)
     if acceptancePipelines == False:
-        return unitTests(ctx)
+        return unitTestPipelines
 
-    return unitTests(ctx) + acceptancePipelines
+    if ("acceptance-tests-only" not in ctx.build.title.lower()):
+        dependsOn(unitTestPipelines, acceptancePipelines)
+
+    return unitTestPipelines + acceptancePipelines
 
 def afterPipelines(ctx):
     return build(ctx) + notify()


pascalwengerter:
@dpakach as before, please kill builds once the acceptance tests start when they shouldn't to reduce CI load😉

dpakach:
@pascalwengerter, I just wanted to check that that builds start as expected.
A lot of times we also create PR with only acceptance tests change where the chances of unit tests failing are highly unlikely. So I thought we could start the acceptance tests and unit tests at the same time in such case. looks like it works as expected.

Another thing is that for unit tests there is a `build-web` step. We dont really need to build it just for running unit tests . Maybe we can remove that to make the unit tests pipeline a bit faster..

phil-davis:
> Another thing is that for unit tests there is a build-web step. We don't really need to build it just for running unit tests . Maybe we can remove that to make the unit tests pipeline a bit faster.

That took 89 seconds. If we don't use it for unit/integration tests then there is no point running the step there.

sonarcloud takes about 2 minutes. Waiting for that delays the start of acceptance tests. 

Note: I will be optimizing the npm-install stuff next, so I could also look at making the start of acceptance tests not depend on waiting for sonarcloud.

phil-davis (approved):
This looks OK to me - it does what is required in the issue.

@pascalwengerter do we merge this, and then we (I) look at optimizations when I split out the npm-install stuff?

pascalwengerter:
> > Another thing is that for unit tests there is a build-web step. We don't really need to build it just for running unit tests . Maybe we can remove that to make the unit tests pipeline a bit faster.
> 
> That took 89 seconds. If we don't use it for unit/integration tests then there is no point running the step there.
> 
> sonarcloud takes about 2 minutes. Waiting for that delays the start of acceptance tests.
> 
> Note: I will be optimizing the npm-install stuff next, so I could also look at making the start of acceptance tests not depend on waiting for sonarcloud.

Actually I think it's fine if acceptance tests don't run if sonarcloud is unhappy. @kulmann opinions?

pascalwengerter:
> This looks OK to me - it does what is required in the issue.
> 
> @pascalwengerter do we merge this, and then we (I) look at optimizations when I split out the npm-install stuff?

Yes please. Let the CI stop if unit/integration tests fail is awesome, and you taking care of the npm-install stuff next is 🥇

ownclouders:
Results for <strong>oCISSharingBasic</strong> https://drone.owncloud.com/owncloud/web/16832/47/1

PR #5420, opened 2021-06-30
Apply CSP rules for onlyoffice and richdocuments

kulmann:
## Description
The iframe integrations of the documentserver coming from onlyoffice / richdocuments need certain CSP rules. With this PR we're extracting the respective documentserver domains from the two apps. We'll work on a generic solution when it becomes relevant.

## Motivation and Context
Enable onlyoffice to build an integration of their documentserver into a web extension.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #5436, opened 2021-07-01
getting onlyoffice address from the app

LinneyS:
These edits can be included in the request #5420

I suggest using the method from the onlyoffice application, instead of duplicating a piece of code into your application

@kulmann


kulmann (approved):
Awesome, thanks! 🚀

PR #5450, opened 2021-07-05
Make view options button less prominent

kulmann:
This PR makes the `view options` button less prominent by giving it a `raw` appearance, making the gap size of the button smaller and shrinking the width of the drop to the actually needed size.

Screenshot 2021-07-05 at 17 06 05
Screenshot 2021-07-05 at 17 06 12

### DIFF
--- a/changelog/unreleased/enhancement-add-view-options
+++ b/changelog/unreleased/enhancement-add-view-options
@@ -5,3 +5,4 @@ Currently, it is possible to toggle visibility of hidden files.
 Changes in view options are persisted in local storage.
 
 https://github.com/owncloud/web/pull/5408
+https://github.com/owncloud/web/pull/5450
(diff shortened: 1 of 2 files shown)


ownclouders:
Results for <strong>oCISSharingBasic</strong> https://drone.owncloud.com/owncloud/web/17078/47/1

pascalwengerter (approved):
LGTM, there might be accessibility concerns regarding it not looking like a proper button though

kulmann:
> LGTM, there might be accessibility concerns regarding it not looking like a proper button though

I did some further reading about this. Seems like the most important part is to use buttons and links in a semantically correct way. That's fulfilled. Then it's helpful if buttons look like buttons for example if someone has an anxiety disorder and feels more comfortable knowing that buttons always look like buttons. But sometimes you need to use e.g. buttons in a text flow (i.e. inline buttons). Those obviously can't look like buttons, otherwise they would rip the text apart. This use case here obviously is no inline text, so the argument is not applicable. But we want the button to be as inconspicuous as possible, which is somewhat destroyed by the hover effect of the outline button...

PR #5600, opened 2021-08-02
[tests-only] make a useless change

phil-davis:
To verify that the new cancel-builds logic works OK when there is no previous build to cancel.


phil-davis:
https://drone.owncloud.com/owncloud/web/17886/2/2

Good - it finds just itself as the only build running for the PR, and does not cancel itself.

Now I will manually cancel this test build.

PR #5617, opened 2021-08-03
Fix overflow in context-menu links

pascalwengerter:
## Description
- Fixes screenshot below

Screenshot 2021-08-03 at 16 20 57

### DIFF
--- a/packages/web-app-files/src/components/FilesList/ContextActions.vue
+++ b/packages/web-app-files/src/components/FilesList/ContextActions.vue
@@ -187,14 +187,20 @@ export default {
 <style lang="scss">
 #oc-files-context-actions {
   text-align: left;
+  white-space: normal;
 
-  > li a,
-  > li a:hover {
-    text-decoration: none;
-    color: var(--oc-color-swatch-passive-default);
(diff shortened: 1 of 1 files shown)


kulmann (approved):
Good for now, thanks 👍 
What do you think about setting `class="uk-width-auto"` to the OcDrop in ODS instead (not for this release), so that the context menu gets wider if needed?

pascalwengerter:
> Good for now, thanks 👍
> What do you think about setting `class="uk-width-auto"` to the OcDrop in ODS instead (not for this release), so that the context menu gets wider if needed?

There's been feedback already that the OcDrop/Contextmenu is (almost) too wide so I wouldn't consider that a good way to go 🤔

kulmann:
> > Good for now, thanks 👍
> > What do you think about setting `class="uk-width-auto"` to the OcDrop in ODS instead (not for this release), so that the context menu gets wider if needed?
> 
> There's been feedback already that the OcDrop/Contextmenu is (almost) too wide so I wouldn't consider that a good way to go 🤔

cool, fine by me 👍

PR #5639, opened 2021-08-06
Improve wording of access denied message to user

rpocklin:
## Description
We have had much feedback that the existing message, while not incorrect is confusing for users where their session has expired or been logged out.  There may be multiple scenarios where users come to this page but it should cater for all of these an explain what has happened in a clear way.

Totally open to alterations on this message - it's an important page to communicate what has happened to the user.

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/pages/accessDenied.vue
+++ b/packages/web-runtime/src/pages/accessDenied.vue
@@ -12,7 +12,7 @@
           <span v-translate>Login Error</span>
         </h3>
         <h4 v-translate class="oc-mb-m">
-          You are not allowed to use this application.
+          Your user session is invalid or has expired.
         </h4>
         <div v-translate class="oc-mb-m" @click="performLogout">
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: there is a new drone pipeline "stop-recent-builds". That is failing for PRs from a fork, like this one.

I raised issue https://github.com/owncloud/web/issues/5640 - we will have a look and sort that out...

pascalwengerter (approved):

pascalwengerter:
@kulmann force-merge if you agree? Doesn't need acceptance tests imho

kulmann (approved):
Good change, thanks. Even nicer that it's based on user feedback 😍

phil-davis:
> Doesn't need acceptance tests imho

Agree - that text string is not referred to explicitly by any test.

PR #5690, opened 2021-08-18
fix: incorrect property name 'buttonConfirmDisabled'

rpocklin:
This was not passing through successfully to the OcModal component because the name was incorrect.

Actually it looks like there are several properties that aren't copied across ie.
`buttonConfirmVariation`
`buttonConfirmAppearance`
`buttonConfirmText`

There are other places that need to change (do not merge this) - I will raise this as an issue instead.
 ## Types of changes


ownclouders:
Results for <strong>oC10SharingInternalGroupsSharingIndicator</strong> https://drone.owncloud.com/owncloud/web/18386/24/1

PR #5709, opened 2021-08-20
Disable actions for the current folder in the right sidebar

kulmann:
This PR disables certain actions entirely when the given resource is the current folder. This is a UX improvement for the right sidebar actions, when the sidebar has the current folder as it's context.


fschade (approved):
nice 👍 LGTM

pascalwengerter (approved):
Awesome!

Issue #5234, opened 2021-06-10
list of components for unit testing


kiranparajuli589:
#### AppBar

- [x] AppBar.vue #5603

##### AppBar/SelectedResources
- [x] SizeInfo.vue #5522
- [x] BatchActions.vue #5282, #5519

##### AppBar/Upload
- [x] FileDrop.vue #5552
- [x] FileUpload.vue #5258
- [x] FolderUpload.vue #5559

#### FilesList
- [x] ListInfo.vue #5495
- [x] ListLoader.vue (test may not be required: has no logic)
- [x] NoContentMessage.vue #5517
- [x] NotFoundMessage.vue #5521
- [x] Pagination #5524 
- [x] QuickActions.vue #5526

#### Search
- [x] List.vue #5693
- [x] Preview.vue #5694

#### SideBar

- [ ] Sidebar.vue #5580
- [ ] FileInfo.vue #5580

##### SideBar/Actions
- [x] FileActions.vue #5557

##### SideBar/Links
- [x] PrivateLinkItem.vue #5638 
- [x] FileLinks.vue #5638
- [x] CopyToClipboardButton #5638

###### SideBar/Links/PublicLinks
- [x] LinkActions.vue #5577
- [x] LinkEdit.vue #5609 
- [x] LinkInfo.vue #5495 
- [x] ListItem.vue #5629

##### SideBar/Shared
- [ ] RoleItem.vue #5580

##### SideBar/Shares
- [ ] FileShares.vue https://github.com/owncloud/web/pull/5699

###### SideBar/Shares/Collaborators
- [x] AdditionalPermissions.vue #5646
- [x] AutocompleteItem.vue #5250
- [ ] CollaboratorsEditOptions.vue #5662
- [ ] ShowCollaborator.vue  #5267  #5295 #5308  WIP @individual-it 
- [ ] EditCollaborator.vue #5662
- [ ] NewCollaborator.vue https://github.com/owncloud/web/pull/5696

##### SideBar/Versions
- [x] FileVersions.vue #5649

#### Upload
- [x] DetailsWidget.vue #5674
- [x] ProgressBar.vue #5682

### packages/web-runtime/src/components/:
- [x] ApplicationsMenu.vue #5717
- [x] Avatar.vue #5718
- [x] FeedbackLink.vue
- [x] MessageBar.vue #5719
- [x] Notifications.vue #5218
- [x] OcAppContent.vue #5721
- [x] SidebarQuota.vue #5229
[...]

PR #5718, opened 2021-08-23
[unit-tests-only] Add unit tests for Avatar component

saw-jan:
## Description
Adds unit  tests for `Avatar` component

## Related Issue
- Part of #5234

## Motivation and Context
- increase unit tests coverage

## How Has This Been Tested?
- test environment: local & CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-runtime/tests/unit/components/Avatar.spec.js
+++ b/packages/web-runtime/tests/unit/components/Avatar.spec.js
@@ -31,13 +31,13 @@ describe('Avatar component', () => {
 
   const spySetUser = jest.spyOn(Avatar.methods, 'setUser')
 
-  it('should call "setUser" method on mount', () => {
+  it('should set user when the component is mounted', () => {
     getShallowWrapper()
     expect(spySetUser).toHaveBeenCalledTimes(1)
     expect(spySetUser).toHaveBeenCalledWith(propsData.userid)
   })
 
-  describe('when loading is set as true', () => {
+  describe('when the component is still loading', () => {
     it('should render oc-spinner but not oc-avatar', () => {
       const wrapper = getShallowWrapper(true)
       const spinner = wrapper.find(ocSpinner)
@@ -51,7 +51,7 @@ describe('Avatar component', () => {
     })
   })
 
-  describe('when loading is set as false', () => {
+  describe('when the component is not loading anymore', () => {
     let wrapper
     beforeEach(() => {
       wrapper = getShallowWrapper()
@@ -71,7 +71,7 @@ describe('Avatar component', () => {
       expect(avatar.props().userName).toEqual(propsData.userName)
     })
 
-    describe('when avatar is not found', () => {
+    describe('when an avatar is not found', () => {
       it('should set empty string to src prop on oc-avatar component', async () => {
         await new Promise(resolve => {
           setTimeout(() => {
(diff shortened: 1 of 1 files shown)


packages/web-runtime/tests/unit/components/Avatar.spec.js
```diff
@@ -0,0 +1,94 @@
+import { shallowMount, createLocalVue } from '@vue/test-utils'
+import Vuex from 'vuex'
+import DesignSystem from 'owncloud-design-system'
+
+import Avatar from '../../../src/components/Avatar.vue'
```
kiranparajuli589, line 5:
```suggestion
import Avatar from '@files/src/components/Avatar.vue'
```

packages/web-runtime/tests/unit/components/Avatar.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    ok: false,
+    status: 404
+  })
+)
+
+const ocSpinner = 'oc-spinner-stub'
+const ocAvatar = 'oc-avatar-stub'
+
+describe('Avatar component', () => {
+  describe('when loading is true', () => {
```
kiranparajuli589, line 28:
```suggestion
  describe('when loading is set as true', () => {
```

kiranparajuli589 (commented):
lgtm 👍🏼

packages/web-runtime/tests/unit/components/Avatar.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+      expect(spinner.exists()).toBeFalsy()
+      expect(avatar.exists()).toBeTruthy()
+    })
+    it('should set props on oc-avatar component', () => {
+      const avatar = wrapper.find(ocAvatar)
+
+      expect(avatar.props().width).toEqual(propsData.width)
+      expect(avatar.props().userName).toEqual(propsData.userName)
+      expect(avatar.props().src).toEqual('')
```
pascalwengerter, line 60: Please also test the case where `src` is set and check that the `setUser` method is being called ;)

packages/web-runtime/tests/unit/components/Avatar.spec.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function createStore() {
+  return new Vuex.Store({
+    getters: {
+      getToken: jest.fn(() => 'GFwHKXdsMgoFwt'),
+      capabilities: jest.fn(() => ({
+        files_sharing: {
+          user: {
+            profile_picture: true
```
pascalwengerter, line 85: Please check the `Avatar component` and have a test for `profile_picture` being both set to true and to false (and checking it causes the desired behavior)!

pascalwengerter (changes requested):
Looks good so far but I'd expect it to test some more functionaity 🤓

packages/web-runtime/tests/unit/components/Avatar.spec.js
pascalwengerter: Edit: Could also happen in an integration test I suppose. For the unit test it should be enough to spy & check that the `setUser` method is successfully being called on mounted

pascalwengerter (approved):
LGTM

PR #5712, opened 2021-08-20
App Provider

elizavetaRa:
## Description
Added the function to open files of certain mimetypes with corresponding apps

## Related Issue
https://github.com/cs3org/reva/issues/1923

## How Has This Been Tested?
- test environment: qa.cernbox.cern.ch
- test case: opening different files with Collabora and CodiMD

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [x ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Done
- Mimetypes list with applications gets retrieved on opening the frontend
- The corresponding to the file mimetype apps are shown in the sidebar actions and the context menu actions
- View for the applications opens in the new tab (new route   apps/:app/:file_id), iframe with the app

## Open tasks:
- make a package for appprovider instead of implementation in the web-app-files package
- Sharing button (opens sidebar sharing menu) in the top-bar in the application tab
- Create files with different extensions and on creation open them with default application
- (backend, in progress) default apps are first in the mimetypes list
- (backend, in progress) fix bugs with Collabora, CodiMD, MS Office

Issue #5460, opened 2021-07-06
Allow extensions to inject views in files app


diocas:
As we discussed before, this feature would allow extensions to hook into the files app in order to add extra buttons in the sidebar menu. These buttons would route to provided views, thus extending the normal files app with extra functionality.

Example, in our case we will create an extension to list our EOS Projects (while we wait for the spaces functionality).

@kulmann feel free to improve the description of this ticket.

fschade:
@diocas thank you for the request, we are currently defining how the api for such a feature could look like. We did something like that by using portals but not sure if this is a good fit in your case.

You have any idea or input how such a api could look like from the perspective of a extension developer?

diocas:
@elizavetaRa do you have suggestions?

JammingBen:
Closing here since this has been implemented via https://github.com/owncloud/web/pull/9814. We will shortly update our dev docs with all the information about the extension system.

Issue #5135, opened 2021-05-21
let extensions register more flexible file open actions


wkloucek:
As an extension developer I want an extension to be able to:
- [X] Register a name per file extension: "open in x", where x is the name, eg: "open in Collabora" and "open in CodiMD" #5805 
- [ ] Register a custom action name: "y in x", where y is the action name, eg: "open in x" or "edit in x" or "view in x" => needs 
- [ ] Register a file action dependent on the permissions on a file, usecase: if I received a share with viewer permissions I see only actions "view in x". Whereas when I received a share with editor permissions I also see the "edit in x" action => needs concepting
- [ ] Load the list of available file extensions from a non static external endpoint (eg. the extensions api) => needs concepting

wkloucek:
@kulmann The answer to all requests is currently not supported I suppose? ;-) Are there already plans to support it?

kulmann:
1) and 2) are easy, we can allow that file handlers define their own name for the action entirely.
For 3) we'd have to extend the first one even more and react differently depending on the permissions on the file. The resource (and with that the permissions) is available anyway, so we just have to come up with a reasonable mapping structure.
4) needs changes in the way we load extensions. As of now only static code is possible in extension loading, but for making your request possible we need access to the store of the runtime. Needs concepting first (we have ideas about this but need to write them down).

wkloucek:
A discussion with CERN today raised the question to not use the file extension but the mimetype from the propfind of a file when looking for the open in app action. This allows oCIS / REVA adminstrators to add custom mimetypes without changing the Web frontend. Should be easily doable, what do you think about this @kulmann?

Example:
mimetype `application/vnd.oasis.opendocument.text` is able to be opened in Collabora and OnlyOffice (Web gets this information from an API endpoint). Then it will get the mimetype for a particular file from the propfind information and can look up, what the user can be offered to open that file with. Currently an Web extension does this by providing the file extensions (eg. `.odt`) that it can open.

CC @labkode @ishank011

kulmann:
Currently we're not using mime types in web, but they are available with the DAV property `{DAV:}getcontenttype`, so at least retrieving it would be easy.

labkode:
cc @elizavetaRa

kulmann:
FTR: we have mime type now on resources: https://github.com/owncloud/web/blob/29341a1c1162e8e1f41fa926b13dde340902df75/packages/web-app-files/src/helpers/resources.js#L37

wkloucek:
[...]

Issue #4735, opened 2021-02-11
Allow to include custom js/html when running ocis web


labkode:
As and admin I will like to configure ocis web to include snippets of html or js code at runtime without having to compile the whole sourcecode.

The goal is to add tracking Js/ feedback forms directly in the UI when running `ocis web --include-js /etc/track.js`.

These scripts can be included in the `<head>` element. 

The same functionality can be used to create announcement/interventions banner for the user of the service.

For example, a JIRA banner to collect feedback:

![Screenshot 2021-02-11 at 12 17 54](https://user-images.githubusercontent.com/5175912/107629988-2e57cc00-6c63-11eb-8b83-36c9d944b9e4.png)

kulmann:
I agree that we need this.

The way we would want to configure this is probably via the config.json. `ocis web` can get its own arguments for configuring this via CLI.

labkode:
@kulmann did I read correctly that when running `ocis server` I can have a custom config  for the web service?

kulmann:
> @kulmann did I read correctly that when running `ocis server` I can have a custom config for the web service?

Yes, you can provide a custom config file for the web service by running it with `WEB_UI_CONFIG="/path/to/config" ocis server`.

labkode:
@kulmann that is nice, do you have at hand what could be changed there? If there are some branding options we can star already playing with it.

kulmann:
> @kulmann that is nice, do you have at hand what could be changed there? If there are some branding options we can start already playing with it.

It is still the same config like for oC Web with oc10. You can find an example here: https://github.com/owncloud/web/blob/master/config/config.json.sample-ocis 

There are plans to refine the way to load extensions. Right now there are three different keys in the json file for loading extensions. We want to have only one.

For branding we will come up with a proper solution for this: https://github.com/owncloud/web/tree/master/packages/web-container/themes which will allow to configure the design tokens as well.

Since we want to maintain compatibility of oC Web with both oC10 and oCIS backends, the config.json is our preferred way for configuration for now.

labkode:
Thanks @kulmann , what is the recommend way to configure web?

I've already defined the following configuration env: 

```
WEB_OIDC_AUTHORITY=https://auth.cern.ch/auth/realms/cern
WEB_OIDC_METADATA_URL=https://auth.cern.ch/auth/realms/cern/.well-known/openid-configuration
WEB_OIDC_CLIENT_ID=xxxx
WEB_UI_CONFIG_SERVER=xxxx
[...]

PR #5752, opened 2021-08-30
Introduce Skeleton-Application, Runtime hooks and demystify runtime boot logic 

fschade:
The main reason for this pr was to provide a skeleton application which showcases how [applications can inject file-handlers](https://github.com/owncloud/web/pull/5712).

The challenge here was that in some cases the application needed to obtain those information from an api which can take time to process. It was not possible to wait for applications and the runtime did proceed even if the application hasn't finished loading what ever it needed (e.g. promises).

To tackle that case it was necessary to introduce such a logic (hooks). The runtime now understand 3 types of hooks.

**initialize**
this hook gets fired once the application is requested and got loaded by the runtime, it is not safe to call other applications from here because it's not guaranteed that all other applications have finished their registration process.

**ready**
all other applications are registered, this is a good place to call other applications or provide needed third party application data

**mounted**
this gets called after the main ui from the runtime got announced to the browser, from here every action will happen AFTER the ui is in place.

we took the opportunity while working on the runtime to refactor and demystify the whole runtime boot logic and rewrite it from the ground up in typescript.

the main reasons for doing this:

- we decided to go with typescript
- it was a long javascript file and hard to understand
- the dependencies where mixed with the application logic
- it will help with the vue 3 migration
- it was hard up to nearly impossible to test

the application boot logic is now class-based and introduces a 'NextApplication' style, it takes care that all current applications still work by doing an automatic conversion between current and next. The next style applications can't be used at the moment, we first need to decide how the public interface could look like.

It also introduces a runtime **API** which only exposes (by DI) allowed functionality to the requested applications.

This can change in the future and for now we only have to decide how the public application runtime API should look like, everything else can be changed later without breaking existing applications.

This is a prerequisite for:
- application don't load in order anymore #2891
- vue 3 migration #5269
- let extensions register more flexible file open actions #5135
- remove loadConfig on extensions #3726 
- capabilities seem to be loaded too late sometimes #3771 (needs api tweak)
- allow to include custom js/html when running ocis web #4735 
- app provider #5712
- allow extensions to inject views in files app #5460
- a new runtime
- doing a backend switch at runtime
- request async information in the boot process of a application 
- boot logic testability

this is wip and the skeleton app needs more attention and functionality


ownclouders:
Results for <strong>oCISSharingPermissions2</strong> https://drone.owncloud.com/owncloud/web/18712/58/1

ownclouders:
Results for <strong>oCISSharingPermissions2</strong> https://drone.owncloud.com/owncloud/web/18713/58/1

packages/web-runtime/src/defaults/vue.js
```diff
@@ -0,0 +1,56 @@
...
+
+Vue.use(Vuex)
+Vue.use(VueRouter)
+Vue.use(VueEvents)
+Vue.use(VueScrollTo)
+Vue.use(MediaSource)
+Vue.use(WebPlugin)
+Vue.use(VueResize)
+Vue.use(VueMeta, {
+  // optional pluginOptions
```
pascalwengerter, line 30: This comment can be removed as far as I'm concerned

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+const requestRouter = (router: VueRouter): VueRouter => {
+  if (isEqual(process.env.NODE_ENV, 'development')) {
+    console.warn('requestRouter // router api is deprecated, use with caution')
+  }
+
+  return router
+}
+
+/**
+ * exposed runtime api, this wraps all available api actions in a closure and provide application
```
pascalwengerter, line 209:
```suggestion
 * exposed runtime api, this wraps all available api actions in a closure and provides application
```

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+ * specific data to the implementations.
+ *
+ * each application get it's own provisioned api!
```
pascalwengerter, line 212:
```suggestion
 * each application gets its own provisioned api!
```

packages/web-runtime/src/container/application/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      err => reject(err)
+    )
+  ).catch(() => {
+    throw new RuntimeError('cannot load application', applicationPath)
+  })
+
+  let application: NextApplication
+
+  try {
+    /** add valuable sniffer to detect nex applications, then implement next application factory */
```
pascalwengerter, line 49:
```suggestion
    /** add valuable sniffer to detect next applications, then implement next application factory */
```

packages/web-runtime/src/container/store.ts
```diff
@@ -0,0 +1,7 @@
+import { NextApplication } from './application'
+
+/**
+ * main propose of this is to keep a reference to all announced applications.
```
pascalwengerter, line 4:
```suggestion
 * main purpose of this is to keep a reference to all announced applications.
```

packages/web-runtime/src/services/api.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class Api {
+  private readonly store: Store<any>
+
+  constructor(store: Store<any>) {
+    this.store = store
+  }
+
+  public registerExtension(app: string, extension: { [key: string]: any }): void {
+    console.log('#############################################################################')
```
pascalwengerter, line 11: Not sure if this is part of the demo (and will get deleted) or if we should be logging here (and then maybe only in `dev` mode?)

pascalwengerter (commented):
Looks pretty awesome! Not sure how we should proceed here (before you're going on holidays next week), maybe it makes sense to pull the runtime/boot-refactoring out of this PR in a separate one already, so the cognitive load of this one shrinks plus I think it also helps with the Vue3 upgrade

ownclouders:
Results for <strong>oC10iPhone1</strong> https://drone.owncloud.com/owncloud/web/18834/45/1
- webUIDeleteFilesFolders/deleteFilesFolders.feature:59

ownclouders:
Results for <strong>oC10Files1</strong> https://drone.owncloud.com/owncloud/web/18837/12/1
- webUIFilesActionMenu/versions.feature:36

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+ * @param routes
+ */
+const announceRoutes = (applicationId: string, router: VueRouter, routes: RouteConfig[]): void => {
+  if (!isArray(routes)) {
+    throw new ApiError("routes can't be blank")
+  }
+
+  const applicationRoutes = routes.map(applicationRoute => {
+    if (!isObject(applicationRoute)) {
+      throw new ApiError("route can't be blank", applicationRoute)
```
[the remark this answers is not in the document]
fschade: in the current version of the "application-manifest" we force to use routes. Int he class based which isn't exported right now the user has to call that api it's wown. So if he calls it without providing routes... it errors, if he skip it all ist fine.

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    const route = clone(applicationRoute)
+    route.name = applicationId === route.name ? route.name : `${applicationId}-${route.name}`
+    route.path = `/${encodeURI(applicationId)}${route.path}`
```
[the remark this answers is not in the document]
fschade: good catch, definitely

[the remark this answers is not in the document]
kulmann: We could only have double slashes if we'd allow leading or trailing slashes in application ids. I'd just forbid that application ids can consist of slahes. ;-) what do you think @LukasHirt ?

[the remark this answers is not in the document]
LukasHirt: How'd we enforce it? Documentation?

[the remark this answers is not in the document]
kulmann: Exactly, that's what I thought. 😁 I'd like to have good documentation for the extension system soonish. A small hint that application ids need to be alphanumeric should be sufficient IMO.

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+ * inject application specific navigation items into runtime
+ *
+ * @param applicationId
+ * @param store
+ * @param navigationItems
+ */
+const announceNavigationItems = (
+  applicationId: string,
+  store: Store<unknown>,
+  navigationItems: ClassicApplicationScript['navItems']
```
kulmann, line 64: Why not 
```suggestion
  navigationItems: ApplicationNavigationItem[]
```
? 🤔

kulmann: Or in other words: you already introduced types, why not use them here? Instead you're creating a dependency to the `ClassicApplicationScript` which should not be needed in the (more) generic runtime API.

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+ * inject application specific translations into runtime
+ *
+ * @param translations
+ * @param appTranslations
+ * @param supportedLanguages
+ */
+const announceTranslations = (
+  supportedLanguages: { [key: string]: string },
+  translations: unknown,
+  appTranslations: ClassicApplicationScript['translations']
```
kulmann, line 101: Same here, why reference the `ClassicApplicationScript`? Would be nicer to introduce a dedicated translation type in `types.ts` and then reuse it here.

packages/web-runtime/src/container/api.ts
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * inject application specific quickActions into runtime
+ *
+ * @param store
+ * @param quickActions
+ */
+const announceQuickActions = (
+  store: Store<unknown>,
+  quickActions: ClassicApplicationScript['quickActions']
```
kulmann, line 122: Same here, why reference the `ClassicApplicationScript`?

kulmann (changes requested):
Found one important thing: configuring an application in the config.json which is not reachable causes the full runtime to stop with a config error page. Extension loading should fail silently (with console error logging, but not by crashing the whole platform). Some other remarks in the comments.

packages/web-runtime/src/container/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  name?: string
+  iconMaterial?: string
+  route?: {
+    name?: string
+    path?: string
+  }
+}
+
+/** application quick action describes a application action that is used in the runtime */
+export interface ApplicationQuickActions {
```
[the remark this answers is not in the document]
kulmann: Hm ok, for compatibility reasons. I'll introduce a singular `ApplicationQuickAction` interface and use that for the values in the `ApplicationQuickActions` interface. Good enough for now.

kulmann (approved):
Took the liberty of fixing some minor things in this PR. Failing hard with the whole system on one application not being able to load is a no-go - but since the next release is 2 weeks ahead we'll have enough time to fix that in a different PR. Fine to merge this now! Thanks for all the urgently needed cleanup! ❤️

ownclouders:
Results for <strong>oCISFiles1</strong> https://drone.owncloud.com/owncloud/web/19075/52/1
- webUIFiles/download.feature:11

ownclouders:
Results for <strong>oC10SharingInternalUsersSharingIndicator</strong> https://drone.owncloud.com/owncloud/web/19079/28/1
- webUISharingInternalUsersSharingIndicator/shareWithUsers.feature:154
- webUISharingInternalUsersToRootSharingIndicator/shareWithUsers.feature:98

ownclouders:
Results for <strong>oC10Files1</strong> https://drone.owncloud.com/owncloud/web/19079/12/1

PR #5759, opened 2021-08-31
Remove unused deps, bump markdown editor version

pascalwengerter:
## Description
Found some unused dependencies when preparing for shared eslint config.
Also, I know there's no priority here but there's a stale dependabot PR about updating the `marked` Markdown editor - I've skimmed the changelog, seemed to be about dropping old Node and IE version and internal changes, let's see what CI says


kulmann (approved):
Nice cleanup 👍

PR #5769, opened 2021-09-02
Update runtime app & usermenu styles

pascalwengerter:
## Description
@kulmann this fixes the user menu & loading spinner styles issues that will "come" with ODS 10.0.0 and are due to the removed orphaned stylesheets


kulmann (approved):
LGTM 👍

Issue #5237, opened 2021-06-10
list of views for unit testing


kiranparajuli589:
#### packages/web-app-files/src/views/:
- [x] Favorites.vue  #5345
- [x] FilesDrop.vue #5680 
- [x] LocationPicker.vue #5697 
- [x] Personal.vue (there are some duplication for setup inside the tests, we can refactor this) (#5588)
- [x] PrivateLink.vue (TODO: #5750) (#5755)
- [ ] PublicFiles.vue wip @kiranparajuli589 #5722 #6181
- [x] PublicLink.vue #5749 #5756 
- [x] SharedViaLink.vue (TODO: #5751) (#5778)
- [x] SharedWithMe.vue #5814 (https://github.com/owncloud/web/pull/5814/commits/90e37cc145655ef362733d3177c1fddbbe8037e7)
- [x] SharedWithOthers.vue wip @kiranparajuli589 #5722 #6175
- [x] Trashbin.vue wip @kiranparajuli589 #5722 #6174

### Motivation
- https://github.com/owncloud/web/issues/5217

individual-it:
blocked waiting for reviews of @kulmann @pascalwengerter

phil-davis:
@kiranparajuli589 is this done, or can progress, or?

kiranparajuli589:
this is done and can be closed.

PR #5778, opened 2021-09-06
[unit-tests-only] Add unit tests for SharedViaLink view

saw-jan:
## Description
Added unit tests for `SharedViaLink` view

## Related Issue
- part of #5237

## Motivation and Context
increase unit test coverage

## How Has This Been Tested?
- test environment: local

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "eslint-plugin-promise": "^5.1.0",
     "eslint-plugin-vue": "^7.13.0",
     "eslint-plugin-vuejs-accessibility": "^0.7.1",
+    "flush-promises": "^1.0.2",
     "focus-trap": "^6.4.0",
     "focus-trap-vue": "^1.1.1",
     "jest": "^26.6.3",
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "eslint-plugin-promise": "^5.1.0",
     "eslint-plugin-vue": "^7.13.0",
     "eslint-plugin-vuejs-accessibility": "^0.7.1",
-    "flush-promises": "^1.0.2",
     "focus-trap": "^6.4.0",
     "focus-trap-vue": "^1.1.1",
     "jest": "^26.6.3",
--- a/packages/web-app-files/tests/unit/views/SharedViaLink.spec.js
+++ b/packages/web-app-files/tests/unit/views/SharedViaLink.spec.js
@@ -1,29 +1,191 @@
+import { shallowMount, mount } from '@vue/test-utils'
+import flushPromises from 'flush-promises'
+import { getStore, localVue, createFile } from './views.setup.js'
+import MixinFilesListPagination from '@files/src/mixins/filesListPagination'
+import FileActions from '@files/src/mixins/fileActions'
+import SharedViaLink from '@files/src/views/SharedViaLink.vue'
+
+const $route = {
+  params: {
+    fileId: '2147491323'
+  },
+  meta: {
+    title: 'Resolving private link'
+  }
+}
+
+const activeFilesCurrentPage = [createFile({ id: 1234, type: 'file' })]
+
+localVue.prototype.$client.requests = {
+  ocs: jest.fn(() =>
+    Promise.resolve({
+      status: 200,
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
     "eslint-plugin-promise": "^5.1.0",
     "eslint-plugin-vue": "^7.13.0",
     "eslint-plugin-vuejs-accessibility": "^0.7.1",
+    "flush-promises": "^1.0.2",
```
saw-jan, line 80: this package helps to flush all promises which means that we don't need to use `wrapper.vm.$nextTick()` twice or more in order to wait for all promises to get resolved.

should we use it?

kiranparajuli589: @dpakach does this helps for the test you're writing for autocomple inputs taking some time to trigger. If such problems with static waits or of course multiple nextTick can be resolved...imo this might be useful

packages/web-app-files/tests/unit/views/SharedViaLink.spec.js
```diff
@@ -1,29 +1,201 @@
...
+  listLoader: 'list-loader-stub',
+  noContentMessage: '#files-shared-via-link-empty',
+  ocTableFiles: '#files-shared-via-link-table',
+  contextActions: 'context-actions-stub',
+  listInfo: 'list-info-stub'
+}
+
 describe('SharedViaLink view', () => {
+  jest
+    .spyOn(SharedViaLink.mixins[3].methods, '$_filesListPagination_updateCurrentPage')
```
kiranparajuli589, line 51: if the order of the mixin get changed, this can fail. what about directly mocking from the mixin like:
```suggestion
    .spyOn(someMixin.methods, '$_filesListPagination_updateCurrentPage')
```

package.json
dpakach: sadly doesn't works in my case

kiranparajuli589 (dismissed):
lgtm 👍🏼

packages/web-app-files/tests/unit/views/SharedViaLink.spec.js
```diff
@@ -1,29 +1,194 @@
...
   })
 
   describe('when the view is still loading', () => {
-    it.todo('should show list-loader component')
+    it('should show list-loader component', () => {
+      const wrapper = getShallowWrapper()
+      const listLoader = wrapper.find(selectors.listLoader)
+
+      expect(listLoader.exists()).toBeTruthy()
+      expect(listLoader).toMatchSnapshot()
```
LukasHirt, line 90: It feels a bit odd testing snapshot of a child component in the parent instead of tests of that component...

packages/web-app-files/tests/unit/views/SharedViaLink.spec.js
```diff
@@ -1,29 +1,194 @@
...
+        stubs['no-content-message'] = false
+        wrapper = getWrapper()
+        await flushPromises()
+      })
+
+      it('should show no-content-message component', () => {
+        const noContentMessage = wrapper.find(selectors.noContentMessage)
+
+        expect(noContentMessage.exists()).toBeTruthy()
+        expect(noContentMessage).toMatchSnapshot()
```
LukasHirt, line 114: It feels a bit odd testing snapshot of a child component in the parent instead of tests of that component...

individual-it:
blocked waiting for review @kulmann @pascalwengerter

package.json
pascalwengerter: This PR needs a rebase and I'm not sure why there was an attempt to use promises so heavily here that it needs an additional library to work around it. Trying a different approach, I'll ping you once I know if it works;)

saw-jan: @pascalwengerter, @kiranparajuli589 has made an update to this PR. Please review.

pascalwengerter (approved):
LGTM, I think we can take a look into adding better tests for the Files App Views but this is a good start 🥅

PR #5823, opened 2021-09-20
fix: loading spinner not rendered correctly in media-viewer app

rpocklin:
Simple one, this fixes the bad spinner below:

image

Might be worth documenting how to center this in the **owncloud-design-system**, as most of the time you want a spinner centered if it represents an area of content loading.

You used to be able to add `uk-position-center` to the same `<oc-spinner>` element, now you must add it to the parent element that contains it.


packages/web-app-media-viewer/src/App.vue
```diff
@@ -27,12 +27,9 @@
         </div>
       </transition>
     </div>
-    <oc-spinner
-      v-if="loading"
-      :aria-label="$gettext('Loading media')"
-      class="uk-position-center"
-      size="xlarge"
-    />
+    <div class="uk-position-center">
```
kulmann, line 30:
```suggestion
    <div v-if="loading" class="uk-position-center">
```

packages/web-app-media-viewer/src/App.vue
```diff
@@ -27,12 +27,9 @@
...
+      <oc-spinner v-if="loading" :aria-label="$gettext('Loading media')" size="xlarge" />
```
kulmann, line 31:
```suggestion
      <oc-spinner :aria-label="$gettext('Loading media')" size="xlarge" />
```

kulmann (changes requested):
Please move the `v-if` to the (new) parent element. :-)

packages/web-app-media-viewer/src/App.vue
rpocklin: ah yep, good point =)

kulmann: Hehe, now you removed it from the spinner but didn't add it to the div 😁

ownclouders:
Results for <strong>oCISSharingPermissions2</strong> https://drone.owncloud.com/owncloud/web/19108/58/1

packages/web-app-media-viewer/src/App.vue
rpocklin: I accepted the propsed code suggestion (from github), sorry let me fix it properly then.

kulmann: Afaik you can only make suggestions on single lines. So I had two different suggestions, one for adding it on the div (which is not committed yet) and one for removing it from spinner (which is committed). Sorry for the confusion

ownclouders:
Results for <strong>oCISSharingPermissions1</strong> https://drone.owncloud.com/owncloud/web/19112/57/1

packages/web-app-media-viewer/src/App.vue
rpocklin: Hmm not sure about that - I have updated the original commit let me know if that's more to your liking.

kulmann (approved):
Awesome, thank you! 😇

ownclouders:
Results for <strong>oC10SharingInternalGroupsSharingIndicator</strong> https://drone.owncloud.com/owncloud/web/19116/24/1
- webUISharingInternalGroupsSharingIndicator/shareWithGroups.feature:114

PR #5837, opened 2021-09-27
fix: change sticky header top offset to reflect actual height on page of #files-app-bar not relative height

rpocklin:
## Description
If the implementation of **web** has a notification bar or anything else consuming space above #files-app-bar, the sticky header will be off by the exact height of anything above it.

**Before**
image

**After**
image

There is one other place where `bottom` is used in place of `height` but I have not tested this so may propose this in another PR.
See https://github.com/owncloud/web/blob/a330d27b7706118d974d219d6d30776e730907ed/packages/web-app-files/src/mixins/filesListScrolling.js#L5

### DIFF
--- a/packages/web-app-files/src/views/LocationPicker.vue
+++ b/packages/web-app-files/src/views/LocationPicker.vue
@@ -293,7 +293,7 @@ export default {
     adjustTableHeaderPosition() {
       this.$nextTick(() => {
         const header = document.querySelector('#files-app-bar')
-        this.headerPosition = header.getBoundingClientRect().bottom
+        this.headerPosition = header.getBoundingClientRect().height
       })
     },
 


ownclouders:
Results for <strong>oC10SharingExternalRoot</strong> https://drone.owncloud.com/owncloud/web/19237/40/1
- webUISharingExternalToRoot/federationSharing.feature:356

ownclouders:
Results for <strong>oC10SharingExternal</strong> https://drone.owncloud.com/owncloud/web/19237/39/1
- webUISharingExternal/federationSharing.feature:363

kulmann (approved):
Thanks for taking care!

I think that the other occurrence that you pointed out also needs to use `height` instead of `bottom`.

PR #5848, opened 2021-09-28
Fix theme paths in docs example/drone configs

pascalwengerter:
## Description
When introducing theming, we changed the way themes are getting loaded to either URL- or path-based. I've recently come across a couple of outdated examples that still use a theme-name and hence fail & fall back to the `ownCloud`-default-theme. Fixing them gives dev-/admin-users a better idea of how to actually use this config option and reduces errors in the JS console


ownclouders:
Results for <strong>oC10SharingInternalGroupsSharingIndicator</strong> https://drone.owncloud.com/owncloud/web/19254/24/1

kulmann (approved):

Issue #5085, opened 2021-05-07
Responses are rendered in different places


labkode:
If I send a request to list a large directory /`A/B/C`.
In the mean time I change path to `/Z`.
Request to `/A/B/C` finished and shows the content of `C` inside `Z`.

The expected behaviour is that the response should not be rendered as this is not the original path it was requested.

pascalwengerter:
@labkode could you check if this is still the case? Should have been fixed by our recent changes

pascalwengerter:
Closed as fixed (due to various performance improvements) unless @labkode can reproduce

labkode:
@kulmann @dragotin 
This is a major blocker for our roll-out and needs to be fixed in short-time (days).
By manually testing it you end up with: browsing into a directory and is empty, browsing into a directory that contains contents from other places, clicking on shared with me and listing your home files rather than your shares, etc ...

kulmann:
> @kulmann @dragotin
> This is a major blocker for our roll-out and needs to be fixed in short-time (days).
> By manually testing it you end up with: browsing into a directory and is empty, browsing into a directory that contains contents from other places, clicking on shared with me and listing your home files rather than your shares, etc ...

Thank you for your comment, I was finally able to reproduce it locally. We only solved edge cases within one view so far (so if you navigate into sub- or parent-folders in the `All files` list). That all works fine now as far as I know, because those requests are resolved sequentially. BUT: across views requests for loading resources can indeed overlap. So a slow loading "shared with me" will render into a fast loading "All files" view and vice versa. "render" is not 100% correct because there are also some error cases, where you'll see "resource not found". With that I'll shift priority back onto https://github.com/owncloud/web/pull/5892 immediately (now).

fschade:
@labkode can you please test https://github.com/owncloud/web/pull/5917. Should fix it, i hope i spotted all places.

fschade:
thanks @diocas

elizavetaRa:
We can reproduce this issue again 
![Files shared with me - CERNBox](https://user-images.githubusercontent.com/7430156/187469460-ad44a684-2148-4529-b51a-9275ebe2520f.gif)

kulmann:
Please provide steps to reproduce. If I do what you show in the screen capture, even with throttled connection and large folders, I can't reproduce it.

dragotin:
Seems fixed, if not @elizavetaRa please reopen.

PR #5892, opened 2021-10-10
WIP resource management in store

kulmann:
### problem statement:
our `files` array in the vuex store of the files app is only capable of representing one state.
multiple PROPFINDs for different folders, fired at similar times, will override each other in that state.
the longest running propfind wins. Cancelling the requests is not solving all edge cases, as the
PROPFINDs might finish at similar times so that we're in postprocessing of the data and request cancellation is not possible anymore (because request already done).

### idea for a solution:
loading resources in the different views should be converted into a two step process, each:
1) set the activeView and activePath immediately after each navigation (`SET_ACTIVE_VIEW`, `SET_ACTIVE_PATH`)
2) do the PROPFIND, convert the response into resources, and set them in a dedicated `folder` store module.
That store module needs to get registered dynamically. The PROPFIND and conversion might be long running.
It's really important that the activePath is set as early as possible!

the views then render the files of the store module that matches the current `activePath`.
as a bonus, requests could be cancelled and data loading method invocations stopped, so that we avoid unnecessary store module creation / data transformation / loading of additional data (e.g. if we know that the response of a PROPFIND is not needed anymore, because the user navigated away from the current location, we could skip loading indicators and previews entirely).

reasons for dynamically registered store modules: we need the reactivity of the individual files.
Which comes built in with the store modules. Another approach would be to build a cache of recent
propfind results (= response converted into a file listing). but that's not reactive by default.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/5085

## Motivation and Context
Reduce glitchiness of the different resource listing views and their respective navigation.

## How Has This Been Tested?
- not yet (WIP)

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] think about when to unregister the folder store modules. they might be big so removal should be happening as soon as not needed anymore.
- [ ] further think about different views: propfind, fetch incoming shares, fetch outgoing shares, trashbin
- [ ] the different `loadResources` methods from the views should live in a service, which then also takes care of creating the respective store module.
- [ ] changelog
- [ ] tests


packages/web-app-files/src/views/Personal.vue
```diff
@@ -1,6 +1,10 @@
 <template>
   <div>
+    <<<<<<< HEAD
     <list-loader v-if="loadResourcesTask.isRunning" />
+    =======
+    <list-loader v-if="isLoading" />
+    >>>>>>> c70e2ee84 (WIP resource management in store)
```
dschmidt, line 7: :trollface:

kulmann:
Uff, yes, I guess this can be closed by now. The resource loading has improved a lot in the meantime. Adapting the remaining parts of this PR doesn't bring too much benefit at the moment.

Issue #4595, opened 2021-01-08
can we clean up the URL path to avoid urlencoded /, eg


butonic:
`https://phoenix.owncloud.com/index.html#files/list/eos%2Fusers%2Fhugo`

It would allow cnp the path and in the future make the url more similar when using other apps, eg for collabora: `https://phoenix.owncloud.com/index.html#collabora/eos%2Fusers%2Fhugo`?

butonic:
might be related to spaces ... but there really is no reason to have `%2F` after the `#` in the url

LukasHirt:
This requires first a update to vue-router@4 which requires vue@3. @fschade thought of a dirty hack which could do this already now but we decided against it and delay this until the update of vue and vue-router.

labkode:
@LukasHirt any idea of the proposed update of the vue components?

LukasHirt:
> any idea of the proposed update of the vue components?

No progress so far. The update is not part of the current sprint. We were focusing on our accessibility tasks at the moment because of the WCAG certification. Is this ticket urgent for you? Last time I heard it wasn't, so we weren't putting too much priority to it.

labkode:
Hi @LukasHirt, it is as we plan to expose the new system, where user would be able to copy/paste the url to access the resources and if the  URL changes this will break behaviour.

LukasHirt:
The current sprint ends in two weeks - should we escalate this and tackle this already @pmaier1 or would it be possible to wait until the next one @labkode ?

labkode:
@LukasHirt it can wait until next one, thanks for following it up

kulmann:
We were thinking about an entirely new URL format recently. There was some concepting done here https://github.com/owncloud/web/issues/5313 but it stalled lately. Needs to be on our agenda again. cc @tbsbdr

pascalwengerter:
At least the %2F encoding should be fixed via #5715 (patched until we switch to Vue3) and just landed in master, closing this issue for now (yet we'll perhaps need some more routing-related ones in the future)

PR #5894, opened 2021-10-10
refactor(vue router patch): apply router patch globally

fschade:
## Description
This patch was already introduced earlier for the files application only.
In the meantime we found out that this is also needed on different places across the ecosystem.

We've refactored the way how the patch get applied to the routes, to do so set the meta.patchCleanPath to true on the individual route.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/4595#issuecomment-938587035

## Motivation and Context
apply router patch not in files app only

## How Has This Been Tested?
- local installation

## Types of changes

## Checklist:

- [X] Code changes

## Open tasks:
- [ ] apply to other known routes

### DIFF
--- a/packages/web-app-media-viewer/src/App.vue
+++ b/packages/web-app-media-viewer/src/App.vue
@@ -211,14 +211,7 @@ export default {
 
     // keep a local history for this component
     window.addEventListener('popstate', this.handleLocalHistoryEvent)
-    const filePath = (filePath => {
-      filePath = filePath.split('/')
-      filePath = filePath.filter(Boolean)
-      filePath = filePath.join('/')
-      filePath = `/${filePath}`
-
-      return filePath
-    })(this.$route.params.filePath)
+    const filePath = `/${this.$route.params.filePath.split('/').filter(Boolean).join('/')}`
     await this.$_loader_loadItems(filePath)
     this.setCurrentFile(filePath)
   },
--- a/packages/web-runtime/src/router/index.js
+++ b/packages/web-runtime/src/router/index.js
@@ -69,7 +69,7 @@ const router = new Router({
   ]
 })
 
-router.beforeEach(function(to, from, next) {
+router.beforeEach(function (to, from, next) {
   const store = Vue.$store
   const isAuthenticated = store.getters.isAuthenticated
   let authRequired = true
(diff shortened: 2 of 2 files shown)


ownclouders:
Results for <strong>oC10Files1</strong> https://drone.owncloud.com/owncloud/web/19521/12/1
- webUIFilesActionMenu/versions.feature:36

ownclouders:
Results for <strong>oC10Files1</strong> https://drone.owncloud.com/owncloud/web/19554/12/1

packages/web-app-media-viewer/src/App.vue
```diff
@@ -201,8 +201,14 @@ export default {
 
     // keep a local history for this component
     window.addEventListener('popstate', this.handleLocalHistoryEvent)
+    const filePath = (filePath => {
```
kulmann, line 204: as discussed, please use chaining to make this a one-liner (which will need line breaks of course 😆 )

packages/web-runtime/src/router/index.js
```diff
@@ -93,4 +94,35 @@ router.beforeEach(function(to, from, next) {
...
+      ...bindMatch,
+      path: cleanPath(bindMatch.path),
+      fullPath: cleanPath(bindMatch.fullPath)
+    }
+  }
+
+  return router
+}
+
+export default patchRouter(router)
```
kulmann, line 128: Please add a comment that the result of this call is a patched, otherwise fully functional router. Keep the name `patchRouter` like it is. Thought about it again, I like it better than something like `patchedRouter`. Name is good.

kulmann (changes requested):

changelog/unreleased/bugfix-router-clean-path
```diff
@@ -0,0 +1,9 @@
+Bugfix: Clean router path handling
+
+This patch was already introduced earlier for the files application only.
+In the meantime we found out that this is also needed on different places across the ecosystem.
+
+We've refactored the way how the patch get applied to the routes, to do so set the meta.patchCleanPath to true on the individual route.
```
pascalwengerter, line 6:
```suggestion
We've refactored the way how the patch gets applied to the routes: It is now possible to set an individual route's `meta.patchCleanPath` to true.
```

PR #5909, opened 2021-10-15
Run CI against latest oC10 release

### LINKED ISSUE #5908: Switch from commit-id to (dockerized) last official release of OC10 in CI
### Problem
We're currently testing the web frontend against an unreleased version of OC10 since we're linking (and updating) the core's commit ID. This potentially leads us to the point of releasing a web version that passes CI but doesn't work with existing client applications since they're running (stable, official) releases.

### Decision
We have decided to switch from the commit-ID based approach to use the dockerized last official release of OC10 in CI. Once a new release of OC10 is within close reach (e.g. there's a code freeze for core and/or release candidates published) and we're getting close to a new web release that will get included in the OC10 release we can switch the docker tag over to the next version of core.

## Checklist:
- [x] Pull and set up OC10 from a docker tag in the OC10 acceptance test pipelines. A reference implementation/configuration can be found [in the local acceptance test setup](https://github.com/owncloud/web/tree/master/dev/docker) - the CI uses the existing `owncloudci/core` docker image to install the required ownCloud server from its release tarball.
- [x] Bring back all the expected failures that have been removed since the official release of OC10.8 and let CI run and see which ones pass unexpectedly - there was no significant change, see #5909 for the minor changes to acceptance test code and expected failures
- [x] Remove all the obsolete configuration of cloning/installing/setting up OC10 from a core commit-ID since this functionality isn't needed anymore

phil-davis:
## Description
- run the pipelines that test with oC10 against oC10 "latest" release (which is currently 10.8.0)
- remove the `CORE_COMMITID` logic - we do not need that any more
- adjust test scenario `webUIFavorites/favoritesFile.feature:74` so that it does not depend on the contents of https://github.com/owncloud/core/tree/master/core/skeleton or whatever "default skeleton" is delivered in a release tarball or daily master QA tarball. That allows the test scenario to pass in all the different oC10 backend "versions".
- skip `webUIFilesActionMenu/versions.feature:36` - that is a known flaky test, and we need to skip that to get reliable CI.
- adjust the accept and decline share code in ` tests/acceptance/helpers/sharingHelper.js` This is used by `Given` steps that set up test scenarios. The various server backends (oCIS, oC10.8 and upcoming oC10.9) provide slightly different things in the share-path of pending and accepted shares. These are `Given` steps and we want them to "just work". The code has been adjusted to just match the name of the resource at the end of the path. So if the path is presented as `/parent/child` or `/Shares/child` or just `/child` we can match the share and then take the accept or decline action. Note: the actual test of the web interface "button-clicking" accept/decline are not impacted by this.

## Related Issue
- Fixes #5908 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


individual-it (approved):

phil-davis:
@kulmann "Looks to me like it would be sufficient to set latest as the default there." - yes, true. I removed the duplication.

PR #5910, opened 2021-10-15
[full-ci] feat(eventbus): resource loading happens multiple times

### LINKED ISSUE #5875: Redirect to homeFolder only
When I open ocis web (I go to `/`), I get redirected first to `/#/files/list/all` and only then I'm redirected to `homeFolder` (in our case, `/#/files/list/all/eos/user/<letter>/<username>`). I should only be redirected to the latter.

We suspect that with our latest update, getting `/` might have become slightly slower. And this became quite noticeable due to #5085, which is causing the view of my home dir to be replaced by the view of the root dir.

fschade:
## Description
the eventBus is able to subscribe many times to the same topic, this
could be problematic in cases where the subscription is build on
component mount hook.

This leads to multiple unwanted subscriptions on
the same topic, which results in unnecessary requests and state
changes.

This is fixed by allowing the eventBus subscriptions to
unsubscribe on a topic and token.

## Related Issue

- Fixes #5875

## Motivation and Context
get rid of unnecessary and duplicated subscriptions

## How Has This Been Tested?

- local installation
- jest

## Types of changes

- [X] Tests

## Checklist:

- [X] Code changes
- [X] Unit tests added

## Open tasks:

### DIFF
--- a/packages/web-app-files/tests/integration/specs/appBar.spec.js
+++ b/packages/web-app-files/tests/integration/specs/appBar.spec.js
@@ -30,7 +30,7 @@ describe('AppBar contains set of actions and informations', () => {
     beforeEach(() => {
       config.mocks.publicPage = () => false
 
-      bus.emit = jest.fn((path) => path)
+      bus.publish = jest.fn((path) => path)
     })
 
     test('user can refresh files list by clicking on last breadcrumb item', async () => {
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/event/bus.ts
```diff
@@ -1,27 +1,43 @@
...
+    if (!this.topics.has(topic)) {
+      return
+    }
 
-    for (let i = 0; i < listeners.length; i++) {
-      const listener = listeners[i]
-      if (!listener) continue
-      listener(evt)
+    if (!token) {
+      this.topics.delete(topic)
```
kulmann, line 34: Hm... this means that one subscriber could unsubscribe all subscribers from a topic. I'd vote for having a different function for killing a topic entirely. Something like `unsubscribeAll`. Is doesn't seem to be used anyway.

kulmann (approved):
LGTM 📞

PR #5917, opened 2021-10-17
fix(view rendering): cancel overlapping resource requests

### LINKED ISSUE #5085: Responses are rendered in different places

fschade:
## Description
While testing the files app with heavy load we've found some cases where it could happen that the ui breaks or displays wrong list of files, this only happened when navigation fast between the different views.

To dig a bit deeper we've used the mounted and created hooks to load current views resources for files app, which can lead to overlapping rendering when view `A` still prepares the resource data and the user quickly switches to view `B`. If then `B` is rendered before `A` is finished vue renders the view `A` into `B`.

This is an overall problem and is present in many projects, where state, data and timing overlaps. This potentially is a memory leak also, in cases where `this` is used and the instance isn't around anymore.

This PR fixes those cases for the files app by superseding the usual vue behavior and use iterable generators from vue-concurrency  and cancel the preparation step (request loading, state loading, view rendering) when not needed anymore.

## Related Issue

- Fixes #5085
- Related #5875 

## Motivation and Context
get a fast and concurrency safe ui

## How Has This Been Tested?

- test suite
- hammering the ui locally

## Types of changes

- [X] Technical debt
- [ ] Tests

## Checklist:

- [X] Code changes
- [X] Unit tests added

## Open tasks:

- [ ] none expect decision and testing ;)

### DIFF
--- a/tests/integration/config/jest.init.js
+++ b/tests/integration/config/jest.init.js
@@ -3,6 +3,7 @@ import { config } from '@vue/test-utils'
 import fetchMock from 'jest-fetch-mock'
 import ODS from 'owncloud-design-system'
 import GetTextPlugin from 'vue-gettext'
+import VueCompositionAPI from '@vue/composition-api'
 
 // eslint-disable-next-line jest/no-mocks-import
 import sdkMock from '../../../__mocks__/sdk'
@@ -15,6 +16,7 @@ window.IntersectionObserver = jest.fn(() => ({
 }))
 
 Vue.use(ODS)
+Vue.use(VueCompositionAPI)
 
 Vue.use(GetTextPlugin, {
   availableLanguages: ['en'],
--- a/yarn.lock
+++ b/yarn.lock
@@ -2934,7 +2934,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@vue/test-utils@npm:^1.1.0":
+"@vue/test-utils@npm:^1.1.0, @vue/test-utils@npm:^1.1.2":
   version: 1.2.1
   resolution: "@vue/test-utils@npm:1.2.1"
   dependencies:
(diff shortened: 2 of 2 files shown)


changelog/unreleased/bugfix-overlapping-view-requests
```diff
@@ -0,0 +1,10 @@
+Bugfix: Fix overlapping requests in files app
+
+In some cases the files app tended to display the wrong resources when navigation quickly through the views. This happened because the resource provisioning step wasn't canceld.
```
kulmann, line 3:
```suggestion
In some cases the files app tended to display the wrong resources when navigating quickly through the views. This happened because the resource provisioning step wasn't canceled.
```

changelog/unreleased/bugfix-overlapping-view-requests
```diff
@@ -0,0 +1,10 @@
...
+This is now fixed by using vue-concurrency which on a high level wraps iterable generators which are cancelable.
```
kulmann, line 4:
```suggestion
This is now fixed by using vue-concurrency which on a high level wraps iterable generators which are cancelable. We're using it to wrap the resource loading and cancel it as soon as the resource set is not needed anymore.
```

packages/web-app-files/src/views/Favorites.vue
```diff
@@ -76,9 +77,19 @@ export default {
     MixinFilesListFilter
   ],
 
-  data: () => ({
-    loading: true
-  }),
+  setup() {
+    const loadResourcesTask = useTask(function* (signal, ref, sameRoute, path = null) {
+      yield new Promise((resolve) => setTimeout(resolve, 2000))
```
kulmann, line 82: dev leftover

packages/web-app-files/src/views/Favorites.vue
```diff
@@ -76,9 +77,19 @@ export default {
     MixinFilesListFilter
   ],
 
-  data: () => ({
-    loading: true
-  }),
+  setup() {
+    const loadResourcesTask = useTask(function* (signal, ref, sameRoute, path = null) {
```
kulmann, line 81: Params `sameRoute` and `path` are not needed here

kulmann (commented):
😍😍😍

kulmann:
Linter is not happy. ;-) Tested it locally and wasn't able to break it. 🚀

kulmann (approved):
❤️

PR #5986, opened 2021-11-05
Performance improvements

diocas:
I'm not very knowledgeable on these things, so please let me know if I'm messing this up...

We had massive performance issues with CERNBox. One of the first things we did was to serve the web behind nginx with caching headers for all these files (this needs to be implemented in `ocis-web`, which for now just has a "TODO" in the code). Since the extensions already have the hash in their name, this works perfectly. I don't know if we can do the same for the css file? For now I just appended the timestamp of the build to ensure the update is properly handled by the browser.

(The second thing we did was to finally separate all our extensions from the fork... They were there for convenience, but the dependencies created a huuuge chunks file. This was our fault, so nothing to fix here :) )

One of the issues our users saw was the browser timing out when loading the chunks file, so we increased the timeout of requirejs. This doesn't seem necessary anymore, as the file is now smaller, but I think it can still be kept.

On the subject of requirejs, we've used a CDN instead of the one copied locally. This, besides potentially being cached already, is a minimized version. If the CDN is not available (running locally without connectivity?), We still fallback to the local one.

The rest of the changes were essentially to parallelize the calls to the backend. Since you await everything immediately, even things without dependencies between themselves, this means that the requests will be sequential. This speeds ups the first loading. So: load early, await when really needed.
When loading the apps, we distinguish between internal and external: we never wait for the externals to be loaded and continue the run, as they will be registered eventually. Please confirm wether this is acceptable or not.

~~When opening for the first time, instead of waiting for everything to load before redirecting to the idp, we just abort the loading and redirect. We've seen errors in the console because there were things trying to use methods that were not available yet... So this is more an idea of what we think needs to be done, but the implementation details still need to be tuned.~~ (removed this after the latest upstream changes messed with it)

Lastly, when PROPFINDING, we decided to show the result as soon as the response arrives, and load the shares information after. But I guess that this info will come with the PROPFIND soon so this fix won't be necessary.

This was just a first attempt but we're quite happy with the performance now. But I guess there are still many improvements that can be done and some issues that were reported and need to be fixed (multiple repeated requests, for example). On this last topic: https://github.com/owncloud/owncloud-sdk/pull/933

### DIFF
--- a/packages/web-app-files/src/views/Personal.vue
+++ b/packages/web-app-files/src/views/Personal.vue
@@ -165,9 +165,12 @@ export default {
         })
 
         // Load quota
-        const user = yield ref.$client.users.getUser(ref.user.id)
-
-        ref.SET_QUOTA(user.quota)
+        const promiseUser = ref.$client.users.getUser(ref.user.id)
+        // The semicolon is important to separate from the previous statement
+        ;(async () => {
+          const user = await promiseUser
+          ref.SET_QUOTA(user.quota)
+        })()
       } catch (error) {
         ref.SET_CURRENT_FOLDER(null)
         console.error(error)
--- a/packages/web-container/index.html.ejs
+++ b/packages/web-container/index.html.ejs
@@ -7,9 +7,14 @@
   <title><%= data.title %></title>
   <link rel="manifest" href="manifest.json" crossorigin="use-credentials">
   <% Object.keys(data.bundle.css).forEach((s) => { %>
-    <link href="<%- data.bundle.css[s] %>" rel="stylesheet">
+    <link href="<%- data.bundle.css[s] %>?<%= data.compilationTimestamp %>" rel="stylesheet">
   <% }); %>
-  <script src="js/require.js"></script>
+  <% if (data.config.cdn) { %>
+  <script src="//cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>
+  <script>if (typeof requirejs === 'undefined') { document.write('<script src="js/require.js?<%= data.compilationTimestamp %>">\x3C/script>') }</script>
+  <% } else { %>
+  <script src="js/require.js?<%= data.compilationTimestamp %>"></script>
+  <% } %>
   <style>
     .splash-banner {
       display: flex;
--- a/packages/web-container/index.html.ejs
+++ b/packages/web-container/index.html.ejs
@@ -68,7 +68,8 @@
     } else {
       requirejs.config({
         baseUrl: <%- JSON.stringify(data.roots.js) %>,
-        paths: <%- JSON.stringify(data.bundle.js) %>
+        paths: <%- JSON.stringify(data.bundle.js) %>,
+        ...<%- JSON.stringify(data.config.requirejs) %>
       })
 
       requirejs(['web-runtime'], function (runtime) {
--- a/packages/web-runtime/src/container/bootstrap.ts
+++ b/packages/web-runtime/src/container/bootstrap.ts
@@ -71,16 +71,20 @@ export const announceStore = async ({
  *
  * @param runtimeConfiguration
  */
-export const announceClient = async (runtimeConfiguration: RuntimeConfiguration): Promise<void> => {
+export const announceClient = (runtimeConfiguration: RuntimeConfiguration): Promise<void> => {
   const { openIdConnect = {} } = runtimeConfiguration
 
   if (!openIdConnect.dynamic) {
     return
   }
 
-  const { client_id: clientId, client_secret: clientSecret } = await registerClient(openIdConnect)
-  openIdConnect.client_id = clientId
-  openIdConnect.client_secret = clientSecret
+  const promiseRegistration = registerClient(openIdConnect)
+
+  promiseRegistration.then(({ client_id: clientId, client_secret: clientSecret }) => {
+    openIdConnect.client_id = clientId
+    openIdConnect.client_secret = clientSecret
+  })
+  return promiseRegistration
 }
 
 /**
@@ -95,7 +99,7 @@ export const announceClient = async (runtimeConfiguration: RuntimeConfiguration)
  * @param translations
  * @param supportedLanguages
  */
-export const announceApplications = async ({
+export const announceApplications = ({
   runtimeConfiguration,
   store,
   router,
@@ -107,7 +111,7 @@ export const announceApplications = async ({
   router: VueRouter
   translations: unknown
   supportedLanguages: { [key: string]: string }
-}): Promise<void> => {
+}): Promise<any> => {
   const { apps: internalApplications = [], external_apps: externalApplications = [] } =
     runtimeConfiguration
 
@@ -116,31 +120,31 @@ export const announceApplications = async ({
     ...externalApplications.map((application) => application.path)
   ].filter(Boolean)
 
-  const applicationResults = await Promise.allSettled(
-    applicationPaths.map((applicationPath) =>
-      buildApplication({
-        applicationPath,
-        store,
-        supportedLanguages,
-        router,
-        translations
+  const registerApp = (applicationPath) => {
+    const promiseBuild = buildApplication({
+      applicationPath,
+      store,
+      supportedLanguages,
+      router,
+      translations
+    })
+
+    promiseBuild
+      .then((application) => {
+        application.initialize().then(() => {
+          application.ready()
+        })
       })
-    )
-  )
+      .catch((reason) => {
+        console.error(reason)
+      })
 
-  const applications = applicationResults.reduce<NextApplication[]>((acc, applicationResult) => {
-    // we don't want to fail hard with the full system when one specific application can't get loaded. only log the error.
-    if (applicationResult.status !== 'fulfilled') {
-      console.error(applicationResult.reason)
-    } else {
-      acc.push(applicationResult.value)
-    }
+    return promiseBuild
+  }
 
-    return acc
-  }, [])
+  const promiseInternalApps = applicationPaths.map(registerApp)
 
-  await Promise.all(applications.map((application) => application.initialize()))
-  await Promise.all(applications.map((application) => application.ready()))
+  return Promise.allSettled(promiseInternalApps)
 }
 
 /**
@@ -152,7 +156,7 @@ export const announceApplications = async ({
  * @param vue
  * @param designSystem
  */
-export const announceTheme = async ({
+export const announceTheme = ({
   store,
   vue,
   designSystem,
@@ -163,13 +167,15 @@ export const announceTheme = async ({
   designSystem: any
   runtimeConfiguration?: RuntimeConfiguration
 }): Promise<void> => {
-  const { theme } = await loadTheme(runtimeConfiguration?.theme)
-  await store.dispatch('loadThemes', { theme })
-  await store.dispatch('loadTheme', { theme: theme.default })
-
-  vue.use(designSystem, {
-    tokens: store.getters.theme.designTokens
+  const promiseLoad = loadTheme(runtimeConfiguration?.theme)
+  promiseLoad.then(async ({ theme }) => {
+    await store.dispatch('loadThemes', { theme })
+    await store.dispatch('loadTheme', { theme: theme.default })
+    vue.use(designSystem, {
+      tokens: store.getters.theme.designTokens
+    })
   })
+  return promiseLoad
 }
 
 /**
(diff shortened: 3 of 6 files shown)


pascalwengerter:
@diocas could you run `yarn lint --fix` and commit the changes so we can see what CI says? :)

diocas:
Had to remove the early bail when no token is available, as it was crashing the startup (it was also preventing the showing of the "logging you in" page).
With your latest changes, you stopped using owncloud-sdk to get the capabilities, so owncloud/owncloud-sdk#933 no longer has any effect and the call will be duplicated. Why did you decide to do this, btw?

fschade:
@diocas can you please rebase your pr? Thanks

packages/web-app-files/src/store/actions.js
```diff
@@ -484,12 +484,13 @@ export default {
...
   },
 
-  async loadIndicators({ dispatch, commit }, { client, currentFolder }) {
+  loadIndicators({ dispatch, commit }, { client, currentFolder }) {
     // kind of bruteforce for now: remove the shares for the current folder and children, reload shares tree for the current folder.
     // TODO: when we refactor the shares tree we want to modify shares tree nodes incrementally during adding and removing shares, not loading everything new from the backend.
     commit('SHARESTREE_PRUNE_OUTSIDE_PATH', dirname(currentFolder))
-    await dispatch('loadSharesTree', { client, path: currentFolder })
-    commit('LOAD_INDICATORS')
+    const promiseLoad = dispatch('loadSharesTree', { client, path: currentFolder })
```
fschade, line 491: the reason why this was blocking is that we do not want to render the current view twice or more often, the idea here is to use more performant propfind solution to get needed share indicator information. We will do that on our end.

packages/web-app-files/src/views/Personal.vue
```diff
@@ -131,9 +131,7 @@ export default {
         })
 
         // Load quota
```
fschade, line 133: its ok to load the quota in the background, but can you please write it as [iife](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) instead of then?

packages/web-container/index.html.ejs
```diff
@@ -7,9 +7,10 @@
   <title><%= data.title %></title>
   <link rel="manifest" href="manifest.json" crossorigin="use-credentials">
   <% Object.keys(data.bundle.css).forEach((s) => { %>
-    <link href="<%- data.bundle.css[s] %>" rel="stylesheet">
+    <link href="<%- data.bundle.css[s] %>?<%= data.compilationTimestamp %>" rel="stylesheet">
```
fschade, line 10: CDN is always problematic because it's a potential attack vector.

packages/web-container/index.html.ejs
```diff
@@ -33,7 +34,8 @@
   <script>
     requirejs.config({
       baseUrl: <%- JSON.stringify(data.roots.js) %>,
-      paths: <%- JSON.stringify(data.bundle.js) %>
+      paths: <%- JSON.stringify(data.bundle.js) %>,
+      waitSeconds: 20
```
fschade, line 38: is this still needed after your vendor.js minifications?

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -37,17 +37,17 @@ export const requestConfiguration = async (path: string): Promise<RuntimeConfigu
...
   runtimeConfiguration,
   store
 }: {
   vue: VueConstructor
   runtimeConfiguration: RuntimeConfiguration
   store: Store<any>
-}): Promise<void> => {
-  await store.dispatch('loadConfig', runtimeConfiguration)
-  await store.dispatch('initAuth')
+}): Promise<any> => {
```
fschade, line 48: we must guarantee that config is loaded before proceeding with initAuth. Thats not the case if we parallelize it with Promis.all

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -70,16 +71,20 @@ export const announceStore = async ({
...
   const { openIdConnect = {} } = runtimeConfiguration
 
   if (!openIdConnect.dynamic) {
     return
   }
 
-  const { client_id: clientId, client_secret: clientSecret } = await registerClient(openIdConnect)
-  openIdConnect.client_id = clientId
-  openIdConnect.client_secret = clientSecret
+  const promiseRegistration = registerClient(openIdConnect)
```
fschade, line 81: this also is a step that must happen before continue the code flow, if we try to use the client on another place it can error

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -106,40 +111,39 @@ export const announceApplications = async ({
   router: VueRouter
   translations: unknown
   supportedLanguages: { [key: string]: string }
-}): Promise<void> => {
+}): Promise<any> => {
```
fschade, line 114: the application initialization happens now in the index.ts and also needs to wait in case if something errors. Besides that please do not use the TS any type

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -162,12 +166,14 @@ export const announceTheme = async ({
   designSystem: any
   runtimeConfiguration?: RuntimeConfiguration
 }): Promise<void> => {
```
fschade, line 168: Loading the theme must also happen before any UI is rendered, if we puth this request in the background UI flickering can happen.

packages/web-runtime/src/store/user.js
```diff
@@ -98,8 +101,13 @@ const actions = {
           return
         }
 
-        const userGroups = await client.users.getUserGroups(login.id)
-        const user = await client.users.getUser(login.id)
+        const promiseUserGroups = client.users.getUserGroups(login.id)
```
fschade, line 104: this should not happen in the background, we have some cases where it's needed to be in place before doing other things. as said before, we will touch the whole bootloader in the future and take care that all needed things happen in one place which then reduces the clutter

packages/web-runtime/src/index.ts
```diff
@@ -21,18 +21,22 @@ import {
 
 export const bootstrap = async (configurationPath: string): Promise<void> => {
   const runtimeConfiguration = await requestConfiguration(configurationPath)
-  announceOwncloudSDK({ vue: Vue, runtimeConfiguration })
-  await announceClient(runtimeConfiguration)
-  await announceStore({ vue: Vue, store, runtimeConfiguration })
-  await announceApplications({
+  const promiseOcSDK = announceOwncloudSDK({ vue: Vue, runtimeConfiguration }) // vue.$client
```
fschade, line 24: can you explain why the awaits not in place anymore and instead happen right after? A possible optimization could be to parallelize the theme and translation loading `Promise.all([])`

rollup.config.js
```diff
@@ -23,6 +23,8 @@ import ts from 'rollup-plugin-ts'
 
 const production = !process.env.ROLLUP_WATCH
 
+const compilationTimestamp = new Date().getTime()
```
fschade, line 26: can we have a switch here to exclude sourcemaps in production

packages/web-app-files/src/store/actions.js
diocas: Ah, makes sense... Yes, I'm waiting for this propfind with the sharing info combined.

packages/web-container/index.html.ejs
diocas: Yes, but is it a recommendation/general industry practice to never use these cdn's? (I'm a n00b, so I'm curious to know)

rollup.config.js
diocas: Yes, I'll add a switch. But should sourceMaps be removed from prod? Is this a good practice? (I ran Google Lighthouse which told me that would be an improvement, that's why I made it a default thing).

diocas:
Thanks for the review. I'll be off until Monday and will try to address the comments right when I'm back. When I ran the tests locally, also saw something failing, but I'm not sure what is the issue and might need some guidance, if you don't mind.

packages/web-container/index.html.ejs
diocas: In normal scenarios, doesn't look like. But network connections in the neighbouring france are still pretty stone age... Do you prefer I remove it?

packages/web-runtime/src/container/bootstrap.ts
diocas: Can you point me where we might need it? I was trying to understand the dependencies, but I got the impression that we have 2 clients: 1 for OIDC and another one which is the owncloud-sdk. So I had the impression that this particular client was only being used after `bootstrap` finishes, which still awaits fot this client registration to happen.

pascalwengerter:
> Thanks for the review. I'll be off until Monday and will try to address the comments right when I'm back. When I ran the tests locally, also saw something failing, but I'm not sure what is the issue and might need some guidance, if you don't mind.

Sure, ping me here or in talk.owncloud.com anytime!

rollup.config.js
diocas: done

packages/web-runtime/src/container/bootstrap.ts
fschade: right now yes but we have plans to do the login page rendering different. Then this can conflict.

packages/web-runtime/src/store/user.js
fschade: sorry i re read my comment and your right, its confusing. Can you explain why we await after the function call?

packages/web-container/index.html.ejs
fschade: some comapanies do not allow to call external scripts, this applies if we use a CDN, at a later point we could add a option to the build process to define the url

packages/web-container/index.html.ejs
fschade: i would prefer to have this as build option

packages/web-runtime/src/container/bootstrap.ts
fschade: LOAD_EXTENSION_CONFIG for example, external apps are able to hook into the auth process (not yet) for example publicFiles needsAuth

packages/web-runtime/src/container/bootstrap.ts
fschade: yes exactly

packages/web-runtime/src/index.ts
fschade: sorry i do not get the point, you await the returned function promise later which is the same es doing it right after calling the function. Can you please explain what you to do?

packages/web-container/index.html.ejs
diocas: I can add this as an option in the build process.

packages/web-runtime/src/container/bootstrap.ts
diocas: ok, so `announceApplications` needs to be called after `announceStore`. I'll change that.

packages/web-runtime/src/container/bootstrap.ts
diocas: Why? As I said, my implementation keeps waiting for `announceTheme` before finishing `bootstrap`. So nothing changes in the behavior of `bootstrap`. If nothing changes, if you have conflicts with what I did, you will have conflicts with what you already have...

One thing that you really need to do in the login workflow is to redirect the user to the idp before trying to load the full application (which takes ages).

packages/web-runtime/src/store/user.js
diocas: Imagine `doA` and `doB` call a network resource. Imagine that the resource of A takes 1ms and the resource of B 1.5ms.
If I do:
```
await doA()
await doB()
```
I have to wait for the full roundtrip of A to finish, before doing the roundtrip of B. This takes 1 + 1.5 ms
If I do:
```
a = doA()
b = doB()
await a
await b
```
The roundtrip of A and B happen at ~ the same time. So it takes me 1.5ms to do both.
If I call these functions inside `doAll`, the external behavior is the same: `doAll` finishes only after `doA` and `doB` have finished, but it takes 40% less time to run it.

packages/web-runtime/src/index.ts
diocas: Please see the previous conversation. Parallelising calls is not the same thing as running them sequentially.

packages/web-runtime/src/container/bootstrap.ts
fschade: Yes, the store can be used in the hooks of a application. And to be safe it needs to happen before that

kulmann:
The most pressing parts of this PR have been solved with https://github.com/owncloud/web/pull/6320 being merged. I'll close here.

Issue #5927, opened 2021-10-19
Context menu by deleted shown inside of the rows (CERNBox)


elizavetaRa:
We face the problem with the context menu by deleted files in https://github.com/cernbox/web/tree/qa, so far didn't find what causes it. Appeared about a month ago.

### Steps to reproduce
1. go to "deleted files" view (not code difference in CERNBox web to owncloud web  https://github.com/cernbox/web/commits/master/packages/web-app-files/src/views/Trashbin.vue)
2. All other views are not affected by this bug

### Expected behaviour
Context menu appears on click

### Actual behaviour
Context menu is shown inside of the row

### Environment general
- latest web https://github.com/cernbox/web/commit/f6465ee8c6f7ca2033be3e98eef3cdcb522a43f1 + our changes
tested with owncloud-design-system: 
- 11.0.0
- 10.0.0
- https://github.com/cernbox/owncloud-design-system/tree/qa (our changes rebased on latest-->11.0.0)

#### Browser log
running in puppet, the html DOM element for the context menu is a different one than in other views
![contextmenu_newqa](https://user-images.githubusercontent.com/7430156/137949879-235e5819-4b9c-45cb-9342-534fce7aa67d.png)

locally the error doesn't appear with the same environment
![contextmenu_local](https://user-images.githubusercontent.com/7430156/137949863-cf11d4bd-aed8-49f5-85d3-898598fa0666.png)

fschade:
we reproduced it in a zoom meeting, the reason why this only happens is that the fileid contains a colon `context-menu-drop-pxid:000...`.

@elizavetaRa checks internally if this is a bug or the id needs to be in this format.
On our side we need to clean the ids and keep care that the format is always valid

elizavetaRa:
@ishank011 @diocas

kulmann:
I'd prefer to create the clean id in web (see helpers/resources.js from the files app) and then just use it as a given field in ODS. Reason: I think that the components in ODS should not modify data (in an ideal world) + we probably need the same clean id in web as well. What do you think?

diocas:
I understand that ODS should not modify data, but in practice we're complementing data. :)
I would counter that the fact that we use id's in the html (on the octable component) is an internal implementation detail and therefore should be hidden (web).

fschade:
@elizavetaRa can you please test https://github.com/owncloud/owncloud-design-system/pull/1766 if fixes the problem?

fschade:
[...]

PR #5998, opened 2021-11-08
sanitize resource id

fschade:
## Description
resource id is used on some places where not all characters are allowed, this tries to clean the id by sanitize it.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5927

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-resource-id
+++ b/changelog/unreleased/bugfix-resource-id
@@ -1,0 +1,6 @@
+Bugfix: Clean resource id
+
+The resource id can contain characters that are not allowed in certain places like the HTML element id, we've fixed this by sanitizing it.
+
+https://github.com/owncloud/web/pull/5998
+https://github.com/owncloud/web/issues/5927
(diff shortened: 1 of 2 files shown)


kulmann (changes requested):
This is not going to work, we need the original id in some places. Please introduce a `cleanId` or `safeId` (or similar naming) prop instead which can be used in html. And please add it to all three resource building helpers. 😅

fschade:
superseded by https://github.com/owncloud/owncloud-design-system/pull/1766

pascalwengerter:
> This is not going to work, we need the original id in some places. Please introduce a `cleanId` or `safeId` (or similar naming) prop instead which can be used in html. And please add it to all three resource building helpers. sweat_smile

What happened to the idea of introducing a `cleanId` or `safeId`? I think this has come up before and I vaguely remember implementing two patchy replace-unsafe-stuff-solutions we could get rid of by adding this to our resource "interface to be" :)

fschade:
I don’t like the idea to have something like a “safeId” in the resource object, I expect that the resource object has only data that describes the object and no data that is used for displaying it. In my opinion we should separate those things strictly.

kulmann:
You're right, fully agreeing with that. However, this means code duplication at the moment, as we need the same helper in both the design system and in web.

PR #6037, opened 2021-11-18
Add tooltips with absolute dates to relative dates

JammingBen:
## Description
* Add tooltips with absolute dates to relative dates
* Make tooltips a11y compliant
* Use absolute date in sidebar file infos

Also see https://github.com/owncloud/owncloud-design-system/pull/1787 for the related changed in the files table.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/5672

image

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/enhancement-relative-date-tooltips
+++ b/changelog/unreleased/enhancement-relative-date-tooltips
@@ -1,6 +1,6 @@
 Enhancement: Add tooltips to relative dates
 
-Relative dates like "1 day ago" now have a tooltip that shows the absolute date. This tooltip is a11y-compliant and can be reached via tab.
+Relative dates like "1 day ago" now have a tooltip that shows the absolute date.
 
 https://github.com/owncloud/web/pull/6037
 https://github.com/owncloud/web/issues/5672
--- a/changelog/unreleased/enhancement-relative-date-tooltips
+++ b/changelog/unreleased/enhancement-relative-date-tooltips
@@ -1,0 +1,6 @@
+Enhancement: Add tooltips to relative dates
+
+Relative dates like "1 day ago" now have a tooltip that shows the absolute date. This tooltip is a11y-compliant and can be reached via tab.
+
+https://github.com/owncloud/web/pull/6037
+https://github.com/owncloud/web/issues/5672
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -211,7 +211,7 @@ export default {
       return this.hasAnyShares && !this.isPublicPage
     },
     displayShareDate() {
-      const date = this.formDateFromNow(new Date(this.sharedTime * 1000), 'JSDate')
+      const date = this.formRelativeDateFromJSDate(new Date(this.sharedTime * 1000))
       return upperFirst(date)
     },
     detailSharingInformation() {
@@ -273,13 +273,9 @@ export default {
     capitalizedTimestamp() {
       let displayDate = ''
       if (this.file.mdate) {
-        displayDate = DateTime.fromRFC2822(this.file.mdate)
-          .setLocale(this.$language.current)
-          .toLocaleString(DateTime.DATETIME_FULL)
+        displayDate = this.formDateFromRFC(this.file.mdate)
       } else {
-        displayDate = DateTime.fromRFC2822(this.file.sdate)
-          .setLocale(this.$language.current)
-          .toLocaleString(DateTime.DATETIME_FULL)
+        displayDate = this.formDateFromRFC(this.file.sdate)
       }
       return upperFirst(displayDate)
     },
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -167,11 +167,6 @@ export default {
     sharedViaLabel() {
       return this.$gettext('Shared via:')
     },
-    shareDateTooltip() {
-      return DateTime.fromSeconds(parseInt(this.sharedTime))
-        .setLocale(this.$language.current)
-        .toLocaleString(DateTime.DATETIME_FULL)
-    },
     sharedViaTooltip() {
       return this.$gettextInterpolate(
         this.$gettext("Navigate to '%{folder}'"),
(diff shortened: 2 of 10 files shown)


changelog/unreleased/enhancement-relative-date-tooltips
```diff
@@ -0,0 +1,6 @@
+Enhancement: Add tooltips to relative dates
+
+Relative dates like "1 day ago" now have a tooltip that shows the absolute date. This tooltip is a11y-compliant and can be reached via tab.
```
kulmann, line 3: Tooltips that get created with the directive `v-oc-tooltip` are aria-hidden by default, see https://github.com/owncloud/owncloud-design-system/blob/469a1c710b46ee8af6d46586ab5b3a41252894e8/src/directives/OcTooltip.js#L74

Afaik we made sure that anything that uses a `v-oc-tooltip` had proper accessible content on its own, via aria-label or aria-describedby. This needs a general decision:
a) make tooltips accessible (enforce tabindex, ideally via v-oc-tooltip directive, get rid of aria-hidden, think about other aria-props involved) or
b) keep tooltips aria-hidden, don't make them reachable via keyboard, make sure that accessible content exists which complements the tooltip content.

JammingBen: Personally I'm fine with both. I mean technically, the tooltips have proper content already. Therefore option b) would be sufficient and easier to achieve for now.

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -58,7 +58,7 @@
         <tr v-if="showShareDate" data-testid="shared-date">
           <th scope="col" class="oc-pr-s" v-text="shareDateLabel" />
           <td>
-            <span v-oc-tooltip="shareDateTooltip" v-text="displayShareDate" />
+            <span v-oc-tooltip="shareDateTooltip" tabindex="0" v-text="displayShareDate" />
```
kulmann, line 61: Same issue with the `tabindex` as in https://github.com/owncloud/owncloud-design-system/pull/1787#discussion_r755966806

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -273,9 +273,13 @@ export default {
     capitalizedTimestamp() {
       let displayDate = ''
       if (this.file.mdate) {
-        displayDate = this.formDateFromNow(this.file.mdate, 'Http')
+        displayDate = DateTime.fromRFC2822(this.file.mdate)
```
kulmann, line 276: Looks like it would be worth adding a `formDate` helper for absolute dates, right? You could accept the format as a param with `DateTime.DATETIME_FULL` as a default value.

packages/web-app-files/src/components/SideBar/FileInfo.vue
```diff
@@ -51,6 +54,17 @@ export default {
   computed: {
     ...mapGetters(['capabilities']),
     modificationTime() {
+      if (this.isTrashbinRoute) {
+        return DateTime.fromRFC2822(this.file.mdate)
```
kulmann, line 58:
```suggestion
        return DateTime.fromRFC2822(this.file.ddate)
```
Also, please use the helper mentioned in the other comment, if you decide to build it.

kulmann (changes requested):

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
JammingBen: You mean just like `formDateFromNow()`, right? This method has a comment `// TODO: Refactor those into their own functions a.k.a relDateFromJSDate() etc`. What do think, let's do it that way and refactor `formDateFromNow()` in the course of this as well? I counted 5 usages, looks doable without too much effort.

packages/web-app-files/src/components/SideBar/FileInfo.vue
kulmann: Yep, good idea, thanks!

kulmann (changes requested):
Hmmm, found two things:
1. timezone from ODS shows `CET`, timezone in web shows `GMT+1` (see screenshot)
2. share date seems to be unix timestamp 0. 😞

Screenshot 2021-11-29 at 18 00 57

JammingBen:
@kulmann I can't reproduce both of these 🤔  Maybe it's related to l10n, Web is always German for me. How do I change that to English?

pascalwengerter:
> @kulmann I can't reproduce both of these thinking Maybe it's related to l10n, Web is always German for me. How do I change that to English?

Running with oCIS you can access the `Settings` in the user menu, with OC10 I'm not sure if we even use OC10's language preference 🤔 
Might even be an issue with System/Browser/OC preference languages getting mixed up, too?

kulmann:
> @kulmann I can't reproduce both of these 🤔 Maybe it's related to l10n, Web is always German for me. How do I change that to English?

In oCIS there is the settings ui (click on the top right account menu -> settings) for changing the language.

Made a PR in ODS to set the language for date formatting there as well. Luxon is stateless and immutable, we have to set the date again on any new instance. That fixes the mismatch of date formats in files table and right sidebar.

kulmann:
Let's wait for the updated ODS 11.3.0-rc3 and update it here before merging.

kulmann:
> > @kulmann I can't reproduce both of these thinking Maybe it's related to l10n, Web is always German for me. How do I change that to English?
> 
> Running with oCIS you can access the `Settings` in the user menu, with OC10 I'm not sure if we even use OC10's language preference 🤔 Might even be an issue with System/Browser/OC preference languages getting mixed up, too?

I think for oc10 it defaults to the browser language

kulmann (approved):
LGTM 💪

ownclouders:
Results for <strong>oC10SharingPubExpAndRoles</strong> https://drone.owncloud.com/owncloud/web/20544/39/1

ownclouders:
Results for <strong>oC10SharingPubExpAndRoles</strong> https://drone.owncloud.com/owncloud/web/20546/39/1

kulmann (changes requested):
Something fishy that you'd need to look into (reason why acceptance tests are failing):
Setting an expiration date on a public link results in an expiration date 1 day short of what was selected in the datepicker. If I select December first it implicitly means "end of day, December 1st", but the public link overview shows `Expires in 10 hours` (now at 2pm on November 30). See screenshots.

Screenshot 2021-11-30 at 13 46 05
Screenshot 2021-11-30 at 13 46 12

kulmann (approved):
Nevermind my latest comment. That's a bug in master o_O

Issue #6048, opened 2021-11-23
Cleanup implicit dependencies


fschade:
In web we use implicit code in several places, the most prominent example are vue mixins.
This introduces some problems regarding code quality, analyzability and also introduces cognitive complexity.

Please read [for a deeper dive](https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html).

In Vue 2 it was common practice to use those and accept the downsides,
with the release of vue 3 and it's composition api the coin has flipped and there are now better options.

This is a bucket issue to group touching PR's.

# Benefits
* Preparation for Vue 3 upgrade
* Increase developer ergonomics 
* Isolated testing (reduce susceptibility to errors)

# Rules
* New Code must be tested (unit)
* Old tests must stay intact (if the test makes sense)
* No Performance regressions
* No implicit code path (must be analyzable)
* All dead code down the road gets removed (getters, mutations, ....)
* A bit duplication here and there is ok while this PR is open, the last PR should cover this and abstract sharable code away

# Bucket
* [X] ~~#6049~~
* [X] ~~#6065~~

dschmidt:
We all agree that we want to get rid of mixins and we heavily use composables now. 
The two linked PRs are merged. 

Can we close this ticket now? 
If not what are acceptance criteria? Last mixin gone?

PR #6049, opened 2021-11-23
convert FileListHeaderPosition mixin to composable

fschade:
## Description
See #6048 

## Related Issue
- Part of #6048

## Motivation and Context
clean code ™️ 

## How Has This Been Tested?
- unit tests

## Types of changes
- [X] Technical debt
- [X] Tests

## Checklist:
- [X] Code changes
- [X] Unit tests added


pascalwengerter (commented):
In general LGTM, nice that it has ~50% test coverage and gets rid of a lot of duplication 🤩 as far as I understood we should have a quick discussion if that approach should be generally followed within the project before merging this?

packages/web-app-files/src/mixins.js
```diff
@@ -155,9 +155,9 @@ export default {
             continue
           }
 
-          this.processDirectoryEntryRecursively(item).then(() => {
-            this.$emit('success', null, item.name)
-          })
+          // this.processDirectoryEntryRecursively(item).then(() => {
```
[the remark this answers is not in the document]
fschade: i first had to revert https://github.com/owncloud/web/pull/6050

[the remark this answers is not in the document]
fschade: rebased and back, works now

packages/web-app-files/src/composables/useFileListHeaderPosition/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    const height = appBar ? appBar.getBoundingClientRect().height : 0
+
+    if (y.value === height) {
+      return
+    }
+
+    y.value = height
+  }
+
+  // vue 3 (also 2 with composition api bridge) has no created hook (uses setup instead), so call it here
```
kulmann, line 19: I don't get your comment about the created hook. Is it meant to be a `FIXME: ` with an explanation why you're using `onMounted` instead of `onCreated`? Or something different?

packages/web-app-files/src/views/PublicFiles.vue
```diff
@@ -124,17 +123,18 @@ export default {
...
   setup() {
-    useMutationObserver(['Files/UPSERT_RESOURCE'], ({ payload }) =>
-      nextTick(() => {
-        fileList.accentuateItem(payload.id)
-      })
-    )
+    const { refresh: refreshFileListHeaderPosition, y: fileListHeaderY } =
+      useFileListHeaderPosition()
+
+    useMutationSubscription(['Files/UPSERT_RESOURCE'], async (mutation) => {
```
kulmann, line 134: What's the reason for switching from exploded object syntax to an object var?

kulmann (changes requested):
I know that there's an ongoing discussion about test folders, but to keep it consistent in the status quo, your test file belong in the unit test folder.

packages/web-app-files/src/composables/useFileListHeaderPosition/index.ts
fschade: we had the listener on the created hook before, this doesn't exist anymore, the comment is just here if someone wonders. You're right it can go away

packages/web-app-files/src/views/PublicFiles.vue
fschade: we now return the whole mutation instead of just the payload like before, this way the type can be used to switch between actions.

https://vuex.vuejs.org/api/#subscribe

ownclouders:
Results for <strong>oC10Files1</strong> https://drone.owncloud.com/owncloud/web/20472/17/1

packages/web-app-files/src/views/PublicFiles.vue
kulmann: yeah understood it that way, but instead of `(mutation)` you could write `({ payload, type })`, i.e. exploded object syntax. I find that easier to read. Is there a reason against it?

kulmann (approved):
:heart:

packages/web-app-files/src/views/PublicFiles.vue
fschade: Done in last commit

PR #6119, opened 2021-12-09
Fix #5068: Add context to date in SideBar/FileInfo

### LINKED ISSUE #5068: Timestamp in file row & sidebar aren't identical
### Steps to reproduce
1. Run oCIS master & web frontend
2. Go to "Shared with me page" and click on a row
3. See that timestamp in table row and on sidebar aren't identical (sidebar always says "a few seconds ago")

### Expected behaviour
Timestamp in the sidebar should reflect the timestamp from the table row

### Actual behaviour

Screenshot 2021-05-05 at 09 32 30

dschmidt:
## Description
The date being shown in the sidebar is not wrong, it just can be confusing if the file list shows e.g., a share date, and the sidebar shows a modification date without further explanation.
This PR adds context.

We need to make the strings lowercase because they are appended after a comma.
We cannot make the part before the comma part of the translation string because we need to show the tooltip only for the currently translated part.

I think we talked about this, but I forgot our conclusion:
Last option I see would be to switch the order of filesize and infoString, then we could always start with a capital letter and for file size capitalization is not an issue. I think we dropped this option, but I don't remember why...

I've added one initial implementation and a refactor commit. I'm not 100% sure which one I prefer:
The initial implementation uses "flatter" more fine grained computed properties but uses the same if check in several separate places, the second implementation creates a timeData object which has everything together in place, which would be much nice as well if we need to add other cases in the future ... what do you think?

When we agreed on one approach, I will rebase, add ChangeLog, ... do we need a test for this? What would that look like?

I would have liked to add a translation hint, but afaict there's no way to add translation hints when `$gettextInterpolate` is used ... 

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5068

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

Screenshot_20211209_133720

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


kulmann:
> I think we talked about this, but I forgot our conclusion: Last option I see would be to switch the order of filesize and infoString, then we could always start with a capital letter and for file size capitalization is not an issue. I think we dropped this option, but I don't remember why...

Hm, I think we didn't have a reason. The way it's displayed in classic is size first, time second. I don't have a preference visually, so I think taking the easier implementation way is good. Let's try having the time info string first and the size second.

> When we agreed on one approach, I will rebase, add ChangeLog, ... do we need a test for this? What would that look like?

You could create a snapshot test for the component, one with a `ddate` and one with an `mdate` / without a `ddate`. We don't want to have handcrafted matches on strings. ;-)

> I would have liked to add a translation hint, but afaict there's no way to add translation hints when `$gettextInterpolate` is used ...

Good one, actually one thing that's missing in your implementation: `$gettextInterpolate` is only a helper for interpolation, not for translation. So you need to translate first with one of the translation helpers, then with the translated string you can interpolate in a second step. See the example here: https://github.com/Polyconseil/vue-gettext#interpolation-support-1

dschmidt:
Fixed and test added. Good to merge from my POV.

ownclouders:
Results for <strong>oCISResharing1</strong> https://drone.owncloud.com/owncloud/web/20855/64/1

ownclouders:
Results for <strong>oC10SharingAccept</strong> https://drone.owncloud.com/owncloud/web/20856/16/1

dschmidt:
Can anybody help with the test failures?

pascalwengerter:
> Can anybody help with the test failures?

Sure. Seems like an unrelated failure though, could you rebase your branch on the current master?

ownclouders:
Results for <strong>oCISFavorites</strong> https://drone.owncloud.com/owncloud/web/20871/55/1

pascalwengerter:
> Results for **oCISFavorites** https://drone.owncloud.com/owncloud/web/20871/55/1 boom The acceptance tests pipeline failed. The build has been cancelled.

Random failure due to the IDP service not being available...perhaps needs another restart

ownclouders:
Results for <strong>oCISBasic</strong> https://drone.owncloud.com/owncloud/web/20884/51/1

ownclouders:
Results for <strong>oC10SharingAccept</strong> https://drone.owncloud.com/owncloud/web/20886/16/1
- webUISharingAcceptShares/acceptShares.feature:111

dschmidt:
Tests are green. Someone needs to approve still :)

pascalwengerter (approved):
LGTM ⏲️

PR #6133, opened 2021-12-11
Fix #4615: Use downloadUrl for 'Open in browser' on public files

### LINKED ISSUE #4615: Opening PDF from public link broken
### Steps to reproduce
1. Share a PDF file via public link
2. Open public link
3. Either click `Open in browser` in the actions accordion of the right sidebar or click on the filename within the file list (`Open in browser` is the default action for pdf files)

### Expected behaviour
The pdf file should be opened in a new browser tab

### Actual behaviour
The browser shows a `blob://` url but shows an error message like `Failed to load PDF document.` (see screenshot). Blob download in the network tab looks ok (has a reasonable content length and shows `200 OK`).

Screenshot 2021-01-14 at 23 29 45

### LINKED ISSUE #6128: Cannot open the pdf file in the browser via public link
### Steps to reproduce

1. Create pdf file into folder
2. Create public link to the folder
3. Open public link 
4. Try to open pdf file in browser

### Expected behaviour
The file can be opened in a browser

### Actual behaviour
The file was downloaded, not opened in a browser

https://user-images.githubusercontent.com/84779829/145585182-f666cd2e-47b9-4dbe-8e75-994b84707434.mov

Env:
up ocis local
OCIS_VERSION=v1.16.0
Web_version= v4.6.1
BRANCH=v1.16.0-rc2

dschmidt:
## Description
When #4615 was created, using "open in browser" on public links indeed resulted in the described behavior.
In https://github.com/owncloud/web/pull/4689 the behavior was changed to open the download url in a new tab, that in turn did not open the file in the browser (for me, Chromium 95) but instead downloaded it with a random/hash filename. 

The code path for private files achieves the correct result: it opens the file in a new tab and displays it.
We can align the behavior for public files with that by simply using the downloadUrl in the XHR download.

## Related Issue

- Fixes #4615
- Fixes #6128

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/mixins/actions/fetch.js
+++ b/packages/web-app-files/src/mixins/actions/fetch.js
@@ -36,18 +36,15 @@ export default {
   },
   methods: {
     $_fetch_trigger({ resources, mimeType, isPublicFile }) {
-      if (isPublicFile) {
-        const url = resources[0].downloadURL
-        window.open(url, '_blank')
-        return
-      }
-
-      // FIXME: use presigned URL instead
-      const url = this.$client.helpers._webdavUrl + resources[0].path
+      let url
       const headers = new Headers()
-
-      headers.append('Authorization', 'Bearer ' + this.getToken)
-      headers.append('X-Requested-With', 'XMLHttpRequest')
+      if (isPublicFile) {
+        url = resources[0].downloadURL
+      } else {
+        url = this.$client.helpers._webdavUrl + resources[0].path
+        headers.append('Authorization', 'Bearer ' + this.getToken)
+        headers.append('X-Requested-With', 'XMLHttpRequest')
+      }
 
       fetch(url, {
         method: 'GET',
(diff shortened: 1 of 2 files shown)


kulmann (approved):
Thanks for bringing in some consistency 👍

packages/web-app-files/src/mixins/actions/fetch.js
```diff
@@ -36,19 +36,16 @@ export default {
   },
   methods: {
     $_fetch_trigger({ resources, mimeType, isPublicFile }) {
+      let url;
```
kulmann, line 39: linter complained about this semicolon, you can just drop it

ownclouders:
Results for <strong>oCISRename</strong> https://drone.owncloud.com/owncloud/web/20940/53/1

Issue #6082, opened 2021-12-02
Smoke tests are flaky


phil-davis:
I noticed a few of these earlier in the week, but couldn't reproduce it yesterday. Just now I got this:

https://drone.owncloud.com/owncloud/web/20591/10/18
```
    Then "Brian" downloads the following files
      │ resource │ from             │
      │ data.zip │ Shares/meal plan │
    ✖ failed
      page.click: Timeout 30000ms exceeded.
      =========================== logs ===========================
      waiting for selector "//span[@data-test-resource-name="data.zip"]/ancestor::tr[contains(@class, "oc-tbody-tr")]//button[contains(@class, "oc-table-files-btn-action-dropdown")]"
      ============================================================
          at Object.open (/var/www/owncloud/web/tests/smoke/support/cta/files/sidebar.ts:5:16)
          at AllFilesPage.downloadFiles (/var/www/owncloud/web/tests/smoke/support/page/files/allFiles.ts:64:43)
          at World.<anonymous> (/var/www/owncloud/web/tests/smoke/features/step_definitions/app.files.ts:75:27)
```

That was running with oC10 server as the back-end. I think that I have only seen it fail when using the oC10 back-end.

phil-davis:
https://drone.owncloud.com/owncloud/web/20603/10/16
```
    Then "Alice" downloads the following files
      │ resource      │ from                                 │
      │ parent.txt    │ groups/Kindergarten Koalas/meal plan │
      │ lorem.txt     │ groups/Kindergarten Koalas/meal plan │
      │ lorem-big.txt │ groups/Kindergarten Koalas/meal plan │
      │ data.zip      │ groups/Pre-Schools Pirates/meal plan │
      │ lorem.txt     │ groups/Pre-Schools Pirates/meal plan │
      │ lorem-big.txt │ groups/Pre-Schools Pirates/meal plan │
      │ data.zip      │ groups/Teddy Bear Daycare/meal plan  │
      │ lorem.txt     │ groups/Teddy Bear Daycare/meal plan  │
      │ lorem-big.txt │ groups/Teddy Bear Daycare/meal plan  │
    ✖ failed
      page.click: Timeout 30000ms exceeded.
      =========================== logs ===========================
      waiting for selector "//span[@data-test-resource-name="data.zip"]/ancestor::tr[contains(@class, "oc-tbody-tr")]//button[contains(@class, "oc-table-files-btn-action-dropdown")]"
      ============================================================
[...]

PR #6135, opened 2021-12-12
Ensure route config is honored for new file handlers

diocas:
If we have more than one new file handler in the extensions, we never check the routes associated with them.
If, for example, we need different routes in order to distinguish internal and public links (i.e #6125), without this fix we would see both entries in the "new" menu.

In the past I only added a single "new handler" to avoid seeing 2 entries, but then when I create a new file, the opening action will match the route it belongs to.

### DIFF
--- a/changelog/unreleased/bugfix-new-file-handler
+++ b/changelog/unreleased/bugfix-new-file-handler
@@ -1,0 +1,5 @@
+Bugfix: Ensure route config is honored for new file handlers
+
+Only display the new file entries for the routes it belongs to.
+
+https://github.com/owncloud/web/pull/6135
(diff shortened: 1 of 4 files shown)


diocas:
I'm not sure why these smoke tests are failing.. Any pointer?

pascalwengerter:
> I'm not sure why these smoke tests are failing.. Any pointer?

Could you rebase? We reduced the flakiness by a bit yesterday, still some OC10 issues remaining, sadly. Context lives in https://github.com/owncloud/web/issues/6082#issuecomment-993328824

diocas:
@pascalwengerter I was following that ticket and rebased today.

There's still the possibility I broke something... I made an assumption that maybe it's not correct.

pascalwengerter:
> @pascalwengerter I was following that ticket and rebased today.
> 
> There's still the possibility I broke something... I made an assumption that maybe it's not correct.

The changes don't look related to the failures in CI. Let's wait for #6149 to be merged and rebase your PR and see :)

diocas:
@pascalwengerter all green now :)

kulmann (approved):

PR #6136, opened 2021-12-12
[full-ci] Apply sorting before pagination

### LINKED ISSUE #5687: Incorrect sort when you use pagination
### **Sort when file more than 100:**

1. Open subfolder "Check Sort/CheckPagination"
2. Choose the sort by name
3. Go to page 2

**Expected behaviour:**
files on page 2: 101, 102, 103, 104

**Actual behaviour:**
files on page 2: the last 4 files from the response (random)

Screenshot 2021-08-03 at 11 53 17
Screenshot 2021-08-03 at 11 56 52

### Server configuration

branch: release-4.0.0
backend: ocis 1.9.0 with docker reachable at https://host.docker.internal:9200/
browser: chrome Version 92.0.4515.107

dschmidt:
## Description
As reported in https://github.com/owncloud/web/issues/5687 sorting is broken for us as soon as we paginate.
The problem is that we paginate in our views and let `OcTable` do the sorting on the current page, which hardly makes any sense. Instead we need to sort first and paginate the sorted data afterwards.

I've extracted sorting from `OcTable` / `SortMixin` and moved it into a new composable.

This approach is great for testing the sorting logic and it also makes it super easy to wire up the sorting mode to the url (🥳 ), but I'm still not completely happy with the code ...

The sorting logic and the `OcTable` basically operate on the same configuration structure now, but the sorting logic lives outside the `ResourceTable` so we cannot pass the fields config from there into the sort composable. Building the whole config outside the `ResourceTable` isn't super straightforward either because the non-searching relevant properties of the fields depend on translation and possibly other data of the `ResourceTable` instance.

For now I've introduced and exported a `determineSortFields` function in the `ResourceTable` file ... which works, but it feels horribly wrong.

This is a PoC, let's discuss it tomorrow or so :) 

## Related Issue
- Fixes #5687 
- Depends on https://github.com/owncloud/owncloud-design-system/pull/1825

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

https://user-images.githubusercontent.com/448487/145723655-94df5cc5-96ed-4c35-9f7e-7ffcbb177448.mp4

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


kulmann:
Only read the description of your Pr so far, but one thing is important: IMO we need to move sorting out of the table entirely. We'll have a grid view soon. And we don't want to reimplement sorting for the grid view a second time but instead just reuse it. Let's discuss the rest tomorrow 👍🏻

packages/web-app-files/src/components/FilesList/ResourceTable.vue
```diff
@@ -268,6 +318,16 @@ export default {
       type: Boolean,
       required: false,
       default: false
+    },
+    sortBy: {
+      type: String,
+      required: false
+    },
+
+    // FIXME: asc or desc
```
kulmann, line 327: Reminder to fix the `FIXME`.

packages/web-app-files/src/composables/useSort/index.ts
```diff
@@ -0,0 +1,117 @@
...
+import orderBy from 'lodash-es/orderBy'
+
+export enum SortDir {
+  Desc = 'desc',
+  Asc = 'asc'
+}
+
+interface SortField {
+  name: MaybeRef<string>
+  sortable?: MaybeRef<boolean>
```
kulmann, line 13: `sortable` can also be a callback which is intended to transform a value into a sortable value.

kulmann (changes requested):
breaks on empty folders 😁 probably because it can't read sortable fields from the "first item" for an empty folder

dschmidt:
> breaks on empty folders grin probably because it can't read sortable fields from the "first item" for an empty folder

I've added a HACK/workaround for this, that I don't quite understand ... maybe someone else can look into why that seems to be neccessary (and only in the case of `[]` as value).

The changes to the Personal page need to be ported to a bunch of other views supporting pagination.

If someone wants to take over this PR, as usual I'm fine with squashing commits etc :)

kulmann:
assigned myself, will take over this PR. Thanks for the work so far! 😍

pascalwengerter (commented):
Great stuff in general, testing locally and seeing some console errors on the Shared with me page (related to dates IIRC), investigating..

pascalwengerter:
Found more problems related to sorting by date on the "Shared with Others" page, too

kulmann:
Share date, yeah... can confirm :-(

kulmann:
Sorting by share date, sort breaking on empty tables in SharedWithMe page and different sorting query params for the two different tables on the SharesWithMe page have been solved now. PR is ready for final review.

pascalwengerter:
Just tried the new "Shared with me" logic, two findings:
- sorting in the "Status" in the Pending table doesn't really make sense - the column only contains actions and items disappear from the table upon interaction
- after accepting/declining shares, they don't show up in the Accepted/Declined tables anymore 🤔 investigating the new route params now, no console errors

pascalwengerter:
> Just tried the new "Shared with me" logic, two findings:
> 
>     * sorting in the "Status" in the Pending table doesn't really make sense - the column only contains actions and items disappear from the table upon interaction
> 
>     * after accepting/declining shares, they don't show up in the Accepted/Declined tables anymore thinking investigating the new route params now, no console errors

Update: View mode didn't change anymore after sorting params were added to the route. Browsing back to a "raw" shared-with-me route makes accepted/declined shares work as expected

kulmann:
> Update: View mode didn't change anymore after sorting params were added to the route. Browsing back to a "raw" shared-with-me route makes accepted/declined shares work as expected

good spot, thank you!

kulmann:
> > Update: View mode didn't change anymore after sorting params were added to the route. Browsing back to a "raw" shared-with-me route makes accepted/declined shares work as expected
> 
> good spot, thank you!

... and fixed. It was a type issue. The view mode computed property was not an int anymore, but a string.

pascalwengerter:
> > > Update: View mode didn't change anymore after sorting params were added to the route. Browsing back to a "raw" shared-with-me route makes accepted/declined shares work as expected
> > 
> > 
> > good spot, thank you!
> 
> ... and fixed. It was a type issue. The view mode computed property was not an int anymore, but a string.

Can confirm, just checked and it's fixed. Not having the status be sortable in the "pending" table could be changed later from my perspective

pascalwengerter (approved):
Code-wise LGTM and works like a charm, kudos for the test coverage!

packages/web-app-files/src/composables/useSort/index.ts
```diff
@@ -40,7 +40,7 @@ export function useSort<T>(options: SortOptions<T>): SortResult<T> {
 
   const items = computed<Array<T>>(() => {
     if (!unref(sortBy)) {
-      return options.items
```
dschmidt: Ah this probably fixed the need for the double unref, nice!

PR #6144, opened 2021-12-13
[test-only]Add api locators to flaky place

ScharfViktor:
I created PR to add more reliability in some flaky places of the test. of course this does not guarantee the elimination of all flaky places, but hopefully will make a step towards the reliability of the test

I added Api locators in two place:
- check number of versions
- Sidebar panel

I added "check response" to the actions:
- move/ copy/ rename object
- create folder
- delete share
- delete calobaration
- accept share

I want to see how works it in CI


pascalwengerter:
@ScharfViktor I think we can close this in favor of #6149?

ScharfViktor:
> @ScharfViktor I think we can close this in favor of #6149?

Damn, we did a similar PR. But still, I don't want to close it, i will drop similar code

ownclouders:
Results for <strong>oC10SharingPublic1</strong> https://drone.owncloud.com/owncloud/web/21030/37/1
- webUISharingPublicExpire/shareByPublicLinkExpiringLinks.feature:11

ScharfViktor:
I have a suspicion that flaky after "check response in accept" in OC10 will continue. 

I repeated it localy:
- Alice creates share
- Brian accepts share
- Brian going to do something but folder Shares is not exist

This test was failed after addition "check response in accept": https://drone.owncloud.com/owncloud/web/21021/10/18
I added message that folder "Shares" is exist in OC10 (I know it's not cool), it's just to check the reason for the error in CI

ScharfViktor:
One more interesting thing, the playwright developers admit that the methods page$, page$$ and page.waitForSelector are flaky and advise use locators:
Screenshot 2021-12-14 at 21 41 58

fschade (approved):
LGTM

PR #6156, opened 2021-12-15
Move common app logic to useAppDefaults composable

dschmidt:
## Description
This extracts some file and folder loading/saving from mediaviewer and drawio apps and finally moves them into a central place.
Functions `getFileInfo`, `putFileContents` and `getFileContents` are based on https://github.com/owncloud/web/pull/6125/commits/68b7332d09184134391f29681c702fa9d93aae92 by @diocas 
Thanks and kudos for that!
I wanted to solve the routing a bit differently and it was easier for me to just copy the relevant parts instead of revoking parts of your commit. I hope this is fine for you, if you prefer I can probably mark the commit as being authored by us two or cherry pick and revoke the routing part. Just let me know, thanks!

Names are not great, typescript types are basically not there, the big composable file can possibly be split into more modules, etc....
This is a PoC and just lays a ground for dicussion.

## Motivation and Context
Creating an app needs deep knowledge of the possible views in oC Web  (public, private, shared, ...) and knowledge about the structure of the store. Also app code should be rather resilient against structural changes in oC Web (as much as reasonably possible at least). That is why we should provide apps a stable-ish API to run against.
This PR adds a composable providing a bunch of helper functions/properties/... usually needed to write applications handling files.
Do we need to/can we/should we consider namespace handling, too?

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] Api Concept
- [ ] Types
- [ ] Proper modularization
- [ ] Tests
- [ ] Remove dependency of `web-pkg` on `web-app-files` (?)


diocas:
All fine for me :)
The only reason I separated into 2 routes was to have a nicer name. But @kulmann explained that by using `contextRouteName` we know how to get back to the previous view... And I guess this doesn't break our (CERNBox's) universal EOS urls, so all good.

Can we apply the same logic to the `external` app? (I opened a similar PR for that as well)

dschmidt:
I will look into it, my goal is certainly to unify URL handling accross all apps. For now I want to port apps shipped inside oC Web, in the long run this should become sort of a universal toolkit also for apps implemented externally.

kulmann (commented):
Nice unification! Didn't look into every detail, but the approach and direction this is going are nice and helpful. 👍

dschmidt:
Naming:
We have a `useDefaults` composable, maybe `useAppDefaults` isn't a too bad name then.
packages/web-runtime/src/container/api.ts uses `applicationName`, so we should probably use that instead of `appName` here.


About the other functions and apis:
As we are not going to release this as public api too soon, we can revisit that at some point.
It should be tested, that mediaviewer and draw-io still work on all routes they support. I've only tested the personal file view and public links.

The composable can (and should) be split into more granular pieces but that can be done at a later point as well.

ownclouders:
Results for <strong>oC10SharingAccept</strong> https://drone.owncloud.com/owncloud/web/21637/15/1
- webUISharingAcceptSharesToRoot/acceptShares.feature:108

ownclouders:
Results for <strong>oC10IntegrationApp1</strong> https://drone.owncloud.com/owncloud/web/21637/68/1

dschmidt:
> I will look into it, my goal is certainly to unify URL handling accross all apps. For now I want to port apps shipped inside oC Web, in the long run this should become sort of a universal toolkit also for apps implemented externally.

I looked at it briefly, but not having a working wopi setup or something like that it's a bit hard for me to test. In the long run we should definitely look to streamline this...

dschmidt:
Okay, drone is green - sonarcloud probably fails because of the `FIXME` (right?), IMHO that's still fine. Adding types for the whole sdk client is way out of scope for this PR and I don't feel like marking it as TODO just to make sonarcloud happy :)

dschmidt:
From my POV this is good to go!

packages/web-pkg/src/composables/useAppDefaults/useAppFileLoading.ts
```diff
@@ -0,0 +1,113 @@
...
+  const getUrlForResource = ({ path, downloadURL }: Resource, query: QueryParameters = null) => {
+    const queryStr = !query ? '' : queryString.stringify(query)
+    if (!unref(isPublicContext)) {
+      const urlPath = ['..', 'dav', 'files', store.getters.user.id, path.replace(/^\//, '')].join(
+        '/'
+      )
+      return [client.files.getFileUrl(urlPath), queryStr].filter(Boolean).join('?')
+    }
+
+    // If the mediaFile does not contain the downloadURL we fallback to the normal
```
dschmidt, line 41:
```suggestion
    // If the resource does not contain the downloadURL we fallback to the normal
```

packages/web-pkg/src/composables/useAppDefaults/useClient.ts
```diff
@@ -0,0 +1,8 @@
+import { getCurrentInstance } from '@vue/composition-api'
+
+// FIXME: use proper type
+export type Client = any
+
+export const useClient = (): Client => {
+  return (getCurrentInstance().proxy as unknown as any).$client
```
kulmann, line 7: The files app has a `ClientService`, I'd like to utilize that here. Would need to be moved to `web-pkg` first, obviously. What do you think?

kulmann (changes requested):

packages/web-pkg/src/composables/useAppDefaults/useAppFileLoading.ts
```diff
@@ -0,0 +1,113 @@
...
+        filePath,
+        unref(publicLinkPassword),
+        davProperties
+      )
+      return fileInfo
+    }
+    return client.files.fileInfo(filePath, davProperties)
+  }
+
+  const putFileContents = (
```
kulmann, line 89: The name of the composable, `useAppFileLoading` seems to be misleading if we also have `putFileContents`.

kulmann (approved):
😍

PR #6165, opened 2021-12-17
Add grouping functionality for tables

elizavetaRa:
## Description
Grouping as well as preview functionality for the tables, web part
ODS part: https://github.com/owncloud/owncloud-design-system/pull/1838

Setup is done with the property "groupingSettings" . Following settings are possible:
- groupingFunctions: Object with keys as grouping options names and functions that get a table data row and return a group name for that row. The names of the functions are used as grouping options. 

- groupingBy: must be either one of the keys in groupingFunctions or 'None'. If not set, default grouping will be 'None'.
- ShowGroupingOptions: boolean value for showing or hinding the select element with grouping options above the table.
- PreviewAmount: Integer value that is used to show only the first n data rows of the table.object

Screenshot from 2021-09-08 10-47-39

## Motivation and Context
UX-friendly option for better overviews of table contents

## Types of changes
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added/updated

## Open tasks:

- Tests

### DIFF
--- a/tests/acceptance/features/webUISharingInternalUsersBlacklisted/shareWithUsers.feature
+++ b/tests/acceptance/features/webUISharingInternalUsersBlacklisted/shareWithUsers.feature
@@ -23,8 +23,8 @@ Feature: Sharing files and folders with internal users
     And user "Alice" has accepted the share "Shares/testimage.jpg" offered by user "Carol" in the server
     And user "Carol" has shared folder "simple-folder" with user "Alice" in the server
     And user "Alice" has accepted the share "Shares/simple-folder" offered by user "Carol" in the server
-    And the administrator has enabled exclude groups from sharing
-    And the administrator has excluded group "grp1" from sharing
+    And the administrator has enabled exclude groups from sharing in the server
+    And the administrator has excluded group "grp1" from sharing in the server
     When user "Alice" logs in using the webUI
     And the user opens folder "Shares" using the webUI
     Then the user should not be able to share file "testimage.jpg" using the webUI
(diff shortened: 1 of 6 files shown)


diocas:
@kulmann is there an ETA for this? We are in a rush to have everything ready by the end of this week.

packages/web-app-files/src/components/FilesList/ResourceTable.vue
```diff
@@ -164,6 +165,17 @@ export default {
       type: Array,
       required: true
     },
+    /**
```
fschade, line 168: the doc-block here is not needed, the table is files-app only and kinda "private"

diocas: I think we discussed you guys would refactor what was necessary and merge this PR (or a replacement of it). Or do you need us to do anything?

kulmann: We can take care. Looking at this PR, I have the impression that the grouping settings for the shared with me page will stay in your fork of web until you have it bulletproof? Asking because this PR only adds a property for the ResourceTable and applies it to the ODS table.

diocas: @elizavetaRa what happened to the configs? Since we now have the same shared with me page, it's ok to have everything merged upstream.

PR #6173, opened 2021-12-20
Remember the UI that was last selected via the application switcher

### LINKED ISSUE #5610: [oc10] Let users choose if they want to use the new or classic design by default
## Acceptance criteria
- Users have a way (e.g. in user settings of oc10 or cookie based) to choose between the classic and the new web UI as their default.

JammingBen:
## Description

This needs https://github.com/owncloud/core/pull/39600 to work.

## Related Issue
- https://github.com/owncloud/enterprise/issues/4694
- Closes #5610

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -75,7 +75,7 @@ endif
 .PHONY: ocx-app-config
 ocx-app-config:
 	cp -R packages/web-integration-oc10/appinfo $(dist_dir)
-	cp -R packages/web-integration-oc10/js $(dist_dir)
+	cp packages/web-integration-oc10/js/* $(dist_dir)/js
 	cp -R packages/web-integration-oc10/lib $(dist_dir)
 
 .PHONY: ocx-app-bundle
--- a/Makefile.release
+++ b/Makefile.release
@@ -75,6 +75,7 @@ endif
 .PHONY: ocx-app-config
 ocx-app-config:
 	cp -R packages/web-integration-oc10/appinfo $(dist_dir)
+	cp -R packages/web-integration-oc10/js $(dist_dir)
 	cp -R packages/web-integration-oc10/lib $(dist_dir)
 
 .PHONY: ocx-app-bundle
--- a/packages/web-integration-oc10/lib/Application.php
+++ b/packages/web-integration-oc10/lib/Application.php
@@ -22,6 +22,7 @@
 namespace OCA\Web;
 
 use OCP\AppFramework\App;
+use OCP\Util;
 
 class Application extends App {
 
@@ -33,4 +34,8 @@ class Application extends App {
 	public function __construct(array $urlParams = []) {
 		parent::__construct(Application::APPID, $urlParams);
 	}
+
+    public function boot(): void {
+        Util::addscript(Application::APPID, 'app');
+    }
 }
--- a/packages/web-integration-oc10/lib/Application.php
+++ b/packages/web-integration-oc10/lib/Application.php
@@ -36,6 +36,6 @@ class Application extends App {
 	}
 
     public function boot(): void {
-        Util::addscript(Application::APPID, 'app');
+        Util::addScript(Application::APPID, 'app');
     }
 }
--- a/packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
+++ b/packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
@@ -24,7 +24,7 @@
       close-on-click
     >
       <oc-list class="applications-list">
-        <li v-for="(n, nid) in applicationsList" :key="`apps-menu-${nid}`">
+        <li v-for="(n, nid) in applicationsList" :key="`apps-menu-${nid}`" @click="clickApp(n)">
           <oc-button
             :key="n.url ? 'apps-menu-external-link' : 'apps-menu-internal-link'"
             :type="n.url ? 'a' : 'router-link'"
@@ -34,7 +34,6 @@
             appearance="raw"
             :class="{ 'oc-background-primary-gradient': n.active }"
             :variation="n.active ? 'inverse' : 'passive'"
-            @click="clickApp(n)"
           >
             <span class="icon-box">
               <oc-icon :name="n.icon" />
@@ -49,6 +48,8 @@
 </template>
 
 <script>
+import { clientService } from 'web-pkg/src/services'
+import { configurationManager } from 'web-pkg/src/configuration'
 import { mapGetters } from 'vuex'
 
 export default {
@@ -60,7 +61,7 @@ export default {
     }
   },
   computed: {
-    ...mapGetters(['configuration', 'getToken']),
+    ...mapGetters('runtime/auth', ['accessToken']),
 
     applicationSwitcherLabel() {
       return this.$gettext('Application Switcher')
@@ -75,22 +76,17 @@ export default {
   methods: {
     async clickApp(appEntry) {
       // @TODO use id or similar
-      if (appEntry.iconMaterial === 'switch_ui') {
+      if (appEntry.url?.endsWith('/apps/files')) {
         await this.setClassicUIDefault()
       }
     },
     setClassicUIDefault() {
-      const endpoint = new URL(this.configuration.server || window.location.origin)
+      const endpoint = new URL(configurationManager.serverUrl || window.location.origin)
       endpoint.pathname =
         endpoint.pathname.replace(/\/$/, '') + '/index.php/apps/web/settings/default'
-      const headers = new Headers()
-      headers.append('Authorization', 'Bearer ' + this.getToken)
-      headers.append('X-Requested-With', 'XMLHttpRequest')
-      return fetch(endpoint.href, {
-        headers,
-        method: 'POST',
-        body: JSON.stringify({ isDefault: false })
-      })
+
+      const httpClient = clientService.httpAuthenticated(this.accessToken)
+      return httpClient.post(endpoint.href, { isDefault: false })
     }
   }
 }
--- a/packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
+++ b/packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
@@ -81,7 +81,7 @@ export default {
       }
     },
     setClassicUIDefault() {
-      const endpoint = new URL(configurationManager.serverUrl || window.location.origin)
+      const endpoint = new URL(configurationManager.serverUrl)
       endpoint.pathname =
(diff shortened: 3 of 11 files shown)


JanAckermann:
Additionally, to say, we could also let the user set the default app by providing a select and let the user choose between several apps in the core, but from our point of view, this is not quite user-friendly.

pmaier1:
Thanks, looks good so far. Still, this approach might be a little hard to find for the average user. Maybe there are some other ideas? @tbsbdr 

> Additionally, to say, we could also let the user set the default app by providing a select and let the user choose between several apps in the core, but from our point of view, this is not quite user-friendly.

Agree. Users do not need to choose other apps.

tbsbdr:
1. **Remember**
UX wise I guess most user would expect that the UI - Classic or Web UI - just stays the way they chose it via the app switcher. So the system should remember the last choice i took via the app switcher regarding the design.
2. **Set explicitly** 
An additional option in the settings would be nice-to-have for those users, who expect to change the appearance from within the settings. But I think that's more of a corner case.

JammingBen:
I agree that 1. is the cleanest approach UX wise, although we'd need to check for the technical details. It makes this feature more appealing and easy to use. 2 issues I still struggle with though:

* There is no "Switch back to the old UI" toggle once the new UI has been activated. IMO that would be nice to have with this approach.
* The settings are always classic UI (currently). Once I am in the settings and want to go back to the files list, I usually go via the app switcher and select "Files". Now I still see the classic UI, even if I've selected the new UI as default. That might be confusing for users... I guess it's ok for now as we will have the settings in the new UI eventually.

tbsbdr:
As just discussed:

The switch back and forth should work via the app switcher for now. 
image

Let's focus on the "Remember" approach. With that, explicit settings are not necessary for a first version.
cc @mcarioscio-ownCloud

packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -84,6 +92,27 @@ export default {
     },
     focusMenuButton() {
       this.$refs.menubutton.$el.focus()
+    },
+    async clickApp(appEntry) {
+      // @TODO use id or similar
+      if (appEntry.iconMaterial === 'switch_ui') {
+        await this.setClassicUIDefault()
+      }
+    },
```
JanAckermann, line 101: Well, this is a bit odd, but we need to have a reserved id or so.

JammingBen:
@kulmann What do you think of the technical solution, especially the web-part? Also I'm not sure about adding tests. I think due to SonarCloud we need at least unit or integration tests.

packages/web-runtime/src/components/ApplicationsMenu.vue
```diff
@@ -22,7 +22,13 @@
     >
       <ul class="uk-grid-small uk-text-center" uk-grid>
         <li v-for="(n, nid) in menuItems" :key="`apps-menu-${nid}`" class="uk-width-1-3">
-          <a v-if="n.url" key="apps-menu-external-link" :target="n.target" :href="n.url">
+          <a
+            v-if="n.url"
+            key="apps-menu-external-link"
+            :target="n.target"
+            :href="n.url"
+            @click="clickApp(n)"
```
pascalwengerter, line 30: Not sure if having a click-handler on an `<a>` tag is optimal here. You could go for the OcButton component and define the type there maybe?

JammingBen: Good point! However, this links to an external page, so I'm not sure if we can use a button here (in regards to a11y).

Buttons *should* be used when triggering some kind of backend logic, which is definitely true here. At the same time, buttons *shouldn't* be used when linking to external pages, which is also true here.

Or do you mean using `type="a"` for the button? But it would be rendered as an `a` element anyway, right?

pascalwengerter: I think we'll have to render a second type of list items here, depending on whether they're supposed to be routing (`<a>`)  or trigger an async action (`<button>`)

```
<li v-for="item of menuItems">
  <a v-if="item.target">...</a>
  <button v-else @click="item.clickHandler()">...</a>
</lI>
```

or go for a more sophisticated solution, but this will most likely get a bit messy with (optional) attributes and click handlers

```
<li v-for="item of menuItems">
  <oc-button :type="menuItemType(item)">...</oc-button>
</lI>

...
  computed: {
    return item.target ? 'a' : ''
  }
```

What do you think?

kulmann: What's the reason for adding the click handler to the link anyway? I'd expect a dedicated item in the user menu (not in the app switcher) for going to the default settings in oc10. Not an implicit magical behaviour when going back to the classic UI. 😉

kulmann: Difficulty with that is, that we don't know if web runs from within an oc10 app at all. So we don't exactly know when to show that menu item...

kulmann:
> @kulmann What do you think of the technical solution, especially the web-part? Also I'm not sure about adding tests. I think due to SonarCloud we need at least unit or integration tests.

In general I like it. 🤩  Just the part with the implicit default handling when switching back to the classic UI is a bit irritating.

pascalwengerter:
> > @kulmann What do you think of the technical solution, especially the web-part? Also I'm not sure about adding tests. I think due to SonarCloud we need at least unit or integration tests.
> 
> In general I like it. star_struck Just the part with the implicit default handling when switching back to the classic UI is a bit irritating.

I think this needs some clarification whether the PR is about remembering the UI that was last used (currently the case code-wise) as opposed to saving the user's preference (as described in some ticket IIRC), no?

packages/web-runtime/src/components/ApplicationsMenu.vue
JanAckermann: @kulmann 
1. Set the default app to oc10
2. forward to oc10
You can see the handler as an interceptor

JanAckermann: @pascalwengerter 
>I think we'll have to render a second type of list items here, depending on whether they're supposed to be routing (<a>) or trigger an async action (<button>)

It is still doing both, I don't think we need to change something.

JanAckermann:
@pascalwengerter AFAIK, this was decided by @pmaier1

ownclouders:
Results for <strong>oCISFiles3</strong> https://drone.owncloud.com/owncloud/web/22530/57/1
- webUIRenameFiles/renameFiles.feature:62

ownclouders:
Results for <strong>oCISSharingPublic1</strong> https://drone.owncloud.com/owncloud/web/22530/67/1
- webUISharingPublicBasic/publicLinkCreate.feature:190

pmaier1:
Let's move on together with https://github.com/owncloud/core/pull/39600.

ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/28173/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/28173/tracing/public-link-for-space-alice-2022-9-8-10-32-22.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/28173/tracing/public-link-for-space-brian-2022-9-8-10-32-50.zip 


</p></details>

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -57,6 +60,8 @@ export default {
     }
   },
   computed: {
+    ...mapGetters(['configuration', 'getToken']),
```
kulmann, line 63: Instead of `getToken` (already deprecated) please map the `runtime/auth` getter `accessToken`

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -66,6 +71,27 @@ export default {
...
+      // @TODO use id or similar
+      if (appEntry.iconMaterial === 'switch_ui') {
+        await this.setClassicUIDefault()
+      }
+    },
+    setClassicUIDefault() {
+      const endpoint = new URL(this.configuration.server || window.location.origin)
+      endpoint.pathname =
+        endpoint.pathname.replace(/\/$/, '') + '/index.php/apps/web/settings/default'
+      const headers = new Headers()
```
kulmann, line 86: instead of using fetch you could use `clientService.httpAuthenticated` which re-uses an axios instance being cached for the current access token.

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -66,6 +71,27 @@ export default {
       onHidden: () => this.$refs.menubutton.$el.focus(),
       onShown: () => this.$refs.menu.$el.querySelector('a:first-of-type').focus()
     })
+  },
+  methods: {
+    async clickApp(appEntry) {
+      // @TODO use id or similar
+      if (appEntry.iconMaterial === 'switch_ui') {
```
kulmann, line 78: as discussed, let's go for `appEntry.url?.endsWith('/apps/files')` for the time being. That seems to be more reliable than an icon prop for now...

packages/web-integration-oc10/lib/Application.php
```diff
@@ -33,4 +34,8 @@ class Application extends App {
 	public function __construct(array $urlParams = []) {
 		parent::__construct(Application::APPID, $urlParams);
 	}
+
+    public function boot(): void {
+        Util::addscript(Application::APPID, 'app');
```
kulmann, line 39:
```suggestion
        Util::addScript(Application::APPID, 'app');
```
don't know if that's required, but the function name is `addScript` with an uppercase `S`.

kulmann (approved):

Makefile.release
```diff
@@ -75,6 +75,7 @@ endif
 .PHONY: ocx-app-config
 ocx-app-config:
 	cp -R packages/web-integration-oc10/appinfo $(dist_dir)
+	cp -R packages/web-integration-oc10/js $(dist_dir)
```
kulmann, line 78:
```suggestion
	cp packages/web-integration-oc10/js/* $(dist_dir)/js
```

kulmann (changes requested):

Makefile.release
kulmann: ... because the `js` folder already exists from all the web js bundles.

kulmann (approved):

PR #6185, opened 2021-12-21
[deployment example] fix proxy config, so that the web sidecar will be used

wkloucek:
## Description
the mount path was changed in #6094, but points to a non-existent file. Therefore the builtin oC Web will be served by oCIS...


pascalwengerter (approved):
THX!

PR #6189, opened 2021-12-23
[tests-only] Use owncloudci/wait-for docker image and PHP 7.4

kiranparajuli589:
### Description<br>Update drone starlark to use owncloudci/wait-for and PHP7.4<br>### Related Issue:<br> - https://github.com/owncloud/QA/issues/707<br>_**Note:** This PR is created with a script. There may be some error which I will follow up later_<br>

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,5 +1,6 @@
-OC_CI_ALPINE = "owncloudci/alpine:latest"
-OC_CI_WAIT_FOR = "owncloudci/wait-for:latest"
+waitForEmailService(testConfig["emailNeeded"]) +
+                             fixPermissions(testConfig["phpVersion"], testConfig["federatedServerNeeded"], params["selUserNeeded"]) +
+                             waitForBrowserService(testConfig["browser"]) +
 OC_CI_NODEJS = "owncloudci/nodejs:14"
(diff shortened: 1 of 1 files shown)


pascalwengerter (changes requested):
Needs a bit of cleanup since we're defining the same variables twice

Also, we have multiple variants of `waitForServer`, e.g.

def waitForOwncloudService():
    return [{
        "name": "wait-for-owncloud-service",
        "image": OC_CI_WAIT_FOR,
        "commands": [
            "wait-for -it owncloud:80 -t 300",
        ],
    }]

pascalwengerter:
> Needs a bit of cleanup since we're defining the same variables twice
> 
> Also, we have multiple variants of `waitForServer`, e.g.
> 
> def waitForOwncloudService(): return [{ "name": "wait-for-owncloud-service", "image": OC_CI_WAIT_FOR, "commands": [ "wait-for -it owncloud:80 -t 300", ], }]

I know it's an automated PR, just figured it would be good to give feedback

phil-davis:
Not needed - this is not a usual o C10 app. The drone starlark is very different, and the wait-for has been done here.

PR #6216, opened 2022-01-05
[full-ci][tests-only] bump middleware version to v1.1.0

kiranparajuli589:
### Description
- bump middleware version to v1.1.0 which contains `filesForUpload` dir
- so remove `filesForUpload` dir from the web repo
- make use of `filesForUpload` from middleware in web acceptance tests
- use middleware in e2e tests too

### Related Issue
- https://github.com/owncloud/web/issues/6209

PR #6249, opened 2022-01-13
[full-ci][tests-only] revert web pr #6216

kiranparajuli589:
## Description
this pr reverts #6216


.drone.star
```diff
@@ -9,7 +9,7 @@ OC_CI_GOLANG = "owncloudci/golang:1.17"
 OC_CI_NODEJS = "owncloudci/nodejs:14"
 OC_CI_PHP = "owncloudci/php:7.4"
 OC_CI_WAIT_FOR = "owncloudci/wait-for:latest"
-OC_TESTING_MIDDLEWARE = "owncloud/owncloud-test-middleware:1.1.0"
+OC_TESTING_MIDDLEWARE = "owncloud/owncloud-test-middleware:1.0.0"
```
pascalwengerter, line 12: Why are we reverting the middleware version? 🤔

kiranparajuli589: fixed

phil-davis: we were going to just revert that whole PR, to make sure that we go back to where we were earlier, then tomorrow think about just bumping the middleware version used by the nightwatch acceptance test suite.

Anyway, if CI passes, then great.

phil-davis (approved):
LGTM, if CI passes. This should remove the middleware from e2e tests.

PR #6269, opened 2022-01-17
Replace filter-obj dep with lodash-es function

pascalwengerter:
Supersedes #6197 by replacing the dependency alltogether

### DIFF
--- a/packages/web-app-files/package.json
+++ b/packages/web-app-files/package.json
@@ -4,10 +4,11 @@
   "description": "ownCloud web files",
   "license": "AGPL-3.0",
   "dependencies": {
-    "@types/semver": "^7.3.8",
     "copy-to-clipboard": "^3.3.1",
-    "filter-obj": "^2.0.1",
     "semver": "^6.1.0",
     "vue2-dropzone": "^3.6.0"
+  },
+  "devDependencies": {
+    "@types/semver": "^7.3.8"
   }
 }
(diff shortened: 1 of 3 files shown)


kulmann (approved):
Nice spot 😎

pascalwengerter:
> Nice spot 😎 

Thanks for the quick review! Thinking about it, why is actions an object of objects and not an array of action objects to begin with? 🤔

ownclouders:
Results for <strong>oC10SharingAccept</strong> https://drone.owncloud.com/owncloud/web/21945/15/1
- webUISharingAcceptSharesToRoot/acceptShares.feature:108

kulmann:
> > Nice spot 😎
> 
> Thanks for the quick review! Thinking about it, why is actions an object of objects and not an array of action objects to begin with? 🤔

Because we reference the quick actions by their names in other parts of the source code. And the quick actions themselves are objects because they're simple key-value-data.

pascalwengerter:
> > > Nice spot sunglasses
> > 
> > 
> > Thanks for the quick review! Thinking about it, why is actions an object of objects and not an array of action objects to begin with? thinking
> 
> Because we reference the quick actions by their names in other parts of the source code. And the quick actions themselves are objects because they're simple key-value-data.

Yeah I took another look this morning, should be fine ;) an array would be nicer to filter (since the underlying key-value-data is pretty uniform) but then we loose the reference-by-name part...

kulmann:
> Yeah I took another look this morning, should be fine ;) an array would be nicer to filter (since the underlying key-value-data is pretty uniform) but then we loose the reference-by-name part...

You could use `Object.values()` and filter the resulting array.

Issue #6291, opened 2022-01-20
E2E Tracing in ci reports too much traces


fschade:
It looks like that the tracing (generation) is enabled by default in the ci.
It should only generate traces (by default) in cases where the ci fails.

See https://github.com/owncloud/web/pull/6288#issuecomment-1017413845 for reference , here we should only have 3 trace zips for the failing step.

Expected behavior:
* if a feature failed it, should create a github comment with the title of the feature and attach the zips
* it should not generate traces if no feature fails (default)
* it should not compose multiple feature traces into one comment (or add a headline to see which feature the trace belongs to)

ScharfViktor:
Okay, I'll try to figure it out.

fschade:
Thanks 🤟  @ScharfViktor, if you need some pairing/brainstorming ping me.

Issue #6260, opened 2022-01-14
Add mor context to the GH Tracing comment


fschade:
Now that we have tracing results posted to gh it would be helpful to see more context regarding the trace.

as an example if you have a look to https://github.com/owncloud/web/pull/6257#issuecomment-1012758661 its hard to understand:
* which feature is the trace for?
* which of the traces failed?
* what backend did fail?
* is it a failing comment or just info ([whit-tracing] gets posted even without any failing step)

pascalwengerter:
@fschade @ScharfViktor is this done / what's the definition of done here? :)

fschade:
this is done, thanks for having an eye on it 💋

PR #6297, opened 2022-01-21
Reduce notification to github 

ScharfViktor:
Step 1:
**reduce traces**

The following logic used to work:
- if the test fails and is retry=n, then trace is enabled. 
- Tracing records the next test run (is correct) but continues on the next tests as well

I changed this by adding retry counter. 
logic:
- first test run does not record trace
- after the test falls and Retry = n, in the "After" function enables trace recording for the next run
- trace records for the next run (regardless of whether test passes or falls)
- after all attempts, the recording is turned off

see block diagram: 

image

Step 2:
**add more details to github comments:**

our trace look like: `npx playwright show-trace https://cache.owncloud.com/owncloud/web/tracing/21580/2022-1-5-09-24-50-Alice.zip` 

- add feature name in drone.star is difficult. We know {DRONE_BUILD_LINK}/${DRONE_JOB_NUMBER}${DRONE_STAGE_NUMBER} but to fall in and get the featureName is too difficult
- we have another, easier way. add  featureName to zip file name. like: `2022-1-5-09-24-50-Kindergarten-can-use-web-to-organize-a-day-Alice.zip`. But the name would be too long. What do you think @fschade ?

 If easer way is okay, I need your help to add `this.feature.name` from `tests/e2e/cucumber/environment/index.ts` correctly to `tests/e2e/support/environment/actor/actor.ts `

### DIFF
--- a/tests/e2e/cucumber/environment/index.ts
+++ b/tests/e2e/cucumber/environment/index.ts
@@ -28,6 +28,7 @@ const logger = pino({
 })
 
 setDefaultTimeout(config.debug ? -1 : config.timeout * 1000)
+let retryCounter = 0
 
 BeforeAll(async function () {
   if (config.ocis) {
@@ -39,6 +40,7 @@ BeforeAll(async function () {
 
 Before(function (this: World, { pickle }: ITestCaseHookParameter) {
   this.feature = pickle
+  console.log('featureName', this.feature.name)
   this.actorsEnvironment.on('console', async (actorId, message): Promise<void> => {
     const msg = {
       actor: actorId,
@@ -93,15 +95,20 @@ After(async function (this: World, { result }: ITestCaseHookParameter) {
 
   await this.attach(`Status: ${result?.status}. Duration:${result.duration?.seconds}s`)
 
-  if (result.status !== Status.PASSED) {
+  if (result.status !== Status.PASSED && retryCounter < config.retry) {
     if (result.willBeRetried) {
       config.reportHar = true
       config.reportTracing = true
-    }
-
-    await this.actorsEnvironment.close()
+      retryCounter += 1
+    } 
+  } else {
+    config.reportHar = false
+    config.reportTracing = false
+    retryCounter = 0
   }
+  await this.actorsEnvironment.close()
 })
+
 
 AfterAll(() => state.browser && state.browser.close())
 
--- a/tests/e2e/cucumber/environment/index.ts
+++ b/tests/e2e/cucumber/environment/index.ts
@@ -100,7 +100,7 @@ After(async function (this: World, { result }: ITestCaseHookParameter) {
       config.reportHar = true
       config.reportTracing = true
       retryCounter += 1
-    } 
+    }
   } else {
     config.reportHar = false
     config.reportTracing = false
@@ -108,7 +108,6 @@ After(async function (this: World, { result }: ITestCaseHookParameter) {
   }
   await this.actorsEnvironment.close()
 })
-
 
 AfterAll(() => state.browser && state.browser.close())
 
--- a/tests/e2e/cucumber/environment/index.ts
+++ b/tests/e2e/cucumber/environment/index.ts
@@ -28,7 +28,6 @@ const logger = pino({
 })
 
 setDefaultTimeout(config.debug ? -1 : config.timeout * 1000)
-let retryCounter = 0
 
 BeforeAll(async function () {
   if (config.ocis) {
@@ -88,6 +87,11 @@ BeforeAll(async (): Promise<void> => {
   }[config.browser]()
 })
 
+const defaults = {
+  reportHar: config.reportHar,
+  reportTracing: config.reportTracing
+}
+
 After(async function (this: World, { result }: ITestCaseHookParameter) {
   if (!result) {
     return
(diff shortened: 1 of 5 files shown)


ScharfViktor:
regarding: https://github.com/owncloud/web/issues/6291
https://github.com/owncloud/web/issues/6260

ScharfViktor:
I forgot to mention one unfortunate thing:
When I drew the diagram I used `new draw.io document` in web
and when I was almost done drawing it, the page refreshed and everything disappeared (apparently 10 minutes had passed). It was a bummer. 
It's the difference when you test a product or work in it

tests/e2e/cucumber/environment/index.ts
```diff
@@ -39,6 +40,7 @@ BeforeAll(async function () {
 
 Before(function (this: World, { pickle }: ITestCaseHookParameter) {
   this.feature = pickle
+  console.log('featureName', this.feature.name)
```
pascalwengerter, line 43: Dev leftover? Or do you want to log it here?

ScharfViktor: No, I'll take it out. left it out since there were questions

tests/e2e/cucumber/environment/index.ts
```diff
@@ -93,14 +95,18 @@ After(async function (this: World, { result }: ITestCaseHookParameter) {
 
   await this.attach(`Status: ${result?.status}. Duration:${result.duration?.seconds}s`)
 
-  if (result.status !== Status.PASSED) {
+  if (result.status !== Status.PASSED && retryCounter < config.retry) {
     if (result.willBeRetried) {
       config.reportHar = true
       config.reportTracing = true
+      retryCounter += 1
```
pascalwengerter, line 102: Why not set it to 1 hardcoded?

tests/e2e/cucumber/environment/index.ts
pascalwengerter: Still needs to be removed?

tests/e2e/cucumber/environment/index.ts
ScharfViktor: then it won't work when RETRY=2 and more

tests/e2e/cucumber/environment/index.ts
ScharfViktor: removed

ScharfViktor:
@pascalwengerter can you please review PR?

tests/e2e/cucumber/environment/index.ts
```diff
@@ -15,8 +15,10 @@ import { api, store } from '../../support'
 import { World } from './world'
 import { state } from './shared'
 import { Browser, chromium, firefox, webkit } from 'playwright'
+import { camelCase } from 'lodash'
```
pascalwengerter, line 18:
```suggestion
import camelCase from 'lodash-es/camelCase'
```
Minor nitpick, but please lodash-es

tests/e2e/cucumber/environment/index.ts
```diff
@@ -39,6 +42,7 @@ BeforeAll(async function () {
 
 Before(function (this: World, { pickle }: ITestCaseHookParameter) {
   this.feature = pickle
+  featureName = camelCase(this.feature.name)
```
pascalwengerter, line 45:
```suggestion
const  featureName = camelCase(this.feature.name)
```
Why not use a const here?

pascalwengerter (changes requested):
Almost good to go!:)

tests/e2e/cucumber/environment/index.ts
ScharfViktor: Ok, I change it now

ScharfViktor: `import camelCase from 'lodash-es/camelCase'`  doesn't work for me. 
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/scharfviktor/Work/web/node_modules/lodash-es/camelCase.js from /Users/scharfviktor/Work/web/tests/e2e/cucumber/environment/index.ts not supported.
Instead change the require of camelCase.js in /Users/scharfviktor/Work/web/tests/e2e/cucumber/environment/index.ts to a dynamic import() which is available in all CommonJS modules.
```

tests/e2e/cucumber/environment/index.ts
ScharfViktor: I created above `export let featureName = ''` 
and then wrote the values into it. 
is that wrong?

pascalwengerter: Not necessary since you don't change the value later on, so removing the `export let featureName = ''` and changing line 45 to `const featureName = camelCase(this.feature.name)` should be fine
`let` is necessary because you change it to keep track of the retries ;)

tests/e2e/cucumber/environment/index.ts
pascalwengerter: Damn, nevermind then ;) just keep it as-is

tests/e2e/cucumber/environment/index.ts
ScharfViktor: I'm exporting featureName
And get featureName in actor.ts. That's what I asked you to help me with. Move correct featureName from index.ts to actor.ts

when i will remove `export let featureName = ''` then i don't get featureName in actor.ts

ScharfViktor: image

fschade:
@ScharfViktor sorry that it took that long! I refactored a bit and please have a look if it fits your needs, @pascalwengerter then can re-review it.

tests/e2e/cucumber/environment/index.ts
pascalwengerter: I can still see it, maybe Florian brought it back :D please remove and check his changes, happy to merge this today/tomorrow if it does what you want 🥳

PR #6298, opened 2022-01-21
Fix #6296: Make app config available to app again

### LINKED ISSUE #6296: Configuration unavailable within extensions
Web is loading the configs **before** registering the apps (which populates `state.fileEditors`). This means that the following code has no effect:

```js
  LOAD_EXTENSION_CONFIG(state, { id, config }) {
    // make available in editors
    const editors = [...state.fileEditors]
    for (const editor of editors) {
      if (editor.app === id) {
        editor.config = config
      }
    }
    state.fileEditors = editors
(...)
```
(https://github.com/owncloud/web/blob/master/packages/web-runtime/src/store/apps.js#L109)

Without this, there are no configurations available within the extension context.

dschmidt:
## Related Issue

- Fixes #6296 

## Motivation and Context
The config is loaded before the actual app, so when the old code tried to persist the setting inside the application object inside the store it was not available. That's why we now cache the config and copy it into the app object, once it becomes available. This should be a true drop in change and I don't expect any side effects as we did not change the internal structure of the object in the store.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-application-config
+++ b/changelog/unreleased/bugfix-application-config
@@ -1,0 +1,6 @@
+Bugfix: Application config not available to application
+
+We fixed a bug in providing config to external apps like draw-io.
+
+https://github.com/owncloud/web/issues/6296
+https://github.com/owncloud/web/pull/6298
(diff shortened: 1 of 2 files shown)


fschade (approved):
good catch ❤️  🤖

dschmidt:
kudos to @diocas for finding the issue :)

PR #6299, opened 2022-01-21
Fix publicLinkPassword in apps

dschmidt:
## Description
Removes a nonsense check from `publicLinkPassword` computed.
The code was just introduced and never released, so we don't need a changelog entry imho.

The second change is unrelated but I stumbled upon this while debugging why files could not be shown in mediaviewer...
It seemed to still work even with the double encoding, so I'm not even sure how to use it to break something, so again no changelog needed imho.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/composables/useAppDefaults/useAppFileHandling.ts
+++ b/packages/web-pkg/src/composables/useAppDefaults/useAppFileHandling.ts
@@ -51,7 +51,7 @@ export function useAppFileHandling(options: AppFileHandlingOptions): AppFileHand
 
     // Since the pre-signed url contains query parameters and the caller of this method
     // can also provide query parameters we have to combine them.
-    const combinedQuery = [queryStr, encodeURIComponent(signedQuery)].filter(Boolean).join('&')
+    const combinedQuery = [queryStr, signedQuery].filter(Boolean).join('&')
     return [url, combinedQuery].filter(Boolean).join('?')
   }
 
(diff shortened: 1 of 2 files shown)


ownclouders:
Results for <strong>oC10Files5</strong> https://drone.owncloud.com/owncloud/web/22092/20/1

packages/web-pkg/src/composables/useAppDefaults/useAppFileHandling.ts
```diff
@@ -51,7 +51,7 @@ export function useAppFileHandling(options: AppFileHandlingOptions): AppFileHand
 
     // Since the pre-signed url contains query parameters and the caller of this method
     // can also provide query parameters we have to combine them.
-    const combinedQuery = [queryStr, encodeURIComponent(signedQuery)].filter(Boolean).join('&')
+    const combinedQuery = [queryStr, signedQuery].filter(Boolean).join('&')
     return [url, combinedQuery].filter(Boolean).join('?')
```
fschade, line 55: can't we use `URL` to construct the url object?

fschade (changes requested):
i would prefer to use URL for construction if nothing speaks against it

fschade (approved):

PR #6316, opened 2022-01-26
[deployment-examples] fix basic auth and remove proxy restart

wkloucek:
## Description
fix basic auth escaping and remove proxy restart workaround (since no longer needed)

## Motivation and Context
sometimes only the proxy started on the continuously deployed server


fschade (approved):
🚀

Issue #6344, opened 2022-01-31
e2e tests. Add restore old version step


ScharfViktor:
TODO:
- [x] add steps "restore old version" to an existing scenario 
- [x] add new versionPage in`web/tests/e2e/support/page/` and move there step `Then "" should see that the resource "" has 1 version` https://github.com/owncloud/web/pull/6354
- [x] add a check that the file is restored. Check response (like https://github.com/owncloud/web/blob/master/tests/e2e/support/page/files/allFiles.ts#L238-L246), check size of file (like https://github.com/owncloud/web/blob/master/tests/e2e/cucumber/step_definitions/app.files.ts#L126)

Location step:
https://github.com/owncloud/web/blob/master/tests/e2e/cucumber/shareFolderJourney.feature#L47

- When "Alice" creates a new version of the following files
- Then "Alice" should see that the resource "folder_to_shared/simple.pdf" has 1 version
- **When "Brian" restores old version**

ScharfViktor:
@individual-it

ScharfViktor:
I thing that it is completed. Thank you

Issue #6361, opened 2022-02-03
Files details page does not update after file version is restored without refresh


grgprarup:
### Steps to reproduce
1. Create a file `version.txt` with content `version 1`
2. Upload a file `version.txt` with content `new version`
3. Restore the old version of file

### Expected behaviour
The UI must be updated with the information of the older version of file without refreshing the page

### Actual behaviour
Before restoring the older version, (file size of current version is 29B, modified 1 day ago on file details page, file size of older version is 19B, time 20 min on versions sidebar)
![version_before_restore](https://user-images.githubusercontent.com/8559121/152336830-8930492e-0370-4122-b5f4-de1d036cd5b6.png)

After restoring the older version, (file size of current version is 29B, modified 1 day ago on file details page is not updated, file size of older version is 29B, time 1 day on versions sidebar which is updated)
![version_after_restore](https://user-images.githubusercontent.com/8559121/152336839-93afc54c-e0fb-4571-9548-ed76c6bb6a56.png)

After refreshing the page, (file size of current version is 19B, modified 30 min ago on file details page is now updated)
![version_after_refresh](https://user-images.githubusercontent.com/8559121/152338046-1f99a565-d5a9-4169-92cf-85601d0739cf.png)

### Environment general
**Operating system**: Linux Ubuntu 20.04

### Environment ownCloud Core
**Web server:** latest master

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/7469

PR #6376, opened 2022-02-07
[tests-only] Added implementation for file old version restore, check file is restored

grgprarup:
## Description
This PR adds steps for file old version restore in feature file and implementation of steps to restore the old version of file. Also, confirms whether file is restored or not.

## Related Issue

- Part of https://github.com/owncloud/web/issues/6344

## How Has This Been Tested?

- locally
- ci

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] E2E tests added

### DIFF
--- a/tests/e2e/support/page/files/version.ts
+++ b/tests/e2e/support/page/files/version.ts
@@ -1,15 +1,26 @@
-import { Actor } from '../../types'
+import { Actor, File } from '../../types'
 import { filesCta } from '../../cta'
-import { Locator } from '@playwright/test'
+import { expect, Locator } from '@playwright/test'
+import path from 'path'
 
 export class VersionPage {
   private readonly actor: Actor
   private readonly fileVersionSidebarXPath: Locator
+  private readonly fileVersionSizeXPath: Locator
+  private readonly fileVersionRevertButtonXPath: Locator
+  private fileSizeXPath: Locator
+  private static olderVersionSize: string
 
   constructor({ actor }: { actor: Actor }) {
     this.actor = actor
     this.fileVersionSidebarXPath = this.actor.page.locator(
       '//*[@id="oc-file-versions-sidebar"]/table/tbody/tr'
+    )
+    this.fileVersionSizeXPath = this.actor.page.locator(
+      '//*[@data-testid="file-versions-file-size"]'
+    )
+    this.fileVersionRevertButtonXPath = this.actor.page.locator(
+      '//*[@data-testid="file-versions-revert-button"]'
     )
   }
 
@@ -27,5 +38,34 @@ export class VersionPage {
 
     const elements = this.fileVersionSidebarXPath
     return await elements.count()
+  }
+
+  async restoreOlderVersion({ files, folder }: { files: File[]; folder?: string }) {
+    const { page } = this.actor
+    const fileName = files.map((file) => path.basename(file.name))
+    await filesCta.navigateToFolder({ page: page, path: folder })
+    await filesCta.sidebar.open({ page: page, resource: fileName[0] })
+    await filesCta.sidebar.openPanel({ page: page, name: 'versions' })
+    const fileSize = await this.fileVersionSizeXPath.textContent()
+    VersionPage.olderVersionSize = fileSize.trim()
+    await Promise.all([
+      page.waitForResponse(
+        (resp) =>
+          resp.url().includes('/v/') && resp.status() === 204 && resp.request().method() === 'COPY'
+      ),
+      await this.fileVersionRevertButtonXPath.click()
+    ])
+    // Files details page does not update after restore button is clicked
+    // This is the issue: https://github.com/owncloud/web/issues/6361
+    await page.reload()
+  }
+
+  async checkOlderVersionRestored({ resource }: { resource: string }) {
+    const { page } = this.actor
+    this.fileSizeXPath = page.locator(
+      `//*[@data-test-resource-name="${resource}"]/ancestor::tr//*[@class="oc-resource-size"]`
+    )
+    const fileSize = await this.fileSizeXPath.textContent()
+    expect(fileSize.length).toBe(VersionPage.olderVersionSize.length)
   }
 }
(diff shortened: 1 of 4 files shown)


tests/e2e/support/page/files/version.ts
```diff
@@ -28,4 +39,33 @@ export class VersionPage {
     const elements = this.fileVersionSidebarXPath
     return await elements.count()
   }
+
+  async restoreOlderVersion({ files, folder }: { files: File[]; folder?: string }) {
+    const { page } = this.actor
+    const { dir: resourceDir, base: resourceBase } = path.parse(folder)
```
ScharfViktor, line 45: I'm a little confused as to why this is necessary?

tests/e2e/support/page/files/version.ts
```diff
@@ -28,4 +39,33 @@ export class VersionPage {
...
+    VersionPage.olderVersionSize = fileSize.trim()
+    await Promise.all([
+      page.waitForResponse(
+        (resp) =>
+          resp.url().includes('/v/') && resp.status() === 204 && resp.request().method() === 'COPY'
+      ),
+      await this.fileVersionRevertButtonXPath.click()
+    ])
+    // Files details page does not update after restore button is clicked
+    await page.reload()
```
ScharfViktor, line 60: I tried running the tests without it. with the backend of oCis and oC10 the tests passed

tests/e2e/support/page/files/version.ts
```diff
@@ -28,4 +39,33 @@ export class VersionPage {
...
+  }
+
+  async checkOlderVersionRestored({ resource }: { resource: string }) {
+    const { page } = this.actor
+    this.fileSizeXPath = this.actor.page.locator(
+      `//*[@data-test-resource-name="${resource}"]/ancestor::tr//*[@class="oc-resource-size"]`
+    )
+    const fileSize = await this.fileSizeXPath.textContent()
+    expect(fileSize.length).toBe(VersionPage.olderVersionSize.length)
```
ScharfViktor, line 69: Nice job) You compared: 
this
Screenshot 2022-02-07 at 22 10 03


with this:
Screenshot 2022-02-07 at 22 10 35

When we create new version, we add the same file (same length). Can you fix it please? 
Use 2 files of different lengths.

for example, change here:
```
When "Alice" creates a new version of the following files
      | resource                   | to               |
      | PARENT/simple.pdf | folder_to_shared |
```

ScharfViktor:
Please tell me how easy it is to checkout to your branch in the git?
image

SwikritiT:
> Please tell me how easy it is to checkout to your branch in the git? image

@grgprarup doesn't have a access to owncloud yet so you might need to add the @grgprarup fork  `git@github.com:grgprarup/web.git` to your remote and maybe fetch the branches and checkout.

tests/e2e/support/page/files/version.ts
grgprarup: Sorry, I forgot to remove that line.

tests/e2e/support/page/files/version.ts
grgprarup: I used the page reload because there is issue on web ( https://github.com/owncloud/web/issues/6361 ), as we are comparing size from UI reload is necessary. May be the test have been passed for you because the size was same on both versions.

tests/e2e/support/page/files/version.ts
grgprarup: Fixed on the recent commit.

tests/e2e/support/page/files/version.ts
individual-it: That is dangerous if we just make this pass, it will cover up the bug.
But I think we also don't want to introduce something like the expected to fail file here because then we would have to mark the whole user-story as "expected to fail" and by that miss all the other checks it does.

@ScharfViktor what to do?

individual-it: Options that I can see:
- have tests that actually don't show the expected behavior but the current behavior, see background [in oc docs](https://doc.owncloud.com/server/10.9/developer_manual/testing/acceptance-tests.html#writing-scenarios-for-bugs) or [this blogpost](https://dev.to/jankaritech/should-you-write-acceptance-tests-for-bugs-2bfm)
- skip any action that has bugs in the e2e tests and first create a unit-test that shows the issue

phil-davis: In the general case, when a user story has steps that currently do not pass because the system-under-test has a bug, then what are we going to do. We need to decide that in general. @individual-it gave some options above. Here are some more that I can think of:

- put the failing step(s) at the end of the user story (as far as possible, make it so that all the other good steps do execute before the failure happens, and
- have something that records the actual step that fails, and the CI can parse the test result and validate that the test really fails at the "expected" test step

Then if a change happens to the system-under-test, and the user story starts passing, or fails earlier (regression), or fails later (the direct bug is fixed but some later step is still a problem) then CI will fail and the developer will be aware of a change in behavior, and can look to see if the behavior change is a regression or improvement.

If the future direction is that the nightwatch test suite is going to reduce (or go away), and the Playwright e2e test suite is going to become the way that we provide some broad "QA anti-regression acceptance end-2-end (insert your favorite word for these kind of tests)" test coverage that helps everyone (QA, developers, users...) have confidence that the functionality is at least as good as previously, and that the functionality is improving, then we need to decide what to do when there are tests that cover functionality that currently does not 100% meet the requirements.

ScharfViktor: I have no quick ideas on how to solve this problem. I created a issue https://github.com/owncloud/web/issues/6385 to don't lock this PR

ScharfViktor (approved):

individual-it:
@grgprarup please add the link to the issue in the comment in https://github.com/owncloud/web/pull/6376/files#diff-a4da461227e06bce50aebd66771ed80de0b9e540dbf548b725d9c477f43855b6R58 and squash the commits and then we can merge it for now

PR #6381, opened 2022-02-07
Ignore routes whitelist and reduce console noise

kulmann:
## Description
This PR kills the `routes` key from the editor app manifests. It has only ever been used as a weird way to disable editors in the trashbin view. Since that logic belongs in the files app, not in the editors, I've converted this to a hardcoded check in the fileActions mixin (all the file actions do hardcoded route checks anyways). This gets us rid of the hassle to update a route whitelist in all applications out there (and internally). Which is the case for the upcoming spaces feature already (i.e. no apps being available in spaces folder listings, which is unnecessary effort).

Also used the chance to get rid of some console noise, namely:
- vue router deprecation warnings of `addRoutes` (use `addRoute` instead)
- warnings about a non-existing `files-spaces-storage` route name (deleted as not needed anymore)
- warning about vuex `mapGetter` usage for the `External/mimeTypes` getter, which only exists if the `external` app is loaded. Which is not always the case.

## Motivation and Context
Improve developer experience.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/AppBar/AppBar.vue
+++ b/packages/web-app-files/src/components/AppBar/AppBar.vue
@@ -134,7 +134,6 @@
 <script>
 import { mapActions, mapGetters, mapState, mapMutations } from 'vuex'
 import pathUtil from 'path'
-import get from 'lodash-es/get'
 
 import Mixins from '../../mixins'
 import MixinFileActions, { EDITOR_MODE_CREATE } from '../../mixins/fileActions'
@@ -179,7 +178,6 @@ export default {
     fileFolderCreationLoading: false
   }),
   computed: {
-    ...mapGetters('External', ['mimeTypes']),
     ...mapGetters([
       'getToken',
       'capabilities',
@@ -192,10 +190,12 @@ export default {
     ...mapState('Files', ['areHiddenFilesShown']),
 
     mimetypesAllowedForCreation() {
-      if (!get(this, 'mimeTypes', []).length) {
+      // we can't use `mapGetters` here because the External app doesn't exist in all deployments
+      const mimeTypes = this.$store.getters['External/mimeTypes']
+      if (!mimeTypes) {
         return []
       }
-      return this.mimeTypes.filter((mimetype) => mimetype.allow_creation) || []
+      return mimeTypes.filter((mimetype) => mimetype.allow_creation) || []
     },
     newButtonTooltip() {
       if (!this.canUpload) {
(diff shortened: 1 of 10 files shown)


ownclouders:
Results for <strong>oCISSharingAndUpload</strong> https://drone.owncloud.com/owncloud/web/22513/66/1
- webUIUpload/upload.feature:94

pascalwengerter (approved):
Had to push another case of mimeTypes being an empty object (but somehow not being empty enough for https://docs-lodash.com/v4/is-empty/, IDK) to avoid further console errors, but otherwise LGTM and very nice improvement!

packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -192,10 +191,12 @@ export default {
     ...mapState('Files', ['areHiddenFilesShown']),
 
     mimetypesAllowedForCreation() {
-      if (!get(this, 'mimeTypes', []).length) {
+      // we can't use `mapGetters` here because the External app doesn't exist in all deployments
+      const mimeTypes = this.$store.getters['External/mimeTypes']
+      if (!mimeTypes || mimeTypes !== {}) {
```
dschmidt, line 195: Isn't this *always* true as `{}` always instantiates a new object which is never the same as `mimeTypes`?

dschmidt (changes requested):
Nice! I was planning on giving this a close look myself, as I always assumed this was a weird workaround for not having to make apps work in all contexts. That's I wanted to look at this after further improvements for the useAppDefaults composable.

Please check the `!== {}` condition I've commented on.

packages/web-app-files/src/components/AppBar/AppBar.vue
pascalwengerter: Great catch, apparently one should go for sth like this https://flaviocopes.com/javascript-check-empty-object/

Turns out we  (=past me) were initializing the `mimeTypes` into an empty object and then overwriting it as an array, fixing that by turning it into an array from the get-go so we can drop this line :)

dschmidt (approved):

ownclouders:
Results for <strong>oCISSharingInternal2</strong> https://drone.owncloud.com/owncloud/web/22536/60/1
- webUISharingInternalUsers/shareWithUsers.feature:342

PR #6417, opened 2022-02-10
Support nested query params in router

dschmidt:
## Description
I had to use file paths for importing `qs` because for some reason `import * from 'qs'` gave me `querystring`. I have no idea how to debug this and would love to hand this over to @fschade . Any ideas?

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Get rid of query-string lib, as we require the more powerful qs library as a peer dependency of owncloud sdk already.
Simplify implementation of `convertRouteParamsToContextQuery` and `convertContextQueryToRouteParams`.
Makes it also easy to switch between [] and . syntax for nested query params

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-draw-io/src/App.vue
+++ b/packages/web-app-draw-io/src/App.vue
@@ -24,7 +24,7 @@
 <script>
 import { mapGetters, mapActions } from 'vuex'
 import { basename } from 'path'
-import qs from 'node_modules/qs/dist/qs.js'
+import qs from 'qs'
 import { DateTime } from 'luxon'
 import { DavPermission, DavProperty } from 'web-pkg/src/constants'
 import { useAppDefaults } from 'web-pkg/src/composables'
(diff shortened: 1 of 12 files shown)


ownclouders:
Results for <strong>oC10SharingPermission</strong> https://drone.owncloud.com/owncloud/web/22697/34/1
- webUISharingPermissionsUsers/sharePermissionsUsers.feature:52

dschmidt:
Now with v5.0.0 released, we need to check whether `qs` still parses `contextRouteParams[...]` even if `allowDots` is `true`, if not this would break links :-\

kulmann:
> Now with v5.0.0 released, we need to check whether `qs` still parses `contextRouteParams[...]` even if `allowDots` is `true`, if not this would break links :-\

Argh. 😞 Could you check?

dschmidt:
https://user-images.githubusercontent.com/448487/154233414-2064f90b-14f7-47b7-8c01-14859761007a.mp4

It does not only work, it even fixes the urls in place - very nice!

So from my POV this is good to merge :)

PR #6287, opened 2022-01-20
Add space front page & bump SDK to `v2.0.0`

JanAckermann:
## Description

* Implemented the front page for spaces
* Use the SDK v2 

To create an image and a readme for a space, you have to do a few things. First, create a yaml file locally with the following content:

```
---
version: "1.0"
description: fooooo ooo
special:
  readme: readme.md
  image: space.png
```

The paths to the readme and the image are relative to the root directory of the space.
Then, run the following curls. Note that you need basic auth enabled in ocis (`PROXY_ENABLE_BASIC_AUTH=true`):

```
curl -k -X MKCOL 'https://localhost:9200/dav/spaces/fe1f3db7-d7bf-4695-b222-91875c79b2e7/.config' -u admin:admin
curl -k -X MKCOL 'https://localhost:9200/dav/spaces/fe1f3db7-d7bf-4695-b222-91875c79b2e7/.config/ocis' -u admin:admin
curl -k -X PUT 'https://localhost:9200/dav/spaces/fe1f3db7-d7bf-4695-b222-91875c79b2e7/.config/ocis/space.yaml' -u admin:admin -v -T space.yaml
```

## Expected failures:

* ~Changing the description of a space from the spaces list does not work~
* Sharing within a space does not work (maybe it shouldn't, but then we would need to hide the actions)
* Some of the breadcrumb actions on a space itself need to be removed/hidden as they make no sense (e.g. move, copy)
* The breadcrumbs currently show the space ID, not the space name
* The space ID in the breadcrumb cannot be clicked when you are on the front page of the space (which would normally trigger a reload)
* Thumbnails do not work within a space

## Related Issue

- https://github.com/owncloud/web/issues/6271
- Fixes #6146 (at least the web parts of it)

## Screenshots:

image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #6449, opened 2022-02-17
Fix file creation

diocas:
When #6287 got merged, you forgot to update `addAppProviderFile`, so creating files via app provider would fail when stating.
edit: the normal creation of files was also broken in public links

### DIFF
--- a/packages/web-app-files/src/components/AppBar/AppBar.vue
+++ b/packages/web-app-files/src/components/AppBar/AppBar.vue
@@ -601,15 +601,20 @@ export default {
           throw new Error(`An error has occurred: ${response.status}`)
         }
 
-        const path = pathUtil.join(this.currentPath, fileName)
         let resource
+        let path = pathUtil.join(this.currentPath, fileName)
+
         if (this.isPersonalLocation) {
+          path = buildWebDavFilesPath(this.user.id, path)
+          resource = await this.$client.files.fileInfo(path, DavProperties.Default)
+        } else if (this.isSpacesProjectLocation) {
+          path = buildWebDavSpacesPath(this.$route.params.spaceId, path)
           resource = await this.$client.files.fileInfo(path, DavProperties.Default)
         } else {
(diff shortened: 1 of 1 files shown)


packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -601,15 +601,20 @@ export default {
...
+          resource = await this.$client.files.fileInfo(path, DavProperties.Default)
+        } else if (this.isSpacesProjectLocation) {
+          path = buildWebDavSpacesPath(this.$route.params.spaceId, path)
           resource = await this.$client.files.fileInfo(path, DavProperties.Default)
         } else {
           resource = await this.$client.publicFiles.getFileInfo(
             path,
             this.publicLinkPassword,
-            DavProperties.PublicLink
+            DavProperties.Default
```
kulmann, line 617: Why not the public link dav properties? Those are all `Default` properties and some additional ones. Are they not needed?

diocas: It's what you use everywhere. And by looking at the properties it looks to me that is the correct one.

diocas:
I've just realized that the creation of files (non app provider) is also broken.. I'll add the fix in this PR.

packages/web-app-files/src/components/AppBar/AppBar.vue
kulmann: Well, no, for public files we usually use `DavProperties.PublicLink`. Looks to me like we should keep it that way, could you revert the line?

diocas:
There.. The public links never get that much love xD

packages/web-app-files/src/components/AppBar/AppBar.vue
diocas: You are correct! I took the opportunity to change it in the other places where it was requesting the default properties as well.

PR #6445, opened 2022-02-16
Fix search context menu

dschmidt:
## Description
Enable/disable certain file actions outside of the file application.

## Related Issue

- Fixes #6293 

## Motivation and Context
Several file actions (currently) work or don't work outside the files app. In general anything that just interacts with the backend works fine, actions that rely on e.g., the sidebar, don't work in the search app for instance.

## How Has This Been Tested?

- needs oC 10

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
Decide whether we want to make the search results selectable in the ResourcesTable, to prepare for multi select and archive download.
Currently we have no search in oCIS and `isDownloadAsArchiveAvailable` is false on oC 10, so currently it's probably just confusing for no benefit... tying this to `isDownloadAsArchiveAvailable` would be possible, but feels weird as well ... happy to hear other opinions :) 

Adding a check whether

PR #6453, opened 2022-02-18
Make fileActions mixin standalone again.

dschmidt:
## Description
In #6445 we introduced a dependency on isFilesAppActive for the fileActions mixin. We added it to
ContextActions and BatchActions, but there are way more places where the mixin is used. That's why
we add a computed to the mixin itself now. Because mixins belong to the options api we unfortunately
cannot use composables directly and extract helper functions for that reason.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-app-files/src/components/AppBar/SelectedResources/BatchActions.vue
```diff
@@ -44,6 +39,11 @@ export default {
     Move,
     Restore
   ],
+  setup() {
+    return {
+      isFilesAppActive: useIsFilesAppActive()
```
kulmann, line 44: I'm confused, can't you delete this again with the helper function now being used in the mixins?

packages/web-app-files/src/mixins/fileActions.js
```diff
@@ -55,6 +58,12 @@ export default {
     ...mapGetters('Files', ['highlightedFile', 'currentFolder']),
     ...mapGetters(['capabilities', 'configuration']),
 
+    // TODO: Replace this with the useIsFilesAppActive composable
+    // once we port this mixin to something using the composition api
+    isFilesAppActive() {
```
kulmann, line 63: Oh wow, is it really sufficient to provide the function here? Are we using the `fileActions` mixin really in all locations that use any of the specific file action mixins?

dschmidt: That's why the composable is still used in BatchActions ... 
We include single actions there and not the fileActions mixin.

... but that's a very good point, this has become hard to reason about. I will add this to all the mixins :(

ownclouders:
Results for <strong>oCISSharingInternal2</strong> https://drone.owncloud.com/owncloud/web/22895/60/1
- webUISharingInternalUsers/shareWithUsers.feature:342

kulmann (approved):
👍

PR #6466, opened 2022-02-22
fix changelog

fschade:
in the Changelog for VUE history mode i mentioned the wrong base-tag attribute.
This fixes it.


fschade:
@diocas, please have a look if i missed something. Thanks :)

pascalwengerter (approved):

Issue #6097, opened 2021-12-07
webUISharingAcceptSharesToRoot/acceptShares.feature:108 fails intermittently


phil-davis:
https://drone.owncloud.com/owncloud/web/20687/16/20
```
  Scenario: unshare auto-accepted shares # features/webUISharingAcceptSharesToRoot/acceptShares.feature:108
- Connecting to selenium on port 4444...

ℹ Connected to selenium on port 4444 (237ms).
  Using: chrome (94.0.4606.61) on Linux platform.

    Given these users have been created with default attributes and without skeleton files:
      │ username │
      │ Alice    │
      │ Brian    │
    And user "Brian" has logged in using the webUI
√ Element <#user> was visible after 765 milliseconds.
√ Element <#user> was not present after 9 milliseconds.
√ Element <button[type=submit]> was visible after 33 milliseconds.
√ Element <button[type=submit]> was not present after 32 milliseconds.
√ Element <.oc-app-container> was visible after 763 milliseconds.
    Given the setting "shareapi_auto_accept_share" of app "core" has been set to "yes"
    And user "Alice" has created folder "/simple-folder"
    And user "Alice" has uploaded file "testavatar.jpg" to "/testimage.jpg"
    And these groups have been created:
      │ groupname │
      │ grp1      │
    And user "Brian" has been added to group "grp1"
    And user "Alice" has shared folder "/simple-folder" with group "grp1"
    And user "Alice" has shared folder "/testimage.jpg" with user "Brian"
    And the user has browsed to the shared-with-me page
√ Element <//*[self::table[contains(@class, "files-table")] or self::div[contains(@class, "files-empty")] or self::div[contains(@class, "files-not-found")]]> was present after 362 milliseconds.
    When the user unshares folder "simple-folder" using the webUI
   NoSuchElementError: An error occurred while running .isVisible() command on <//*[@id="files-sidebar"]>: 
   {"sessionId":"15ecc616aaeabf6c75f62aaa5cc68669","status":0,"value":[]}
       at Proxy.isSideBarOpen (/var/www/owncloud/web/tests/acceptance/pageObjects/FilesPageElement/appSideBar.js:30:18)
       at Proxy.isSideBarOpenForResource (/var/www/owncloud/web/tests/acceptance/pageObjects/FilesPageElement/appSideBar.js:43:24)
       at Proxy.openSideBar (/var/www/owncloud/web/tests/acceptance/pageObjects/FilesPageElement/filesList.js:232:28)
[...]

PR #6112, opened 2021-12-08
[not-for-merge][tests-only] Run flaky test (webUISharingAcceptSharesToRoot/acceptShares.feature:108) more often

kiranparajuli589:
### Description
Run flaky scenario mentioned in the issue below is run more often in the CI

### Related
https://github.com/owncloud/web/issues/6097

> This PR is just created for demonstration purpose

PR #6483, opened 2022-02-25
[test-only] check in shared with me page first after unsharing

kiranparajuli589:
## Description
after unsharing resources, with this pr:
- check for the resources state in the shared with me declined view
- then only browse to the files page and check for the presence of the unshared resource
with this order, the tests are passing solid. see demonstration at https://github.com/owncloud/web/pull/6112

## Related Issue
- Fixes https://github.com/owncloud/web/issues/6097


pascalwengerter (approved):
Fine with me, glad if this makes CI more stable 🙏🏽

phil-davis (approved):
OK - that means that there is some timing thing with the regular files list getting updated/refreshed and the declined  received shares being remove from the list. But it was intermittent, so the entries were being removed, but there must have been just a timing thing sometimes about when the test looked.

This re-ordering is fine - I don't think that we are hiding any real bug in the web app.

PR #6531, opened 2022-03-04
Implement the spaces permission concept

JammingBen:
## Description
We've implemented the spaces permission concept and improved the code structure for further permission changes.

Note that this PR does not yet include the permission check for the "Create New Space"-button. The permission for this is bound to the current user role, which needs some more in-depth work.

## Related Issue

- Related to https://github.com/owncloud/web/issues/6283

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #6545, opened 2022-03-07
Update the stored space object after members have been changed

JammingBen:
## Description
We now update the stored space after its members have been changed. Also, the permission-object of a built space has been simplified in the course of this.

Currently the only way to update the space object is by fetching the space again because the response from the OCS-API gives us no usable information. As soon as UUIDs have been introduced to the OCS-API, we could use those instead.

Note that there is still a caching bug in OCIS. E.g. when navigating between space's share panels via the share indicator, and adding/removing members, and then switching back and forth between the share panels again, the change seems to be lost. A reload reveals that this is actually not the case.

## Related Issue
- Groundwork for https://github.com/owncloud/web/pull/6531

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/helpers/resources.js
+++ b/packages/web-app-files/src/helpers/resources.js
@@ -111,7 +111,10 @@ export function buildResource(resource) {
 }
 
 export function buildSpace(space) {
-  let spaceImageData, spaceReadmeData, spacePermissions
+  let spaceImageData, spaceReadmeData
+  let spaceViewers = []
+  let spaceEditors = []
+  let spaceManagers = []
   let disabled = false
 
   if (space.special) {
@@ -120,7 +123,19 @@ export function buildSpace(space) {
   }
 
   if (space.root) {
-    spacePermissions = space.root.permissions
+    for (const permission of space.root.permissions) {
+      if (permission.roles.includes(spaceRoleViewer.name)) {
+        spaceViewers = spaceViewers.concat(permission.grantedTo.map((el) => el.user.id))
+      }
+
+      if (permission.roles.includes(spaceRoleEditor.name)) {
+        spaceEditors = spaceEditors.concat(permission.grantedTo.map((el) => el.user.id))
+      }
+
+      if (permission.roles.includes(spaceRoleManager.name)) {
+        spaceManagers = spaceManagers.concat(permission.grantedTo.map((el) => el.user.id))
+      }
+    }
 
     if (space.root.deleted) {
       disabled = space.root.deleted?.state === 'trashed'
@@ -154,7 +169,11 @@ export function buildSpace(space) {
     ownerId: '',
     disabled,
     spaceQuota: space.quota,
-    spacePermissions,
+    spaceRoles: {
+      viewers: spaceViewers,
+      editors: spaceEditors,
+      managers: spaceManagers
+    },
     spaceImageData,
     spaceReadmeData,
     canUpload: function () {
--- a/packages/web-app-files/src/helpers/resources.js
+++ b/packages/web-app-files/src/helpers/resources.js
@@ -9,6 +9,7 @@ import {
   SharePermissions,
   ShareStatus,
   ShareTypes,
+  SpacePeopleShareRoles,
   spaceRoleEditor,
   spaceRoleManager,
   spaceRoleViewer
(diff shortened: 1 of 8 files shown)


ownclouders:
Results for <strong>oCISFiles1</strong> https://drone.owncloud.com/owncloud/web/23351/55/1
- webUIDeleteFilesFolders/deleteFilesFolders.feature:59

ownclouders:
Results for <strong>oCISTrashbinUploadMoveJourney</strong> https://drone.owncloud.com/owncloud/web/23351/69/1
- webUITrashbinRestore/trashbinRestore.feature:84

packages/web-app-files/src/helpers/resources.js
```diff
@@ -154,7 +169,11 @@ export function buildSpace(space) {
     ownerId: '',
     disabled,
     spaceQuota: space.quota,
-    spacePermissions,
+    spaceRoles: {
```
kulmann, line 172: This introduces knowledge about the available space sharing roles by hardcoding their names here. You have everything at hand to build this `spaceRoles` map from looping over the result of `SpacePeopleShareRoles.list()`

packages/web-app-files/src/store/actions.js
```diff
@@ -157,18 +164,30 @@ export default {
     if (space) {
       const promises = []
       const spaceShares = []
+      const userRoles = [
```
kulmann, line 167: Same here, can be collected dynamically by iterating over the `space.spaceRoles` entries.

kulmann (changes requested):

packages/web-app-files/src/helpers/resources.js
JammingBen: 👍 

Not a 100% sure if I did it the most convenient way though...

packages/web-app-files/src/helpers/resources.js
```diff
@@ -111,16 +112,39 @@ export function buildResource(resource) {
 }
 
 export function buildSpace(space) {
-  let spaceImageData, spaceReadmeData, spacePermissions
+  let spaceImageData, spaceReadmeData
   let disabled = false
 
+  const spaceRoles = {}
```
kulmann, line 118: from array to map is easiest with `reduce`.

packages/web-app-files/src/helpers/resources.js
```diff
@@ -111,16 +112,39 @@ export function buildResource(resource) {
...
+
   if (space.special) {
     spaceImageData = space.special.find((el) => el.specialFolder.name === 'image')
     spaceReadmeData = space.special.find((el) => el.specialFolder.name === 'readme')
   }
 
   if (space.root) {
-    spacePermissions = space.root.permissions
+    for (const permission of space.root.permissions) {
+      if (permission.roles.includes(spaceRoleViewer.name)) {
```
kulmann, line 130: Please loop over the space roles, otherwise you'll still have hardcoded references of the respective roles.

kulmann (changes requested):

packages/web-app-files/src/helpers/resources.js
```diff
@@ -111,16 +112,28 @@ export function buildResource(resource) {
...
     spaceImageData = space.special.find((el) => el.specialFolder.name === 'image')
     spaceReadmeData = space.special.find((el) => el.specialFolder.name === 'readme')
   }
 
   if (space.root) {
-    spacePermissions = space.root.permissions
+    for (const permission of space.root.permissions) {
+      for (const role of SpacePeopleShareRoles.list()) {
+        if (permission.roles.includes(role.name)) {
+          spaceRoles[role.name] = spaceRoles[role.name].concat(
```
kulmann, line 131:
```js
spaceRoles[role.name].push(...permission.grantedTo.map((el) => el.user.id))
```
should do the trick in one line

kulmann: (or maybe it also breaks into 3 🙈 )

JammingBen: Nope, works fine 🙂

ownclouders:
Results for <strong>oCISSharingAndUpload</strong> https://drone.owncloud.com/owncloud/web/23365/66/1
- webUIUpload/upload.feature:94

PR #6579, opened 2022-03-11
Federated and lightweight accounts

elizavetaRa:
## Description
Implementation of the interface for federated and lightweight accounts at CERNBox

## Motivation and Context
At CERNBox users with federated and lightweight accounts don't have their own quota. It's possible to share with them as well as include them in projects. Based on these capabilities we implemented following elements:

- Home view that appears instead of the resources table if the account is federated or lightweight and the user is in his home directrory (functions isLightweight & isHomeRoute-> CERNBox only). 
- Navitems are changing if the account is federated or lightweight: the user doesn't see "Shared with others", "Shared by link", "Trashbin"

## Screenshots:
Screenshot from 2022-03-11 10-45-27

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] the isHomeRoute function needs reimplementation. Currently it's based on the url part like "/eos/user/r/ragozina/" and checks if the userid matches the loggedin user. 
- [ ] We also want to use "home" route instead of "files-personal" as there can't be any files in home directory of f.&l. users. But we didn't find a way to dynamically change home route, so we are currently showing home inside personal view instead

### DIFF
--- a/packages/web-app-files/src/views/Personal.vue
+++ b/packages/web-app-files/src/views/Personal.vue
@@ -1,74 +1,77 @@
 <template>
   <div>
-    <app-bar
-      :has-bulk-actions="true"
-      :breadcrumbs="breadcrumbs"
-      :breadcrumbs-context-actions-items="[currentFolder]"
-      :show-actions-on-selection="true"
-    >
-      <template #actions>
-        <create-and-upload />
+    <div v-if="isLightweight && isHomeRoute"><home /></div>
+    <div v-else>
+      <app-bar
+        :has-bulk-actions="true"
+        :breadcrumbs="breadcrumbs"
+        :breadcrumbs-context-actions-items="[currentFolder]"
+        :show-actions-on-selection="true"
+      >
+        <template #actions>
+          <create-and-upload />
+        </template>
+      </app-bar>
+      <app-loading-spinner v-if="loadResourcesTask.isRunning" />
+      <template v-else>
+        <not-found-message v-if="folderNotFound" class="files-not-found oc-height-1-1" />
+        <no-content-message
+          v-else-if="isEmpty"
+          id="files-personal-empty"
+          class="files-empty"
+          icon="folder"
+        >
+          <template #message>
+            <span v-translate>There are no resources in this folder</span>
+          </template>
+          <template #callToAction>
+            <span v-translate>
+              Drag files and folders here or use the "New" or "Upload" buttons to add files
+            </span>
+          </template>
+        </no-content-message>
+        <resource-table
+          v-else
+          id="files-personal-table"
+          v-model="selectedResources"
+          class="files-table"
+          :class="{ 'files-table-squashed': !sidebarClosed }"
+          :are-thumbnails-displayed="displayThumbnails"
+          :resources="paginatedResources"
+          :target-route="resourceTargetLocation"
+          :header-position="fileListHeaderY"
+          :drag-drop="true"
+          :sort-by="sortBy"
+          :sort-dir="sortDir"
+          @fileDropped="fileDropped"
+          @fileClick="$_fileActions_triggerDefaultAction"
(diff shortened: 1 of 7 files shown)


diocas:
@elizavetaRa can you put this PR as draft?
The idea of this PR, as discussed last week, is to understand if the functionality can be merged. This is not final code and it's very hardcoded, and we can/will improve it if there's a path to merger.

elizavetaRa:
@pascalwengerter @fschade could you comment on this pr? Anything that could be improved?

elizavetaRa:
@pascalwengerter @fschade sorry for repeating, could you give some comments here

packages/web-app-files/src/views/Personal.vue
```diff
@@ -149,6 +154,16 @@ export default defineComponent({
     ]),
     ...mapGetters(['user', 'homeFolder', 'configuration']),
 
+    isLightweight() {
+      const windowVue = window.Vue as any
+      return windowVue.$store.getters.user.usertype === 'lightweight'
```
pascalwengerter, line 159:
```suggestion
      return this.user.usertype === 'lightweight'
```
Should be enough? 🤔

packages/web-app-files/src/views/Personal.vue
```diff
@@ -149,6 +154,16 @@ export default defineComponent({
...
+    },
+    isHomeRoute() {
+      const windowVue = window.Vue as any
+      return this.$route.fullPath.includes(
+        `/${windowVue.$store.getters.user.id.charAt(0)}/${windowVue.$store.getters.user.id}`
```
pascalwengerter, line 164:
```suggestion
        `/${this.user.user.id.charAt(0)}/${this.user.id}`
```
Should be enough?

packages/web-app-files/src/views/Personal.vue
```diff
@@ -1,74 +1,77 @@
...
-    </app-bar>
-    <app-loading-spinner v-if="loadResourcesTask.isRunning" />
-    <template v-else>
-      <not-found-message v-if="folderNotFound" class="files-not-found oc-height-1-1" />
-      <no-content-message
-        v-else-if="isEmpty"
-        id="files-personal-empty"
-        class="files-empty"
-        icon="folder"
+    <div v-if="isLightweight && isHomeRoute"><home /></div>
```
pascalwengerter, line 3: Why render the `Home` view into the `Personal` view and not just make it a dedicated view inkl. route?

pascalwengerter (changes requested):
Sorry just took a quick look with some minor questions. 

Is there any "quick" way of running this code with our setup? Building from your branch is easy, but is there a dockerized CERN backend we can use with a local `dist` folder?

packages/web-app-files/src/views/Personal.vue
elizavetaRa: This is a little bit of a workaround. I don't find a way to remove "Personal" nav item as there should be at least one in export in web/packages/web-app-files/index.js. And this export happens before the user information arrives. Any better ideas on that? I would otherwise love to have an extra route for lightweight personal view

elizavetaRa:
@pascalwengerter unfortunately currently there is no possibility to test with dockerized CERN backend for you. You can test of course by creating a lightweight account in https://new.cernbox.cern.ch/ by signing in with some social network. 

Question about another workaround. So as we have the url to access user's home dir like "/home/eos/user/r/ragozina" but the user (also lightweight) is able to go to other users' directories, I implemented isHomeRoute() in Personal to check if user is in his own dir. It's based on url pattern letter/username but this can of course match some other urls inside other users' dirs. Do you have any better suggestion here as well?

PR #6593, opened 2022-03-14
[full-ci] Sharing jail

kulmann:
Switches over to the new webdav endpoints for a) personal files, i.e. personal space and b) shares, i.e. share jail.

To be tested with oCIS from https://github.com/owncloud/ocis/pull/3626 (or later on with master when it's merged) because of a new capability that enables the share jail.

## needs clarification
- [x] propfind to any share only ever loads content of one and the same share ("first one"?). can't manage to get files of another share in propfind response. => turns out this was a caching issue. resolved in https://github.com/cs3org/reva/pull/2784
- [x] preview request broken in shares => no thumbnails or right-sidebar-previews (also broken for `preview` app) => see https://github.com/owncloud/ocis/issues/3596 for details
- [x] additional capability. sub-capability of the `spaces` capability might be a good idea, see https://github.com/owncloud/ocis/issues/3615 for discussion, https://github.com/owncloud/ocis/pull/3626 for implementation

## open tasks
- [x] finish resource loader
- [x] hack webDavPath
- [x] breadcrumbs for new `SharedResource` view/route
- [x] fix file/folder creation inside share
- [x] make files app file actions available in `SharedResource` route/view
- [x] include existing url query params in breadcrumbs (otherwise navigation to parent in a share fails with a `Resource not found` because there is no `resourceId` available anymore)
- [x] `contextRouteQuery` for app defaults (`SharedResource` view needs share id in the url query because share name in the path is non-unique and doesn't work for propfind)
- [x] fix path and webdav path after `accept share` action
- [x] left sidebar active route highlighting
- ~~check back with CERN if their shares still work as expected~~ not needed because we hide the share jail behind a new capability. if they set that to `false` they have the old behaviour.
- [x] switch to new personal space webdav endpoint for location picker
- [x] repair that quick actions and sharing context menu items are not available anymore (broke it in this PR)
- [x] `Shared with me` doesn't show an accepted share in the right sidebar anymore, because it has a path of `/` and that is treated as a root element (which are not shown in the right sidebar).
- [x] fix unit tests
- [x] fix acceptance tests 🤯 
- [x] fix e2e tests
- [x] adapt `packages/web-app-files/src/mixins/actions/createQuicklink.js` regarding share capabilities after https://github.com/owncloud/web/pull/6820 is merged
- [x] adapt new `hasSpaces` occurrences after https://github.com/owncloud/web/pull/6841 is merged
- [x] fix uploads in shares (needs different endpoint, still points to `/dav/files/<user>`)

## followup after merge (create issues once this PR is merged or close to it)
- get rid of `/Shares` path prefix for ocs `shares` responses from ocis. at the moment this is rewritten client side.
- make accepted shares and project spaces available on the location picker page https://github.com/owncloud/web/issues/6892
- make `copy` and `move` actions available in `SharedResource` route/view https://github.com/owncloud/web/issues/6892
- concepting: rename shares https://github.com/owncloud/ocis/issues/3721

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-app-files/src/services/folder/loaderSharedWithMe.ts
```diff
@@ -3,27 +3,30 @@ import Router from 'vue-router'
...
-      store,
-      clientService: { owncloudSdk: client }
-    } = context
+    const { store, clientService } = context
 
     return useTask(function* (signal1, signal2) {
       store.commit('Files/CLEAR_CURRENT_FILES_LIST')
 
-      let resources = yield client.requests.ocs({
+      let resources = yield clientService.owncloudSdk.requests.ocs({
```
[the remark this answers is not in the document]
kulmann: I needed the graph client from the client service as well and didn't want to deconstruct into 1 property and 1 callback. Felt more natural to me to use the `clientService` as is again... open for discussion!

packages/web-app-files/src/services/folder/loaderFavorites.ts
```diff
@@ -4,9 +4,15 @@ import { useTask } from 'vue-concurrency'
 import { DavProperties } from 'web-pkg/src/constants'
 import { buildResource } from '../../helpers/resources'
 import { isLocationCommonActive } from '../../router'
+import { Store } from 'vuex'
+import get from 'lodash-es/get'
 
 export class FolderLoaderFavorites implements FolderLoader {
-  public isEnabled(router: Router): boolean {
+  public isEnabled(store: Store<any>): boolean {
+    return get(store, 'getters.capabilities.files.favorites', true)
```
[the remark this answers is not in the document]
kulmann: That's highly individual. Some capabilities exist in oc10, some don't. In general we can define new capabilities in oCIS easier than in oc10 (because of release cycles). So most of the time we have to put the expected oc10 value as default. Feels like too many comments to have that on every capability usage. :-/

JammingBen:
At a first glance, renaming files as a share recipient does not work for me most of the time. Strangely it works when declining a share, then accepting that share again and renaming it. But when I freshly browse to the page with accepted shares, renaming fails for me.

pascalwengerter (commented):
I don't think `tests/e2e/cucumber/reportGenerator/cucumber_report.json` should have been committed? 🤔 wondering why it's not in the .gitignore

pascalwengerter:
(Indirect) link shares seem to cause trouble, so files in a subfolder in a "people" share that also have a (quick)link

Screenshot 2022-05-05 at 11 05 40

JammingBen:
Issues I've found:

* Shared with me page: Rename actions for declined or not-yet-accepted shares should be hidden
* Shared with me page: Rename actions for accepted shares don't work most of the time (tbd with ocis team)
* Shared with me page: The "Open folder" action (via right-click) has a false location on both a shared folder itself as well as on folders inside a shared folder.
* The breadcrumb actions on a share itself have some issues:
    * "Details"-action does not set the current share as selected/highlighted file
    * "Rename"-action doesn't even open
    * "Delete"-action declines the share -> better naming?
image

packages/web-app-files/src/views/shares/SharedWithMe.vue
```diff
@@ -416,7 +433,9 @@ export default defineComponent({
 
     isResourceInSharesSelection(resource) {
       return this.sharesSelected?.includes(resource)
-    }
+    },
+
+    createRouteTarget() {}
```
JammingBen, line 438: Looks like this is not needed?

packages/web-app-files/src/views/LocationPicker.vue
```diff
@@ -145,11 +148,28 @@ export default {
...
+              1,
+              DavProperties.Default
+            )
+          } else {
+            resources = yield ref.$client.files.list(
+              buildWebDavFilesPath(ref.user.id, target),
+              1,
+              DavProperties.Default
+            )
+          }
```
JammingBen, line 172: Since only the user ID differs, it may be nicer to save that in a variable and then call `yield ref.$client.files.list()` outside the if-block.

JammingBen (approved):
2 minor things, definitely no "Must-Haves" though. LGTM overall 🚀 

Also, `hasSpaces` is still being used in `FileLinks.vue`.

packages/web-app-files/src/views/shares/SharedWithMe.vue
kulmann: 👍

packages/web-app-files/src/views/LocationPicker.vue
kulmann: 👍

kulmann:
Commented about the upcoming changes of this PR in https://github.com/owncloud/owncloud-test-middleware/issues/116

pascalwengerter (commented):
Just figured I have a pending review, let's tackle this in a follow-up 🤦🏽

PR #6618, opened 2022-03-17
[full-ci] Multiple fixes

### LINKED ISSUE #5913: User/group filters type does not work
In reva the user and group provider implement some filters which allow a user to filter by the user/group type (primary, secondary, lightweight account), when creating a share of a folder.
These filters are specified prepending a letter to the actual query name, i.e. `<letter>:<user/group>`, and can be `a` to get all the primary, service and secondary accounts, `l` to get only the lightweight accounts, and no filter for only primary accounts.

In web side only the last works because the list got from reva is re-filtered according to the user query, and the filters are indeed broken.

### LINKED ISSUE #6324: "empty trashbin" as capability
At CERN we want to disable the empty trashbin button to avoid the permanent delete action (per policy decision).
Right now we have commented the buttons, but can we add this as a capability?

diocas:
As we discussed, I'm trying to push some of our changes, in particular I refactored the code to make some parts optional. 

Renaming a share, permanently deleting files and showing the custom permissions role are now optional. By default, it keeps the same behavior as the current master.

Added the possibility to add "tags" while searching for users (as we discussed).

Also as discussed, I've added the "editor" role for single files public links. This seems not to be possible with oc10, so I've added a check to enable it only for oCIS.
(this needs another fix... I saw the "rename" button on the file and I shouldn't, but at least if fails when I tried to do it)

@elizavetaRa is still working on the css for the drop down menus.

Fixes #6324
Fixes #5913

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Links/PublicLinks/LinkEdit.vue
+++ b/packages/web-app-files/src/components/SideBar/Links/PublicLinks/LinkEdit.vue
@@ -190,7 +190,7 @@ export default {
   },
   computed: {
     ...mapGetters('Files', ['highlightedFile']),
-    ...mapGetters(['getToken', 'capabilities']),
+    ...mapGetters(['getToken', 'capabilities', 'isOcis']),
     ...mapState('Files', ['publicLinkInEdit']),
 
     selectedRole() {
@@ -219,7 +219,7 @@ export default {
     },
 
     availableRoleOptions() {
-      return LinkShareRoles.list(this.highlightedFile.isFolder).map((r) =>
+      return LinkShareRoles.list(this.highlightedFile.isFolder, this.isOcis).map((r) =>
         this.convertRoleToSelectOption(r)
       )
     },
--- a/packages/web-app-files/src/components/SideBar/Links/PublicLinks/LinkEdit.vue
+++ b/packages/web-app-files/src/components/SideBar/Links/PublicLinks/LinkEdit.vue
@@ -190,7 +190,7 @@ export default {
   },
   computed: {
     ...mapGetters('Files', ['highlightedFile']),
-    ...mapGetters(['getToken', 'capabilities', 'isOcis']),
+    ...mapGetters(['getToken', 'capabilities']),
     ...mapState('Files', ['publicLinkInEdit']),
 
     selectedRole() {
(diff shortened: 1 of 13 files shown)


diocas:
Still need to have a look at the tests

kulmann:
hey @diocas could you run `yarn lint --fix` on your branch and commit the result?

packages/web-app-files/src/components/SideBar/Links/PublicLinks/LinkEdit.vue
```diff
@@ -190,7 +190,7 @@ export default {
   },
   computed: {
     ...mapGetters('Files', ['highlightedFile']),
-    ...mapGetters(['getToken', 'capabilities']),
+    ...mapGetters(['getToken', 'capabilities', 'isOcis']),
```
kulmann, line 193: The `isOcis` helper has been removed in https://github.com/owncloud/web/pull/6619 in favour of checking the resharing capability - do you mind looking into the usage of that? it needs to be used in a `setup` call (composition api).

packages/web-app-files/src/helpers/share/role.ts
```diff
@@ -282,12 +289,13 @@ export abstract class LinkShareRoles {
     linkRoleUploaderFolder
   ]
 
-  static list(isFolder: boolean): ShareRole[] {
-    return this.all.filter((r) => r.folder === isFolder)
+  static list(isFolder: boolean, isOcis: boolean = false): ShareRole[] {
```
kulmann, line 292: Can you rename that usage of `isOcis` to a capability?

kulmann (changes requested):

packages/web-app-files/src/helpers/share/role.ts
diocas: I've changed to `capabilities.files_sharing?.public?.can_edit` is this ok?

packages/web-app-files/src/components/SideBar/Links/PublicLinks/LinkEdit.vue
diocas: I do not understand what you asked, sorry. I've replaced with just a check from a capability.

packages/web-app-files/tests/unit/mixins/actions/delete.spec.js
```diff
@@ -15,6 +15,14 @@ const Component = {
   mixins: [Delete]
 }
 
+const store = new Vuex.Store({
+  getters: {
+    capabilities: () => {
+      return {}
```
pascalwengerter, line 21: My gripe with this is that it doesn't check the newly added logic via `this.capabilities?.files?.permanent_deletion` (as in it should now test both cases, one where permanent_deletion is true and one where it's false)

kulmann (approved):

PR #6622, opened 2022-03-18
[full-ci] Remove contextRouteName from app params and add it to the query instead

dschmidt:
## Description
This removes `contextRouteName` from app parameters. It's handled fully automatically in the url query, so apps can just use the provided `useAppDefaults` composable and don't need to mess with urls themselves for something they really should not have to care about.

Nice sideeffect is that it's now visible further in the back of the url and hopefully less confusing for users.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Make everything more streamlined and straightforward for developing file editors/viewers.
I've also taken care that changing the query item name is also just a two line patch, as we plan to adjust the naming in the near future.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/composables/appDefaults/useAppDefaults.ts
+++ b/packages/web-pkg/src/composables/appDefaults/useAppDefaults.ts
@@ -1,5 +1,5 @@
 import { computed, unref, Ref } from '@vue/composition-api'
-import { useRouter } from '../router'
+import { useRouter, useRoute } from '../router'
 import { useStore } from '../store'
 import { ClientService, clientService as defaultClientService } from '../../services'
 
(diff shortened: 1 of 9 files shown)


pascalwengerter (approved):
Fine for me if CI is happy, too

packages/web-pkg/src/composables/appDefaults/useAppDefaults.ts
```diff
@@ -33,24 +34,29 @@ type AppDefaultsResult = AppConfigResult &
...
+  const currentRoute = useRoute()
   const clientService = options.clientService || defaultClientService
 
-  const currentRoute = computed(() => {
-    return router.currentRoute
-  })
-
   const isPublicLinkContext = computed(() => {
-    return unref(currentRoute).params.contextRouteName === 'files-public-files'
+    return unref(currentRoute).query[contextRouteNameKey] === 'files-public-files'
```
kulmann, line 41: Did you see the `useRouteQuery` composable? 🤔

kulmann, line 41: Did you see the `useRouteQuery` composable? 🤔

dschmidt: 😅

PR #6653, opened 2022-03-23
Add quick rename

lookacat:
## Description
See https://github.com/owncloud/web/issues/6626

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6626

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-runtime/package.json
+++ b/packages/web-runtime/package.json
@@ -18,7 +18,7 @@
     "luxon": "^2.3.0",
     "marked": "^4.0.12",
     "oidc-client": "1.11.5",
-    "owncloud-design-system": "^13.0.0",
+    "owncloud-design-system": "^13.0.1-rc.1",
     "owncloud-sdk": "~3.0.0-alpha.4",
     "p-queue": "^6.1.1",
     "popper-max-size-modifier": "^0.2.0",
--- a/changelog/5.2.0_2022-03-03/enhancement-add-quick-rename-button
+++ b/changelog/5.2.0_2022-03-03/enhancement-add-quick-rename-button
@@ -1,0 +1,6 @@
+Enhancement: Add quick rename button
+
+We've added a button for quick editing a resource name
+
+https://github.com/owncloud/web/pull/6645
+https://github.com/owncloud/web/issues/6626
--- a/packages/web-app-files/src/components/FilesList/ResourceTable.vue
+++ b/packages/web-app-files/src/components/FilesList/ResourceTable.vue
@@ -45,17 +45,26 @@
       />
     </template>
     <template #name="{ item }">
-      <oc-resource
-        :key="`${item.path}-${resourceDomSelector(item)}-${item.thumbnail}`"
-        :resource="item"
-        :is-path-displayed="arePathsDisplayed"
-        :parent-folder-name-default="defaultParentFolderName"
-        :is-thumbnail-displayed="areThumbnailsDisplayed"
-        :is-resource-clickable="isResourceClickable(item.id)"
-        :folder-link="folderLink(item)"
-        :parent-folder-link="parentFolderLink(item)"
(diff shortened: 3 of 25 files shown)


dschmidt:
Could we add a keyboard shortcut for this? Dolphin in KDE uses F2 for example

dschmidt:
Yeah, fine with me. That's how it works in dolphin as well. And I think explorer on Windows does the same.

tbsbdr:
## Which key?
Agree, reference would be "F2" as most people might know it from windows file explorer

## Avoid cannibalisation
Remember, we are not alone :-) If the user already expects a specific action on "F2" we must not hijack this key. Quick research results:
  - Firefox: "F2" seems to be currently unassigned [FF shortcuts](https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly)
  - Chrome: "F2" seems to be currently unassigned [Chrome shortcuts](https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop)
  - Safari "F2" seems to be currently unassigned [Safari shortcuts](https://support.apple.com/de-de/guide/safari/cpsh003/mac)
  - Screenreader NVDA: "F2" seems to be currently unassigned [NVDA shortcuts](https://dequeuniversity.com/screenreaders/nvda-keyboard-shortcuts)

Having a F2 shortcuts is fine imo for a start.
We need to inform & educate the user about the shortcut, so `F2`  should be mentioned in the tooltip and in the context menu next to the item "rename".

@michaelstingl I would be interested in your opinion if we should should have F2 as a rename shortcut in Web.

michaelstingl:
> @michaelstingl I would be interested in your opinion if we should should have F2 as a rename shortcut in Web.

Maybe with an option in the user settings. Similar to a setting to disable the right-click hijacking for our context menu.

Do we have more keyboard shortcuts? Is there a help button somewhere, where users could display all the keyboard shortcuts? iPad apps for example, show an overview when you long-press the Command-key:

5FB89BF8-80BA-41BB-863B-9374273119A4

kulmann:
I rebased (conflicted files) and squashed what you've done so far @lookacat 
Also added a check if the rename action is available at all for the respective file. That is a must have. There are tons of situations where the rename action is not available. 😉

kulmann:
I'd be fine with the keyboard shortcut being added in another PR by the way

kulmann:
Reviewed the part that @lookacat built, that looked good to me (except for the check if the rename action is available). Will not approve the PR because I was too much involved. Can someone else please take over the review?

tbsbdr:
@michaelstingl 
> Do we have more keyboard shortcuts? Is there a help button somewhere, where users could display all the keyboard shortcuts?

Currently we have no other shortcuts and there is also no help button to have a overview of the shortcuts.

As soon as time permits I'd like to have more shortcuts eg. for selection, sharing, search, maybe even copy/paste. Then we also need a dedicated shortcuts overview.

packages/web-runtime/package.json
```diff
@@ -18,7 +18,7 @@
     "luxon": "^2.3.0",
     "marked": "^4.0.12",
     "oidc-client": "1.11.5",
-    "owncloud-design-system": "^13.0.1-rc.1",
+    "owncloud-design-system": "13.0.1-rc.1",
```
pascalwengerter, line 21: Smarrrrt

pascalwengerter: Ah damn, but you're missing the updated lockfile here?

pascalwengerter: Could you drop the `try` commit if the "use fake timers" commit works? 😇

pascalwengerter (approved):

Issue #5958, opened 2021-10-29
Meaningful tab title for external apps


wkloucek:
## Feature Request

Having a meaning full title in the tab would be great. The current text is not really helpfull:
![image](https://user-images.githubusercontent.com/34452982/135329054-9bf40f6c-f79a-416f-9802-e0fd7000b276.png)

_Originally posted by @wkloucek in https://github.com/owncloud/web/issues/5805#issuecomment-930444495_

PR #6655, opened 2022-03-23
Show filename and app name in external app tab title

elizavetaRa:
## Description
The pr changes external apps route to take filePath instead of fileId. It extracts fileName from filePath and sets document.title to <fileName> - <appName>

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5958

## Motivation and Context
Current document title for external apps is not helpful in understanding, which file is open and with which app. By several open files, it's hard to find the right one.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -82,8 +82,7 @@ export default {
     formParameters: {}
   }),
   computed: {
-    ...mapGetters(['getToken', 'capabilities', 'configuration']),
-    ...mapGetters('Files', ['publicLinkPassword']),
+    ...mapGetters(['capabilities', 'configuration']),
 
     pageTitle() {
       const translated = this.$gettext('"%{appName}" app page')
@@ -116,22 +115,6 @@ export default {
       const filePath = this.currentFileContext.path
       const fileId = this.fileId || (await this.getFileInfoResource(filePath)).fileId
 
-      // build headers with respect to the actual auth situation
-      const publicToken = (filePath || '').split('/')[1]
-      const publicLinkPassword = this.publicLinkPassword
-      const accessToken = this.getToken
-      const headers = {
-        'X-Requested-With': 'XMLHttpRequest',
-        ...(!accessToken && publicToken && { 'public-token': publicToken }),
-        ...(publicLinkPassword && {
-          Authorization:
-            'Basic ' + Buffer.from(['public', publicLinkPassword].join(':')).toString('base64')
-        }),
-        ...(accessToken && {
-          Authorization: 'Bearer ' + accessToken
-        })
-      }
-
       // fetch iframe params for app and file
       const configUrl = this.configuration.server
       const appOpenUrl = this.capabilities.files.app_providers[0].open_url.replace(/^\/+/, '')
@@ -141,10 +124,7 @@ export default {
         `?file_id=${fileId}` +
         (this.appName ? `&app_name=${this.appName}` : '')
 
-      const response = await fetch(url, {
-        method: 'POST',
-        headers
-      })
+      const response = await this.makeRequest('POST', url)
 
       if (response.status !== 200) {
         this.errorMessage = response.message
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -36,6 +36,9 @@
 import { mapGetters } from 'vuex'
 import ErrorScreen from './components/ErrorScreen.vue'
 import LoadingScreen from './components/LoadingScreen.vue'
+import { DavProperties } from 'web-pkg/src/constants'
+import { buildResource } from '../../web-app-files/src/helpers/resources'
+import { useAppDefaults } from 'web-pkg/src/composables'
 
 // FIXME: hacky, get rid asap, just a workaround
 // same as packages/web-app-files/src/views/PublicFiles.vue
@@ -44,7 +47,7 @@ const unauthenticatedUserReady = async (router, store) => {
     return
   }
 
-  const publicToken = router.currentRoute.query['public-token']
+  const publicToken = (router.currentRoute.params.filePath || '').split('/')[0]
   const publicLinkPassword = store.getters['Files/publicLinkPassword']
 
   await store.dispatch('loadCapabilities', {
@@ -61,6 +64,13 @@ export default {
   components: {
     ErrorScreen,
     LoadingScreen
+  },
+  setup() {
+    return {
+      ...useAppDefaults({
+        applicationName: 'external'
+      })
+    }
   },
 
   data: () => ({
@@ -88,73 +98,92 @@ export default {
       })
     },
     appName() {
-      return this.$route.params.appName
+      return this.$route.query.app
     },
     fileId() {
-      return this.$route.params.fileId
+      return this.$route.query.fileId
     }
+  },
+  mounted() {
+    const appNameTitle = this.appName ? `${this.appName} - ` : ''
+    document.title = `${this.currentFileContext.fileName} - ${appNameTitle}${this.configuration.currentTheme.general.name}`
   },
   async created() {
     await unauthenticatedUserReady(this.$router, this.$store)
 
     this.loading = true
+    try {
+      const filePath = this.currentFileContext.path
+      const fileId = this.fileId || (await this.getFileInfoResource(filePath)).fileId
 
-    // build headers with respect to the actual auth situation
-    const { 'public-token': publicToken } = this.$route.query
-    const publicLinkPassword = this.publicLinkPassword
-    const accessToken = this.getToken
-    const headers = {
-      'X-Requested-With': 'XMLHttpRequest',
-      ...(publicToken && { 'public-token': publicToken }),
-      ...(publicLinkPassword && {
-        Authorization:
-          'Basic ' + Buffer.from(['public', publicLinkPassword].join(':')).toString('base64')
-      }),
-      ...(accessToken && {
-        Authorization: 'Bearer ' + accessToken
+      // build headers with respect to the actual auth situation
+      const publicToken = (filePath || '').split('/')[1]
+      const publicLinkPassword = this.publicLinkPassword
+      const accessToken = this.getToken
+      const headers = {
+        'X-Requested-With': 'XMLHttpRequest',
+        ...(!accessToken && publicToken && { 'public-token': publicToken }),
+        ...(publicLinkPassword && {
+          Authorization:
+            'Basic ' + Buffer.from(['public', publicLinkPassword].join(':')).toString('base64')
+        }),
+        ...(accessToken && {
+          Authorization: 'Bearer ' + accessToken
+        })
+      }
+
+      // fetch iframe params for app and file
+      const configUrl = this.configuration.server
+      const appOpenUrl = this.capabilities.files.app_providers[0].open_url.replace(/^\/+/, '')
+      const url =
+        configUrl +
+        appOpenUrl +
+        `?file_id=${fileId}` +
+        (this.appName ? `&app_name=${this.appName}` : '')
+
+      const response = await fetch(url, {
+        method: 'POST',
+        headers
       })
-    }
 
-    // fetch iframe params for app and file
-    const configUrl = this.configuration.server
-    const appOpenUrl = this.capabilities.files.app_providers[0].open_url.replace(/^\/+/, '')
-    const url =
-      configUrl +
-      appOpenUrl +
-      `?file_id=${this.fileId}` +
-      (this.appName ? `&app_name=${this.appName}` : '')
+      if (response.status !== 200) {
+        this.errorMessage = response.message
+        this.loading = false
+        this.loadingError = true
+        console.error('Error fetching app information', response.status, this.errorMessage)
+        return
+      }
 
-    const response = await fetch(url, {
-      method: 'POST',
-      headers
-    })
+      const data = await response.json()
 
-    if (response.status !== 200) {
-      this.errorMessage = response.message
+      if (!data.app_url || !data.method) {
+        this.errorMessage = this.$gettext('Error in app server response')
+        this.loading = false
+        this.loadingError = true
+        console.error('Error in app server response')
+        return
+      }
+
+      this.appUrl = data.app_url
+      this.method = data.method
+      if (data.form_parameters) this.formParameters = data.form_parameters
+
+      if (this.method === 'POST' && this.formParameters) {
+        this.$nextTick(() => this.$refs.subm.click())
+      }
+      this.loading = false
+    } catch (error) {
+      this.errorMessage = this.$gettext('Error retrieving file information')
+      console.error('Error retrieving file information', error)
       this.loading = false
       this.loadingError = true
-      console.error('Error fetching app information', response.status, this.errorMessage)
-      return
     }
-
-    const data = await response.json()
-
-    if (!data.app_url || !data.method) {
-      this.errorMessage = this.$gettext('Error in app server response')
-      this.loading = false
-      this.loadingError = true
-      console.error('Error in app server response')
-      return
+  },
+  methods: {
+    async getFileInfoResource(path) {
+      const file = await this.getFileInfo(path, DavProperties.Default)
+      return buildResource(file)
     }
-
-    this.appUrl = data.app_url
-    this.method = data.method
-    if (data.form_parameters) this.formParameters = data.form_parameters
-
-    if (this.method === 'POST' && this.formParameters) {
-      this.$nextTick(() => this.$refs.subm.click())
-    }
-    this.loading = false
   }
 }
 </script>
--- a/packages/web-app-external/src/index.js
+++ b/packages/web-app-external/src/index.js
@@ -2,11 +2,6 @@ import translations from '../l10n/translations'
 import App from './App.vue'
 import store from './store'
 
-// just a dummy function to trick gettext tools
-function $gettext(msg) {
-  return msg
-}
-
 const appInfo = {
   name: 'External',
   id: 'external'
@@ -15,11 +10,11 @@ const appInfo = {
 const routes = [
   {
     name: 'apps',
-    path: '/:fileId/:appName?',
+    path: '/:filePath*',
     component: App,
     meta: {
       auth: false,
-      title: $gettext('External app')
+      patchCleanPath: true
     }
   }
 ]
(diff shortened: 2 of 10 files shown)


elizavetaRa:
To be discussed: 
instead of retrieving fileId from filePath through another request `this.$client.files.fileInfo` in packages/web-app-external/src/App.vue the backend endpoint could be changed to accept filePath

packages/web-app-external/src/App.vue
```diff
@@ -84,15 +87,27 @@ export default {
...
+    else if (this.$route.params.filePath) document.title = `${this.$route.params.filePath}`
+  },
   async created() {
     await unauthenticatedUserReady(this.$router, this.$store)
 
     this.loading = true
 
+    // get fileId by filePath
+    const resource = await this.$client.files.fileInfo(this.filePath, DavProperties.Default)
+    this.fileId = buildResource(resource).id
```
dschmidt, line 109: In the short term: this should also go through the `useAppDefaults` composable and use getFileInfo there.

In the long term: If I understood @kulmann correctly, we will add the fileId to the query params anyway, so we could also add it to the `FileContext` and it would make this call completely obsolete.

dschmidt (changes requested):

diocas:
I've also changed something: made the app name a query parameter instead of part of the url (this is what we suggested in the app url PR).
I don't really like having the app name in the url, I think it would make more sense to have an appId instead. Also, the call to the backend without an app opens the default one, but then we don't have its name. Maybe the backend should return it as well.
Let me know if you agree/disagree.

Sorry for the previous commits that didn't use `useAppDefaults`, there was a miscommunication on our side.
Also took the opportunity to clean old auth code that should not be necessary anymore.

I've tested this PR locally with OCIS and when opening from a public link with password, I got a basic auth dialog from the backend. It was sending `public:pwd` correctly, so not sure what happened.

I have 2 questions: 
1. Can we make `getFileInfo` return a `Resource` (or `Promise<Resource>`)? I don't see why the app should be dealing with webdav params.
2. Is there a way we can move the request logic (essentially the headers construction) to a central place that could be re-used by others?

packages/web-app-external/src/App.vue
```diff
@@ -82,16 +74,20 @@ export default {
       })
     },
     appName() {
-      return this.$route.params.appName
+      return this.$route.query.app
     },
-    fileId() {
-      return this.$route.params.fileId
+    fileName() {
+      return this.currentFileContext.path.split('/').pop()
```
dschmidt, line 80: You could `import { basename } from 'path'` and use that instead. Or we might even add a `fileName()` function to `FileContext` using that for convenience

packages/web-app-files/src/mixins/fileActions.js
```diff
@@ -245,20 +247,23 @@ export default {
           class: `oc-files-actions-${app.name}-trigger`,
           isEnabled: () => true,
           canBeDefault: defaultApplication === app.name,
-          handler: () => this.$_fileActions_openLink(app.name, fileId),
+          handler: () => this.$_fileActions_openLink(app.name, webDavPath, app.name),
```
dschmidt, line 250: Why do you pass in `app.name` a second time?

packages/web-app-external/src/index.js
```diff
@@ -15,11 +10,11 @@ const appInfo = {
 const routes = [
   {
     name: 'apps',
-    path: '/:fileId/:appName?',
+    path: '/:filePath*',
     component: App,
     meta: {
       auth: false,
-      title: $gettext('External app')
```
dschmidt, line 22: Is this nowhere used?

dschmidt:
Nice, I really like where this is going now!

> I've also changed something: made the app name a query parameter instead of part of the url (this is what we suggested in the app url PR). I don't really like having the app name in the url, I think it would make more sense to have an appId instead. Also, the call to the backend without an app opens the default one, but then we don't have its name. Maybe the backend should return it as well. Let me know if you agree/disagree.

I'm not so familiar with how the external apps work, so I can't comment on this right now :)

> Sorry for the previous commits that didn't use `useAppDefaults`, there was a miscommunication on our side. Also took the opportunity to clean old auth code that should not be necessary anymore.

No worries, that's what reviews are for :) 

> I've tested this PR locally with OCIS and when opening from a public link with password, I got a basic auth dialog from the backend. It was sending `public:pwd` correctly, so not sure what happened.

And it doesn't happen on master? Then you could compare requests - in doubt "copy as curl" from browser tools and diff it...

 
> I have 2 questions:
> 
> 1. Can we make `getFileInfo` return a `Resource` (or `Promise<Resource>`)? I don't see why the app should be dealing with webdav params.

*full ack* regarding dav params.
I'm not sure whether we can always safely assume we're dealing with resources, possibly yes.
In this case the fix is probably easier and avoids the request completely. As mentioned in one comment already, I think there are some changes to filename handling in urls in the pipeline.
If I understood @kulmann correctly, we will add the fileId to the query params anyway, so we can add that to the FileContext. If you take care of that now, you can avoid this request in the first place and we can delay the discussion about proper return types to a later point in time ;-)


> 2. Is there a way we can move the request logic (essentially the headers construction) to a central place that could be re-used by others?

Why not `useAppDefaults.fetch` or something like that?

I think we're completely on the same page, we should make apps as dumb as possible. I really really want to keep any knowledge of whether we're in a personal, space or public (or whatever) context out of apps. As an app developer I don't care at all how my file is retrieved, it should just work and not break every other day because there's a new context I don't handle. So yes, please let us move everything that is not strictly app specific to a central place and make app development as easy as possible :)

packages/web-app-external/src/index.js
diocas: No, we overwrite the title in the code to include the file and the app name (I think this should be centralized somehow as well...)

packages/web-app-external/src/App.vue
```diff
@@ -84,15 +87,27 @@ export default {
...
     }
   },
+  mounted() {
+    if (this.$route.params.filePath && this.$route.params.appName)
+      document.title = `${this.$route.params.filePath.split('/').pop()} - ${
+        this.$route.params.appName
+      }`
+    else if (this.$route.params.appName) document.title = `${this.$route.params.appName}`
+    else if (this.$route.params.filePath) document.title = `${this.$route.params.filePath}`
+  },
```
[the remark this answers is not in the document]
diocas: Yes, I forgot that in my comment, but I wanted to discuss it. Because all apps have a pretty horrible name. I'll try to do it.

packages/web-app-files/src/mixins/fileActions.js
diocas: 🤦

packages/web-app-external/src/index.js
dschmidt: We should keep apps in mind that don't handle files, so the filename part should be optional imho :)

diocas:
Did some improvements based on your comments. At the moment I don't have an idea on how to centralize the title, so I kept as is.

I've removed the public token from the query params, since we can extract it from the path, like other extensions.
Also created this central `makeRequest` method and tried to clean stuff. 

I still need to do a bit more testing, but it looks everything is working, except opening external apps with protected public links. The issue that I'm getting is `ReferenceError: Buffer is not defined` from the new `makeRequest` method... Any idea how to solve it?

> If I understood @kulmann correctly, we will add the fileId to the query params anyway, so we can add that to the FileContext.

This is coming, but I think we agreed to have it optional. I believe CERNBox will set this config to false, so it's still needed. Either way, the id will be complementary to the path, and there needs to be a logic to choose one or the other, so I imagine a request will be necessary either way (inclusive there were plans to match them, but I don't know the details of the implementation).

diocas:
Btw, as a consequence of using this new method, drawio should have its visio import function working in external links as well.

diocas:
Quick fix for creation of files. I'm pretty sure you will have plenty of comments on these things.
(I promise I'll learn vue this year..)

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -150,7 +151,10 @@ export default {
       isPersonalLocation: useActiveLocation(isLocationSpacesActive, 'files-spaces-personal-home'),
       isPublicLocation: useActiveLocation(isLocationPublicActive, 'files-public-files'),
       isSpacesProjectsLocation: useActiveLocation(isLocationSpacesActive, 'files-spaces-projects'),
-      isSpacesProjectLocation: useActiveLocation(isLocationSpacesActive, 'files-spaces-project')
+      isSpacesProjectLocation: useActiveLocation(isLocationSpacesActive, 'files-spaces-project'),
+      ...useAppDefaults({
```
diocas, line 155: @kulmann this was my "fix" for having `makeRequest` in here and the reason I said it was "quick" and that you will "have comments" on it.
I tried separating things to not need to apply the full "defaults", but it was becoming messy, I don't think I managed to make it work and didn't have more time to spend on it. So if you have suggestions, they are welcomed. (if this is ok for you, even better).

diocas:
@dschmidt @kulmann I've made the tests green (there's a FIXME in the code I can remove). When can you review our PRs?

packages/web-app-files/src/mixins/fileActions.js
```diff
@@ -260,30 +259,25 @@ export default {
...
         },
+        filePath,
         undefined,
-        resourceId,
         undefined
       )
 
-      routeOpts.params.appName = appName
-
       routeOpts.query = {
```
kulmann, line 278: If you only need to push the `app` var into the query, then `routeOpts.query.app = appName` does the same and only needs one line of code.

packages/web-app-files/src/mixins/fileActions.js
```diff
@@ -260,30 +259,25 @@ export default {
...
-        ...routeOpts.query,
-        // public-token retrieval is weak, same as packages/web-app-files/src/index.js:106
-        ...(!isAuthenticatedRoute(this.$route) && {
-          'public-token': (this.$route.params.item || '').split('/')[0]
```
kulmann, line 285: Can you clarify if you really can delete the `public-token` here?!

kulmann, line 285: Can you clarify if you really can delete the `public-token` here?!

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
kulmann: The refactoring into `makeRequest` feels like a huge scope creep for this PR and makes it much harder to review. Also we have some different ideas about making network requests. Will allow it for this PR but will refactor it soon...

kulmann: The refactoring into `makeRequest` feels like a huge scope creep for this PR and makes it much harder to review. Also we have some different ideas about making network requests. Will allow it for this PR but will refactor it soon...

packages/web-app-files/src/mixins/fileActions.js
```diff
@@ -260,30 +259,25 @@ export default {
...
 
-    $_fileActions_openLink(appName, resourceId) {
+    $_fileActions_openLink(appName, filePath) {
       const routeOpts = this.$_fileActions__routeOpts(
         {
           routeName: 'external-apps'
         },
+        filePath,
         undefined,
-        resourceId,
```
kulmann, line 275: Please keep the `resourceId`. We want to have it in the route query.

packages/web-app-external/src/App.vue
```diff
@@ -88,73 +80,69 @@ export default {
...
-    const response = await fetch(url, {
-      method: 'POST',
-      headers
-    })
-
-    if (response.status !== 200) {
-      this.errorMessage = response.message
+    try {
+      const filePath = this.currentFileContext.path
+      const fileInfo = await this.getFileInfoResource(filePath)
```
kulmann, line 94: We would already have the file id in the route query. Feels unnecessary to fetch the whole file info again just to extract the id and don't do anything else with it.

packages/web-app-external/src/App.vue
```diff
@@ -36,24 +36,9 @@
...
-
-  const publicToken = router.currentRoute.query['public-token']
-  const publicLinkPassword = store.getters['Files/publicLinkPassword']
-
-  await store.dispatch('loadCapabilities', {
-    publicToken,
-    ...(publicLinkPassword && { user: 'public', password: publicLinkPassword })
-  })
-
-  store.commit('SET_USER_READY', true)
```
kulmann, line 55: Loading the capabilities and setting the user ready is important. Wondering how you still have the app working at all without it...

kulmann (changes requested):
Added some comments. The PR does a lot of things which are potentially error prone while it's goal was initially to show the filename and app name in the page title. Is that goal still the only purpose of this PR? Because then I'd ask @dschmidt to develop that in a different way if you're fine with that...

diocas:
@kulmann the scope of this PR was to do general improvements to the external apps (the title is not aligned with my initial intentions). 

I'm totally fine for @dschmidt to take it and change things. I can also try to split `makeRequest` into a diff PR, if you'd like. This was just to avoid having to keep dealign with headers and replicating logic in different places.

I'll go through your comments, but I can already give some explanations: the main idea was to make the external apps behave just like the other apps (which is def not the case now). 
This is why I replaced the URL with the file path (which was discussed as part of the better url's schema), and why I took the public-token from the URL as well. The other apps don't use it already.

For the readiness of the user and capabilities, no other application does this (I don't remember if this is done as part of the `useAppDefaults`, which we now use, or if it's being done anywhere else). So, again, this PR makes the external app aligned with all the others. And it wouldn't make sense to keep this logic in the app itself, IIRC this was only a workaround for public links.

I can re-add the `resourceId`, but this is not being used with the other apps. This would be inconsistent and that's why I deleted it. As I commented above, I know this will be coming back in the future, but in our discussions this was going to be **optional** and there's some logic to conciliate both that will be required. So I would propose that those get done all together in a separate PR. This one uniforms all the apps, and a new PR brings that functionality, hopefully in a central place.
I also commented why a call to the backend to get the ID is important: if this `resourceId` will be optional (and we are planning not to use it), this will still be required. Also, since you need to conciliate path and id somehow, this might still be necessary, but maybe (hopefully) not inside the app itself. Again, I would leave that for the separate PR I've mention above.
But... Maybe I would prefer to remove the need to get the ID by making the app registry take the path instead of the ID... I think it would be more consistent, but I'm not sure this aligns with the future direction of the frontend/backend communication.

Let me know your comments.

kulmann:
> @kulmann the scope of this PR was to do general improvements to the external apps (the title is not aligned with my initial intentions).

Please adjust title and description accordingly then. 😇 


> I'm totally fine for @dschmidt to take it and change things. I can also try to split `makeRequest` into a diff PR, if you'd like. This was just to avoid having to keep dealign with headers and replicating logic in different places.

Not exactly happy with the fact that `makeRequest` landed in the app defaults composables. You hinted towards that yourself and I agree... since composables are, well, composable, the `makeRequest` could be it's own composable (also the 4 auth related props that you needed could be small composables on their own). Can easily be added to the useAppDefaults set of composables as well so that apps that deal with files and use the composable wouldn't need to import the composable from somewhere else. There's more to it, we want to shift from using `fetch` to using `axios` (for various reasons, e.g. cancellation of requests) and there is already a `clientService` which gives you easy to use functions for getting a client. Discussed a possible use of this with @dschmidt and he will come up with some refactoring of that in the near future. For the time being I'm fine with getting your changes merged and @dschmidt refactoring them afterwards or @dschmidt taking over the changes you intended, as long as it doesn't take long. 😅 @dschmidt agree with merge and then refactor later on? (merge not yet possible for other reasons, see below)


> I'll go through your comments, but I can already give some explanations: the main idea was to make the external apps behave just like the other apps (which is def not the case now). This is why I replaced the URL with the file path (which was discussed as part of the better url's schema), and why I took the public-token from the URL as well. The other apps don't use it already.

Using the file path in the URL makes sense, sorry for the confusion. But at the moment we need the file id as well - which can easily live in the URL query.
- since the app provider resolves files via file id (which is a good thing in my opinion) I don't understand why you'd accept an otherwise unnecessary request for retrieving the file info by path just to read the id from it.
- you need (!) the file id for identifying the space. If you remove it from URL without going the extra mile of prefixing the path with the drive alias you won't be able to use the app provider for any file that lives outside the home space of the user. This will fall on your feet in just a couple of weeks. And would already now for us if we merge this, since our project spaces couldn't use the app provider anymore.


> For the readiness of the user and capabilities, no other application does this (I don't remember if this is done as part of the `useAppDefaults`, which we now use, or if it's being done anywhere else). So, again, this PR makes the external app aligned with all the others. And it wouldn't make sense to keep this logic in the app itself, IIRC this was only a workaround for public links.

The other apps don't strictly need the capabilities in public links. The external app needs it because the availability of the app provider is announced via capabilities and we can't load the available mimeType/application mapping from the app provider without the capability. This will break the external app on page reload. Might make sense to do the capability loading and firing the `user ready` event inside the app defaults, still. Having capabilities available is useful. But that should be a separate PR, thus just deleting it here is not an option.


> I can re-add the `resourceId`, but this is not being used with the other apps. This would be inconsistent and that's why I deleted it. As I commented above, I know this will be coming back in the future, but in our discussions this was going to be **optional** and there's some logic to conciliate both that will be required. So I would propose that those get done all together in a separate PR. This one uniforms all the apps, and a new PR brings that functionality, hopefully in a central place. I also commented why a call to the backend to get the ID is important: if this `resourceId` will be optional (and we are planning not to use it), this will still be required. Also, since you need to conciliate path and id somehow, this might still be necessary, but maybe (hopefully) not inside the app itself. Again, I would leave that for the separate PR I've mention above. But... Maybe I would prefer to remove the need to get the ID by making the app registry take the path instead of the ID... I think it would be more consistent, but I'm not sure this aligns with the future direction of the frontend/backend communication.

I do agree with centralizing common functionality and then having refactoring being simpler later on. But see comments above, this PR would break crucial things at the moment.

kulmann:
Uff.... just realizing now that we have the full webdav path in the URL (thank you @dschmidt ) so at least loading files from other spaces works already. Sorry for that mistake...
Then I guess it should be fine. But I still don't prefer consistency over the extra request. Having the same URL layout as the other apps is good. But we could accept it being just cosmetic for the moment and still relying on the resourceId from an URL parameter for the moment.

packages/web-app-external/src/App.vue
diocas: You are totally right, sorry. I didn't realize that it only worked because this was cached.

kulmann (approved):
Cool! Sorry for the long ping-pong and thank you that you pushed through it. We could add a changelog item about the page title in a separate PR. Also, like I said, @dschmidt will take care of taking the `makeRequest` abstraction a bit further. For now this is good to merge. Thanks again!

Issue #6585, opened 2022-03-11
[QA] new appswitcher has empty label for richdocuments


jnweiger:
Seen in web-5.2.0 and web 5.3.0-rc1

* Enable also richdocuments
* Click on the 9-squares menu
* While the mouse pointer hovers on the blank area, the url at the bottom shows that this is richdocuments 
![image](https://user-images.githubusercontent.com/1108546/157925748-b4c4f9f5-5f15-45d5-9a18-9d78741df15b.png)

pascalwengerter:
@JanAckermann @JammingBen could you share some insights on what might be missing? I checked for something obvious like a missing title/name in https://github.com/owncloud/richdocuments/blob/master/appinfo/info.xml, but that's not the case

pascalwengerter:
Needs a clarification on whether it's a `web` issue in general, and might be moved to a different repo in the process

JammingBen:
AFAIK those titles come from the Web `config.json` file, see https://owncloud.dev/clients/web/deployments/oc10-app/#configure-owncloud-web. @jnweiger Is a title set for this application?

pascalwengerter:
> AFAIK those titles come from the Web `config.json` file, see https://owncloud.dev/clients/web/deployments/oc10-app/#configure-owncloud-web. @jnweiger Is a title set for this application?

Aren't we pulling them from the backend? Seeing classic apps in the new frontend was implemented here https://github.com/owncloud/web/pull/5996/files

JammingBen:
@pascalwengerter is absolutely right, I totally forgot about this feature. https://github.com/owncloud/web/pull/6656 should fix it. I hardly believe it was working in the past for richdocuments... At least not when fetching the app automatically.

PR #6656, opened 2022-03-24
Use oC10 navigation entry names

JammingBen:
## Description
When fetching navigation entries from oC10, we previously used the app's names. This caused issues when the navigation entry ID and the app ID differ. Also, the navigation entries did not match with the ones in the classic UI.

This has been fixed as we now use the navigation entry name, which falls back to the app name if not given.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6585

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


pascalwengerter (approved):
Thanks for the swift reaction!

Issue #6661, opened 2022-03-24
App routes/extensions configurable/callable


diocas:
It would be great to have an app configuration when we configure its routes/extensions.

For example, you mentioned that you wanted to make some apps open in the same tab, while we think it's better to keep opening in a new tab. If this was a configuration, it would be possible.

Also, we would prefer to use your text/md editor instead of our text-editor, but we don't want it to open/create .md files. This could become a config. We also want to configure the extensions it can open, to easily add others.

If the routes/extensions was a callable, I guess it would also be possible to retrieve the mimetypes and configure the routes like any other app, thus removing the need to have it hardcoded in the files app.

diocas:
We don't have a way to configure individually for now, but we have an alternative in #8458 
So I think we can close it.

PR #6667, opened 2022-03-25
[full-ci] Improvements to text/md editor

diocas:
I'm trying to push upstream some improvements that we have in our own text-editor so that we can drop it. This PR is to discuss the direction and to have some pointers on how to implement things (sorry, I'm not well versed in Vue, and even less in composition api).

What this PR adds:
* Prevents the browser from unloading the page if there's still unsaved changes
* Allows ctrl/cmd+s for saving
* Displays proper error messages when saving
* Asks the user if he wants to save changes when exiting (either by clicking X or go back/forward in history, which might be by mistake)

Still in early stages, I'm also trying to make the app configurable, because we might want to remove the preview pane, since we also want to disable the markdown support (check #6661 )


packages/web-app-markdown-editor/src/App.vue
```diff
@@ -98,14 +143,44 @@ export default {
       return loadFileTask.isRunning || saveFileTask.isRunning
     })
 
+    const config = computed(() => {
+      const { showPreview = true } = getCurrentInstance().proxy.applicationConfig
```
dschmidt, line 147: I would recommend something like this:
```javascript
const defaults = useAppDefaults({
  applicationName: 'markdown-editor'
})
const { applicationConfig } = defaults

const config = computed(() => {
      const { showPreview = true } = unref(applicationConfig)
})

...

return {
  ...defaults

 ...
}
```

There's no need to go through the current instance

packages/web-app-markdown-editor/src/App.vue
```diff
@@ -98,14 +143,44 @@ export default {
...
+      return { showPreview }
+    })
+
     onMounted(() => {
-      loadFileTask.perform(getCurrentInstance().proxy)
+      const that = getCurrentInstance().proxy
+      loadFileTask.perform(that)
```
dschmidt, line 153: Again there's no need to use the instance imho, as all we need is `getFileContents` which is available from the instance via `useAppDefaults`, so we can use that directly.

Just pass `defaults` (from my recommendation above) to perform.

packages/web-app-markdown-editor/src/App.vue
```diff
@@ -98,14 +143,44 @@ export default {
...
+      window.addEventListener('beforeunload', handleUnload)
+    })
+
+    onBeforeUnmount(() => {
+      window.removeEventListener('beforeunload', handleUnload)
+      document.removeEventListener('keydown', handleSKey, false)
     })
 
     const save = function () {
       saveFileTask.perform(this)
```
dschmidt, line 167: This seems weird, what is `this` in this context? I doubt it provides `putFileContents`, better pass `defaults` here as well

dschmidt (changes requested):
Nice start, I have some hints for improvements, but I like where this is going :)

diocas:
@dschmidt thanks for your feedback, it helped me implementing what was missing as well.
I've done what you suggested and finalized what was pending:
* Support for read-only files (the user cannot edit the file, only see. I've opened this bug as well: [owncloud-design-system#2053](https://github.com/owncloud/owncloud-design-system/issues/2053)
* Design polish
* Configurable preview pane (it's enabled for md files and disabled for others by default)
* No auth in route, for compatibility with public links
* More extensions (it would be great if we could open no extension files..)

I've also renamed it from markdow-editor to simple-editor. I have the impression this naming is clearer for normal users, but if you prefer I can revert that commit.

What is missing is for an option to disable md creation, since we would prefer to do it with codimd.
We could also group extensions in the new menu.. But ok, that's a new ticket on its own.

I might need another iteration with the tests.

packages/web-app-simple-editor/src/index.js
```diff
@@ -43,6 +45,13 @@ const appInfo = {
   ]
 }
 
+for (const ext of ['js', 'json', 'xml', 'py', 'php', 'yaml']) {
+  appInfo.extensions.push({
+    extension: ext,
+    ...fileExtensionConfig
+  })
+}
+
```
dschmidt, line 54: Not a fan of this, maybe we can register editors for mimetypes in the long run? 

As we probably don't have a mechanism for that currently, it's totally out of scope for this PR. Maybe you can open a ticket?

dschmidt (approved):
Much better! I like!

packages/web-app-simple-editor/src/index.js
kulmann: Actually, I built registering apps on mime types in https://github.com/owncloud/web/pull/6514 but that one got stale...

packages/web-app-simple-editor/src/index.js
```diff
@@ -17,8 +19,8 @@ const fileExtensionConfig = {
 }
 
 const appInfo = {
-  name: 'MarkdownEditor',
-  id: 'markdown-editor',
+  name: 'SimpleEditor',
```
kulmann, line 22: totally fine with the rename, it's scoped way better than `markdown-editor`. 👍

packages/web-app-simple-editor/src/index.js
diocas: 👍  for that! I would like to be able to open files with no extension.

ownclouders:
Results for <strong>oCISFiles1</strong> https://drone.owncloud.com/owncloud/web/24506/55/1

.drone.star
```diff
@@ -78,7 +78,7 @@ config = {
                 "oC10Files1": [
                     "webUIFilesCopy",
                     "webUIFavorites",
-                    "webUIMarkdownEditor",
+                    "webUISimpleEditor",
```
pascalwengerter, line 81: @diocas that would need a rename of https://github.com/owncloud/web/tree/master/tests/acceptance/features/webUIMarkdownEditor in order to work :) since it's mapping feature files/folders

packages/web-app-simple-editor/src/index.js
kulmann: I hope you don't get too overexited with this. Turns out that mime type guessing from the backend isn't always what you'd expect. E.g. ogg video files still get announced with `audio/ogg` as mime type. But I continued the PR and it's close to being merged.

kulmann:
@diocas could you search&replace various combinations of the word `markdown editor`? I found at least one more `markdowneditor` in an acceptance test feature file. That should probably be `simpleeditor` now in that case. But maybe there's more.
Also I added `[full-ci]` to the PR title so that upon your next push the CI won't stop on the first failed acceptance test pipeline but will push through with all pipelines. That should give us a clear picture of what's missing for merge.

diocas:
Btw, what about the disabling of md file creation?

pascalwengerter:
> Btw, what about the disabling of md file creation?

What do you mean?

kulmann:
> Btw, what about the disabling of md file creation?

Can you open an issue please?

tests/acceptance/features/webUIFilesActionMenu/fileFolderActionMenu.feature
```diff
@@ -18,7 +18,7 @@ Feature: User can see the file or folder actions menu options
     Then the app-sidebar for file "lorem.txt" should be visible on the webUI
     And only the following items with default items should be visible in the actions menu on the webUI
       | items                  |
-      | open in markdowneditor |
+      | open in simpleditor |
```
kulmann, line 21:
```suggestion
      | open in simpleeditor |
```

kulmann (changes requested):
Found a typo. Fixing that should finally turn the test suite green 🥳

kulmann (approved):

PR #6669, opened 2022-03-25
Enable typescript in single file components

dschmidt:
## Description
This enables typescript in `.vue` files and adjusts the linter to deal with it properly.

As you can see we need to wrap components in `defineComponent` from `composition-api` to make the type checking work at all. Before you ask: yes, it's completely valid to wrap components that use only options api.

## Motivation and Context
After I have replaced the old Vetur extension with the new [Volar](https://marketplace.visualstudio.com/items?itemName=johnsoncodehk.volar) extension in VS Code (this PR is mostly based on its README), I have some basic autocompletion and type checking working in `.vue` files. 
It works very well in the templates, not so much in the `script` part of the SFC. For some reason autocompletion and typechecking seems to only really work in the `data()` function, not in setup(), not in methods, not in computeds, just in data.

We probably can now have better types with vuetype or vue-tsc, but that might require more involved changes, so I'd leave that for later. I think this is already a great improvement :) 

Screenshot_20220325_204252

Notice:
- `fetchResources` (a method) can be autocompleted, just like any other vue instance members: Vue api, props, data members, methods ....
- `highlightedFile` is accordingly recognized as a valid reference
- `foobardoobardoo` is marked invalid

Interestingly my IDE marks `foobardoobardoo` as not being defined, but the compilation does not fail 🤷🏻
We might be able to tune tsconfig.json to have stricter warnings/errors. But again I'd leave that for later.

### DIFF
--- a/packages/web-runtime/src/App.vue
+++ b/packages/web-runtime/src/App.vue
@@ -29,15 +29,16 @@
     />
   </div>
 </template>
-<script>
+<script lang="ts">
 import { mapGetters, mapState, mapActions } from 'vuex'
 import SkipTo from './components/SkipTo.vue'
 import LayoutApplication from './layouts/Application.vue'
 import LayoutLoading from './layouts/Loading.vue'
 import LayoutPlain from './layouts/Plain.vue'
 import { getBackendVersion, getWebVersion } from './container/versions'
+import { defineComponent } from '@vue/composition-api'
 
-export default {
+export default defineComponent({
   components: {
     SkipTo
   },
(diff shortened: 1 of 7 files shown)


pascalwengerter (approved):
Fine with me, waiting for a technical decision by @kulmann and @fschade

kulmann (approved):
Love it! 🥇

fschade (changes requested):
lat me take a look later

packages/web-runtime/src/App.vue
```diff
@@ -130,7 +131,7 @@ export default {
   },
 
   metaInfo() {
-    const metaInfo = {}
+    const metaInfo: any = {}
```
fschade, line 134: ```const metaInfo: { [key: string]: unknown } = {}```

fschade (approved):
LOVING it, thank you.

I would say adopting it to other places in the codebase can be done incremental. 

i only found 2 optional improvements

Issue #6549, opened 2022-03-08
Cached parts of the interface by switching between accounts not updating


elizavetaRa:
### Steps to reproduce
1. In CERNBox login with a normal account
2. Logout and login with the lightweight account, that gets another list of navItems and doesn't have quota

Implementation of different lists of navItems here: https://github.com/cernbox/web/blob/0dfa4717578cedfaa0cf976e027b20c212f76b6f/packages/web-app-files/src/index.js#L169-L171

### Expected behaviour
The navItems and the quota are changed corresponding to the account

### Actual behaviour
The list of navItems for the lightweight accounts is mixed with cached navItems from previous account. Quota is shown where it doesn't exist. By opening the app in new tab the problem disappears.

![Untitled_ Mar 8, 2022 11_49 AM](https://user-images.githubusercontent.com/7430156/157224653-c176c900-401c-467d-8a74-64e3c18d5425.gif)

elizavetaRa:
@kulmann I found the issue with quota: if the backend returns no quota (our case with lightweight accounts), the quota in store is not reset from the previous logged in user inside one browser tab

diocas:
@elizavetaRa I guess this can be reproduced in master if you - for example - add a new menu entry on the left sidebar?

PR #6694, opened 2022-03-31
Reset store on logout

elizavetaRa:
## Description
When logging out, only some parts of vuex store were reset to default. This caused bugs by switching to another account that has some other/missing settings. For example, if the account has no quota, the quota of the previously logged in account was shown. We have fixed this by resetting the store on logout with reset function (vuex extensions library).

## Related Issue
- Fixes [6549](https://github.com/owncloud/web/issues/6549)

## How Has This Been Tested?
 logging in, logging out, logging in
- test case 1: logging in with account with quota, logging out, logging in with account without quota --> quota not shown
- test case 2:  logging in with account with navItems, logging out, logging in with account with another set of navItems --> they are displayed correctly
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-runtime/src/store/user.js
+++ b/packages/web-runtime/src/store/user.js
@@ -25,12 +25,11 @@ const actions = {
     if (context.state.id === '') {
       return
     }
-    // reset user to default state
-    context.commit('SET_USER', state)
-    // reset capabilities to default state
-    context.commit('SET_CAPABILITIES', { capabilities: null, version: null })
-    // set userReady to false
-    context.commit('SET_USER_READY', false)
+    // reset user
+    this.reset({ self: false, nested: false, modules: { user: { self: true } } })
+
+    // clear dynamic navItems
+    context.dispatch('clearDynamicNavItems')
 
     // clear oidc client state
     vueAuthInstance.clearLoginState()
@@ -41,9 +40,6 @@ const actions = {
       dispatch('cleanUpLoginState')
       dispatch('hideModal')
       dispatch('loadSettingsValues')
-
-      // Reset store
-      this.reset()
 
       // Force redirect to login
       if (forceRedirect) {
--- a/packages/web-runtime/src/store/user.js
+++ b/packages/web-runtime/src/store/user.js
@@ -42,6 +42,9 @@ const actions = {
       dispatch('hideModal')
       dispatch('loadSettingsValues')
 
+      // Reset store
+      this.reset()
+
       // Force redirect to login
       if (forceRedirect) {
         router.push({ name: 'login' })
--- a/changelog/unreleased/bugfix-resetting-store-on-logout
+++ b/changelog/unreleased/bugfix-resetting-store-on-logout
@@ -1,0 +1,6 @@
+Bugfix: Resetting store on logout
+
+When logging out, only some parts of vuex store were reset to default. This caused bugs by switching to another account that has some other/missing settings. For example, if the account has no quota, the quota of the previously logged in account was shown. We have fixed this by resetting the user store module on logout with reset function (vuex extensions library) and creating an action to reset dynamic nav items.
+
+https://github.com/owncloud/web/pull/6694
+https://github.com/owncloud/web/issues/6549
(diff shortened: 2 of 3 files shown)


ownclouders:
Results for <strong>oC10Files3</strong> https://drone.owncloud.com/owncloud/web/24286/18/1

<p>


### webUILogin-oauthLogin_feature-L22.png
  


</p></details>

pascalwengerter:
> Results for **oC10Files3** https://drone.owncloud.com/owncloud/web/24286/18/1
> 💥 The acceptance tests failed on retry. Please find the screenshots inside ...
> 💥 The acceptance tests pipeline failed. The build has been cancelled.

Fails as the redirect with OAuth2 in combination with OC10 fails with these changes. I'll try to take a look with @kulmann tomorrow

packages/web-runtime/src/store/user.js
```diff
@@ -42,6 +42,9 @@ const actions = {
       dispatch('hideModal')
       dispatch('loadSettingsValues')
 
+      // Reset store
+      this.reset()
```
kulmann, line 46: To be blunt, I don't understand where the `reset` is coming from. 🤔 at least it's not something that `vuex` provides.
What you could use is `replaceState`, but that's probably too much for what you want to achieve.
There are some actions dispatched in the lines above. Maybe it makes more sense for the time being to use those for resetting state. For example the `SET_USER` mutation doesn't reset the quota, you could fix that. And a new action in the `navigation` store module for clearing the dynamic nav items would help as well.

kulmann (changes requested):

packages/web-runtime/src/store/user.js
elizavetaRa: The reset function is part of the functionality of vuex-extensions library that is used in web for example to use createStore. It uses replaceState: https://github.com/huybuidac/vuex-extensions/blob/c80fda5e0f1bed3362cc2064086499d2b2e80fa5/src/index.js#L18

I thought, resetting the whole store state might prevent possible future bugs by login switches between different account types. 

If parts of store need to be keeped on logout it's possible to do partial reset of selected children/grandchildren, for example cleanUpLoginState could be implemented like this
`this.reset({ 
  self: false,
  nested: false,
  modules: { user: { self: true} }
})`
instead of manually cleaning user, his capabilities and userReady.

What do you think of it?

kulmann: Hm... found the issue now why resetting the whole store is not going to work. The `navigation` module has dynamic and static nav items. The `static` nav items are pushed into the store during bootstrap. But the bootstrap doesn't run again after login. Maybe that should. 😅 We can look into that during our beta phase.

I like your approach of only resetting particular modules. For the user it definitely makes sense like you proposed it.
For solving another issue that you had after relogin you need to reset the dynamic nav items only from the `navigation` store in the runtime.

kulmann: For the record, this is what happens on re-login: tries to access on of the static nav items and since they are cleared this results in an exception.
Screenshot 2022-04-14 at 09 41 13

elizavetaRa:
@kulmann implemented an action for clearing dynamic items as you asked, testing by us with switching between different accounts works correctly.

kulmann:
> @kulmann implemented an action for clearing dynamic items as you asked, testing by us with switching between different accounts works correctly.

Hm. 🤔  it needs at least a rebase to pass CI. But I see another issue: The oCIS settings service UI adds dynamic nav items and there is no logic in your PR for re-adding certain dynamic nav items. See https://github.com/owncloud/ocis/blob/44e8b6e7be947efcabda1074776c045164021cf3/extensions/settings/ui/components/SettingsApp.vue#L106
In other words, while your PR fixes what you do, it kills functionality of the settings UI.
I'm starting to think that we need a complete re-bootstrapping...

kulmann (approved):
Turns out that clearing the dynamic nav items is ok. The `ADD_NAV_ITEM` mutation in the web runtime is implemented in a way that allows setting the same nav item again over and over. In fact, all known apps use the dynamic nav item init during render time, thus they already get re-created when needed.

Thank you for the contribution! 🤩

PR #6715, opened 2022-04-01
Move compiler and linter settings to subpackages

dschmidt:
## Description
This PR moves config files for eslint, ts and prettier (and their respective dependency declarations) into subpackages, which can be reused in other repos later on.

# Related Issue

- Fixes <issue_link>

## Motivation and Context
I asked @fschade and @pascalwengerter whether we could update our linters in oC Web to align it with the versions I had already upgraded to in ownBrander. Pascal brought up a wip repo he had been working on for centralizing configs. After Florian suggested I should create subpackages with their best practice names (and after he helped me debug some issues), I talked to @kulmann and he suggested we should simply do everything mono-repo style in oC Web and it turns out: it works quite fine.

## How Has This Been Tested?
- just a tooling restructuring, if CI is green, we're good

## Types of changes
- [x] Technical debt
- [ ] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:
- [ ] update versions
- [ ] make sure dependabot sees the package.json and sends prs to update deps
- [ ] carefully review where what dependencies have been moved where, whether they should be marked as devDependencies, as dependencies or peerDependencies etc.  (@fschade to the rescue! 🦸🏻 )
- [ ] publish config packages to npm on release
- [ ] figure out a versioning strategy / automation (?) (@fschade to the rescue again, please! 🦸🏻 )
- [ ] validate babel settings are properly picked up  (easier for eslint, prettier and ts configs as the build fails without them as linting changes or by enabling strict in tsconfig which makes our builds fail)
- [ ]  should we add yarn.lock files for every package (with deps?)?


fschade:
regarding deps:
things like eslint or eslint-rules should go into peerDependencies, as a result the consuming application is responsible providing them.

In our monorepo it's not that important (!!only here, not the published version of those rules) because they all end up in roots node_modules.

a good read is https://withblue.ink/2020/06/07/is-this-a-dependency-or-devdependency.html

fschade:
@dschmidt can you explain what you mean with version strategy? i think first and foremost we have to decide how we publish those kind of individual packages.... e.g. lerna, ....

then the versions should be independent and we could follow semver.

packages/tsconfig/tsconfig.json
```diff
@@ -0,0 +1,96 @@
+{
```
fschade, line 1: would love o have the schema here

```
...
  "$schema": "https://json.schemastore.org/tsconfig",
...
```

.prettierrc.json
```diff
@@ -1,6 +1 @@
-{
-  "printWidth": 100,
-  "singleQuote": true,
-  "semi": false,
-  "trailingComma": "none"
-}
+"@owncloud/prettier-config"
```
kulmann, line 1:
```suggestion
"@ownclouders/prettier-config"
```

kulmann (changes requested):
Found one missed `@owncloud` to `@ownclouders` rename. ;-) Could you also rebase this PR? it has some conflicting files.

PR #6718, opened 2022-04-04
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
## Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## Motivation and Context

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated .

## How Has This Been Tested?
I tested the return of each statement to make sure it's still the same as before

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/enhancement-replace-deprecated-substr
+++ b/changelog/unreleased/enhancement-replace-deprecated-substr
@@ -1,0 +1,7 @@
+Enhancement: Replace deprecated String.prototype.substr()
+
+We've replaced all occurrences of the deprecated String.prototype.substr()
+function with String.prototype.slice() which works similarly but isn't
+deprecated.
+
+https://github.com/owncloud/web/pull/6718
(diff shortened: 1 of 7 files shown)


pascalwengerter:
@CommanderRoot thanks for the effort, could you rebase and squash your commits so we can merge? :)

kulmann (approved):
Thank's a lot for your contribution 🤩

PR #6727, opened 2022-04-04
Vertically centered loading spinner or error message on startup

diocas:
We discussed this some weeks ago. Noop change, since when the app loads it will change it.
But with this, while the app is not loaded, we see the spinner centered.

---
Can we also commit a hover effect to the quick actions on the files table? The idea is to hide them and only show on hover, like so:
image

### DIFF
--- a/packages/web-container/index.html.ejs
+++ b/packages/web-container/index.html.ejs
@@ -16,6 +16,10 @@
   <script src="js/require.js?<%= data.compilationTimestamp %>"></script>
   <% } %>
   <style>
+    html,
+    body {
+      height: 100%
+    }
     .splash-banner {
       display: flex;
       flex-direction: row;


kulmann:
@wkloucek could you check why this PR fails in drone? It fails in the `skip-if-unchanged` step and I'd suspect that it's failing because the previous commit from the respective fork is unknown in our repo. See https://drone.owncloud.com/owncloud/web/24411/9/2 for the drone run. Error message is
```
time="2022-04-04T13:19:07Z" level=error msg="execution failed: commit from DRONE_COMMIT_BEFORE not found: object not found"
```

wkloucek:
> @wkloucek could you check why this PR fails in drone? It fails in the `skip-if-unchanged` step and I'd suspect that it's failing because the previous commit from the respective fork is unknown in our repo. See https://drone.owncloud.com/owncloud/web/24411/9/2 for the drone run. Error message is
> 
> ```
> time="2022-04-04T13:19:07Z" level=error msg="execution failed: commit from DRONE_COMMIT_BEFORE not found: object not found"
> ```

nope, actually it clones the fork: https://drone.owncloud.com/owncloud/web/24411/9/1. So that has needs to be known (if it exists). We've also a pending issue https://github.com/owncloud-ci/drone-skip-pipeline/issues/18.

I can look into it tomorrow

wkloucek:
@kulmann this PR will be unblocked by https://github.com/owncloud/web/pull/6743, but needs to be rebased. Problem is that the commit is not on the fork, but on our repo only (outdated master on the fork) but we don't have pulled our repo...

kulmann:
> @kulmann this PR will be unblocked by #6743, but needs to be rebased. Problem is that the commit is not on the fork, but on our repo only (outdated master on the fork) but we don't have pulled our repo...

Nice, thank you so much for taking care! 🤗

kulmann:
@diocas could you rebase after the PR mentioned above is merged? Otherwise I can make a new PR and cherry-pick your commit if you want to.

Regarding your question about showing quick actions on hover: we had some user testing with a click dummy of that and it wasn't perceived in a good way. Long story short, for the time being we want to keep the quick actions visible. Could be a configurable option, but that would have to wait a bit or be developed by you. In any case, a ticket would be good.

diocas:
@kulmann yes, I had already subscribed to that PR to get the merge notification :)

PR #6736, opened 2022-04-05
Add selection range for rename to select resource name without extension

lookacat:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-app-files/src/mixins/actions/rename.js
```diff
@@ -81,13 +81,16 @@ export default {
           : this.$gettext('Rename file %{name}'),
         { name: resources[0].name }
       )
+      const resourceName = resources[0].name
+      const nameWithoutExtensionLength = resourceName ? resources[0].name.split('.')[0].length : -1
```
kulmann:
```suggestion
      const nameWithoutExtensionLength = resourceName ? resourceName.split('.')[0].length : -1
```

JanAckermann (approved):

packages/web-app-files/src/mixins/actions/rename.js
```diff
@@ -81,13 +81,16 @@ export default {
           : this.$gettext('Rename file %{name}'),
         { name: resources[0].name }
       )
+      const resourceName = resources[0].name
+      const nameWithoutExtensionLength = resourceName ? resourceName.split('.')[0].length : -1
```
kulmann, line 85: This doesn't work for file names that have more than one `.`. It is possible that the filename part as any amount of `.`s. And there are also some file extensions  that consist of more than one segment, e.g. `.tar.gz`. For a start I'd be fine that you select everything until the last dot. That would break `archive.tar.gz` as it will select too much (`archive.tar` instead of `archive`). In a followup PR you could add a mapping table for file extensions that are known to have more than one segment (like e.g. the beforementioned `.tar.gz`).

kulmann (changes requested):
Functionality wise there is an issue, see comment.
Apart from that I'd like to see some test coverage here, ideally as unit tests for the rename action. It makes sense to test the selection range for:
- folder name without `.` -> selects full name
- folder name with `.` -> selects full name (because the `extension` part is not an actual file extension)
- file -> select everything but the **last** `.` and whatever comes after it, e.g. `file.name.with.dots.txt` selects `file.name.with.dots` and doesn't select the `.txt`.

kulmann:
> Functionality wise there is an issue, see comment. Apart from that I'd like to see some test coverage here, ideally as unit tests for the rename action. It makes sense to test the selection range for:
> 
> * folder name without `.` -> selects full name
> * folder name with `.` -> selects full name (because the `extension` part is not an actual file extension)
> * file -> select everything but the **last** `.` and whatever comes after it, e.g. `file.name.with.dots.txt` selects `file.name.with.dots` and doesn't select the `.txt`.

For easy as possible unit test writing you could extract the code that calculates the selection, based on the resource, into a helper and then only test that helper. There also is a ts interface for the `Resource` which you could make use of.

packages/web-app-files/src/mixins/actions/rename.js
lookacat: I see, yes thought of that too but realized we dont have the file extension on the resource, but creating a mapping file sounds good to me. Will do that in a second PR.

lookacat:
@kulmann yes sound good to me 👍🏼

kulmann (dismissed):

ownclouders:
Results for <strong>oCISSharingAndUpload</strong> https://drone.owncloud.com/owncloud/web/24481/66/1

<p>


### webUIUpload-upload_feature-L94.png
  


</p></details>

lookacat:
Dont merge yet, will implement a better extension function for resources.ts

packages/web-app-files/src/helpers/resource/resource.ts
```diff
@@ -57,3 +58,24 @@ export const extractStorageId = (id?: string): string => {
   }
   return id.indexOf('!') >= 0 ? id.split('!')[0] : ''
 }
+
+export const extractNameWithoutExtension = (resource?: Resource): string => {
+  const extension = resource.extension || ''
+  const name = resource.name || ''
+  if (!extension.length || !name.length) return name
```
kulmann, line 65:
```suggestion
  if (!extension.length) return name
```
should be sufficient, right?

packages/web-app-files/src/helpers/resource/resource.ts
```diff
@@ -57,3 +58,24 @@ export const extractStorageId = (id?: string): string => {
...
+  const extensionIndexInName = name.lastIndexOf(`.${extension}`)
+  return name.substring(0, extensionIndexInName)
+}
+
+export const extractExtensionFromFile = (file: string): string => {
+  const parts = file.split('.')
+  for (let i = 0; i < parts.length; i++) {
+    const possibleExtension = parts.slice(i, parts.length).join('.')
+    if (!fileExtensions.complex.includes(possibleExtension)) continue
```
kulmann, line 74: please invert to
```js
if (fileExtensions.complex.includes(possibleExtension)) {
  return possibleExtension
}
```
the `continue` is not needed.

packages/web-app-files/src/helpers/resource/resource.ts
```diff
@@ -57,3 +58,24 @@ export const extractStorageId = (id?: string): string => {
...
+
+export const extractNameWithoutExtension = (resource?: Resource): string => {
+  const extension = resource.extension || ''
+  const name = resource.name || ''
+  if (!extension.length || !name.length) return name
+  const extensionIndexInName = name.lastIndexOf(`.${extension}`)
+  return name.substring(0, extensionIndexInName)
+}
+
+export const extractExtensionFromFile = (file: string): string => {
```
kulmann, line 70: Since this is in the `resource.ts` helper, I'd expect that you don't work with a name string, but with a `Resource` here, like in the other function.
First step: check if `resource.type === 'folder'`, if yes return empty string.
Second step: split the file name on a `.`, if it only has two parts you can immediately return the second part. With those two steps we return super quick 99% of the time.
Only after that it makes sense to go into the complex extension checks.

kulmann (changes requested):

kulmann:
> Dont merge yet, will implement a better extension function for resources.ts

You can enforce this by converting your PR back to draft mode. 😉

kulmann (approved):
LGTM 👍

PR #6746, opened 2022-04-06
Fix #5716: 

### LINKED ISSUE #5716: Provide peerDependencies to standalone apps
### Steps to reproduce
1. Create a standalone app (one that is not built inside this repo)
2. Declare `vuex` or `luxon` as `peerDependencies`
3. Try to load app
4. See it fail because `requirejs` does not know how to provide `vuex` or `luxon`

```
failed to load app http://localhost:8082/assets/standalone-app.js Error: Script error for "vuex", needed by: http://localhost:8082/assets/standalone-app.js
https://requirejs.org/docs/errors.html#scripterror
    at makeError (require.js:168)
    at HTMLScriptElement.onScriptError (require.js:1738)
```

### Expected behaviour
Apps load fine without shipping their own copy of `vuex` or `luxon`.
I'm not sure if we want to *somehow* provide all modules to apps or just some whitelisted modules. 
I think it makes sense for at least `vuex` and `luxon`.

### Actual behaviour
Apps fail to load because `requirejs` does not know about these modules.

Related code:
https://github.com/owncloud/web/blob/b9a8989b51ff462c32b608a67785230f21285a5c/packages/web-runtime/src/index.js#L269-L314
https://github.com/owncloud/web/blob/b9a8989b51ff462c32b608a67785230f21285a5c/packages/web-runtime/src/index.js#L117-L124

dschmidt:
## Description
This defines `@vue/composition-api`, `vuex`and `luxon` as modules in `requirejs` and provide them that way to applications.
That allows us to bundle apps without those libraries by marking them as `external` in rollup config.

## Related Issue

- Fixes #5716

## Motivation and Context
This way externally built applications don't need to bundle their own versions of libs. While it worked for `vuex` and `luxon` to ship copies of them in applications, bundling `@vue/composition-api` broke our whole runtime in weird ways (i.e. getCurrentInstance()` returned `null` in composables used in the application layout). Moreover, it's obviously undesirable to ship libraries multiple times to reduce file size and loading time.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


tests/unit/config/jest.init.js
```diff
@@ -2,6 +2,8 @@ import { config } from '@vue/test-utils'
 import fetchMock from 'jest-fetch-mock'
 import 'regenerator-runtime/runtime'
 
+window.define = jest.fn()
```
fschade, line 5: is this needed only for unit tests or should we have it inside the integration tests too?

fschade (approved):
LGTM, one small thing which is optional if the tests pass without erroring

PR #6763, opened 2022-04-08
[full-ci] Rename SimpleEditor to TextEditor

kulmann:
## Description
As hinted by a lot of user feedback from todays reviews event we re-renamed the markdown editor to `Text Editor`.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


kulmann:
cc @diocas

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -157,7 +157,7 @@ const rewriteDeprecatedAppNames = (
 ): RuntimeConfiguration => {
   const appAliases = [
     { name: 'preview', oldName: 'media-viewer' },
-    { name: 'simple-editor', oldName: 'markdown-editor' }
+    { name: 'text-editor', oldName: 'markdown-editor' }
```
pascalwengerter, line 160: Could have been handy to make oldName an array of strings to check against multiple old aliases but not relevant for not I'd say

pascalwengerter (approved):

packages/web-runtime/src/container/bootstrap.ts
kulmann: since it's not exposed we can always do so if we need it. prefer to keep it as simple as possible for now. :-)

Issue #6764, opened 2022-04-11
Unit test failure in recent nightly build and needs investigation


SwikritiT:
https://drone.owncloud.com/owncloud/web/24623/8/6
https://drone.owncloud.com/owncloud/web/24624/8/6

```console
Summary of all failing tests
FAIL packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js (29.552 s)
  ● SharedWithOthers view › when the wrapper is not loading anymore › when length of paginated resources is greater than zero › should load the resource table with correct props

    expect(received).toMatchSnapshot()

    Snapshot name: `SharedWithOthers view when the wrapper is not loading anymore when length of paginated resources is greater than zero should load the resource table with correct props 1`

    - Snapshot  - 4
    + Received  + 4

    @@ -32,11 +32,11 @@
            <td class="oc-td oc-table-cell oc-table-cell-align-right oc-table-cell-align-middle oc-table-cell-width-auto oc-text-nowrap oc-table-data-cell oc-table-data-cell-sharedWith">
              <oc-button-stub type="button" size="medium" submit="button" variation="inverse" appearance="raw" justifycontent="center" gapsize="medium">
                <oc-avatars-stub items="[object Object]" stacked="true" istooltipdisplayed="true" maxdisplayed="3" accessibledescription=" This folder is shared via 1 link" class="resource-table-people"></oc-avatars-stub>
              </oc-button-stub>
            </td>
    -       <td class="oc-td oc-table-cell oc-table-cell-align-right oc-table-cell-align-middle oc-table-cell-width-auto oc-text-nowrap oc-table-data-cell oc-table-data-cell-sdate" aria-label="9 months ago (June 9, 2021, 12:44 PM UTC)"><span tabindex="0">9 months ago</span></td>
    +       <td class="oc-td oc-table-cell oc-table-cell-align-right oc-table-cell-align-middle oc-table-cell-width-auto oc-text-nowrap oc-table-data-cell oc-table-data-cell-sdate" aria-label="10 months ago (June 9, 2021, 12:44 PM UTC)"><span tabindex="0">10 months ago</span></td>
            <td class="oc-td oc-table-cell oc-table-cell-align-right oc-table-cell-align-middle oc-table-cell-width-auto oc-text-nowrap oc-table-data-cell oc-table-data-cell-actions oc-pr-m">
              <div class="resource-table-actions">
[...]

PR #6765, opened 2022-04-11
[unit-tests] fix failing `packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js`

kiranparajuli589:
## Description
unit test at `packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js`
>SharedWithOthers view › when the wrapper is not loading anymore › when length of paginated resources is greater than zero › should load the resource table with correct props

started failing in recent ci builds. check the related issue for more details

Imo the problem is with the relative sdate used in the `ResourceTable` component used by `SharedWithOthers` view. We use resources from the fixtures with constant stime set.

The calculated relative sdate was 9 months previously and is 10 months now. It will keep on rising every month. So maybe it's better to only assert for props rather matching the whole html snapshot.

With this PR:
- match only props for the `ResourceTable`
- update related screenshots

### Related
Fixes https://github.com/owncloud/web/issues/6764

### DIFF
--- a/packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js
+++ b/packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js
@@ -8,7 +8,7 @@ import { buildSharedResource } from '../../../../src/helpers/resources'
 import { Settings, DateTime } from 'luxon'
 const resourcesList = SharedData.json().ocs.data.map((resource) => buildSharedResource(resource))
 
-const expectedNow = DateTime.local(2022, 1, 1, 23, 0, 0);
+const expectedNow = DateTime.local(2022, 1, 1, 23, 0, 0)
 
 const router = {
   push: jest.fn(),
--- a/packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js
+++ b/packages/web-app-files/tests/unit/views/shares/SharedWithOthers.spec.js
@@ -5,8 +5,10 @@ import SharedWithOthers from '@files/src/views/shares/SharedWithOthers.vue'
 import SharedData from '@/__fixtures__/sharedFiles.js'
 import { createLocationShares } from '../../../../src/router'
 import { buildSharedResource } from '../../../../src/helpers/resources'
-import { Settings } from 'luxon'
+import { Settings, DateTime } from 'luxon'
 const resourcesList = SharedData.json().ocs.data.map((resource) => buildSharedResource(resource))
+
+const expectedNow = DateTime.local(2022, 1, 1, 23, 0, 0);
 
 const router = {
   push: jest.fn(),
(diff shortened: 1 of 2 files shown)


kiranparajuli589:
oh, for me, locally the tests generates `"sdate": "Wed, 09 Jun 2021 18:22:24 +0545"` but in ci, `"sdate": "Wed, 09 Jun 2021 18:22:24 +0000"`

kiranparajuli589:
```console
    -       "sdate": "Wed, 09 Jun 2021 18:29:17 +0000",
    +       "sdate": "Wed, 09 Jun 2021 12:44:17 +0000",
```
now the timestamp is not matching :smile:

kiranparajuli589:
now the tests do not pass locally in my local setup. Is this a locale problem for my local setup or something else? afaik, while date conversion, the locale is set to `this.currentLanguage` in the `ResourceTable` component, which is mocked as
```
$language: {current: "en"}
```
in the `jest.init.js` file for tests. should be the same for ci and local setup.

cc @kulmann @pascalwengerter @fschade

kulmann:
IMO we should mock the current date/time for luxons relative date calculation. first search result: https://dev.to/thealexkates/how-to-mock-luxon-s-datetime-now-when-using-tdd-3j6o (didn't look into the article). As this is blocking any CI run now - do you have time for this as high prio @kiranparajuli589 ? or should someone from ownCloud look into it?

kiranparajuli589:
@kulmann, i need you one more suggestion here :) the relative date calculation is part of `ResourceTable` component which is being used by `SharedWithOthers` view. imo, tests for the relative date should be inside `ResourceTable.spec.js` where we may have to mock luxon datetime as you've suggested. Here, i've refactored the tests to completely stub `ResouceTable` component used within and just assert the feed props for it. Should I try passing tests with mocking here?

kulmann:
> @kulmann, i need your one more suggestion here :) the relative date calculation is part of `ResourceTable` component which is being used by `SharedWithOthers` view. imo, tests for the relative date should be inside `ResourceTable.spec.js` where we have to mock luxon datetime as you've suggested. Here, i've refactored the tests to completely stub `ResouceTable` component used within and just assert the feed props for it. Should I try passing tests with mocking here?

Tests should only ever check real input and output. The output of the components is html, so intercepting that by directly checking the props is not favourable I guess. Could you try with mocking instead?

kiranparajuli589:
> > @kulmann, i need your one more suggestion here :) the relative date calculation is part of `ResourceTable` component which is being used by `SharedWithOthers` view. imo, tests for the relative date should be inside `ResourceTable.spec.js` where we have to mock luxon datetime as you've suggested. Here, i've refactored the tests to completely stub `ResouceTable` component used within and just assert the feed props for it. Should I try passing tests with mocking here?
> 
> Tests should only ever check real input and output. The output of the components is html, so intercepting that by directly checking the props is not favourable I guess. Could you try with mocking instead?

sure :rocket:

kulmann (approved):
😍

PR #6774, opened 2022-04-12
add demo users flag to docker dev env

fschade:
## Description
Literally the same as, #6773 but for the local dev env we use. 
`demo users are now opt in and therefore re-enabled in this PR`

## Related Issue
- extends #6773


pascalwengerter (approved):
Damn yes, it felt like we were missing a spot in the other PR 🥵

PR #6779, opened 2022-04-12
Add search to usersManagement

### LINKED ISSUE #6674: Filtering and sorting of users list
# Description

## User Stories

* > As an admin I want to identify a secific user in the users list, so that I can later modify their data.

## Value
- user list filtering & sorting

## Acceptance Criteria
- Columns can be sorted (exept actions column)
- Filter
  - Placeholder text and aria-label: "Search users" 
  - userlist gets filtered with every keystroke
  - filtered columns are: Username, Email, Name. ~~Status~~
  - matching searchphrase gets highlighted in every result in the respective fields (bkulmann hat codeexamples for this)

## Definition of ready
[ ] everybody needs to understand the value written in the user story
[ ] acceptance criteria has to be defined
[ ] all dependencies of the user story need to be identified
[ ] feature should be seen from an end user perspective
[ ] user story has to be estimated
[ ] story points need to be less then 20

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] codre review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

fschade:
handover @JanAckermann 

## Description
image

## Related Issue

- Fixes #6674

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-user-management/src/components/Users/UsersList.vue
+++ b/packages/web-app-user-management/src/components/Users/UsersList.vue
@@ -1,5 +1,6 @@
 <template>
   <oc-table
+    ref="table"
     :sort-by="sortBy"
     :sort-dir="sortDir"
     :fields="fields"
@@ -52,6 +53,7 @@
 import { onBeforeUnmount, ref } from '@vue/composition-api'
 import { Registry } from '../../services'
 import Fuse from 'fuse.js'
+import Mark from 'mark.js'
 
 export default {
   name: 'UsersList',
@@ -81,7 +83,8 @@ export default {
   data() {
     return {
       sortBy: 'onPremisesSamAccountName',
-      sortDir: 'asc'
+      sortDir: 'asc',
+      markInstance: null
     }
   },
   computed: {
@@ -145,6 +148,24 @@ export default {
       return this.selectedUsers.map((user) => user.id)
     }
   },
+  watch: {
+    searchTerm() {
+      if (!this.markInstance) {
+        return
+      }
+      this.markInstance.unmark()
+      this.markInstance.mark(this.searchTerm, {
+        element: 'span',
+        className: 'highlight-mark',
+        exclude: ['th *', 'tfoot *']
+      })
+    }
+  },
+  mounted() {
+    this.$nextTick(() => {
+      this.markInstance = new Mark(this.$refs.table.$el)
+    })
+  },
   methods: {
     filter(users, searchTerm) {
       if (!(searchTerm || '').trim()) {
@@ -188,3 +209,9 @@ export default {
   }
 }
 </script>
+
+<style lang="scss">
+.highlight-mark {
+  background: yellow;
+}
+</style>
--- a/packages/web-app-user-management/src/components/Users/UsersList.vue
+++ b/packages/web-app-user-management/src/components/Users/UsersList.vue
@@ -213,5 +213,6 @@ export default {
 <style lang="scss">
 .highlight-mark {
   background: yellow;
+  color: var(--oc-color-text-muted);
 }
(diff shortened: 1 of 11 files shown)


packages/web-app-user-management/src/components/Users/UsersList.vue
```diff
@@ -125,13 +138,27 @@ export default {
...
   methods: {
+    filter(users, searchTerm) {
+      if (!searchTerm) {
+        return users
+      }
+      const usersSearchEngine = new Fuse(users, {
+        includeScore: true,
+        useExtendedSearch: true,
+        threshold: 0.3,
+        keys: ['displayName', 'mail', 'onPremisesSamAccountName']
```
kulmann, line 157: Would also make sense to filter by role names if possible

packages/web-app-user-management/src/components/Users/UsersList.vue
```diff
@@ -125,13 +138,27 @@ export default {
...
+      const orderedUsers = this.orderBy(this.users, this.sortBy, this.sortDir === 'desc')
+      return this.filter(orderedUsers, this.searchTerm)
     },
     highlighted() {
       return this.selectedUsers.map((user) => user.id)
     }
   },
   methods: {
+    filter(users, searchTerm) {
+      if (!searchTerm) {
```
kulmann, line 150:
```suggestion
      if (!(searchTerm || '').trim()) {
```

packages/web-app-user-management/src/search/filter/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor(router: VueRouter) {
+    super()
+
+    this.id = 'usersManagement.filter'
+    this.router = router
+  }
+
+  public get label(): string {
+    return $gettext(`Search ${kind(this.router.currentRoute)} ↵`)
```
kulmann, line 32: This doesn't work in transifex. I'm afraid you have to go the repetitive way here with:
```js
switch(kind(this.router.currentRoute)) {
  case "users": return $gettext("Search users ↵")
  case "groups": return $gettext("Search groups ↵")
}
```
otherwise those strings will never land in transifex. additionally extraction from .ts files doesn't work with our current setup, so it's probably broken anyway 🙈

kulmann (changes requested):

packages/web-app-user-management/src/components/Users/UsersList.vue
JanAckermann: Will be done later

kulmann (changes requested):
one weird effect: mark.js highlights all occurrences of the search term, while the filtering doesn't work that way. This leads to filtering by `admin`, not seeing all users that have the admin role (only those who have admin in their username, name or email) but still getting the role `Admin` highlighted. Another indicator that we should refactor the role into the user, but of course for a separate PR and not something that blocks this PR. Could you open a bug ticket please?

Filtering and highlighting works really well.

I see a contrast issue though: in dark mode the text is barely readable. Please open a bug ticket for that as well.

Changelog item is missing (for that I have to `request changes` 😅 will approve once you added it)

kulmann (approved):
😍

Issue #5634, opened 2021-08-05
[Bug] Guests are displayed as "Users"


tbsbdr:
### Steps to reproduce
1. OC10.8: login to classic ui
2. share a file with a guest (enter an email)
3. switch to Web UI
4. select the same, shared file and open "People" section the sidebar
5. guest share with email gets displayed as "User" share (which is wrong, its a guest)

![image](https://user-images.githubusercontent.com/26610733/128336173-98e93e36-331c-4d03-8eae-5ab89203c457.png)

### Expected behaviour
Display "Guest" for guest shares

### Actual behaviour
Guest shares are displayed as "User" shares

## Background
- The Web UI relys on the share-type of oc10
- oC10 provides sharetype = 0 for guest shares which we won't change any more (sharetype 0 = user; sharetype 4 = guest)
- @kulmann mentioned that there should be a flag llike "isGuest" which we could use to display the right role in WebUI+oC10

tbsbdr:
@JanAckermann do you maybe have an idea how to distinguish guests from users? @kulmann mentioned some flag or tag "isguest" do you maybe know more?

OCS api doesnt know about guests:

> The share’s type. This can be one of:
> 
> ```
> - 0 = user
> - 1 = group
> - 3 = public link
> - 6 = federated cloud share
> ```

https://doc.owncloud.com/server/next/developer_manual/core/apis/ocs-share-api.html

JanAckermann:
@tbsbdr we only know this by config flag in oc_preferences. We can learn the API

kulmann:
In oc10 you can do `$this->config->getUserValue($uid, 'owncloud', 'isGuest', false);` to determine in the php code if a user is a guest user. The classic UI is utilizing that information. To my knowledge there is no API that makes this information available. Best would be to use the proper `guest` share type in OCS responses. 😞 but I don't know the implications of such a change.

JanAckermann:
@kulmann With ownCloud 10.7 I implemented user types that are `user` or `guest`.
We can emit this information via the share API, for example by adding a  new field called `<share_with_user_type/>`

I think breaking changes to the share api are not favorable.

JanAckermann:
![image](https://user-images.githubusercontent.com/26169327/164727845-34a5bbea-c48c-4571-bc2d-80a24e2ddf4f.png)

JanAckermann:
Oc10 Backend part: https://github.com/owncloud/core/pull/40013

JanAckermann:
[...]

PR #6813, opened 2022-04-26
Guest share enhancements

JanAckermann:
## Description
**NOTE: For oc10, this requires a higher ownCloud 10 version than 10.9.1** 

Bugfix: Indicate guest shares

We've fixed a bug, where guest shares were not correctly indicated and shown as users shares
at the share panel in the right sidebar.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5634

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

image
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-app-files/src/services/folder/loaderSharedWithOthers.ts
```diff
@@ -51,6 +51,9 @@ export class FolderLoaderSharedWithOthers implements FolderLoader {
         )
       }
 
+      const user = store.state.user
+      resources = resources.filter((r) => r.shareOwner === user.id)
```
JanAckermann, line 55: shared_with_me = false seems not to work, therefore filtering

packages/web-app-files/src/helpers/resources.ts
```diff
@@ -279,41 +279,41 @@ function addSharedWithToShares(shares) {
       previousShare?.storage_id === share.storage_id &&
       previousShare?.file_source === share.file_source
     ) {
-      if (ShareTypes.containsAnyValue(ShareTypes.authenticated, [share.share_type])) {
+      if (ShareTypes.containsAnyValue(ShareTypes.authenticated, [parseInt(share.share_type)])) {
```
JanAckermann, line 282: Note: owncloud-sdk will return properties casted to string, therefore we need to parse a lot

packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
```diff
@@ -13,6 +13,13 @@
         :width="48"
         class="files-collaborators-collaborator-indicator"
       />
+      <oc-avatar-guest
```
kulmann, line 16: Was this not working with the `oc-avatar-item` below? That one was intended to reflect the correct data based on the given share type. Which should work in theory 😅

packages/web-app-files/src/services/folder/loaderSharedViaLink.ts
```diff
@@ -23,14 +23,12 @@ export class FolderLoaderSharedViaLink implements FolderLoader {
     return useTask(function* (signal1, signal2) {
       store.commit('Files/CLEAR_CURRENT_FILES_LIST')
 
-      let resources = yield client.requests.ocs({
-        service: 'apps/files_sharing',
-        action: '/api/v1/shares?format=json&share_types=3&include_tags=false',
-        method: 'GET'
+      let resources = yield client.shares.getShares('', {
+        share_types: '3',
```
kulmann, line 27: Please use `ShareTypes.link.value` here (cast it to string)

packages/web-app-files/src/services/folder/loaderSharedWithOthers.ts
kulmann: Did you manage to confirm that this happens with oCIS backend only? If yes please create an issue in the ocis repo and add the link here together with a `FIXME` comment.

kulmann: As far as I understood a request to `shares` should only return the outgoing shares, having `shared_with_me=true` flips it around to only incoming shares. But it's also possible that I just got that wrong. 😅

packages/web-app-files/tests/unit/components/SideBar/Shares/Collaborators/InviteCollaborator/AutocompleteItem.spec.js
```diff
@@ -25,7 +25,13 @@ describe('AutocompleteItem component', () => {
       expect(wrapper.find('oc-icon-stub').attributes().name).toEqual('group')
       expect(wrapper.find('oc-icon-stub').rootNode.key).toEqual('avatar-group')
     })
-    it.each([3, 4, 6])(
+    it('should display a guest icon for guest shares', () => {
+      const wrapper = createWrapper({ shareType: 4 })
```
kulmann, line 29: I know that ints were used before, but please use the respective `ShareType.xyz.value` here and in line 34 instead of magic numbers.

kulmann (changes requested):

packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
JanAckermann: As mentioned before in chat: 
:icon="shareTypeKey" would be "guest",  but we want to have the icon "global"

packages/web-app-files/src/services/folder/loaderSharedWithOthers.ts
JanAckermann: Done

packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
kulmann: What about a computed prop `shareTypeIcon` where you return `global` if share type is guest, and in all other cases just the value of `shareTypeKey`? then you can keep the template clean and only have a different case for icon selection.

JanAckermann: I have the feeling, that adding the icon to the share type, would be much cleaner and future proof. 
Do you disagree?

kulmann (approved):
nice! 😎

PR #6831, opened 2022-04-28
Increase amount of retries when a tus upload fails

JammingBen:
## Description
1 retry on failed chunk-uploads seems to be too little (although we shouldn't have failed uploads in an ideal world). However, the default value in Uppy is 4 retries, probably for a good reason...

## Types of changes

- [x] Technical debt
- [ ] Tests


JammingBen:
Let's see if it has an effect on the flakiness of our pipelines.

fschade (approved):
i wonder a bit why we have that option there and not just rely on the defaults!? I#m sure there's a reason... if yes LGTM 🚀 

`When uploading a chunk fails, automatically try again after the millisecond intervals specified in this array. By default, we first retry instantly; if that fails, we retry after 1 second; if that fails, we retry after 3 seconds, etc.`

JammingBen:
> i wonder a bit why we have that option there and not just rely on the defaults

This value does not only affect the amount of retries, it also specifies the time after which a failed upload will be retried. My intention was to keep that a bit lower than the default.

PR #6834, opened 2022-04-28
[full-ci] Fixed some occurences of invalid $gettext usage

kulmann:
Fixes some invalid `$gettext` usages.

### Todo
- [x] fix remaining 1 empty msgid in web-app-files
- [x] fix remaining 2 empty msgids in web-runtime

### DIFF
--- a/packages/web-app-files/src/views/LocationPicker.vue
+++ b/packages/web-app-files/src/views/LocationPicker.vue
@@ -299,18 +299,12 @@ export default {
     currentHint() {
       if (this.currentAction === 'move') {
         return this.$gettext(
-          `Navigate into the desired folder and move selected resources into it.
-          You can navigate into a folder by clicking on its name.
-          To navigate back, you can click on the breadcrumbs.
-          Resources will be moved into the folder where you are currently located.`
+          'Navigate into the desired folder and move selected resources into it. You can navigate into a folder by clicking on its name. To navigate back, you can click on the breadcrumbs. Resources will be moved into the folder where you are currently located.'
         )
       }
 
       return this.$gettext(
-        `Navigate into the desired folder and copy selected resources into it.
-        You can navigate into a folder by clicking on its name.
-        To navigate back, you can click on the breadcrumbs.
-        Resources will be copied into the folder where you are currently located.`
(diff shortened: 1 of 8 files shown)


packages/web-app-files/src/views/LocationPicker.vue
```diff
@@ -299,18 +299,12 @@ export default {
...
 
       return this.$gettext(
-        `Navigate into the desired folder and copy selected resources into it.
-        You can navigate into a folder by clicking on its name.
-        To navigate back, you can click on the breadcrumbs.
-        Resources will be copied into the folder where you are currently located.`
+        'Navigate into the desired folder and copy selected resources into it. You can navigate into a folder by clicking on its name. To navigate back, you can click on the breadcrumbs. Resources will be copied into the folder where you are currently located.'
       )
     }
   },
```
dschmidt, line 310: Just for the record: this is probably not invalid per se, just our lib and/or transifex just don't deal well with this.

dschmidt (approved):
🆒

packages/web-app-files/src/views/LocationPicker.vue
fschade: i wonder why, cant it handle \n?

dschmidt: I *think* our lib/tool creates files that look like this

``` 
msgid ""
"You can navigate into a folder by clicking on its name."
"..."
```

which is valid, but apparently transifex does not like the empty `""` prefix string, if I got it all right in my discussion with @kulmann

fschade (approved):

packages/web-app-files/src/views/LocationPicker.vue
kulmann: ... which sounds really weird. Honestly I'm not sure what to believe anymore regarding transifex 🙈

kulmann: Ah, I remember why I changed this 😆 it was never actual multiline text so we can also have it as one line to avoid the empty msgid start.

ownclouders:
Results for <strong>oCISSharingPublic1</strong> https://drone.owncloud.com/owncloud/web/25057/68/1

PR #6835, opened 2022-04-28
Switching over to a git tag of easygettext

kulmann:
we had weird checksum issues with the commit hash. trying with an
archive now.


kulmann:
install step was successful in https://drone.owncloud.com/owncloud/web/25013/6/5
force merging...

Issue #6819, opened 2022-04-27
Open tasks for Uppy


JammingBen:
This issue contains a list of Uppy-related open tasks after https://github.com/owncloud/web/pull/6202 has been merged.

- [x] Polish upload-info-box -> https://github.com/owncloud/web/pull/6837
    - Parent folder links are hard to read in dark mode
    - "Upload successful" message is displayed even on upload errors
- [x] ~Add possibility to remove successful uploads from the upload-info-box -> we decided against this~
- [x] Fix links in the upload-info-box when resources were uploaded to the personal space root -> https://github.com/owncloud/web/pull/6846
- [x] Remove `ProgressBar` component and `inProgress` state  -> https://github.com/owncloud/web/pull/6825
- [x] `UppyService` shouldn't extend `Vue` - use own event implementation instead (https://github.com/owncloud/web/pull/6202/files#r850141613) -> https://github.com/owncloud/web/pull/6853
- [x] Refactor file- and folder-upload-components to one component -> https://github.com/owncloud/web/pull/6859
- [x] Add unit tests -> https://github.com/owncloud/web/pull/6871
- [x] Use tus for uploads to password protected public pages when possible (needs https://github.com/owncloud/ocis/issues/3588 to be fixed first) -> https://github.com/owncloud/web/pull/6890
- [ ] a11y

## needs product decision first:
- [x] Thumbnails for files in the upload-info-box (?) // not sure, yet, if this makes the UI too chaotic. seeing it (with a quick and dirty implementation or with a mockup) side by side with and without thumbnails would be beneficial for decision making @kulmann -> not for now

pascalwengerter:
I'd like to promote fixing #2443 here - folders uploaded via drop-links are getting "flattened"

pascalwengerter:
Also, maybe we can give #2222 another look in the process and finally resolve it through our E2E tests

PR #6202, opened 2021-12-30
[full-ci] Switch upload logic to Uppy

pascalwengerter:
## Description

We've implemented Uppy as a library for handling uploads. This concludes the following features and changes:

- Resumable uploads when the backend supports the Tus-protocol
- A nice looking overview for all files that have been uploaded successfully or failed to upload
- Navigation across Web while uploads are in progress
- Improved rendering of uploadProgress-visualization
- Removed `vue2-dropzone` and `vue-drag-drop` libraries

## Related Issue
- Fixes #6268
- Fixes _tbd_ (maybe there's more)

## Screenshots

image

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] add changelog item(s)
- [x] decide where to initialize/instantiate uppy -> currently not possibly globally due to our architecture.
- [x] make uppy work for OC10 (private&public context, upload-buttons&dropzone)
- [x] make uppy work for oCIS (private&public context, upload-buttons&dropzone)
- [x] make uppy create new folders (if none existing on the backend)
- [x] take care of UploadProgress-visualization
- [x] take care of Resume-button
- [x] make uppy work on FilesDrop.vue
- [x] take care of l10n
- [x] careful code cleanup
- [x] Fix Acceptance tests
- [x] fix unit tests
- [x] File conflict on same name

Open tasks/enhancements: https://github.com/owncloud/web/issues/6819

PR #6846, opened 2022-04-29
Fix root folder links in the upload overlay

JammingBen:
## Description
We've fixed a bug where the folder links in the upload overlay were broken when navigating into another app or file list.

Follow up for https://github.com/owncloud/web/pull/6202

## Related Issue

- Part of https://github.com/owncloud/web/issues/6819

### DIFF
--- a/packages/web-runtime/src/components/UploadInfo.vue
+++ b/packages/web-runtime/src/components/UploadInfo.vue
@@ -120,7 +120,7 @@ export default {
         this.closeInfo()
       }
     })
-    this.$uppyService.subscribe('uploadedFileFetched', (uppyResource, fetchedFile) => {
+    this.$uppyService.subscribe('uploadedFileFetched', ({ uppyResource, fetchedFile }) => {
       this.successfulUploads.push({
         ...fetchedFile,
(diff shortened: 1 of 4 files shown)


ownclouders:
Results for <strong>oCISSharingPublic1</strong> https://drone.owncloud.com/owncloud/web/25081/68/1

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -169,7 +160,8 @@ export default {
         name: targetRoute.name,
         query: targetRoute.query,
         params: {
-          ...(storageId && path && { storageId })
+          ...(storageId && path && { storageId }),
+          storage: targetRoute.params?.storage
```
dschmidt, line 164: Is this really always available? Shouldn't it be like the line above so it gets added conditionally?

JammingBen: I couldn't find a scenario where this wasn't set. But you're right, better be safe :) Need to rebase and push again anyways.

kulmann (approved):
Code LGTM. CI is unhappy though

PR #6856, opened 2022-05-02
[full-ci] Remove hideSearchBar config option in favor of capabilities

### LINKED ISSUE #6806: get rid of hideSearchBar option by using existing capability
Currently web checks a dedicated option to disable the search bar, introduced wcith https://github.com/owncloud/web/pull/3817

However, the capabilities do contain the list of reports that clients can fire against webdav.

For backend search (aka sdk search) the `capabilities.dav.reports` string array has to contain `"search-files"`:
```json
...
{
  "dav":{
    "chunking":"",
    "trashbin":"1.0",
    "reports": ["search-files"],
    "chunkingParallelUploadDisabled":false
  }
}
...
```

Web should rely on that `dav.reports` property in the capabilities instead of an option in the config.

Another report is a `filter-files` report which is used to filter files by favorites and tags ... but that is a different issue.

pascalwengerter:
## Description
We should debate if we really want to remove this (e.g. think of ownBrander) before merging

## Related Issue
- Fixes #6806


ownclouders:
Results for <strong>oCISSharingPublic1</strong> https://drone.owncloud.com/owncloud/web/25136/68/1
- webUISharingPublicBasic/publicLinkEdit.feature:57

pascalwengerter:
@kulmann could be force-merged since it failed with an unrelated false positive in the acceptance tests in https://drone.owncloud.com/owncloud/web/25136. I'll use the PR to update some minor dependencies, too

kulmann (commented):
What about hiding the search field if no search providers are available?

kulmann (approved):

Issue #6878, opened 2022-05-05
Cannot run `oc10` using docker-compose


saw-jan:
Running oc10 with: `docker-compose up oc10`
exists with error:
```bash
...
oc10_1             | Cloning into '/mnt/data/apps/testing'...
oc10_1             | testing enabled
oc10_1             | Cloning into '/var/www/owncloud/apps/oauth2'...
oc10_1             | /usr/bin/entrypoint: line 25: make: command not found
web_oc10 exited with code 127
```

individual-it:
does it depend on what `OC10_IMAGE` is set to?

kulmann:
Can't confirm this 🤔 

```
web_oc10           | make: Entering directory '/var/www/owncloud/apps/oauth2'
web_oc10           | /usr/bin/composer install --no-dev
web_oc10           | Installing dependencies from lock file
web_oc10           | Verifying lock file contents can be installed on current platform.
web_oc10           | Package operations: 1 install, 0 updates, 0 removals
web_oc10           |   - Downloading rowbot/url (1.1.1)
web_oc10           |   - Installing rowbot/url (1.1.1): Extracting archive
web_oc10           | Generating autoload files
web_oc10           | make: Leaving directory '/var/www/owncloud/apps/oauth2'
web_oc10           | oauth2 enabled
```

pascalwengerter:
@saw-jan could you provide more details on your OS, docker (compose) versions and whether you have set a `OC10_IMAGE` env variable?

saw-jan:
> Can't confirm this thinking
> 
> ```
> web_oc10           | make: Entering directory '/var/www/owncloud/apps/oauth2'
> web_oc10           | /usr/bin/composer install --no-dev
> web_oc10           | Installing dependencies from lock file
> web_oc10           | Verifying lock file contents can be installed on current platform.
> web_oc10           | Package operations: 1 install, 0 updates, 0 removals
> web_oc10           |   - Downloading rowbot/url (1.1.1)
> web_oc10           |   - Installing rowbot/url (1.1.1): Extracting archive
> web_oc10           | Generating autoload files
> web_oc10           | make: Leaving directory '/var/www/owncloud/apps/oauth2'
> web_oc10           | oauth2 enabled
> ```

Have you tried removing `web_oc10` and `owncloud/server` docker images?

saw-jan:
My system info:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
[...]

PR #6879, opened 2022-05-05
Install make library

saw-jan:
Fixes https://github.com/owncloud/web/issues/6878
While running `docker-compose up oc10`, the command fails with `make: command not found`
Added `make` so that the command will be executed

### DIFF
--- a/dev/docker/oc10.Dockerfile
+++ b/dev/docker/oc10.Dockerfile
@@ -4,5 +4,6 @@ FROM ${OC10_IMAGE}
 RUN apt -qqy update \
   && apt -qqy --no-install-recommends install \
     bash \
+    make \
   && rm -rf /var/lib/apt/lists/* \
   && apt -qyy clean


individual-it:
how did that ever work?

saw-jan:
> how did that ever work?

Do you mean previously?
I think oc10 image got updated

pascalwengerter:
I can't recall anyone else running into this problem, ever. @fschade can you take a look?

saw-jan:
> I can't recall anyone else running into this problem, ever. @fschade can you take a look?

I removed all the docker images and tried docker-compose and ran into this problem.
Some of us are having the same issue.

saw-jan:
I also tried with `10.9.1-rc2` tag that was last pushed `4 months ago` and I can find `make` command in the container.

pascalwengerter:
> > I can't recall anyone else running into this problem, ever. @fschade can you take a look?
> 
> I removed all the docker images and tried docker-compose and ran into this problem. Some of us are having the same issue.

Could you try and collect more data from what OS are being used and what env variables are set (see my comment in https://github.com/owncloud/web/issues/6878 also)? I can't recall ever running into problems here in the past on different machines, and never came across this with other people either

pascalwengerter (approved):
LGTM, thanks for the swift fix

pascalwengerter:
Side note: I would have suggested a more informative commit message - `install make library` could mean a lot of things, `add missing installation of make in oc10 docker setup step` scopes it more nicely and tells someone in >3 months what happens at a glance :)

PR #6888, opened 2022-05-05
Move unit test coverage calculation to CI && Remove dead code in files store

pascalwengerter:
## Description
Not much to see here, feels good nonetheless


kulmann (approved):
👍 🕵️

Issue #3984, opened 2020-08-21
Notification messages are not being translated


kulmann:
Our current code base suggests, that notification messages will be translated automagically, see:
https://github.com/owncloud/phoenix/blob/51d860e5cb9773f1a4a08ca5c564283285de366b/src/components/MessageBar.vue#L8
However, this is not how gettext works - it works based on extraction of strings from the codebase, pushing them to a translation platform, and making the translated strings available for internal lookup. Pushing a variable into `$gettext` has no effect, as the extraction cannot take place.

We need to
- remove the wrapping `$gettext` function calls from the `MessageBar` code (easy part)
- make sure that all the places that are pushing messages into the vuex store for the `MessageBar` are translated (i.e. use $gettext directly on the strings) (hard part, as there are supposedly a lot of places)

PR #6893, opened 2022-05-06
Remove $gettext func from MessageBar.vue

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/3984

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/AppBar/CreateSpace.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateSpace.vue
@@ -117,9 +117,9 @@ export default {
           })
         })
         .catch((error) => {
+          console.error(error)
           this.showMessage({
             title: this.$gettext('Creating space failed…'),
-            desc: error,
             status: 'danger'
           })
         })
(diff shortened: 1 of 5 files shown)


pascalwengerter (approved):

kulmann:
> we sometimes pass the backend error to the desc property, which I don't think is a good idea, I could take care of that in this pr.

Good spot, we never want to show backend errors in the web ui. Instead they should be logged to the browser console (with `console.error` until we come up with logging lib). If the error makes it possible to show a good description to the user, then that's always a better idea than just having a title. But showing a backend error as description is not good UX.

kulmann:
> @kulmann I might be missing something but went through all 'showMessage' calls and it seems to me, that all titles are translated already

Agreed, your change makes sense to me 👍

kulmann (approved):

fschade (approved):

Issue #6918, opened 2022-05-09
Opening the sidebar for a space removes the "Create Space" button


JammingBen:
### Steps to reproduce
1. Create a space
2. Open the sidebar for this space -> the "Create Space" button will magically disappear

PR #6919, opened 2022-05-09
Prevent 'Create Space'-button from disappearing when sidebar is open

JammingBen:
## Description
We've fixed a bug where the "Create Space"-button would disappear when opening the sidebar for a space.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6918

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/AppBar/AppBar.vue
+++ b/packages/web-app-files/src/components/AppBar/AppBar.vue
@@ -33,7 +33,7 @@
       </div>
       <div class="files-app-bar-actions">
         <div class="oc-flex-1 oc-flex oc-flex-start" style="gap: 15px">
-          <slot v-if="selectedFiles.length === 0" name="actions" />
+          <slot v-if="showActionsOnSelection || selectedFiles.length === 0" name="actions" />
           <size-info v-if="showSelectionInfo" class="oc-visible@l" />
           <batch-actions v-if="showBatchActions" />
         </div>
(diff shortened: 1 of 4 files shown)


ownclouders:
Results for <strong>oCISFiles1</strong> https://drone.owncloud.com/owncloud/web/25432/56/1

<p>


### webUICreateFilesFolders-createFile_feature-L31.png
  


</p></details>

packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -73,7 +73,8 @@ export default {
     hasBulkActions: { type: Boolean, default: false },
     hasSharesNavigation: { type: Boolean, default: false },
     hasSidebarToggle: { type: Boolean, default: true },
-    hasViewOptions: { type: Boolean, default: true }
+    hasViewOptions: { type: Boolean, default: true },
+    showActionsOnSelection: { type: Boolean, default: false }
```
kulmann, line 77: wouldn't setting `hasBulkActions` to `false` achieve the same result? 🤔

kulmann: hm... no, it wouldn't. but should it? 🤔

JammingBen: Not quite, as "Create Space" is not a bulk action. The button would still disappear. Currently, we don't have bulk actions for spaces because you can only select on space at a time.

kulmann: The thing is, it will still show the selection info which makes the create space button disappear.

JammingBen: Yeah you're right... I mean, we currently only need this for spaces. They are kind of a "special" case because they don't have multiselect and a selected space is not visibly selected for the user (other than in the sidebar). Therefore `showActionsOnSelection` would be the easiest fix without touching the rest.

kulmann: You're right, that's probably the most generic we can get and (semantically) merging with other props doesn't make too much sense. Thanks for discussing it 😃

kulmann (approved):

Issue #6913, opened 2022-05-09
show unlimited quota / respect `quota.definition: none`


kulmann:
### Steps to reproduce
1. login to oc10 classic ui as admin
2. set unlimited quota for a user or self
3. switch to new web ui
4. open user menu (click on user name, top right)

### Expected behaviour
Personal quota information shows the used space + that the quota is unlimited

### Actual behaviour
Personal quota information shows everything as if a quota was set, but with the free disk space as quota.

### Spec
the `user.quota.definition` field is set to `none` with both oc10 and ocis as backend for unlimited quota. In that case the `total`, `free` and `relative` fields should be ignored as they don't make sense.

### bonus points
- don't fail hard when `user.quota` doesn't exist
- don't fail hard when fields of `user.quota` don't exist

PR #6923, opened 2022-05-09
fix account-info quota rendering

fschade:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/6913
- https://github.com/owncloud/ocis/issues/3748

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/components/Topbar/UserMenu.vue
+++ b/packages/web-runtime/src/components/Topbar/UserMenu.vue
@@ -58,14 +58,14 @@
           <oc-icon name="cloud" fill-type="line" class="oc-p-xs" />
           <div class="storage-wrapper-text">
             <p class="oc-my-rm">
-              <template v-if="!isNaN(quota.relative) && quota.definition !== 'none'">
+              <template v-if="limitedPersonalStorage">
                 <span v-text="personalStorageLabel" />
                 <br />
               </template>
               <span class="oc-text-small" v-text="personalStorageDetailsLabel" />
             </p>
             <oc-progress
-              v-if="!isNaN(quota.relative) && quota.definition !== 'none'"
+              v-if="limitedPersonalStorage"
               :value="quota.relative"
               :max="100"
               size="small"
@@ -111,6 +111,9 @@ export default {
           total: filesize(total)
         }
       )
+    },
+    limitedPersonalStorage() {
+      return !isNaN(this.quota.relative) && this.quota.definition !== 'none'
     },
     quotaEnabled() {
       return !!this.quota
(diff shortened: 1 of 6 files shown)


packages/web-runtime/tests/unit/components/Topbar/__snapshots__/UserMenu.spec.js.snap
```diff
@@ -29,8 +29,10 @@ exports[`User Menu component when basic quota is set renders a danger quota prog
       <li class="storage-wrapper oc-pl-s">
         <oc-icon-stub name="cloud" filltype="line" accessiblelabel="" type="span" size="medium" variation="passive" color="" class="oc-p-xs"></oc-icon-stub>
         <div class="storage-wrapper-text">
-          <p class="oc-my-rm"><span>Personal storage (91.0% used)</span> <br> <span class="oc-text-small">910 B of 1 kB used</span></p>
-          <oc-progress-stub value="91" max="100" size="small" variation="danger"></oc-progress-stub>
+          <p class="oc-my-rm">
+            <!----> <span class="oc-text-small">910 B of 1 kB used</span>
```
butonic, line 33: Is this a comment? A line comment or until the next occurrence? Looks fishy

pascalwengerter: It's the `<template v-if="xy" .. />` which doesn't fulfill the condition and is hence rendered as an "empty comment" in snapshot tests

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -58,12 +58,15 @@
           <oc-icon name="cloud" fill-type="line" class="oc-p-xs" />
           <div class="storage-wrapper-text">
             <p class="oc-my-rm">
-              <span v-text="personalStorageLabel" />
-              <br />
+              <template v-if="!isNaN(quota.relative) && quota.definition !== 'none'">
+                <span v-text="personalStorageLabel" />
```
pascalwengerter, line 62: They way I read this it can have `personalStorageLabel` followed by `personalStorageDetailsLabel` which provides the same information twice?

butonic (approved):

packages/web-runtime/src/store/user.js
```diff
@@ -263,7 +263,7 @@ const mutations = {
   SET_QUOTA(state, quota) {
     // Turn strings into ints
     quota.free = parseInt(quota.free)
-    quota.relative = parseInt(quota.relative)
+    quota.relative = parseFloat(quota.relative)
```
pascalwengerter, line 266: @fschade took me a bit longer because I needed to investigate why the relative used storage (which the backend reports correctly) always ends up as `0` in the store 💩 also fixed the fixtures where this was reflected, too

Issue #2443, opened 2019-11-12
Upload in file-drop public link flattens folder hierarchy


skshetry:
### Steps to reproduce
1. Create a files-drop link.
2. Open the files-drop link.

### Expected behaviour
Some way to upload folder

### Actual behaviour
No way to upload a folder. Dragging a folder uploads the contents of the folder (recurse through it and flattens them).
Also, using `setValue` and trying to upload folder through `Nightwatch` creates an empty file.

PVince81:
@skshetry mind trying this with OC 10 ?

as far as I remember we don't support subdirectories in files drop mode so this might be expected behavior

skshetry:
@PVince81, Folder uploads in OC 10 are not supported at all. Close this or is this a requirement (it is listed on #2222)?

Also, `phoenix` should not upload folder as an `empty` file and nor should it flatten them and upload..

PVince81:
not a requirement imo, it would mess up the concept and make it more complex as we'd need to avoid file name collisions even across folders.

@pmaier1 thoughts ?

PVince81:
@pmaier1 please reopen if you think we want to change how it works in Phoenix

skshetry:
@PVince81, then, I would not expect the contents of sub-folders getting uploaded in the root of the publicly-shared folder (at least they should be in their respective sub-folders).

PVince81:
@skshetry you're right. We need to block this with a warning.

pascalwengerter:
@skshetry could you retry this with all the changes that have occured in the last ~1,5 years? Should work fine now 😉

dpakach:
@pascalwengerter, uploading folders seems to work but all the files in the folder will be flattened.
for example, if we upload a folder with the following structure
```
├── upload
│   ├── file1.pdf
│   ├── file.ttf
│   └── sub
│       └── file2.pdf
```
it will upload all 3 files `file1.pdf`, `file.ttf`, and `file2.pdf` will be uploaded in the same folder. Is that correct behavior?

Uploading a similar folder in the main files section keeps the folder structure of the originally uploaded folder.

saw-jan:
@pascalwengerter 
Is the current behavior the expected one?

pascalwengerter:
> @pascalwengerter
> Is the current behavior the expected one?

No, should upload the folder including files & subfolders the way they were structured before

kulmann:
[...]

PR #6935, opened 2022-05-10
Create folders that have been uploaded to public drop pages

JammingBen:
## Description
Recursively create folders that have been uploaded to public drop pages instead of flattening the hierarchy.

Note that we currently don't care if a folder already exists. Is this the case, the request for creating folders will fail, hence no upload happens.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/2443

## Types of changes

- [ ] Technical debt
- [ ] Tests


JammingBen:
Currently on hold because of https://github.com/owncloud/web/issues/2443#issuecomment-1122179578. Needs a decision how the expected behavior should be.

JammingBen:
Closing as this is massively outdated. In the future, oCIS will probably handle the folder creation, therefore no need to continue here.

Issue #5886, opened 2021-10-07
Logging out of oC web does not log out of oC classic


Deaddy:
This might be just an annoyance, internally we are also still unsure whether we see this as a huge problem or if it is just a  meh issue for power users during the transition away from oC classic.

 
### Steps to reproduce
1. Run ownCloud web as app for owncloud classic
2. Log into ownCloud web via the menu in the upper left corner
3. Authenticate the app
4. Log out of owncloud web
5. Be still logged in classic owncloud

### Expected behaviour from a technical perspective
Everything works as expected. We have authorized a app, which should be independent of being logged into the classical webinterface or not.
You are being redirected to the authorization page for the webapp, where you can just click again re-authorize.

### Expected behaviour from a user or admin perspective
To be logged out in the classic webinterface as well, I suspect. And maybe not being able to authorize without 

### Actual behaviour
See expected behaviour from a technical perspective. ;-)

### relevant parts of the current setup
owncloud 10.8 with 4.2.0 web app
**skipping all the other stuff because as it is not really relevant to the discussion**

### Possible solutions and discussion
I think before we look at a technical solution the question is whether or not this much of an issue.
- somehow hack around so that logging out the web app also logs out the classic interface
- (attempting to) redirect to the original webinterface where you are possibly still logged in after logout
- accept this behaviour and move on, find a clearer way to communicate to the user
- ???

tbsbdr:
@kulmann in some usecases this is a severe bug if the logout is done for security reasons. though there is a workaround (switch to oc10 and logout here), the workaround might not be obvious to every user. imo this p1 bug should be amongst top 3 in the queue ;-) would you agree?

kulmann:
Apparently the backend session handling is really broken for the combination of oauth2 and classic login. There is https://github.com/owncloud/web/issues/7018 as the other facet of the "client logout issue" (logout in classic doesn't logout in web).

PR #6939, opened 2022-05-11
Logout in oc10 while click logout in web 

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/5886

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/store/user.js
+++ b/packages/web-runtime/src/store/user.js
@@ -34,17 +34,23 @@ const actions = {
     // clear oidc client state
     vueAuthInstance.clearLoginState()
   },
-  async logout({ dispatch }) {
-    const logoutFinalizer = (forceRedirect = false) => {
-      // Remove signed in user
-      dispatch('cleanUpLoginState')
-      dispatch('hideModal')
-      dispatch('loadSettingsValues')
+  async logout(context) {
+    const logoutFinalizer = (isOauth2 = false) => {
+      // Force redirect
+      if (isOauth2) {
+        const logoutUrl = context.getters?.configuration?.auth?.logoutUrl
 
-      // Force redirect to login
-      if (forceRedirect) {
+        if (logoutUrl) {
+          return (window.location = logoutUrl)
+        }
+
         router.push({ name: 'login' })
       }
+
+      // Remove signed in user
+      context.dispatch('cleanUpLoginState')
+      context.dispatch('hideModal')
+      context.dispatch('loadSettingsValues')
     }
     const u = await vueAuthInstance.getStoredUserObject()
 
--- a/packages/web-runtime/src/store/user.js
+++ b/packages/web-runtime/src/store/user.js
@@ -38,7 +38,13 @@ const actions = {
     const logoutFinalizer = (isOauth2 = false) => {
       // Force redirect
       if (isOauth2) {
-        const logoutUrl = context.getters?.configuration?.auth?.logoutUrl
+        let logoutUrl
+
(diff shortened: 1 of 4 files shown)


packages/web-runtime/src/store/user.js
```diff
@@ -34,17 +34,23 @@ const actions = {
...
-    const logoutFinalizer = (forceRedirect = false) => {
-      // Remove signed in user
-      dispatch('cleanUpLoginState')
-      dispatch('hideModal')
-      dispatch('loadSettingsValues')
+  async logout(context) {
+    const logoutFinalizer = (isOauth2 = false) => {
+      // Force redirect
+      if (isOauth2) {
+        const logoutUrl = context.getters?.configuration?.auth?.logoutUrl
```
kulmann, line 41: Would it make sense to use configuration.server concatenated with "index.php/logout" as fallback?

packages/web-runtime/src/store/user.js
```diff
@@ -34,17 +34,23 @@ const actions = {
...
+        if (logoutUrl) {
+          return (window.location = logoutUrl)
+        }
 
-      // Force redirect to login
-      if (forceRedirect) {
         router.push({ name: 'login' })
       }
+
+      // Remove signed in user
```
kulmann, line 50: Don't we need the store cleanups before the redirect?

packages/web-runtime/src/store/user.js
JanAckermann: good idea!

packages/web-runtime/src/store/user.js
JanAckermann: Well, this is something I can't answer yet, we do a redirect, so I thought the store should be cleaned while accessing the page the next time. 
I moved this code path, as there is already a redirect happening in one of these dispatched actions

ownclouders:
Results for <strong>oCISSharingAndUpload</strong> https://drone.owncloud.com/owncloud/web/25502/67/1

packages/web-runtime/src/store/user.js
```diff
@@ -31,18 +31,26 @@ const actions = {
     // clear dynamic navItems
     context.dispatch('clearDynamicNavItems')
 
-    // clear oidc client state
-    vueAuthInstance.clearLoginState()
+    if (options.clearOIDCLoginState) {
+      // clear oidc client state
+      vueAuthInstance.clearLoginState()
+    }
```
JanAckermann, line 37: @kulmann this code path triggers a redirect already and interferes with redirecting to the oc10 log out,
I don't call ` vueAuthInstance.clearLoginState()`anymore if we are logged in via oauth2, is this ok?

kulmann: Is it really sufficient for oauth2 to skip that step? o_O

kulmann (approved):
LGTM 🚀 
FYI, found some other issues with cyclic action dispatching... will tackle that in a subsequent PR that refactors most of the auth related code. 😅

pascuflow:
Pulled latest from master to test this, after logging out of Web, if you then log in with different credentials, you are still logged in as previous user.

pascuflow:
Fixed by adding `sessionStorage.clear()` before logout() is called in web/packages/web-runtime/src/components/Topbar/UserMenu.vue

Issue #4839, opened 2021-03-23
Reload files list of current view after accepting a share via notification


kulmann:
We'll introduce a regression when https://github.com/owncloud/web/pull/4627 gets merged:

### Steps to reproduce
1. login as alice
2. share a file or folder with bob
3. login as bob
4. have a notification about the new share
5. accept the share with the respective button within the notification

### Expected behaviour
The file list of the current view (personal, shared with me, others are not relevant) updates and shows the accepted resource / reflects that the resource is now accepted

### Actual behaviour
The view doesn't update

### Server configuration
**ownCloud version:** [10.6,)

### Thoughts about when to solve it
When we migrated to vue3 and typescript we can refactor the code for loading resources. That is very repetitive as of writing this issue and involves 3-4 steps in every view (fetching the resources, transforming it into a format that is understood by e.g. the `oc-table-files` component, loading previews, loading indicators). When we refactor that we should also think about how to utilize resource loading from components outside of the files app, or if it even makes sense to move parts of the code to the runtime, so that accepting notifications doesn't have to rely on code of an extension.

Issue #4247, opened 2020-10-27
"Shares" folder does not appear when using `share_folder` and accepting remote share via notification


individual-it:
### Steps to reproduce
1. set the share_folder to Shares (./occ config:system:set share_folder --value=Shares)
2. from a remote server share a folder via federation sharing
3. reload page on the receivers side
4. accept share via notification

### Expected behaviour
"Shares" folder should appear

### Actual behaviour
"Shares" folder is not visible till the page gets reloaded

![accept-notification](https://user-images.githubusercontent.com/2425577/97291276-d5118b80-1871-11eb-8ddb-84728525df86.gif)

Issue #1774, opened 2019-08-12
after accepting a remote share the files-list is not reloaded


individual-it:
### Steps to reproduce
1. accept an incoming remote share

### Expected behaviour
files list should be reloaded to show the new resource immediately like its doe with local shares that need to be accepted

### Actual behaviour
file list is not reloaded, user has to reload the page to see the new resource

PVince81:
in OC 10 this was tricky because we had to have some kind of event and listener that detects that the notification is of a certain type and was dismissed.

LukasHirt:
Yeah, we have the same. There is already implemented refresh for accepting pending share which was done in the same way - not a nice solution.

pascalwengerter:
@kulmann @LukasHirt can one of you check if this has been fixed by #5130, too?

PR #6942, opened 2022-05-11
[full-ci] Reload file list after accepting a remote share

JammingBen:
## Description
We've fixed a bug where the file list would not reload after accepting a remote share.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/1774
- Fixes https://github.com/owncloud/web/issues/4247
- Fixes https://github.com/owncloud/web/issues/4839

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-accept-remote-share-reload
+++ b/changelog/unreleased/bugfix-accept-remote-share-reload
@@ -4,3 +4,5 @@ We've fixed a bug where the file list would not reload after accepting a remote 
 
 https://github.com/owncloud/web/pull/6942
 https://github.com/owncloud/web/issues/1774
+https://github.com/owncloud/web/issues/4247
+https://github.com/owncloud/web/issues/4839
--- a/changelog/unreleased/bugfix-accept-remote-share-reload
+++ b/changelog/unreleased/bugfix-accept-remote-share-reload
@@ -1,0 +1,6 @@
+Bugfix: Reload file list after accepting a remote share
+
+We've fixed a bug where the file list would not reload after accepting a remote share.
+
+https://github.com/owncloud/web/pull/6942
+https://github.com/owncloud/web/issues/1774
--- a/packages/web-runtime/src/components/Topbar/Notifications.vue
+++ b/packages/web-runtime/src/components/Topbar/Notifications.vue
@@ -59,18 +59,11 @@
 </template>
 <script>
 import { mapGetters, mapActions } from 'vuex'
-import { useAppDefaults } from 'web-pkg/src/composables'
+import { bus } from 'web-pkg/src/instance'
 
 export default {
-  setup() {
-    return {
-      ...useAppDefaults({
-        applicationName: 'runtime'
-      })
-    }
-  },
   computed: {
-    ...mapGetters(['activeNotifications', 'configuration', 'user']),
+    ...mapGetters(['activeNotifications', 'configuration']),
 
     notificationsLabel() {
       return this.$gettext('Notifications')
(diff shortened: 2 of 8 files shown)


packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -86,26 +93,29 @@ export default {
...
+                if (shareRoot === currentPath) {
+                  this.loadFolder(`files/${this.user.id}/${currentPath}`)
+                  return
+                }
+              }
+
+              if (path) {
+                const itemPath = path.slice(0, path.lastIndexOf('/') + 1)
+                if (itemPath === currentPath) this.loadFolder(`files/${this.user.id}/${itemPath}`)
+              }
```
JammingBen, line 115: Not 100% sure if we should include such logic here, or if we should move that to the files app and just trigger an event here.

ownclouders:
Results for <strong>oC10iPhoneNotifications</strong> https://drone.owncloud.com/owncloud/web/25507/48/1

JammingBen:
I'm not sure about the best solution here. I've now implemented it to always reload the file list when accepting a federated share. If we only want to reload when necessary, it gets a bit more complicated:

* Are we currently in the directory where the federated file will be located?
* Is there a default share folder like `Shares`? Does this folder already exist or does it need to be created (see https://github.com/owncloud/web/issues/4247)?

Especially the latter would require some more logic. Logic, that shouldn't exist in the web runtime. Hence we would need to write something in the files app.

@kulmann Any opinions here?

changelog/unreleased/bugfix-accept-remote-share-reload
```diff
@@ -0,0 +1,6 @@
+Bugfix: Reload file list after accepting a remote share
+
+We've fixed a bug where the file list would not reload after accepting a remote share.
+
+https://github.com/owncloud/web/pull/6942
+https://github.com/owncloud/web/issues/1774
```
pascalwengerter, line 6: Fixes two more issues, no?

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -86,27 +86,22 @@ export default {
...
             json.ocs.data.forEach((item) => {
-              const path = item.path.slice(0, item.path.lastIndexOf('/') + 1)
-              const absolutePath = this.$route.params.item ? this.$route.params.item : '/'
-              if (path === absolutePath) this.reloadFilesList(path)
+              const currentPath = this.$route.params.item ? `/${this.$route.params.item}` : '/'
+              const { state, path, file_target: fileTarget } = item
+
+              // accepted federated share
+              if (state === 0 && fileTarget) {
+                bus.publish('app.files.list.load')
```
pascalwengerter, line 94: Do we want to return here to avoid publishing `app.files.list.load` twice?

pascalwengerter (commented):
In general LGTM, would like to hear @fschade & @kulmann 's opinions first though

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -86,27 +86,23 @@ export default {
...
           res.json().then((json) => {
             json.ocs.data.forEach((item) => {
-              const path = item.path.slice(0, item.path.lastIndexOf('/') + 1)
-              const absolutePath = this.$route.params.item ? this.$route.params.item : '/'
-              if (path === absolutePath) this.reloadFilesList(path)
+              const currentPath = this.$route.params.item ? `/${this.$route.params.item}` : '/'
+              const { state, path, file_target: fileTarget } = item
+
+              // accepted federated share
+              if (state === 0 && fileTarget) {
```
kulmann, line 93: Please use the `ShareStatus.accepted` constant here

kulmann (approved):
Nice improvement to use the event bus for this. 👍  Small nitpick regarding a state value comparison.

kulmann:
> * Is there a default share folder like `Shares`? Does this folder already exist or does it need to be created (see ["Shares" folder does not appear when using `share_folder` and accepting remote share via notification #4247](https://github.com/owncloud/web/issues/4247))?
> 
> Especially the latter would require some more logic. Logic, that shouldn't exist in the web runtime. Hence we would need to write something in the files app.

That should happen in the backend anyway. We don't create the default `share_folder` location in the frontend code. Or did I misread your comment?

JammingBen:
> That should happen in the backend anyway. We don't create the default share_folder location in the frontend code. Or did I misread your comment?

The backend creates it, yes, but the frontend would need to know that to accordingly reload the page. Anyway, I think we're fine by always reloading when accepting a share, no matter in which directory you're currently in.

kulmann:
> > That should happen in the backend anyway. We don't create the default share_folder location in the frontend code. Or did I misread your comment?
> 
> The backend creates it, yes, but the frontend would need to know that to accordingly reload the page. Anyway, I think we're fine by always reloading when accepting a share, no matter in which directory you're currently in.

Yes, same opinion 😇 Fine this way for the time being.

PR #6946, opened 2022-05-11
[full-ci] Bump oCIS commit ID after flock fix & glauth/accounts service removal

pascalwengerter:
## Description
Should make CI way more reliable since the personal space flock issue has been resolved in Reva.

Also, accounts&glauth service have been removed from oCIS. Now, the oc10 acceptance tests relying on OIDC use an old/outdated oCIS instance for their glauth service, which we should eventually replace with a proper solution


rhafer:
Somewhat related: https://github.com/owncloud/ocis/issues/3770

JammingBen (approved):

Issue #6849, opened 2022-04-29
Break long words


diocas:
I'm not sure where to put this ticket, if here or ods.
If I have a very long path (without spaces, which I would say is pretty normal), there are some elements of the UI that don't break the words and this causes elements to become scrollable (and this might cause the UI to flicker).

Places where I've noticed this: metadata table on the right sidebar (i.e "shared via"), and tippy tooltips.

pascalwengerter:
Tooltips have been fixed in https://github.com/owncloud/owncloud-design-system/pull/2137 (needs another ODS RC and bump in web), I guess with #6959 we can close this ticket for now

pascalwengerter:
Closing via #6959

Issue #6702, opened 2022-03-31
Customization of UI elements


elizavetaRa:
It would be great to have the possibility to customize the following elements:

- feedback link (href, ariaLabel, description)
- tooltip help for "invite" searchfield (description text). The tooltip can be shown, if the description text exists.

kulmann:
> * feedback link (href, ariaLabel, description)
Customizing the `href` is easy to implement via `options` object in the `config.json`. We can allow adding text for `aria-label` and `description` to that as well, but that won't be translatable then.

> * tooltip help for "invite" searchfield (description text). The tooltip can be shown, if the description text exists.
IMO overriding text in the ui needs a concept first. Now starting to add overrides to some config or theme file and using that througout the code base will result in a big mess to be frank. :-/
More context: we currently use english text directly in the code, export that to transifex nightly, import translations from transifex nightly and store them in a json file (one per app). The english text is the `key` in that json file / translations mapping and thus not stable (because it might be adjusted in the code at any time, e.g. because of typos, new better wording, etc). IMO we need a stable key for our translations. Then and only then it would be easy / easier to let strings be overridden via theming.

> * Sidebar should have the possibility to inject content/extensions into the "Details" section (like [cernbox@52bbf24e](https://github.com/cernbox/web/commit/52bbf24e))
This one would deserve it's own ticket. ;-) This is one use case for a fine grained extension system we envision for web. The core team at ownCloud won't have time to work on that before our oCIS GA.

kulmann:
> * tooltip help for "invite" searchfield (description text). The tooltip can be shown, if the description text exists.

If you can limit that requirements for just a handful of locations in the code then we can have a somewhat dirty patch here upstream.

kulmann:
> * Sidebar should have the possibility to inject content/extensions into the "Details" section (like [cernbox@52bbf24e](https://github.com/cernbox/web/commit/52bbf24e))

Compromise at this point (also for the string adjustment above) could be that you introduce an `eos` capability and we hardcode some stuff that will only be shown when `eos=true`.

kulmann:
@elizavetaRa discussed internally again today, an additional option in the config.json would be better than a capability as it is only relevant for the web ui / has no implications for the backend or other clients.

dj4oC:
Before this can be acted on, the following needs to be resolved:

[...]

PR #6959, opened 2022-05-12
Changes to links in sidebar, fix tooltips

javfg:
This PR addresses some additions and fixes to the sidebar links:

* Adds an `isEos` setting into `config.json` as discussed in https://github.com/owncloud/web/issues/6702#issuecomment-1117256159
* Adds direct link to the side bar (maybe related to https://github.com/owncloud/web/issues/6840)
* Adds an eos path to the side bar, shown if `isEos`
* Ellipsizes long links so the sidebar does not get a scrollbar. Along with https://github.com/owncloud/owncloud-design-system/pull/2137, these two should close https://github.com/owncloud/web/issues/6849.
* Normalized some sidebar table method names (some of them were `titleSomething`, all are now `labelSomething`)
* Removed the colons and made the headers of the table `nowrap`. This makes the table looks better, as there is no need to have colons in headers; and they are not extremely long

Current looks:

image

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -23,7 +23,7 @@
       </div>
       <table class="details-table" :aria-label="detailsTableLabel">
         <tr v-if="hasTimestamp" data-testid="timestamp">
-          <th scope="col" class="oc-pr-s" v-text="timestampTitle" />
+          <th scope="col" class="oc-pr-s" v-text="timestampLabel" />
           <td>
             <oc-button
               v-if="showVersions"
@@ -51,7 +51,7 @@
           </td>
         </tr>
         <tr v-if="ownerDisplayName" data-testid="ownerDisplayName">
-          <th scope="col" class="oc-pr-s" v-text="ownerTitle" />
+          <th scope="col" class="oc-pr-s" v-text="ownerLabel" />
           <td>
             <p class="oc-m-rm">
               {{ ownerDisplayName }}
@@ -63,11 +63,11 @@
           </td>
         </tr>
         <tr v-if="showSize" data-testid="sizeInfo">
-          <th scope="col" class="oc-pr-s" v-text="sizeTitle" />
+          <th scope="col" class="oc-pr-s" v-text="sizeLabel" />
           <td v-text="getResourceSize(file.size)" />
         </tr>
         <tr v-if="showVersions" data-testid="versionsInfo">
-          <th scope="col" class="oc-pr-s" v-text="versionsTitle" />
+          <th scope="col" class="oc-pr-s" v-text="versionsLabel" />
           <td>
             <oc-button
               v-oc-tooltip="seeVersionsLabel"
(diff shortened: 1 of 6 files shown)


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -131,16 +198,85 @@ export default defineComponent({
...
 
+    sharedWithLabel() {
+      return this.$gettext('Shared with')
+    },
     file() {
       return this.displayedItem.value
     },
 
+    isEos() {
+      return !!this.configuration.options?.eos
```
pascalwengerter, line 223: Please add the eos config options to https://owncloud.dev/clients/web/getting-started/#options (markdown file in the `docs/` directory in this repo) and ideally also with a default value in https://github.com/owncloud/web/blob/72fba4481b9d3b9592ed064b0f39d8503518df66/packages/web-runtime/src/store/config.js#L40

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
             />
           </td>
         </tr>
+        <tr v-if="runningOnEos">
+          <th scope="col" class="oc-pr-s" v-text="eosPathLabel" />
+          <td>
+            <div class="oc-flex oc-flex-middle oc-flex-between oc-width-1-1">
+              <div class="oc-flex oc-flex-middle">
```
pascalwengerter, line 85: The wrapping div around the `<p>` tag is not necessary

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                <oc-icon v-else key="oc-copy-to-clipboard-copy" name="clipboard" />
+              </oc-button>
+            </div>
+          </td>
+        </tr>
+        <tr>
+          <th scope="col" class="oc-pr-s" v-text="directLinkLabel" />
+          <td>
+            <div class="oc-flex oc-flex-middle oc-flex-between oc-width-1-1">
+              <div class="oc-flex oc-flex-middle">
```
pascalwengerter, line 117: The wrapping div around the `<p>` tag is not necessary

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                <p
+                  ref="filePath"
+                  v-oc-tooltip="file.path"
+                  class="oc-my-rm oc-text-truncate"
+                  v-text="file.path"
+                />
+              </div>
+              <oc-button
+                oc-tooltip="Copy EOS path"
```
pascalwengerter, line 94: Please use a `label` prop and `this.$gettext()` here

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                appearance="raw"
+                aria-label="Copy EOS path"
```
pascalwengerter, line 96: Please use a `label` prop and `this.$gettext()` here

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                class="oc-files-private-link-copy-url"
```
pascalwengerter, line 97: The class looks wrong

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+              <div class="oc-flex oc-flex-middle">
+                <p
+                  ref="directLink"
+                  v-oc-tooltip="directLink"
+                  class="oc-my-rm oc-text-truncate"
+                  v-text="directLink"
+                />
+              </div>
+              <oc-button
+                oc-tooltip="Copy direct link"
```
pascalwengerter, line 126: Please use a `label` prop and `this.$gettext()` here

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                appearance="raw"
+                aria-label="Copy direct link"
```
pascalwengerter, line 128: Please use a `label` prop and `this.$gettext()` here

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                class="oc-files-private-link-copy-url"
```
pascalwengerter, line 129: The class looks wrong

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+                :variation="copiedDirect ? 'success' : 'passive'"
+                @click="copyDirectLinkToClipboard"
+              >
+                <span text="EOS Path" />
```
pascalwengerter, line 133: Please use a `label` prop and `this.$gettext()` here

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -78,6 +78,70 @@
...
+              </div>
+              <oc-button
+                oc-tooltip="Copy EOS path"
+                appearance="raw"
+                aria-label="Copy EOS path"
+                class="oc-files-private-link-copy-url"
+                :variation="copiedEos ? 'success' : 'passive'"
+                @click="copyEosPathToClipboard"
+              >
+                <span text="EOS Path" />
```
pascalwengerter, line 101: Please use a `label` prop and `this.$gettext()` here

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
pascalwengerter: Actually already exists as `eosPathLabel`, just needs to be used

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -131,16 +198,85 @@ export default defineComponent({
...
+    },
     file() {
       return this.displayedItem.value
     },
 
+    runningOnEos() {
+      return !!this.configuration?.options?.runningOnEos
+    },
+
+    hasSharees() {
```
pascalwengerter, line 226: From here to line 279 are not necessary in this component (possibly rebase-gone-wrong?), could you double-check and delete if you agree?

pascalwengerter (changes requested):
Just tried and it mostly looks good, have some code hygiene/a11y points to fix but that should be a quick one :)

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
javfg: I'm thinking I should remove these (this and line 101). They are not showing because `text` is not a valid attribute so they are effectively empty. If I make them have something inside, we get this:

image

I assume there is no reason to have text there, but correct me if I'm wrong (pinging @diocas).

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
javfg: Same as https://github.com/owncloud/web/pull/6959#discussion_r874611497.

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
javfg: Correct, this was included by mistake. I'll remove it.

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
pascalwengerter: You could use `v-text` and a computed property to show text (`text` itself won't work), but agreed that you could use an icon for "visual" usage and then both aria-label and tooltip on the wrapping button for contextual/screenreader help

javfg:
I've applied all the suggestions (also removed some more stuff which didn't belong, like a getter used in the unneeded code you mentioned).

Also, rebased to latest master.

Let me know if I missed something!

PR #6994, opened 2022-05-17
[full-ci] Enhancement: Implement Copy/Move Conflict Dialog

lookacat:
## Description
See https://github.com/owncloud/web/issues/6996

## Related Issue
- Works towards https://github.com/owncloud/web/issues/6996

## Open tasks:
- [x] https://github.com/owncloud/owncloud-design-system/pull/2149


lookacat:
Needs https://github.com/owncloud/owncloud-design-system/pull/2149 to be linked

packages/web-app-files/src/views/Personal.vue
```diff
@@ -220,84 +217,36 @@ export default defineComponent({
...
   methods: {
     ...mapActions('Files', ['loadPreview']),
-    ...mapActions(['showMessage']),
+    ...mapActions(['showMessage', 'createModal', 'hideModal']),
     ...mapMutations('Files', ['REMOVE_FILE', 'REMOVE_FILE_FROM_SEARCHED', 'REMOVE_FILE_SELECTION']),
 
-    fetchResources,
+    async fetchResources(path, properties) {
+      try {
+        return await this.$client.files.list(path, 1, properties)
```
kulmann, line 225: `fetchResources` comes from the folder loader service. What's the reason for switching to a custom implementation here? only for development and to be reverted?

packages/web-app-files/src/helpers/resource/move.ts
```diff
@@ -0,0 +1,166 @@
...
+    hideModal
+  )
+  const movedResources = []
+
+  for (const resource of resourcesToMove) {
+    const hasConflict = resolvedConflicts.some((e) => e.resource.id === resource.id)
+    if (hasConflict) {
+      const resolveStrategy = resolvedConflicts.find((e) => e.resource.id === resource.id)?.strategy
+      if (resolveStrategy === ResolveStrategy.CANCEL) {
+        continue
```
kulmann, line 149: `cancel` sounds like it should abort the whole move process, not only the current file, i.e. I'd expect a `return` statement here.
`continue` would be the case for `skip`, which might be an additional resolve strategy actually. What do you think? 🤔

packages/web-app-files/src/helpers/resource/move.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+      if (resolveStrategy === ResolveStrategy.REPLACE) {
+        await client.files.delete(join(targetFolder.webDavPath, resource.name))
```
kulmann, line 152: Do we need to delete via client code? I'd expect this to be handled by the backend 🤔 Not sure how it's implement. If we need to handle it via client code then it might be more safe to first rename the file, then move the other one and lastly delete the previously renamed file.

kulmann (changes requested):
I like where this is going 🤩  some conceptual questions in the comments.

packages/web-app-files/src/views/Personal.vue
lookacat: didnt know better to be honest, will try it out 👍🏼

packages/web-app-files/src/helpers/resource/move.ts
lookacat: Cancel equals skip, i think i messed up the wording ^^, but yes cancel implies to stop all

packages/web-app-files/src/helpers/resource/move.ts
lookacat: I've seen an overwrite flag for restoring files in the sdk, not sure if its possible with move, have looked it up and didn't find a way

lookacat: ok fixed it

ownclouders:
Results for <strong>oC10Sharing2</strong> https://drone.owncloud.com/owncloud/web/25916/25/1

<p>


### webUIResharingToRoot-reshareUsers_feature-L25.png
  


</p></details>

changelog/unreleased/enhancement-copy-move-conflict-dialog
```diff
@@ -0,0 +1,6 @@
+Enhancement: Copy/Move conflict dialog
+
+We've added a move conflict dialog for drag and drop
```
pascalwengerter, line 3:
```suggestion
We've added a conflict dialog for moving resources via drag&drop in the files list
```

pascalwengerter (changes requested):
Works fine and is a great improvement!

Only suggestion I have is a differentiation when it's only one file (left) - either use `ngettext` and a separate translation, or remove the checkbox entirely?

Screenshot 2022-06-02 at 13 07 30

lookacat:
Checkbox for single conflict is removed, good idea 👍🏼

pascalwengerter (approved):
I think it's good to merge now from my POV

packages/web-runtime/package.json
```diff
@@ -25,8 +25,8 @@
     "luxon": "^2.4.0",
     "marked": "^4.0.12",
     "oidc-client": "1.11.5",
-    "owncloud-design-system": "^13.1.0-rc.7",
-    "owncloud-sdk": "~3.0.0-alpha.9",
+    "owncloud-design-system": "^13.1.0-rc.8",
```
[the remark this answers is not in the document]
kulmann: Please note that ODS is at rc.9 already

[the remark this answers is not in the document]
pascalwengerter: Rebased, no worries

packages/web-app-files/src/helpers/resource/move.ts
```diff
@@ -0,0 +1,220 @@
...
+  errors,
+  movedResources,
+  showMessage,
+  $gettext,
+  $gettextInterpolate
+) => {
+  if (errors.length === 0) {
+    const count = movedResources.length
+    const title = $gettextInterpolate(
+      $gettext('%{count} item was moved successfully'),
```
pascalwengerter, line 139: I fear we need to use `$ngettext()` here
Screenshot 2022-06-02

lookacat: done 👍🏼

packages/web-app-files/src/helpers/resource/move.ts
```diff
@@ -0,0 +1,223 @@
...
+  // if we implement MERGE, we need to use 'infinity' instead of 1
+  const targetFolderItems = await client.files.list(targetFolder.webDavPath, 1)
+  const targetPath = targetFolder.path
+  const index = targetFolder.webDavPath.lastIndexOf(targetPath)
+  const webDavPrefix = targetFolder.webDavPath.substring(0, index)
+
+  // Collect all conflicting resources
+  const allConflicts = []
+  for (const resource of resourcesToMove) {
+    const potentialTargetWebDavPath = join(webDavPrefix, targetFolder.path, resource.path)
```
kulmann, line 86:
```suggestion
    const potentialTargetWebDavPath = join(webDavPrefix, targetFolder.path, resource.name)
```

kulmann (changes requested):
Found one issue when moving inside subfolders of the respective webdav root (see comment).
Potential followups:
- trigger the conflict dialog for copy & move from location picker (only create followup ticket please. implementation would probably be too hacky right now)
- repeating the same action again leads to naming conflicts that are not handled: (1) move "asdf" to "target/asdf" and select "keep both" -> creates "target/asdf copy" when "target/asdf" already existed before. (2) again, move "asdf" to "target/asdf" and select "keep both" -> tries to create "target/asdf copy", but fails because it already exists from previous move with "keep both".

kulmann (approved):
Nice!

PR #6997, opened 2022-05-17
Port EOS details to details tab in files sidebar

pascalwengerter:
## Description
This ports #6959 over, with slight adoptions:
- unit tests now run with default disabled EOS, and have cases with eos to make sure the component does what we expect
- "direct links" now also are only visible if EOS ins enabled in the config (pretty sure `web` should have dedicated private links for OC10 and a strong incentive for users to use quicklinks/public links)

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -23,7 +23,7 @@
       </div>
       <table class="details-table" :aria-label="detailsTableLabel">
         <tr v-if="hasTimestamp" data-testid="timestamp">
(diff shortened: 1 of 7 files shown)


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -157,10 +220,10 @@ export default defineComponent({
       return this.$gettext('Overview of the information about the selected file')
     },
     shareDateLabel() {
-      return this.$gettext('Shared:')
+      return this.$gettext('Shared')
```
pascalwengerter, line 223: Discussed with PM that omitting the `:` is fine regarding wording

Issue #7006, opened 2022-05-18
Upload via Drag'n Drop doesn't work, if a resource is selected 


tbsbdr:
### Steps to reproduce
1. go to https://ocis.ocis-wopi.latest.owncloud.works/
2. create new folder
3. select this folder ☑
4. drag a file ex. from your desktop and drop it on the fileslist
5. upload does not start - (instead: browser tries to open the droped resource)

https://user-images.githubusercontent.com/26610733/169082651-edcff05d-b139-47b9-a0a3-fb510b22aca0.mp4

### Expected behaviour
upload should start

### Actual behaviour
Upload does not start

PR #7036, opened 2022-05-23
Fix drag and drop upload when a file is selected

JammingBen:
## Description
We've fixed a bug where uploading via drag and drop wouldn't work if one or more files were selected. This happened because the `CreateAndUpload` component was destroyed on a file selection.

This is the "easy" solution to the problem. Another, more complex solution would be to move all that upload logic from the `CreateAndUpload` component to somewhere else (`useUploadHelpers` composable e.g.) and initialize the Dropzone in file list directly. That would require quite some work though.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7006

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -181,7 +181,9 @@ export default defineComponent({
     },
 
     showActions() {
```
pascalwengerter, line 183: I was most likely involved in settings this up, but having a v-if on the wrapping div in a component seems like an antipattern versus having the parent component define whether the child should get rendered at all (I may miss the whole upload logic being handled here, maybe that needs to happen individually?)

packages/web-app-files/src/views/Personal.vue
```diff
@@ -4,6 +4,7 @@
       :has-bulk-actions="true"
       :breadcrumbs="breadcrumbs"
       :breadcrumbs-context-actions-items="[currentFolder]"
+      :show-actions-on-selection="true"
```
pascalwengerter, line 7: How haven't we realized this wasn't the case before? 🤔

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
JammingBen: I know 😞 Hence my note in the initial post, this is the "Easy-but-not-so-nice"-solution. Moving the upload logic is probably nicer but also way more complicated. But if we all decide to go this way, I'll rework this PR.

pascalwengerter (approved):

Issue #7038, opened 2022-05-23
Slow loading state of file list


kulmann:
### Steps to reproduce
1. Navigate into a folder with 1000 files

### Expected behaviour
When the network request (propfind) is done the file list is visible and usable in a timely manner.

### Actual behaviour
The file list takes ages to render and even a little bit longer until the first interaction is possible (e.g. context menu).

### Action items / hints
- [ ] check if loading previews blocks the UI (especially on scroll)
- [ ] analyse what exactly takes so long after the initial propfind is done. we thought that the lazy loading of cells would improve render speed a lot, but it did not
- [ ] There seem to be a lot of events, causing (partial) re-renders (-> have a look at events in vue dev tools)
- [ ] Collecting all the different tr/td attributes in `OcTable` is apparently not cached, this might be a huge issue

kulmann:
Probably best to solve this in a narrow scope: https://github.com/owncloud/owncloud-design-system/pull/1867

pascalwengerter:
I've done some digging and also opened #7056 (which doesn't even seem to work now, apparently caching made it seem like it worked the other day), pretty convinced that while the OcTable/ResourceTable rendering adjustments do have some positive impact the real culprint seems to be the `loadIndicators` logic. From how I understood it, we
- load resources (network request)
- save resources to store (triggers Vue rendering)
- purge parts of the store related to share status of resources (triggers Vue rendering)
- load indicators (network request)
- save indicators to store by running a for-loop over all available resource (triggers Vue rendering)

There's a [comment](https://github.com/owncloud/web/blob/f4320705d53ba2aea4b79728c042cdeac04e95d3/packages/web-app-files/src/store/actions.js#L636) hinting at room for performance improvements by changing that logic, but there's a lower-hanging fruit from what I've seen: 

[We don't seem to run into the `continue` condition when saving indicators to the store](https://github.com/owncloud/web/blob/716a475d90ff40e6121fedcca29cd41b3bc70c11/packages/web-app-files/src/store/mutations.js#L233) - this leads to an insane amount of store commits, see screenshot below

<img width="334" alt="Screenshot 2022-05-26 at 12 33 35" src="https://user-images.githubusercontent.com/16822008/170471219-ff3277f4-03b7-49e4-8872-ff6290d0839b.png">

Ideally for now, we can reduce the continue-logic to `if (!indicators.length) { continue }` to not modify the store if there is no shareIndicators to be added/loaded

[...]

PR #7056, opened 2022-05-24
Increase ResourceTable render performance

pascalwengerter:
## Description
- Use v-once
  - in ResourceTable (aka "freeze" static parts of each row to avoid expensive re-rendering)
  - on selected components in the ODS PR (to be linked)
- Refactor OcTable to use more native HTML elements instead of (expensive, unnecessary) Vue components in the ODS PR (to be linked)
- More _tbd_

## Related Issue
- Works towards #7038


pascalwengerter:
Closing here, the approach of using `v-once` doesn't work (and only did in the past due to a caching "issue" which made it seem like it works). I'll follow up with de-VUE-ing the OcTable though

Issue #7047, opened 2022-05-24
webUIResharing2/reshareUsers.feature:29 fails in nightly


Talank:
failed in https://drone.owncloud.com/owncloud/web/25812/21/20

```
Scenario: Reshare a folder without share permissions using API and check if it is listed on the collaborators list for resharer # features/webUIResharing2/reshareUsers.feature:29
   ✔ Before # setup.js:41
   ✔ Before # setup.js:45
   ✔ Before # setup.js:49
   ✔ Before # setup.js:66
   ✔ Before # setup.js:73
   ✔ Before # setup.js:84
   ✔ Before # stepDefinitions/filesContext.js:17
   ✔ Before # stepDefinitions/generalContext.js:199
   ✔ Before # stepDefinitions/generalContext.js:237
   ✔ Before # stepDefinitions/middlewareContext.js:46
   ✔ Given the setting "shareapi_auto_accept_share" of app "core" has been set to "no" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And the administrator has set the default folder for received shares to "Shares" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And these users have been created with default attributes and without skeleton files in the server: # stepDefinitions/middlewareContext.js:76
       | username |
       | Alice    |
       | Brian    |
       | Carol    |
   ✔ And user "Brian" has created folder "simple-folder" in the server # stepDefinitions/middlewareContext.js:66
   ✔ Given user "Brian" has shared folder "simple-folder" with user "Alice" with "read, share" permissions in the server # stepDefinitions/middlewareContext.js:66
   ✔ And user "Alice" has accepted the share "Shares/simple-folder" offered by user "Brian" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And user "Alice" has shared folder "Shares/simple-folder" with user "Carol" with "read" permissions in the server # stepDefinitions/middlewareContext.js:66
   ✔ And user "Carol" has accepted the share "Shares/simple-folder" offered by user "Alice" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And user "Alice" has logged in using the webUI # stepDefinitions/loginContext.js:64
   ✔ And the user opens folder "Shares" using the webUI # stepDefinitions/filesContext.js:155
   ✔ When the user opens the share dialog for folder "simple-folder" using the webUI # stepDefinitions/sharingContext.js:588
[...]

PR #7015, opened 2022-05-19
[full-ci] Fix inheritance of share permissions on reshares

JammingBen:
## Description
We've fixed a bug where the permissions of a share were not inherited when trying to reshare a resource. We've also disabled the role-select-dropdown if only one role is available for sharing.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/2963

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

Issue #7063, opened 2022-05-27
Collaborator role in side-bar is sometimes a drop-down list and sometimes a plain text


saw-jan:
### Steps to reproduce
1. Create users `user1`, `user2`, and `user3`
2. As `user1`, create a folder `FOLDER` and share as `Viewer` with user `user2`
3. As `user2`, reshare the received share with `user3` with `read` permission only
![custom-read](https://user-images.githubusercontent.com/52366632/170638637-e414fecf-d38d-4695-b394-e7641cc185bb.png)
4. As `user2`, refresh the side-bar panel using share indicator (`Show invited people`) button from files list
![share-indicator](https://user-images.githubusercontent.com/52366632/170639261-c9cfcdad-98a1-4661-9c8a-afdba259b25c.png)

### Expected behaviour
The collaborator role sidebar to have consistent web elements

### Actual behaviour
The collaborator role sometimes have `drop-down` and sometimes a `plain text`

drop-down:
![drop-role](https://user-images.githubusercontent.com/52366632/170640167-d10715ff-6761-488e-92ec-5ef373c00f25.png)

Plain text:
![plain-text-role](https://user-images.githubusercontent.com/52366632/170640182-306814a1-3678-488e-acc6-3a0230177baf.png)

Video:

https://user-images.githubusercontent.com/52366632/170640858-1e1b2334-3086-4c61-969d-895ae4c4ebbe.mp4

### Environment general
**Operating system**:

**Backend (ownCloud Core or Infinite Scale)**:

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

### Environment ownCloud Core
version: `10.10.1 prealpha (git)`

saw-jan:
Acceptance Tests failing: https://github.com/owncloud/web/issues/7047

JanAckermann:
@saw-jan can you check again, issue might be fixed

saw-jan:
Tried to reproduce with the latest web but cannot reproduce.
Seems like the issue has been fixed.

Closing here!

PR #7064, opened 2022-05-27
[tests-only][full-ci] Get collaborator role using either button or span selector

saw-jan:
## Description
The collaborator role can either be a dropdown or a plain text according to the share permissions. In this PR, I have refactored the `getCollaboratorsList` function to get the collaborator role either from a span or a button element.

Also, added the code that gets the permissions from a dropdown inside a try-catch block and returned empty permission if there is no dropdown elements.

## Related Issue
Changes made in https://github.com/owncloud/web/pull/7015
Closes https://github.com/owncloud/web/issues/7047

## How Has This Been Tested?
 - local

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
+++ b/tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
@@ -118,7 +118,7 @@ module.exports = {
       if (subSelectors === null) {
         subSelectors = {
           displayName: this.elements.collaboratorInformationSubName,
-          role: this.elements.collaboratorInformationSubRole,
+          role: this.elements.collaboratorInformationSubRoleBtn,
           additionalInfo: this.elements.collaboratorInformationSubAdditionalInfo,
           shareType: this.elements.collaboratorShareType
         }
@@ -142,16 +142,15 @@ module.exports = {
         const collaboratorResult = {}
         for (const attrName in subSelectors) {
           let attrElementId = null
-          await this.api.elementIdElement(
-            collaboratorElementId,
-            'css selector',
-            subSelectors[attrName],
-            (result) => {
-              if (result.status !== -1) {
-                attrElementId = result.value.ELEMENT
-              }
-            }
-          )
+
+          if (attrName === 'role') {
+            attrElementId = await this.getCollaboratorRoleElemId(collaboratorElementId)
+          } else {
+            attrElementId = await this.getElementIdElement(
+              collaboratorElementId,
+              subSelectors[attrName]
+            )
+          }
 
           if (attrElementId) {
             await this.api.elementIdText(attrElementId, (text) => {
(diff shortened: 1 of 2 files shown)


tests/acceptance/pageObjects/FilesPageElement/sharingDialog.js
```diff
@@ -324,15 +324,20 @@ module.exports = {
...
+        // read the permissions from the checkboxes
+        currentSharePermissions = await this.getSharePermissions()
+
+        // Hide role select dropdown
+        await this.moveToElement('@customPermissionsDrop', -9, 0)
+        await this.api.mouseButtonClick()
+        await this.waitForElementNotPresent('@customPermissionsDrop', 1000)
+      } catch (e) {
+        console.info('Collaborator Role is not editable!')
+      }
```
saw-jan, line 340: Since there is this https://github.com/owncloud/web/issues/7063 issue on the web, I am less confident about this change.

And we don't have a test to check that after resharing with read-only permission, the collaborator role is listed as plain text. Also, I am not sure whether the role should be listed as plain text or a drop-down after resharing with read-only permission.

saw-jan: CC @kulmann @pascalwengerter @phil-davis

phil-davis: It would be nice if we can work out why this different behavior happens! 

@saw-jan do you see the problem manually? (Is it possible to observe both behaviors when slowly using the web UI manually?) (I want to rule out the chance that there is some transient display of UI elements which the automated test sometimes "sees" but for real users maybe the correct UI elements end up being displayed after a "flicker" of the incorrect UI elements)

saw-jan: Yes, the issue can be reproduced manually. Steps to reproduce are listed here https://github.com/owncloud/web/issues/7063

individual-it: I actually don't like to have this kind of workarounds in the test.
We had a similar discussion for the e2e tests in https://confluence.owncloud.com/pages/viewpage.action?pageId=28807613

what to do with bugs discovered by e2e tests? 
- create an issue
- if possible run the journey without the step that fails
- if not then don't merge the test-PR that demonstrate the bug till the bug is fixed

We might want to make an exception here, because its not a failing test but a flaky one, but then we have to make super-sure that the workaround will be taken out  when the bug is fixed

@kulmann what do you think?

individual-it (changes requested):

tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
```diff
@@ -168,6 +167,39 @@ module.exports = {
...
+    getCollaboratorRoleElemId: async function (parentElemId) {
+      let elementId = null
+
+      elementId = await this.getElementIdElement(
+        parentElemId,
+        this.elements.collaboratorInformationSubRoleBtn
+      )
+
+      if (!elementId) {
+        this.api.pause(50 * 1000)
```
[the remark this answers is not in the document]
saw-jan: That is a leftover. Must be removed

fschade:
@saw-jan, thanks for the PR, the linked bug is already in our [Beta board](https://github.com/owncloud/web/projects/7), lets see when this will be fixed, from my side i would love to see it fixed inside web which then would make this pr obsolete.

cc.: @phil-davis

phil-davis:
I moved this to "blocked" and removed review requests. We will see this week if the underlying problem with the web code can be fixed.

saw-jan:
The main issue has been fixed in the source code: https://github.com/owncloud/web/issues/7063
Closing this PR

PR #7078, opened 2022-06-01
[full-ci] Implement cut, copy, paste keyboard shortcuts

lookacat:
## Description
See https://github.com/owncloud/web/issues/7071

## Related Issue
- Fixes [<issue_link>](https://github.com/owncloud/web/issues/7071)

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] https://github.com/owncloud/owncloud-sdk/pull/1087


packages/web-app-files/src/helpers/clipboardActions.js
```diff
@@ -0,0 +1,4 @@
+export const clipboardActions = {
+  cut: 'cut',
```
kulmann, line 2: typescript please :-) check for examples in `web-pkg/src/constants`

packages/web-runtime/src/App.vue
```diff
@@ -159,6 +166,26 @@ export default defineComponent({
 
   methods: {
     ...mapActions(['fetchNotifications']),
+    ...mapActions(['showMessage', 'createModal', 'hideModal']),
+    ...mapActions('Files', ['copySelectedFiles', 'cutSelectedFiles', 'pasteSelectedFiles']),
```
kulmann, line 170: This is a no-no, sorry. The runtime should not have knowledge about apps and even needs to avoid imports from other apps. Otherwise we'll have hard or soft errors when web is being run without the files app (which is a valid use case!).
Could you please think about a way to register keyboard handlers in the runtime? I.e. the files app implements the keyboard handlers and uses an API from the runtime to register those keyboard handlers for certain keycodes and while being the active app?

lookacat: oh okay i see, yes should be doable

packages/web-app-files/src/components/FilesList/ResourceTable.vue
```diff
@@ -397,6 +397,7 @@ export default defineComponent({
   computed: {
     ...mapGetters(['configuration']),
     ...mapState('Files', ['areFileExtensionsShown', 'spaces']),
+
```
fschade, line 400: unnecessary paragraph, else no changes in that file

packages/web-app-files/src/helpers/resource/move.ts
```diff
@@ -175,6 +177,55 @@ export const move = async (
...
+  targetFolder,
+  client,
+  createModal,
+  hideModal,
+  showMessage,
+  $gettext,
+  $gettextInterpolate,
+  $ngettext
+) => {
+  return await copyMoveResource(
```
fschade, line 205: no await needed here, not a problem just a hint ;)

packages/web-app-files/src/helpers/resource/move.ts
```diff
@@ -203,14 +255,25 @@ export const move = async (
       }
       if (resolveStrategy === ResolveStrategy.KEEP_BOTH) {
         targetName = $gettextInterpolate($gettext('%{name} copy'), { name: resource.name }, true)
+        resource.name = targetName
       }
     }
+    resource.path = join(targetFolder.path, resource.name)
```
fschade, line 261: join from 'path' is a node module, please take care that it is polyfilled

packages/web-app-files/src/store/actions.js
```diff
@@ -36,6 +38,76 @@ export default {
...
+        createModal,
+        hideModal,
+        showMessage,
+        $gettext,
+        $gettextInterpolate,
+        $ngettext
+      )
+    }
+    for (const resource of movedResources) {
+      const loadedResource = await client.files.fileInfo(resource.webDavPath, DavProperties.Default)
```
fschade, line 107: is it needed that the op runs one after the other or can it run simultaneous `Promise.all([...])`?

packages/web-app-files/src/App.vue
```diff
@@ -55,13 +55,46 @@ export default defineComponent({
...
+    ...mapActions(['showMessage', 'createModal', 'hideModal']),
+    ...mapActions('Files', ['copySelectedFiles', 'cutSelectedFiles', 'pasteSelectedFiles']),
+    ...mapMutations('Files', ['UPSERT_RESOURCE']),
+
+    handleShortcut(event) {
+      console.log(event)
+      const key = event.keyCode || event.which
+      const ctr = window.navigator.platform.match('Mac') ? event.metaKey : event.ctrlKey
+      if (!ctr /* CTRL | CMD */) return
+      if (key === 67) {
```
fschade, line 81: not an showstopper, but for readability I would prefer something like `const isCopyAction = key === 67`

fschade (changes requested):
just some minor findings, thanks for the pr

packages/web-app-files/src/helpers/resource/move.ts
lookacat: it is imported via `import { join } from 'path'` ^^ or do you mean something else?

packages/web-app-files/src/App.vue
```diff
@@ -55,13 +55,49 @@ export default defineComponent({
...
       closeSidebar: 'close',
       setActiveSidebarPanel: 'setActivePanel'
     }),
-    ...mapActions(['showMessage']),
+    ...mapActions(['showMessage', 'createModal', 'hideModal']),
+    ...mapActions('Files', ['copySelectedFiles', 'cutSelectedFiles', 'pasteSelectedFiles']),
+    ...mapMutations('Files', ['UPSERT_RESOURCE']),
+
+    handleShortcut(event) {
+      console.log(event)
```
fschade, line 77: dev leftover?

packages/web-app-files/src/helpers/resource/move.ts
fschade: nah its fine, i had a look. Path is a node module and is not available in browsers context, thats the reason why we have to polyfill it.... anyway i had a look and were already polyfilled it :)

fschade (commented):
found just one small dev leftover, expect that.... LGTM 🚀

fschade (approved):
LGTM 🚀

Issue #7066, opened 2022-05-30
Upload-overlay is slow and not robust


tbsbdr:
### Steps to reproduce
1. Download und unpack Linux Kernel 5.17.8 : https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.17.8.tar.xz
2. Login to https://ocis.ocis-wopi.released.owncloud.works/
3. select "Upload" > "Folder" "linux-5.17.8/arch/`arm64`" (1.741 items, 12.75 MB)
4. wait until List "XX uploads completed" shows up
5. XX counts up and gets slower until it stops completely.

https://user-images.githubusercontent.com/26610733/170960507-d2b19371-9098-4416-9184-89d7d1b547c6.mp4

### Expected behaviour
- upload overlay should not stop working for upload of many files (> 1.000 files). Please stresstest with linux-kernel files (Note: final goal is, that the upload of the whole linux kernel should be possible with 1.1 GB and ~80.000 files)
- Simple overlay that tells me if the upload is
  - `In progress`
  - `Completed`
  - `Failed`
 - possibility to show `details` of the upload
 
## Simple overlay:
![image](https://user-images.githubusercontent.com/26610733/170957412-d4b6e27e-2398-444e-b0f7-e00b714de18f.png)

## Details overlay
The scribble shows
- progress per single file
- cancel per single file
--> both info would be nice to have, but are costly to implement. So @JammingBen and I discussed to tackle this nice-to-have not within this ticket.
--> Details-Scribble is subject to change until we know how to tackle the performance issue in the upload process.

![image](https://user-images.githubusercontent.com/26610733/170958038-339b1a50-fc0e-46c5-981e-765ef2d334e5.png)

## Simplified time format
- upload estimation time should be simplified (rounded)
- mocks show a proposal - imo can be adapted if rounding-functions already exist

![image](https://user-images.githubusercontent.com/26610733/172338056-e5011c37-015c-4c43-8107-516581621fa2.png)

### Actual behaviour
- upload overlay stops working if many files are getting uploaded.

tbsbdr:
@JammingBen any new insights where the performance bottle necks are? If yes, we should talk about what solution we can provide with cost/benefits in mind.

tbsbdr:
@JammingBen in the scribble i used a "counter-clockwise" orientated arrow but noticed now in your draft, that this is confusing.
[...]

PR #7111, opened 2022-06-10
Use 'uploadDataDuringCreation' option when uploading via tus

JammingBen:
## Description
Uploading via tus now uses the `uploadDataDuringCreation` option which saves up one request. Also, we've fixed a serialization error during uploads, saving up another request.

## Related Issue
- https://github.com/owncloud/web/issues/7066

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-upload-data-during-creation
+++ b/changelog/unreleased/enhancement-upload-data-during-creation
@@ -1,0 +1,6 @@
+Enhancement: Upload data during creation
+
+Uploading via tus now uses the `uploadDataDuringCreation` option which saves up one request. Also, we've fixed a serialization error during uploads, saving up another request.
+
+https://github.com/owncloud/web/pull/7111
+https://github.com/owncloud/web/issues/7066
--- a/changelog/unreleased/enhancement-upload-data-during-creation
+++ b/changelog/unreleased/enhancement-upload-data-during-creation
@@ -1,6 +1,6 @@
 Enhancement: Upload data during creation
 
-Uploading via tus now uses the `uploadDataDuringCreation` option which saves up one request. Also, we've fixed a serialization error during uploads, saving up another request.
+Uploading via tus now uses the `uploadDataDuringCreation` option which saves up one request.
 
 https://github.com/owncloud/web/pull/7111
 https://github.com/owncloud/web/issues/7066
--- a/packages/web-runtime/src/composables/upload/useUpload.ts
+++ b/packages/web-runtime/src/composables/upload/useUpload.ts
@@ -1,5 +1,6 @@
 import { ClientService } from 'web-pkg/src/services'
 import {
+  useCapabilityFilesTusExtension,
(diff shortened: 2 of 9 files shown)


packages/web-runtime/src/composables/upload/useUpload.ts
```diff
@@ -9,7 +9,6 @@ import { computed, Ref, unref, watch } from '@vue/composition-api'
 import { useActiveLocation } from 'files/src/composables'
 import { isLocationPublicActive } from 'files/src/router'
 import { UppyService } from '../../services/uppyService'
-import { Location } from 'vue-router/types/router'
 import * as uuid from 'uuid'
 
 export interface UppyResource {
```
kulmann, line 14: Could you please add a comment that this interface MUST never contain complex types / MUST always consist of primitive types only? With the serialization reasoning. I guess that's a fair warning that should stay in the code to not run into the same trap again in the future.

packages/web-runtime/src/composables/upload/useUpload.ts
```diff
@@ -22,11 +21,15 @@ export interface UppyResource {
...
-    route: Location
     tusEndpoint: string
     webDavBasePath: string
     uploadId: string
     topLevelFolderId?: string
+    routeName?: string
+    routeItem?: string
+    routeShareName?: string
+    routeStorage?: string
+    routeStorageId?: string
```
kulmann, line 32: For the share jail / `SharedResource.vue` we also need the `shareId` as additional param. That doesn't come from `route.params` but `route.query` though.

changelog/unreleased/enhancement-upload-data-during-creation
```diff
@@ -0,0 +1,6 @@
+Enhancement: Upload data during creation
```
kulmann, line 1: I think the bugfix is worth a separate changelog item. 🚀

kulmann (changes requested):

packages/web-runtime/src/services/uppyService.ts
```diff
@@ -51,7 +51,8 @@ export class UppyService {
       chunkSize: chunkSize,
       removeFingerprintOnSuccess: true,
       overridePatchMethod: !!tusHttpMethodOverride,
-      retryDelays: [0, 500, 1000]
+      retryDelays: [0, 500, 1000],
+      uploadDataDuringCreation: true
```
kulmann, line 55: can we use capabilities here? the tus plugin is announced in the capabilities AFAIR.

kulmann (approved):
Nice 💪

ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/26156/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/26156/tracing/unstructured-collection-of-testable-space-interactions-alice-2022-6-13-08-46-58.zip 


</p></details>

PR #7112, opened 2022-06-10
E2e test.  set space image

ScharfViktor:
- Created e2e test for setting space image

### DIFF
--- a/tests/e2e/support/objects/app-files/spaces/actions.ts
+++ b/tests/e2e/support/objects/app-files/spaces/actions.ts
@@ -1,7 +1,9 @@
 import { Page } from 'playwright'
 import { sidebar } from '../utils'
+import { File } from '../../../types'
 import util from 'util'
 import { inviteMembers, inviteMembersArgs } from '../share/actions'
+import { expect } from '@playwright/test'
 
 const newSpaceMenuButton = '#new-space-menu-btn'
 const spaceNameInputField = '.oc-modal input'
@@ -10,6 +12,7 @@ const spaceIdSelector = `[data-space-id="%s"]`
 const spacesRenameOptionSelector = '.oc-files-actions-rename-trigger'
 const editSpacesSubtitleOptionSelector = '.oc-files-actions-edit-description-trigger'
 const editQuotaOptionSelector = '.oc-files-actions-edit-quota-trigger'
+const editImageOptionSelector = '.oc-files-actions-upload-space-image-trigger'
 const spacesQuotaSearchField = '.oc-modal .vs__search'
 const selectedQuotaValueField = '.vs--open'
 const quotaValueDropDown = `.vs__dropdown-option :text-is("%s")`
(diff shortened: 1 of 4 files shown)


tests/e2e/support/objects/app-files/spaces/actions.ts
```diff
@@ -158,3 +160,38 @@ export const changeQuota = async (args: {
...
+    page.waitForResponse(
+      (resp) =>
+        resp.url().endsWith(resource.name) &&
+        resp.status() === 207 &&
+        resp.request().method() === 'PROPFIND'
+    ),
+    fileChooser.setFiles(resource.path)
+  ])
+
+  await sidebar.close({ page: page })
```
kulmann, line 196: Could we also check that the `href` of the `img` embedded into the project space root view changed accordingly?

ScharfViktor: @SwikritiT maybe could you please give me advise how do it better?

Screenshot 2022-06-20 at 20 29 16


check that class contains src atribute, but I need first get image_id 
or more easer: that class ".space-overview-image" appears

```
const locator = await page.locator('.space-overview-image')
expect(locator).toHaveAttribute('src', 'blob:https://host.docker.internal:9200/ebf3d7fd-da5c-4cbb-9216-132c8b999e1d')
```

ScharfViktor:
> Should we add a "real" update case to the tests as well? The way I read them (even though it says `updates` in the .feature file) is that we only set an image for each space initially, but don't update them later?

no problem, step `And "Alice" updates the space "team.1" image to "testavatar.png"
` works for first image setting and for each subsequent image setting. I just add one step `And "Alice" updates the space "team.1" image to "another image"
`

tests/e2e/support/objects/app-files/spaces/actions.ts
SwikritiT: I think checking that the class contains the `src` attribute is better
I could get the values of the src attribute like this
```  
const img = await page.locator('.space-overview-image')
const src = await img.evaluate((e) => (e as HTMLImageElement).src)
console.log(src)
```
Maybe you can assert that it's not empty? In case of update store the previous `src` and assert it with new, confirming that they aren't the same? 
Does this make sense or will be of any help to you?

ScharfViktor: Thanks for the suggestion: 
```
const img = await page.locator('.space-overview-image')
const src = await img.evaluate((e) => (e as HTMLImageElement).src)
expect(src).not.toBe(spaceImageId)
spaceImageId = src
```

first line checks that image exist on web
3 line checks that src ≠ '' in first image setting and src ≠ old spaceImageId  subsequent image setting

tests/e2e/support/objects/app-files/spaces/actions.ts
```diff
@@ -206,3 +211,43 @@ export const openSpaceTrashBin = async (args: openSpaceTrashBinArgs): Promise<vo
...
+    page.waitForResponse(
+      (resp) =>
+        resp.url().endsWith(resource.name) &&
+        resp.status() === 207 &&
+        resp.request().method() === 'PROPFIND'
+    ),
+    fileChooser.setFiles(resource.path)
+  ])
+
+  const img = await page.locator('.space-overview-image')
```
SwikritiT, line 247: Maybe we can store this selector in some variable like others

ScharfViktor: You're right.  done

ScharfViktor:
@kulmann @pascalwengerter could you please review it. looks like everything is in order

pascalwengerter (approved):
Works like a charme

Issue #7089, opened 2022-06-03
Bugfix: Copy / move repeating action


lookacat:
Followup to https://github.com/owncloud/web/pull/6994:
 - Repeating copy or move action with the same name should be handled correctly: (1) move "asdf" to "target/asdf" and 
select "keep both" -> creates "target/asdf copy" when "target/asdf" already existed before. (2) again, move "asdf" to "target/asdf" and select "keep both" -> tries to create "target/asdf copy", but fails because it already exists from previous move with "keep both".

tbsbdr:
Please note:
I'm thinking about if we can replace the whole "location picker" with a "clipboard" based concept as users might know it from native os. (for oc10 we already have an app: https://marketplace.owncloud.com/apps/files_clipboard)
--> this ticket and work on the location picker might get obsolete if we decide to use a "clipboard" based concept. (Needs proper concept)

PR #7119, opened 2022-06-13
Bugfix: Copy / move file name exists count duplicate

lookacat:
## Description
See https://github.com/owncloud/web/issues/7089

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7089

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/helpers/resource/copyMove.ts
+++ b/packages/web-app-files/src/helpers/resource/copyMove.ts
@@ -1,4 +1,4 @@
-import { Resource } from './index'
+import { Resource, extractNameWithoutExtension } from './index'
 import { join } from 'path'
 import { buildResource } from '../resources'
 
@@ -221,8 +221,7 @@ export const resolveFileNameDuplicate = (name, extension, existingFiles, iterati
   if (extension.length === 0) {
     potentialName = `${name} (${iteration})`
   } else {
-    const extensionIndexInName = name.lastIndexOf(`.${extension}`)
-    const nameWithoutExtension = name.substring(0, extensionIndexInName)
+    const nameWithoutExtension = extractNameWithoutExtension({ name, extension } as Resource)
     potentialName = `${nameWithoutExtension} (${iteration}).${extension}`
   }
   const hasConflict = existingFiles.some((f) => f.name === potentialName)
--- a/packages/web-app-files/src/helpers/resource/copyMove.ts
+++ b/packages/web-app-files/src/helpers/resource/copyMove.ts
@@ -216,11 +216,18 @@ export const copy = async (
   )
 }
 
-export const resolveFileNameDuplicate = (name, existingFiles, iteration = 1) => {
-  const potentialName = `${name} (${iteration})`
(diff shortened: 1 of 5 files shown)


packages/web-app-files/src/helpers/resource/copyMove.ts
```diff
@@ -241,7 +252,9 @@ export const copyMoveResource = async (
   )
   const movedResources = []
 
-  for (const resource of resourcesToMove) {
+  for (let resource of resourcesToMove) {
+    // because javascripts way of cloning is awesome
```
kulmann, line 256: please be more precise why a shallow copy is not sufficient 😁

packages/web-app-files/src/helpers/resource/copyMove.ts
```diff
@@ -215,6 +215,14 @@ export const copy = async (
     true
   )
 }
+
+export const resolveFileNameDuplicate = (name, existingFiles, iteration = 1) => {
+  const potentialName = `${name} (${iteration})`
```
kulmann, line 220: this only works for folders, not for files, right?

packages/web-app-files/tests/unit/helpers/resource/copyMove.spec.ts
```diff
@@ -23,6 +23,22 @@ describe('copyMove', () => {
       webDavPath: '/target'
     }
   })
+
+  it('should name duplicate file correctly', async () => {
```
kulmann, line 27: please also add test cases with files / file extensions. probably a good idea to make it a table test (look for `it.each` examples in existing unit tests)

kulmann (changes requested):

packages/web-app-files/src/helpers/resource/copyMove.ts
lookacat: ah yes, right

packages/web-app-files/src/store/mutations.js
```diff
@@ -324,8 +324,11 @@ export default {
 // eslint-disable-next-line camelcase
 function $_upsertResource(state, resource, allowInsert) {
   const files = [...state.files]
-  const index = files.findIndex((r) => r.id === resource.id)
-  const found = index > -1
+  let index = files.findIndex((r) => r.id === resource.id)
+  if(resource.webDavPath && resource.webDavPath.length) {
```
kulmann, line 328: Please add `index === -1` as an additional condition to the `if` clause. You check that in the body and basically do nothing in that case, so you could as well skip the block entirely, which makes it easier to read.

kulmann: is it possible at all the the `webDavPath` is empty?

lookacat: yes sadly in some unittests it can happen that webDavPath is empty

kulmann (approved):
LGTM 👍

PR #7120, opened 2022-06-14
Use tus chunksize from backend

JammingBen:
## Description
Instead of using the tus chunksize from the web config, we now use the chunksize coming from the backend (if set).

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-tus-backend-chunksize
+++ b/changelog/unreleased/enhancement-tus-backend-chunksize
@@ -1,5 +1,5 @@
 Enhancement: Use tus chunksize from backend
 
-Instead of using the tus chunksize from the web config, we now use the chunksize coming from the backend (if set).
+We now use the chunksize value coming from the backend for uploading via tus. If no chunksize is given, it will default to `Infinity`. The Web config `uploadChunkSize` has been removed as a result.
 
 https://github.com/owncloud/web/pull/7120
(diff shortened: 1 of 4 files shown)


packages/web-runtime/src/services/uppyService.ts
```diff
@@ -43,11 +43,7 @@ export class UppyService {
     tusExtension: string
     headers: { [key: string]: string }
   }) {
-    const chunkSize =
-      tusMaxChunkSize > 0 && uploadChunkSize !== Infinity
-        ? Math.max(tusMaxChunkSize, uploadChunkSize)
-        : uploadChunkSize
-
+    const chunkSize = tusMaxChunkSize > 0 ? tusMaxChunkSize : uploadChunkSize
```
kulmann, line 46:
```suggestion
    const chunkSize = tusMaxChunkSize || uploadChunkSize
```
should do the same

kulmann (approved):

changelog/unreleased/enhancement-tus-backend-chunksize
```diff
@@ -0,0 +1,5 @@
+Enhancement: Use tus chunksize from backend
+
+Instead of using the tus chunksize from the web config, we now use the chunksize coming from the backend (if set).
```
pascalwengerter, line 3: Can we remove the config chunksize entirely then?

kulmann: if we can have a sane fallback value without config, then yes

JammingBen: The current fallback is `infinity`, I'm not sure if this can be seen as "sane" 😅

kulmann (approved):

PR #7130, opened 2022-06-15
Drop version from sample configs and ci configs (it's unused)

kulmann:
The version in our `config.json` is not used. At least not anymore. Removing it from all sample configs and CI configs.


pascalwengerter (approved):
🥳

PR #7172, opened 2022-06-23
[not-for-merge][tests-only]Use new nightwatch just for the dry run feature

kiranparajuli589:
- use Nightwatch v2 for dry runs


kiranparajuli589:
here is the check https://drone.owncloud.com/owncloud/web/26396/6/3

with this, yarn cache is affected. this case also has to be investigated.

PR #7174, opened 2022-06-23
Allow blocking file extensions of text-editor

kulmann:
## Description
Allow to block file extensions of the text-editor app

## Related Issue
- Works towards https://github.com/owncloud/web/issues/6661

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-text-editor/src/index.js
+++ b/packages/web-app-text-editor/src/index.js
@@ -6,6 +6,8 @@ function $gettext(msg) {
   return msg
 }
 
+const appId = 'text-editor'
+
 const routes = [
   {
     path: '/:filePath*',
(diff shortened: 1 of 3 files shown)


kulmann:
I've added the docs to the oc10 deployment section because that one already has such a section. But for the `text-editor` it works the same way for ocis.

pascalwengerter (approved):

diocas:
It would be great to be able to disable the default opening + creation, while keeping the possibility to open a file with the editor, but I think is good enough for now. (I just realized that I mention "opening" instead of "default opening" in #6661...)

packages/web-app-text-editor/src/index.js
```diff
@@ -19,41 +21,69 @@ const routes = [
...
   icon: 'file-text',
   isFileEditor: true,
-  extensions: [
-    {
-      extension: 'txt',
+  extensions: fileExtensions().map((extensionItem) => {
+    return {
+      extension: extensionItem.extension,
       newFileMenu: {
         menuTitle($gettext) {
```
diocas, line 81: Since we have all file extensions in the same level (in the "new" menu), the fact that we allow creating this many types will generate a huge list. I prefer the previous behavior: we can open many, but create just txt  - users can still rename the extension. Ofc it would be even better if we could just open any text mimetype instead of file extensions.

packages/web-app-text-editor/src/index.js
```diff
@@ -19,41 +21,69 @@ const routes = [
...
+      extension: 'php',
+      label: $gettext('PHP file')
+    },
+    {
+      extension: 'yaml',
+      label: $gettext('YAML file')
+    }
+  ]
+  let blackList =
+    window.Vue.$store.getters.extensionConfigByAppId(appId).blockedFileExtensions || []
```
diocas, line 60: This seems a workaround. I hope in the future we can have the callables I mention in the other ticket.

kulmann: Yes, that's clearly a workaround. We want to provide the feature for you guys, but we can't offer it in a technically nice way right now, as the extension framework is not on the short term roadmap.

packages/web-app-text-editor/src/index.js
kulmann: Updated, could you please check out if the new behaviour is better for your needs?

diocas: I guess you still need to set `canBeDefault` to false explicitly? Other wise, [this](https://github.com/owncloud/web/blob/master/packages/web-runtime/src/store/apps.js#L57).

kulmann: 👍

diocas (approved):

PR #7188, opened 2022-06-28
Improve performance of share indicators

JammingBen:
## Description
We've improved the performance of share indicators when loading resource tables as well as when adding or removing shares.

In detail:

* Accessing a shared folder: The share indicators now get loaded initially with the resources. The store then gets updated once, including all indicators. Before, the store was updated after the initial load. Also, this happened for each resource, resulting in a re-render of each table row.
* Creating shares & public links: Instead of loading share indicators for all resources in the store, we now only load those for the current file (and files inside a folder).

## Related Issue

- https://github.com/owncloud/web/issues/7038

## Performance measurements

Tests were done with a shared folder including 800 files. Note that the measured time is generally longer when the measurement tool is running. This means the tool is fine for a before- & after-comparison, but not for evaluating the total time as it would be much faster without the measurement.

**Accessing the folder:**

Without optimization (~ 8,5 s):

Pasted Graphic 1

With optimization (~ 6,5 s):

Pasted Graphic

**Creating a public link for a resource within the folder:**

Without optimization (~ 11 s):

image

With optimization (~ 6,5 s):

image

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-share-indicators-performance
+++ b/changelog/unreleased/enhancement-share-indicators-performance
@@ -1,0 +1,6 @@
+Enhancement: Improve performance of share indicators
+
+We've improved the performance of share indicators when loading resource tables as well as when adding or removing shares.
+
+https://github.com/owncloud/web/issues/7038
+https://github.com/owncloud/web/pull/7188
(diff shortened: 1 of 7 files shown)


kulmann (approved):
LGTM 💪

Issue #7181, opened 2022-06-27
Non-Admin should not see "User Management" in the App Switcher


tbsbdr:
### Steps to reproduce
1. login as Albert on: https://ocis.ocis-web.latest.owncloud.works
2. Select "User Management" in the App Switcher
3. Non-Admin sees User Management UI

### Expected behaviour
- User should not see "User Management" in the App Switcher, if she/he has not the appropriate permissions

### Actual behaviour
- Non-Admin sees User Management UI

PR #7197, opened 2022-06-29
Don't show user management in app switcher if user has not role 'admin'

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/7181

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-user-management/src/index.js
+++ b/packages/web-app-user-management/src/index.js
@@ -9,6 +9,8 @@ function $gettext(msg) {
   return msg
 }
 
+const store = window.Vue.$store
+
 const appInfo = {
   name: $gettext('User management'),
   id: 'user-management',
@@ -41,6 +43,9 @@ const navItems = [
     icon: 'user',
     route: {
       path: `/${appInfo.id}/users?`
+    },
+    enabled: () => {
+      return store.getters.user.role.name === 'admin'
     }
   },
   {
@@ -48,6 +53,9 @@ const navItems = [
     icon: 'group-2',
     route: {
       path: `/${appInfo.id}/groups?`
+    },
+    enabled: () => {
+      return store.getters.user.role.name === 'admin'
     }
   }
 ]
--- a/packages/web-runtime/src/store/user.js
+++ b/packages/web-runtime/src/store/user.js
@@ -5,6 +5,7 @@ import { router } from '../router'
 import { clientService } from 'web-pkg/src/services'
 
 import { setUser as sentrySetUser } from '@sentry/browser'
+import axios from 'axios'
 
 let vueAuthInstance
 
@@ -20,7 +21,9 @@ const state = {
   groups: [],
   userReady: false,
   quota: null,
-  language: null
+  language: null,
+  role: {},
+  roles: []
 }
 
 const actions = {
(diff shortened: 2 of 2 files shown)


ownclouders:
Results for <strong>e2e-tests oC10</strong> https://drone.owncloud.com/owncloud/web/26516/11/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/26516/tracing/alice-can-share-this-weeks-meal-plan-with-all-parents-alice-2022-6-29-09-14-13.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/26516/tracing/file-alice-2022-6-29-09-13-04.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/26516/tracing/folder-alice-2022-6-29-09-11-56.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/26516/tracing/public-link-alice-2022-6-29-09-10-48.zip 


</p></details>

packages/web-app-user-management/src/index.js
```diff
@@ -9,6 +9,8 @@ function $gettext(msg) {
   return msg
 }
 
+const permissionManager = window.Vue.$permissionManager
```
kulmann, line 12: Please move the variable inside both `enabled` callbacks. Reason: safer regarding timing of accessing the `window.Vue.$permissionManager` object. At least for when the `enabled` callbacks are being called we are absolutely sure that the `$permissionManager` instance exists.

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -253,6 +253,24 @@ export const announceClientService = ({
...
+ */
+export const announcePermissionManager = ({
+  vue,
+  store
+}: {
+  vue: VueConstructor
+  store: Store<any>
+}): void => {
+  const permissionManager = new PermissionManager(store)
+  vue.prototype.$permissionManager = permissionManager
```
kulmann, line 270: You don't need to do both. Either set it on `prototype` or use reactive `set`. They both have the same result, so one is sufficient.

packages/web-runtime/src/store/user.js
```diff
@@ -98,12 +101,44 @@ const actions = {
 
         const userGroups = await client.users.getUserGroups(login.id)
         const user = await client.users.getUser(login.id)
+        const {
```
kulmann, line 104: needs a capability check as well. API is not available in oc10.

kulmann (changes requested):

packages/web-runtime/src/store/user.js
kulmann: There is none. Use the `spaces.enabled` capability for now, like it's already done for loading the user. We need to come up with better fitting capabilities though.

packages/web-runtime/src/container/bootstrap.ts
JanAckermann: I tried that.
If I only use set(vue, '$permissionManager', permissionManager), the permissionManager is accessible in /home/jan/working/web/packages/web-app-user-management/src/index.js but not in 
/home/jan/working/web/packages/web-app-files/src/views/spaces/Projects.vue

and vice versa, I have no clue why this is the case :x

PR #7226, opened 2022-07-06
Unselect on focus outside of the table

elizavetaRa:
## Description
Selected files get unselected on focus on navigation sidebar and filesListWrapper outside of the table (but not if focusable elements inside like buttons get focused)

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Usability improvement: allows more performant unselecting one or multiple selected files with one click. The user doesn't need to move the mouse to X button or checkboxes (Fitt's law implication: smaller targets require more time to click on them, clicking somewhere outside is quicker).

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

unselect

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/App.vue
+++ b/packages/web-app-files/src/App.vue
@@ -1,6 +1,12 @@
 <template>
   <main id="files" class="oc-flex oc-height-1-1">
-    <div ref="filesListWrapper" tabindex="-1" class="files-list-wrapper oc-width-expand">
+    <div
+      id="files-list-wrapper"
+      ref="filesListWrapper"
+      tabindex="-1"
+      class="files-list-wrapper oc-width-expand"
+      @click="unselectOnClick"
+    >
       <router-view id="files-view" tabindex="0" />
     </div>
     <side-bar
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/components/SidebarNav/SidebarNav.vue
```diff
@@ -72,28 +87,19 @@ export default {
...
-        : 'toggle-sidebar-button-expanded oc-pr-s'
-    },
-
-    toggleSidebarButtonIcon() {
-      return this.navigation.closed ? 'arrow-drop-right' : 'arrow-drop-left'
-    }
-  },
   methods: {
     ...mapActions(['openNavigation', 'closeNavigation']),
+    ...mapActions('Files', ['resetFileSelection']),
```
kulmann, line 92: We can't introduce a dependency on the files app from the runtime, since it's not required to load the files app at all (yes, we do have a production use case for that). Apart from that I'm not convinced that focussing the left sidebar should result in clearing the file selection. 🤔 What's the reasoning behind this?

packages/web-app-files/src/App.vue
```diff
@@ -68,6 +74,15 @@ export default defineComponent({
         to: this.$refs.filesSidebar?.$el,
         revert: event === 'beforeDestroy'
       })
+    },
+    /* on click instead of on focus since child element got the higher tabindex and focuses first */
+    unselectOnClick(e) {
```
kulmann, line 79: What do you think about only interpreting clicks **below** the file list for clearing the selection model? In the area above the file list I'd prefer to treat a click on whitespace as a `whoops I missed that button that I wanted to click`. In that case I wouldn't want my current selection to be gone.

kulmann (changes requested):

pascalwengerter:
@elizavetaRa you could try and set a `ref` on the FilesView and work with https://vueuse.org/core/onClickOutside/ (which is available in the web-app-files)

PR #7263, opened 2022-07-11
Enable tracing nightly

ScharfViktor:
When we run tests at night and the tests fail, we don't get a trace https://drone.owncloud.com/owncloud/web/26735/10/12

To find out what happened, I enabled night tracing. 
I didn't make the tracing result public, we can get the tracing from `uploadTracingResult`.


SwikritiT (approved):
LGTM

Issue #5937, opened 2021-10-22
Fix rendering of error in case of no quota space


ishank011:
When web calls the `/ocs/v1.php/cloud/users/$username` endpoint, it returns back the home quota of the user. If the user doesn't have enough quota, the 'Upload' button is disabled, but no error message is shown. If you inspect the element, you can see the error 'You don't have enough quota to upload' but it's not displayed to the user.

pascalwengerter:
@tbsbdr needs concepting (not a lot, but anything to make it actionable 🙏🏽 )

tbsbdr:
proposal:

*   show notification on click of the disabled button like "no quota ...."
*   add contextual helper with same message as in the notification

JanAckermann:
@tbsbdr 
We currently have a tooltip on the button which shows the message, I'd vote for this is enough?
![image](https://user-images.githubusercontent.com/26169327/181908904-f0b9c153-e920-41d1-b330-907693c65132.png)


There is still a little error tough, that's why the tooltip wasn't shown in the first place

tbsbdr:
Tooltip-information should be optional as they are not accessible. I treat them as optional, , meaning: they are nice to have and enhance ux, but the ui must be usable if they were not there. this is at least my understanding, is this correct @kulmann 

If we can't rely on tooltip information, we would need an accessible solution which would contextual help or notification.

JanAckermann:
@tbsbdr If it comes down to a11y, we also have area descriptors on those buttons.

kulmann:
Well, the linked PR fixes that the tooltips are shown (they were there before but not rendered on disabled buttons). Good first step. For making the information accessible we should just add the information via `aria-describedby` like we already do in other situations. @JanAckermann could you do those adjustments as well please?

kulmann:
> If we can't rely on tooltip information, we would need an accessible solution which would contextual help or notification.

That's not needed. Tooltip for visuals, aria-describedby for screen readers solves the job well.

PR #7376, opened 2022-07-30
Fix tooltips are not shown on disabled buttons

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/5937

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-tooltips-not-shown-on-disabled-create-and-upload-button
+++ b/changelog/unreleased/bugfix-tooltips-not-shown-on-disabled-create-and-upload-button
@@ -1,0 +1,7 @@
+Bugfix: Tooltips not shown on disabled create and upload button
+
+We've fixed a bug where tooltips that contain important information for example quota exceeded message weren't
+shown on hovering over the create or upload button.
+
+https://github.com/owncloud/web/pull/7376
+https://github.com/owncloud/web/issues/5937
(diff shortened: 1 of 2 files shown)


kulmann (approved):
Woah, I'm a bit surprised, that this is actually the way to go. My first thought was `why don't we allow the tooltip directive / tippy.js to work on disabled buttons` but then found e.g. this: https://github.com/atomiks/tippyjs/issues/104#issuecomment-332318229
Which is exactly what you did here. Good job! 💪

ishaiavrahami:
Hello, i noticed that its mentioned that's fixed in beta 6, however, i'm running beta 20 and its still not working. Any advice?

kulmann:
> Hello, i noticed that its mentioned that's fixed in beta 6, however, i'm running beta 20 and its still not working. Any advice?

Hey, could you state the exact ocis version you're running? There is no `beta 20`, we only had `2.0.0-beta.8` as last beta version, after that `2.0.0-rc.1`, `2.0.0-rc.2` and the final `2.0.0` "general availability release".

Issue #7400, opened 2022-08-03
Regular member should not be able to change Space quota


tbsbdr:
### Steps to reproduce
1. create a space
2. login as regular space member
3. regular member has the option in the context menu to change the quota 

### Expected behaviour
Regular member should not be able to change Space quota

### Actual behaviour
regular member has the option in the context menu to change the quota

micbar:
There is a permission in the backend `SetSpaceQuota` which should be respected by the frontend.

PR #7401, opened 2022-08-03
Respect space quota permission

JammingBen:
## Description
By taking the space quota permission into account, we've fixed a bug where a regular space member could see the "Edit space quota" action.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7400

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-pkg/src/services/permissionManager.ts
+++ b/packages/web-pkg/src/services/permissionManager.ts
@@ -1,5 +1,5 @@
 import { Store } from 'vuex'
-interface RoleSetting {
+interface Permission {
   description: string
   displayName: string
   id: string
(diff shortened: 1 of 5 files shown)


JanAckermann (approved):

packages/web-pkg/src/services/permissionManager.ts
```diff
@@ -1,6 +1,21 @@
 import { Store } from 'vuex'
+interface RoleSetting {
```
kulmann, line 2:
```suggestion
interface Permission {
```
technically, backend side, it's a setting of the role. But semantically, in the context of a role, it's a permission.

kulmann (approved):
👍

Issue #7304, opened 2022-07-20
Scoped links: Can't create private-ish links


tbsbdr:
### Steps to reproduce
1. Login to https://ocis.ocis-web.latest.owncloud.works/
2. click on "copy quicklink"
3. quicklink shows "Recipients can view and download contents."
4. there is no option to create a link that works only as a pointer for invited people

### Expected behaviour
- Option to have a link that works only as a pointer for invites people/space members
- actual feature description: https://github.com/owncloud/web/issues/6844

### Actual behaviour
- I can only create public-ish links

kulmann:
Creating them is implemented in https://github.com/owncloud/web/pull/7133 (not merged yet, because it has no test coverage, yet). Resolving them is not possible, yet.

kulmann:
Reminder to self and @dschmidt : for public links we also want to resolve the user into the correct context. For that we need an active user session. We wanted to look at the `session monitor` feature from `oidc-client-ts` to see if we can determine the login state without a roundtrip to the IdP.

PR #7405, opened 2022-08-04
[full-ci] Resolve alias links

JammingBen:
## Description
Needs the alias link capability to be set to `true` in ocis. See https://github.com/owncloud/ocis/pull/4324 (needs to be reverted after merge).

## Related Issue
- Works towards https://github.com/owncloud/web/issues/7304
- Works towards https://github.com/owncloud/web/issues/7707
- Works towards https://github.com/owncloud/web/issues/6844

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Open tasks:
- [x] Password protection -> not supported
- [x] Include SDK update after https://github.com/owncloud/owncloud-sdk/pull/1126 has been merged
- [x] Changelog item
- [ ] Acceptance Tests (create ticket after code review/merge)
- [ ] Undo https://github.com/owncloud/ocis/pull/4324 after merge
- [x] Further adjustments regarding file names? (see https://github.com/owncloud/web/pull/7405#issuecomment-1217796951)
- [x] oC10 compatibility
- [ ] Make it work for Shares


ownclouders:
Results for <strong>oCISSharingAndUpload</strong> https://drone.owncloud.com/owncloud/web/27350/67/1

packages/web-app-files/src/components/SideBar/Shares/Links/DetailsAndEdit.vue
```diff
@@ -378,6 +378,10 @@ export default defineComponent({
 
     passwortProtectionTooltip() {
       return this.$gettext('This link is password-protected')
+    },
+
+    isAliasLink() {
+      return this.link.permissions === '0'
```
kulmann, line 384: No magic numbers please :-) Please have a look at `_buildLink` in `web-app-files/src/helpers/resources.ts`. That already loads the correct role object. You could add the role to the json that the function returns and utilize it here.

packages/web-runtime/src/pages/resolvePublicLink.vue
```diff
@@ -91,6 +113,18 @@ export default defineComponent({
...
     })
+    const loadTokenInfoTask = useTask(function* (signal, ref) {
+      let tokenInfo
+      try {
+        if (unref(isLoggedInUser)) {
+          tokenInfo = yield owncloudSdk.shares.getProtectedTokenInfo(unref(token))
+        } else {
+          tokenInfo = yield owncloudSdk.shares.getUnprotectedTokenInfo(unref(token))
+        }
+      } catch (e) { } // oC10
```
kulmann, line 124:
```suggestion
      } catch (e) { } // backend doesn't support the token info endpoint
```

packages/web-runtime/src/pages/resolvePublicLink.vue
```diff
@@ -134,14 +171,37 @@ export default defineComponent({
...
+          if (error.statusCode !== 404) {
+            throw error
+          }
+          // alias link
+          if (!this.isLoggedInUser) {
+            await authService.loginUser(queryItemAsString(`/files/spaces/personal/${fullPath}`))
+            return this.$router.push({ name: '/login' })
+          }
+
+          return this.$router.push({ name: 'files-spaces-personal', params: { storageId: fullPath } })
```
kulmann, line 199: Is it not possible to resolve alias links to project spaces? I only see personal space routes in your code.

kulmann (changes requested):

packages/web-runtime/src/pages/resolvePublicLink.vue
JammingBen: Ohh good point! I need to check this, resolving for spaces does not work at all currently...

JammingBen: After some debugging I found out that we have another issue when sharing single files (not folders) via alias links. That's not related to spaces however. Actually, spaces resources also need to be resolved in the personal space. Otherwise we would have a problem with unauthenticated users, because in this case, the token info does not reveal if we're dealing with a personal or a space resource.

JammingBen:
I've fixed the issue with resolving single files. However, currently there is no way to retrieve the file name in such case, so the IDs get used:

image

Backend team is aware and working on it. I don't know if it needs further adjustments in Web (probably yes), but IMO the PR is fine as alias links are disabled by default anyways.

kulmann (changes requested):

packages/web-app-files/src/helpers/resources.ts
```diff
@@ -49,6 +49,16 @@ export function buildResource(resource): Resource {
...
+  const idComponent = resource.name.split('/')[2]
+  let webDavPath
+  if (idComponent?.includes('!')) {
+    // idComponent includes the opaqueID which represents the resource name.
+    // For the webdav path, we need to replace the opaqueID with the actual resource name.
+    const withoutOpaqueId = resource.name.split('!')[0]
+    webDavPath = `${withoutOpaqueId}/${name}`
+  } else {
+    webDavPath = resource.name
+  }
```
JammingBen, line 61: I don't know here... maybe there is a more elegant way to solve this issue? With alias links on single resources, we don't have the actual resource name in the path, but the opaque id. Actions like `rename` or editing via the text editor require the actual name in the `webDavPath` though.

kulmann: the sdk has a function to retrieve a webdav path by file id via `meta` request. maybe you need to do that...

JammingBen: But that would be one additional request per resource, right...?

kulmann: ah, per resource... that's not an option.

JammingBen: I adjusted the condition a bit so it only affects alias links on single resources. This is a very special case as described above, I don't see another way right now besides moving this logic to all places where needed (rename, text editor app, ...).

kulmann:
Currently blocked until https://github.com/cs3org/reva/pull/3179 has been brought to ocis

packages/web-app-files/src/helpers/resources.ts
dschmidt: Uhm, what logic would you want to add to text editor? What makes the text editor special?

packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -571,6 +554,25 @@ export default defineComponent({
             status: 'danger'
           })
         })
+    },
+
+    getAvailableRoleOptions(link) {
```
dschmidt, line 559: Can you add a type to `link`?

packages/web-app-files/src/components/SideBar/Shares/Links/DetailsAndEdit.vue
```diff
@@ -226,15 +230,15 @@ export default defineComponent({
     }
   },
   computed: {
+    currentLinkRole() {
+      return LinkShareRoles.getByBitmask(parseInt(this.link.permissions), this.isFolderShare)
+    },
```
dschmidt, line 235: iThe parseInt here looks a bit misplaced. Again it would be cool to have a type/interface for the link - then parsing would probably happen in a more appropriate place? 

I see that you are just moving code but now might be a good time for a little cleanup while you touch the code anyway

packages/web-app-files/src/helpers/resources.ts
```diff
@@ -50,6 +50,16 @@ export function buildResource(resource): Resource {
   }
 
   const id = resource.fileInfo[DavProperty.FileId]
+  const pathParts = resource.name.split('/')
+  let webDavPath
+  if (pathParts.length === 3 && !isFolder && pathParts[2].includes('!')) {
```
dschmidt, line 55: Can you add a comment why you check for length === 3?

packages/web-app-files/src/services/folder/spaces/loaderPersonal.ts
```diff
@@ -37,15 +37,18 @@ export class FolderLoaderSpacesPersonal implements FolderLoader {
...
         resources = resources.map(buildResource)
 
-        const currentFolder = resources.shift()
-        const hasShareJail = useCapabilityShareJailEnabled(store)
-        yield store.dispatch('Files/loadSharesTree', {
-          client: clientService.owncloudSdk,
-          path: currentFolder.path
-        })
+        const currentFolder =
+          resources.length > 1 || resources[0].isFolder ? resources.shift() : undefined
```
dschmidt, line 41: Can you add braces here (and the legacy copy of the code)? 
Had to look multiple times what actually was going on with the or and ternary operators...

packages/web-pkg/src/composables/capability/useCapability.ts
```diff
@@ -52,11 +46,15 @@ export const useCapabilityFilesTusExtension = createCapabilityComposable<string>
   'files.tus_support.extension',
   ''
 )
+export const useCapabilityFilesSharingResharing = createCapabilityComposable(
+  'files_sharing.resharing',
+  true
+)
 export const useCapabilityFilesSharingPublicCanEdit = createCapabilityComposable(
-  'files.files_sharing.public.can_edit',
+  'files_sharing.public.can_edit',
```
dschmidt, line 54: 🙈

packages/web-runtime/src/pages/resolvePublicLink.vue
```diff
@@ -91,7 +112,19 @@ export default defineComponent({
...
+    const loadTokenInfoTask = useTask(function* () {
+      let response = {}
+      try {
+        if (unref(isUserContext)) {
+          response = yield owncloudSdk.shares.getProtectedTokenInfo(unref(token))
+        } else {
+          response = yield owncloudSdk.shares.getUnprotectedTokenInfo(unref(token))
+        }
+      } catch (e) { } // backend doesn't support the token info endpoint
+      debugger
```
dschmidt, line 124: :trollface:

packages/web-runtime/src/pages/resolvePublicLink.vue
```diff
@@ -69,18 +81,27 @@ import {
...
     const { owncloudSdk } = useClientService()
+    const store = useStore()
     const token = useRouteParam('token')
+    const isUserContext = useUserContext({ store })
     const password = ref('')
-    const isPasswordRequiredTask = useTask(function* (signal) {
+    const tokenInfo = ref({})
+    const isPasswordRequiredTask = useTask(function* () {
+      if (!isEmpty(unref(tokenInfo))) {
+        return unref(tokenInfo).password_protected === 'true'
```
dschmidt, line 102: Does the sdk really return an object with a property that contains true as a string? I don't even

kulmann (approved):
LGTM 👍

packages/web-app-files/src/helpers/resources.ts
```diff
@@ -50,6 +50,16 @@ export function buildResource(resource): Resource {
   }
 
   const id = resource.fileInfo[DavProperty.FileId]
+  const pathParts = resource.name.split('/')
+  let webDavPath
```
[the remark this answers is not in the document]
kulmann: I don't remember this :D good that I set a reminder...

PR #7425, opened 2022-08-09
[full-ci] Load spaces in runtime

dschmidt:
This converts a bunch of files to TypeScript and fixes one or two potential errors that became apparent after the conversion.

I'm sorry this got so huge, but splitting it into separate commits would have been a big effort and we would have merged it to master anyway, so I didn't want to invest that much time... 

## Description
This moves the spaces loading into web-runtime (from web-app-files) in order to prepare `driveAlias` handling that we will need in files app and other file handling apps.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


JanAckermann (changes requested):
IHMO the UPDATE_SPACE_FIELD mutation should be moved to

packages/web-client/src/helpers/space/functions.ts
```diff
@@ -0,0 +1,128 @@
+import { User } from '../user'
```
kulmann, line 1: What about a file name like e.g. `conversion.ts`? 🤔  For now the only function that this file has is a conversion from a drive response to a resource. On the other hand we could build a class for the space resource in a followup... your choice.

packages/web-app-files/src/store/mutations.js
```diff
@@ -6,9 +6,6 @@ import { renameResource } from '../helpers/resources'
 import { ShareTypes } from '../helpers/share'
 
 export default {
-  LOAD_SPACES(state, spaces) {
-    state.spaces = spaces
-  },
   /**
    * Updates a single space field. If the space with given id doesn't exist nothing will happen.
    *
```
kulmann, line 11: Like @JanAckermann already posted, the following mutations `UPDATE_SPACE_FIELD`, `UPSERT_SPACE`, `REMOVE_SPACE`, `CLEAR_SPACES` should move over to the new store module as well.

packages/web-app-files/src/services/folder/loaderPublicFiles.ts
```diff
@@ -5,9 +5,9 @@ import { DavProperties, DavProperty } from 'web-pkg/src/constants'
 import { buildResource } from '../../helpers/resources'
 import { isLocationPublicActive, createLocationPublic } from '../../router'
 
-import { linkRoleUploaderFolder } from '../../helpers/share'
 import { Store } from 'vuex'
 import { authService } from 'web-runtime/src/services/auth'
+import { linkRoleUploaderFolder } from 'web-client/src/helpers'
```
kulmann, line 10: You're sometimes importing from `web-client/src/helpers` and sometimes from `web-client/src/helpers/share`. IMO the latter is more intuitive. So you'd need to drop the export statement in `web-client/src/helpers/index.ts`.

packages/web-app-files/src/mixins/actions/navigate.ts
```diff
@@ -75,7 +75,7 @@ export default {
       }
 
       if (isLocationSpacesActive(this.$router, 'files-spaces-project')) {
-        return createLocationPublic('files-spaces-project')
+        return createLocationSpaces('files-spaces-project')
```
kulmann, line 78: Probably worth a bugfix changelog item.

packages/web-app-files/src/helpers/share/index.ts
```diff
@@ -1,6 +1,2 @@
 export * from './link'
-export * from './permission'
-export * from './role'
-export * from './share'
-export * from './status'
-export * from './type'
+export * from 'web-client/src/helpers/share/type'
```
kulmann, line 2: is this intended? 🤔  I thought you now import everything share related from the new location.

kulmann (changes requested):

packages/web-app-files/src/helpers/share/index.ts
dschmidt: Nope, automatic refactor failure -_-'

packages/web-client/src/helpers/space/functions.ts
dschmidt: I find conversion strange tbh, it's more of a deserialize function to me ... then again the function is called buildSpace ... which is also a bit of a weird name, so I rather wouldn't call the file `build.ts` :shrug:

packages/web-app-files/src/services/folder/loaderPublicFiles.ts
dschmidt: 🤷🏻‍♂️

packages/web-app-files/src/store/mutations.js
dschmidt: ✅

packages/web-app-files/src/mixins/actions/navigate.ts
dschmidt: Would have done that if I had understood what the code is good for 😂 
just fixed it based on the type...

dschmidt:
```typescript
export const createLocationPublic = (name: shareTypes, location = {}): Location =>
  createLocation(name, location)
  
export const createLocationSpaces = (name: spaceTypes, location = {}): Location =>
  createLocation(name, location)
``` 
I think it's just a code fix without any implication for the user. But at least investigating this made me find the "open folder" issue...

Issue #6648, opened 2022-03-23
Enable navigation by driveAlias in the Web UI


labkode:
Given the following space:

```
{
   "value":[
      {
         "driveAlias":"/eos/user/h/hugo",
         "driveType":"personal",
         "id":"aaaa-bbbb-cccc-dddd",
         "webdavUrl":"https://cernbox.cern.ch/dav/spaces/eos/user/h/hugo"
   ]
}
````

Navigation works by `id`:

`https://cernbox.cern.ch/files/spaces/aaaa-bbbb-cccc-dddd/relative/path/to/file.txt`

We expect navigation to work with `driveAlias`:
`https://cernbox.cern.ch/files/spaces/eos/user/h/hugo/relative/path/to/file.txt`

Is this implemented?
If yes, how can we enable it?

**Comment by kulmann** [comment]
This is not implemented, yet. We're currently working on the sharing jail implementation (see https://github.com/owncloud/web/pull/6593 and https://github.com/owncloud/web/issues/6448 for details). Expected to be done approx. 4th of April. Having the drive alias in the URL is planned after that. Together with showing custom drives with their own nav items and routes in the web ui that will at least take another week.

Could you please share your static spaces config with us within the next few days? We'll probably hardcode titles and icons for you for the start. For that we need a reference. Happy to receive that via mail, share, chat, ... if it contains confidential data. ;-)

**Comment by labkode** [comment]
> 


We're taking about 700 project spaces, 35k home spaces and 200K share spaces ...
My understanding is that the call to /drives/me only returns the spaces the user has access to, so we'll be only returning a subset (1 personal space, multiple share spaces and multiple project spaces).
Our implementation will be dynamic, meaning that we won't have the spaces registry loaded from a static file.

**Comment by kulmann** [comment]
Awesome to hear that it's dynamic. Is it still possible for you to provide a list of required nav items with a) icon, b) title, c) drive type to be mapped to a url? Or are the drive types dynamic as well? If the drive types are dynamic then that's far beyond the scope that we discussed. :-/

Do you already have an ETA for having the space listing backend side?

**Comment by micbar** [comment]
@labkode I was assuming that we could also use a static approach. Let me change your snippet

```json
{
   "value":[
      {
         "driveAlias":"/eos/user/",
         "driveType":"personal",
         "id":"aaaa-bbbb-cccc-dddd",
         "webdavUrl":"https://cernbox.cern.ch/dav/spaces/eos/user/"
   ]
}
```

[...]

PR #7430, opened 2022-08-10
  [full-ci] Drive aliases

dschmidt:
## Description
We changed the URL format to not use storageIds in the URL path anymore to identify spaces, but instead use drive aliases of spaces in the URL path.

BREAKING CHANGE for users: this breaks existing bookmarks - they won't resolve anymore.
BREAKING CHANGE for developers: the appDefaults composables from web-pkg now work with drive aliases, concatenated with relative item paths, instead of webdav paths. If you use the appDefaults composables in your application it's likely that your code needs to be adapted.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/6648
- Possible follow up: https://github.com/owncloud/web/issues/6247
- Fixes https://github.com/owncloud/web/issues/7627

## Types of changes

- [x] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:
- [x] use drive aliases in trash routes
- [x] use drive aliases in app routes
- [x] interpret public link token as drive alias
- [x] replace usages of the `storageId` route param with the id of the resolved drive item
- [x] get rid of all occurrences of route names `files-spaces-personal` and `files-spaces-project`
- [x] check and fix right sidebar in all views
- [x] check and fix file actions in all views
- [x] grep for routes with `driveAliasAndItem` and add `shareId` to route query
- [x] discuss the public link files route (duplicates `public` path segment) -> `files/link/${driveAliasAndItem}` & `files/upload/${token}`
- [x] Personal space initial load -> `New` and `Upload` buttons are disabled
- [x] double check upload composables (some dead code?)
- [x] grep for `storageId` or `currentStorageId` or `item` in context of routes (should not exist anymore)
- [x] UploadInfo.vue -> `createFolderLink` fix wrong route params

## Followups
- [ ] fix view titles (spaces, trash)
- [ ] write unit tests for generic space view
- [ ] web-app-draw-io: modifies filePath for visio files but not the fileContext: wrong route, saving and permission checks possibly don't work... 
- [ ] review/improve types for web-client.webdav/appDefaults composables
- [ ] webDavPath handling: spaces should have a function to join webDavPath/Url with a path (could also strip public-files for sdk usage)
- [ ] redirect public links to upload/link URL if their role was changed later (resolve public link, change role, reload page)
- [ ] folder upload for "keep both" choice (wrong path in POST requests) -> already broken on master, not to be fixed in this PR
- [ ] Navigate from the trashbin to personal space -> `PROPFIND 404` -> this happens because of the sidebar resource fetching. We do this as a follow up. Also see https://github.com/owncloud/web/pull/7546 which should fix this
- [ ] port loadPreview to space + path
- [ ] cleanup mitigate storageId absence in oc10


packages/web-app-files/src/components/FilesList/NotFoundMessage.vue
```diff
@@ -55,16 +55,27 @@ import { useRouter, useStore } from 'web-pkg/src/composables'
...
-      showHomeButton: isLocationSpacesActive(router, 'files-spaces-personal'),
-      showSpacesButton: isLocationSpacesActive(router, 'files-spaces-project'),
-      homeRoute: createLocationSpaces('files-spaces-personal', {
-        params: { item: store.getters.homeFolder }
+      showHomeButton:
+        isLocationSpacesActive(router, 'files-spaces-generic') &&
+        props.space?.driveType !== 'project',
+      showSpacesButton:
+        isLocationSpacesActive(router, 'files-spaces-generic') &&
+        props.space?.driveType === 'project',
```
dschmidt, line 76: Maybe we should put this in a helper var and negate it once

packages/web-app-files/src/views/spaces/Generic.vue
```diff
@@ -82,79 +86,83 @@ import { mapGetters, mapState, mapActions, mapMutations } from 'vuex'
...
     MixinFilesListScrolling,
     MixinFilesListFilter
   ],
+
   props: {
     space: {
       type: Object,
-      required: false
+      required: true
     }
```
dschmidt, line 153: Does this work with oc10?

dschmidt: Or is that relevant because of the legacy view?

kulmann: I want to introduce a kind of "virtual space" object for when spaces are not available in the backend. Currently have the impression that it would reduce quite some code complexity. That being said, it only works for ocis at the moment 😅

packages/web-client/src/webdav/listFiles.ts
```diff
@@ -1,12 +1,33 @@
+import { buildResource } from 'files/src/helpers/resources'
+import { DavProperties, DavProperty } from 'web-pkg/src/constants'
 import { OwnCloudSdk } from 'web-pkg/src/services'
 import { isPublicSpaceResource, Resource, SpaceResource } from '../helpers'
 
-export const ListFilesFactory = (sdk: OwnCloudSdk) => {
+export const GetListFilesFactory = (sdk: OwnCloudSdk) => {
```
dschmidt, line 6: Oh, doh - I tried to stick to the naming scheme of GetCapabilitiesFactory, but the Get did not refer to Factory there hehe. Let's revert this naming change tomorrow...

packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -137,7 +142,14 @@ export default defineComponent({
     DetailsAndEdit,
     NameAndCopy
   },
-  inject: ['incomingParentShare'],
+  inject: ['displayedItem', 'incomingParentShare'],
+  props: {
+    space: {
+      type: Object,
```
dschmidt, line 148: Type

packages/web-app-files/src/components/SideBar/Shares/Links/DetailsAndEdit.vue
```diff
@@ -218,6 +220,15 @@ export default defineComponent({
     link: {
       type: Object,
       required: true
+    },
+    file: {
+      type: Object,
+      required: true
+    },
+    space: {
+      type: Object,
```
dschmidt, line 229: Type

packages/web-client/src/helpers/space/functions.ts
```diff
@@ -1,8 +1,20 @@
...
+  SpaceResource
+} from '../resource'
 import { SpacePeopleShareRoles, spaceRoleEditor, spaceRoleManager } from '../share'
 
-export function buildSpace(space) {
+export function buildPublicSpaceResource(space): PublicSpaceResource {
+  const s = buildSpace(space) as PublicSpaceResource
+  s.publicLinkPassword = space.publicLinkPassword
+  return s
+}
```
dschmidt, line 15: Object.assign should work without a typecast

packages/web-pkg/src/composables/appDefaults/useAppDefaults.ts
```diff
@@ -48,11 +52,32 @@ export function useAppDefaults(options: AppDefaultsOptions): AppDefaultsResult {
...
-    const path = `/${unref(currentRoute).params.filePath?.split('/').filter(Boolean).join('/')}`
+    let path
+    if (unref(space)) {
+      if (unref(space).driveType === 'public') {
+        path = buildWebDavPublicPath(unref(space).id, unref(item))
+      } else if (unref(hasShareJail)) {
+        path = buildWebDavSpacesPath(unref(space).id, unref(item))
+      } else {
+        path = buildWebDavFilesPath(unref(space).id, unref(item))
+      }
```
dschmidt, line 70: Simplify

packages/web-pkg/src/composables/driveResolver/useDriveResolver.ts
```diff
@@ -0,0 +1,85 @@
...
+        path = item.join('/')
+      } else if (driveAliasAndItem.startsWith('share/')) {
+        const [shareName, ...item] = driveAliasAndItem.split('/').slice(1)
+        matchingSpace = buildSpace({
+          id: [SHARE_JAIL_ID, unref(shareId)].join('!'),
+          driveAlias: `share/${shareName}`,
+          driveType: 'share',
+          name: shareName,
+          shareId
+        })
```
dschmidt, line 50: Add webDavPath?

packages/web-runtime/src/pages/resolvePublicLink.vue
```diff
@@ -73,16 +73,25 @@ import {
...
+      id: unref(token),
+      driveAlias: `public/${unref(token)}`,
+      driveType: 'public',
+      webDavPath: buildWebDavPublicPath(unref(token), ''),
+      ...(unref(password) && { publicLinkPassword: unref(password) })
+    }))
+    const isPasswordRequiredTask = useTask(function* () {
       try {
-        yield owncloudSdk.publicFiles.list(unref(token), '', DavProperties.PublicLink, '0')
+        yield webdav.getFileInfo({ ...unref(publicLinkSpace), publicLinkPassword: null })
```
dschmidt, line 94: Why the shallow copy?

packages/web-client/src/webdav/listFiles.ts
```diff
@@ -1,20 +1,26 @@
...
+      { path }: { path?: string },
+      { depth = 1, davProperties }: ListFilesOptions = {}
     ): Promise<Resource[]> {
       let resources: Resource[]
       if (isPublicSpaceResource(space)) {
         resources = await sdk.publicFiles.list(
           `${space.webDavPath.replace(/^\/public-files/, '')}/${path || ''}`,
           space.publicLinkPassword,
-          DavProperties.PublicLink,
+          [DavProperties.PublicLink].concat(davProperties),
```
dschmidt, line 23: Does this make sense? Do we always want to add this? Is it harmful to pass it to private files? 
If we don't add it by default, I'm wondering how API consumers are supposed to add it if they need it?

kulmann: no, if you provide dav properties they should completely replace the default dav properties. Otherwise it would not be possible to do a propfind with just the id prop for example. All other requested data would be useless and bloat the response payload.

dschmidt: Should we remove the PublicLink property from davProperties for non PublicLink spaces?
Or doesn't it harm to request it even if it does not make sense?

I'm think of requesting additional properties for a resource and not having to think about whether we're dealing with a public link... does that make sense at all? I haven't worked with the webdav api really, so it's hard for me to judge...

packages/web-app-files/src/composables/upload/useUploadHelpers.ts
```diff
@@ -31,53 +33,38 @@ interface inputFileOptions {
...
+        return buildWebDavFilesPath(unref(user)?.id, unref(currentPath))
+      }
 
-    if (unref(hasShareJail)) {
-      return buildWebDavSpacesPath(unref(personalDriveId), unref(currentPath))
+      return buildWebDavSpacesPath(unref(options.space)?.id, unref(currentPath))
     }
-
-    return buildWebDavFilesPath(unref(user)?.id, unref(currentPath))
   })
```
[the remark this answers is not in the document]
dschmidt: I bet we can, but it's hard to reason about the code ...

packages/web-client/src/webdav/listFiles.ts
dschmidt: I've replaced the code with `davProperties || [DavProperties.PublicLink]`. Still wondering, should we filter the PublicLink property out for the private files?

ownclouders:
Results for <strong>oC10XGAPortraitNotifications</strong> https://drone.owncloud.com/owncloud/web/28566/44/1
:boom: The oC10XGAPortraitNotifications tests pipeline failed. The build has been cancelled.

packages/web-runtime/src/pages/resolvePublicLink.vue
dschmidt: Ah, to spread it into a new object with the publicLinkPassword overridden...

packages/web-app-files/src/components/Spaces/SpaceHeader.vue
```diff
@@ -0,0 +1,293 @@
...
+        }
+        const decodedUri = decodeURI(data.webDavUrl)
+        const webDavPathComponents = decodedUri.split('/')
+        const idComponent = webDavPathComponents.find((c) => c.startsWith(props.space.id as string))
+        if (!idComponent) {
+          return
+        }
+        const path = webDavPathComponents
+          .slice(webDavPathComponents.indexOf(idComponent) + 1)
+          .join('/')
```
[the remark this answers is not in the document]
dschmidt: This strips the server url and webDavPath from the url.

packages/web-pkg/src/composables/driveResolver/useDriveResolver.ts
kulmann: not needed, already handled internally in `buildSpace` :-)

JammingBen:
My findings so far:

- Indirect share indicators are missing in files table
- Console errors on „Shared with me page“: `Missing required prop: "space"`
- „Shared via link“ view is is broken (endless reload)
- Sidebar is broken on „Shared with me/others“ page
- Resources within a space don’t show space members in sidebar
- „New/Upload“-buttons are missing on a public link with „edit“-permissions on root
- Reload on personal space root -> „New/Upload“-buttons are disabled

JammingBen:
Some more:

- [ ] Open sidebar in/for the root of personal space: sidebar for a project space is showing (space image, members)
- [ ] Sidebar for public link root shows space sidebar (same as on personal view)
- [ ] Sidebar for resources on a public link shows „Select a file or folder to view details.“
- [ ] A strange one that only happens from time to time: right click file -> `Delete` -> right click another file -> context menu breaks: `TypeError: Cannot read properties of undefined (reading '$el')`
- [ ] Navigate from the trashbin to personal space -> `PROPFIND 404`
- [ ] Navigate into a subfolder, delete the subfolder via the context menu in the breadcrumbs -> redirect to parent folder is broken
- [ ] Sidebar & context menu on search result page is broken
- [ ] Breadcrumb in the trashbin shows context menu -> should not show
- [ ] „Delete“-action in share jail results in `AxiosError` in console (delete succeeds however)

JanAckermann:
- [ ] Batch download doesn't work in public link view

JammingBen:
Findings with oc10:

- [ ] "Create/Upload"-buttons on public link page are missing
- [ ] Links to resources (and parent folders) in "Shared with others/via links" views are broken ("Open"-action as well)
- [ ] Shares for resources in "Shared with others/via links" views are missing in sidebar
- [ ] Favorites View is broken
- [ ] Upload to oC10 doesn't work
- [ ] Rename in "Shared with me" page doesn't work
- [ ] Search links don't work (search result modal & search file list)

JanAckermann:
OCIS followup findings:
- [ ] Copy paste in space root, doesn't work -> conflict dialog not shown + precondition failed  
- [ ] Shared via link -> shared folder in root doesn't open click, neither parent folder is clickable (wrong uri)

packages/web-app-files/src/router/deprecated.ts
```diff
@@ -54,8 +55,14 @@ export const buildRoutes = (): RouteConfig[] =>
...
           ...to,
-          params: { ...to.params, driveAliasAndItem: 'personal/home' }
-        })
+          params: {
+            ...to.params,
+            driveAliasAndItem: urlJoin('personal/home', to.params.item, { leadingSlash: false })
+          }
+        }),
+      meta: {
+        patchCleanPath: true
```
dschmidt, line 64: Is this really needed? Or just a left over from trying?

kulmann: indeed not needed. removed it again.

pascalwengerter:
Congrats on merging this beast 🥳

PR #7445, opened 2022-08-12
[full-ci] Improve user-management-app and add max-quota to QuotaSelect.vue component

JanAckermann:
## Description
- Added AppTemplate.vue component as design template
- General refactoring (less code, bugfixing)
- Add maxQuota property to QuotaSelect.vue

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-user-management-app-close-sidebar-throws-error
+++ b/changelog/unreleased/bugfix-user-management-app-close-sidebar-throws-error
@@ -1,0 +1,6 @@
+Bugfix: User management app close side bar throws error
+
+We've fixed a bug, where closing the side bar in the user management app thrown an error,
+when a group or user was selected.
+
+https://github.com/owncloud/web/pull/7445
--- a/changelog/unreleased/bugfix-user-management-app-close-sidebar-throws-error
+++ b/changelog/unreleased/bugfix-user-management-app-close-sidebar-throws-error
@@ -1,6 +1,6 @@
 Bugfix: User management app close side bar throws error
 
-We've fixed a bug, where closing the side bar in the user management app thrown an error,
+We've fixed a bug, where closing the side bar in the user management app threw an error,
 when a group or user was selected.
 
 https://github.com/owncloud/web/pull/7445
--- a/packages/web-app-user-management/src/components/AppTemplate.vue
+++ b/packages/web-app-user-management/src/components/AppTemplate.vue
@@ -25,9 +25,8 @@
       </div>
       <side-bar
         v-if="sideBarOpen"
-        class="user-management-side-bar oc-width-1-1 oc-width-1-3@m oc-width-1-4@xl"
-        :available-panels="availableSideBarPanels"
-        :sidebar-active-panel="activePanel"
+        :available-panels="sideBarAvailablePanels"
+        :sidebar-active-panel="sideBarActivePanel"
         :loading="false"
         @selectPanel="(panel) => $emit('selectPanel', panel)"
         @close="$emit('closeSideBar')"
@@ -60,11 +59,11 @@ export default defineComponent({
       required: true,
       type: Boolean
     },
-    availableSideBarPanels: {
+    sideBarAvailablePanels: {
       required: true,
       type: Array
     },
-    activePanel: {
+    sideBarActivePanel: {
       required: true,
       type: String
     }
@@ -95,11 +94,5 @@ export default defineComponent({
 
 #user-management-wrapper {
   overflow-y: auto;
-}
-
-.user-management-side-bar {
-  position: relative;
-  overflow: hidden;
-  border-left: 1px solid var(--oc-color-border);
 }
 </style>
--- a/packages/web-app-user-management/src/components/AppTemplate.vue
+++ b/packages/web-app-user-management/src/components/AppTemplate.vue
@@ -25,8 +25,7 @@
       </div>
       <side-bar
         v-if="sideBarOpen"
-        id="user-management-side-bar"
-        class="oc-width-1-1 oc-width-1-3@m oc-width-1-4@xl"
+        class="user-management-side-bar oc-width-1-1 oc-width-1-3@m oc-width-1-4@xl"
         :available-panels="availableSideBarPanels"
         :sidebar-active-panel="activePanel"
         :loading="false"
@@ -98,7 +97,7 @@ export default defineComponent({
   overflow-y: auto;
 }
 
-#user-management-side-bar {
+.user-management-side-bar {
   position: relative;
   overflow: hidden;
   border-left: 1px solid var(--oc-color-border);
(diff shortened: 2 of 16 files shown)


kulmann (commented):
Getting a `[vue-select warn]: Label key "option.displayValue" does not exist in options object {}.` in the browser console. Could you look into fixing that?

kulmann (commented):
If the height of the viewport is not sufficient, scrolling to the bottom in the right sidebar only shows the input field, but not the description message below.

Screenshot 2022-08-15 at 16 00 32
Screenshot 2022-08-15 at 16 00 47

JanAckermann:
>Getting a [vue-select warn]: Label key "option.displayValue" does not exist in options object {}. in the browser console. Could you look into fixing that?
Fixed

packages/web-app-user-management/src/views/Groups.vue
```diff
@@ -240,62 +208,46 @@ export default defineComponent({
 
   methods: {
     ...mapActions(['showMessage']),
+
     calculateListHeaderPosition() {
-      this.listHeaderPosition = this.$refs.appBar.getBoundingClientRect().height
+      this.listHeaderPosition = this.$refs.template.$refs.appBar.getBoundingClientRect().height
```
kulmann, line 213: Soft-fails in unit tests because `this.$refs.template.$refs` is undefined in the unit tests of the view (-> gives an ugly error message in the unit tests)

changelog/unreleased/bugfix-user-management-app-close-sidebar-throws-error
```diff
@@ -0,0 +1,6 @@
+Bugfix: User management app close side bar throws error
+
+We've fixed a bug, where closing the side bar in the user management app thrown an error,
```
kulmann, line 3:
```suggestion
We've fixed a bug, where closing the side bar in the user management app threw an error
```

packages/web-app-user-management/src/components/AppTemplate.vue
```diff
@@ -0,0 +1,105 @@
...
+              </oc-button>
+            </div>
+          </div>
+          <slot name="topbarActions" class="oc-flex-1 oc-flex oc-flex-start" />
+        </div>
+        <slot name="mainContent" />
+      </div>
+      <side-bar
+        v-if="sideBarOpen"
+        class="user-management-side-bar oc-width-1-1 oc-width-1-3@m oc-width-1-4@xl"
```
kulmann, line 28: Sidebar in the files app has a fixed width by now (except for mobile resolutions, where it takes 100% of the width). Worth to apply that here as well?

packages/web-app-user-management/src/components/AppTemplate.vue
```diff
@@ -0,0 +1,105 @@
...
+    },
+    breadcrumbs: {
+      required: true,
+      type: Array
+    },
+    sideBarOpen: {
+      required: true,
+      type: Boolean
+    },
+    availableSideBarPanels: {
```
kulmann, line 63:
```suggestion
    sideBarAvailablePanels: {
```

packages/web-app-user-management/src/components/AppTemplate.vue
```diff
@@ -0,0 +1,105 @@
...
+      required: true,
+      type: Array
+    },
+    activePanel: {
```
kulmann, line 67:
```suggestion
    sideBarActivePanel: {
```

packages/web-app-user-management/src/views/Groups.vue
```diff
@@ -1,5 +1,64 @@
 <template>
   <div>
+    <app-template
+      ref="template"
+      :loading="loadResourcesTask.isRunning"
```
kulmann, line 5:
```suggestion
      :loading="loadResourcesTask.isRunning || !loadResourcesTask.last"
```
`isRunning` doesn't evaluate to `true` if the task was never started (yet).

packages/web-app-user-management/src/views/Users.vue
```diff
@@ -1,5 +1,64 @@
 <template>
   <div>
+    <app-template
+      ref="template"
+      :loading="loadResourcesTask.isRunning"
```
kulmann, line 5:
```suggestion
      :loading="loadResourcesTask.isRunning || !loadResourcesTask.last"
```
same as for groups: `isRunning` doesn't evaluate to true if the task was never started (yet).

packages/web-app-user-management/src/views/Users.vue
```diff
@@ -356,7 +303,7 @@ export default defineComponent({
     ...mapMutations('runtime/spaces', ['UPDATE_SPACE_FIELD']),
 
     calculateListHeaderPosition() {
-      this.listHeaderPosition = this.$refs.appBar.getBoundingClientRect().height
+      this.listHeaderPosition = this.$refs.template.$refs.appBar.getBoundingClientRect().height
```
kulmann, line 306: Same as for groups, soft-fails in unit tests because this.$refs.template.$refs is undefined in the unit tests of the view (-> gives an ugly error message in the unit tests)

packages/web-pkg/src/components/sidebar/types.ts
```diff
@@ -8,6 +8,7 @@ export interface Panel {
   title: string
   component: Component
   componentAttrs: any
+  componentHandlers: any
```
kulmann, line 11:
```suggestion
  componentListeners: any
```
I know, ultra nitpick. But from my understanding the handler is only the callback, while the listener maps a handler to a certain event, so what you're adding here are listeners, not handlers. Do you agree, or is that flawed?

Since this is not "some application" but `web-pkg` (which we want to publish at some point in the future) I'd like to be strict with naming.

kulmann (changes requested):
Still found some nitpicks, please have a look (see comments).

Apart from that, but not blocking the PR: Unit tests for the `AppTemplate` don't exist at all at the moment. Can happen in a followup PR, as I understand that this one should go into master soonish. But still, needs to be tackled. At the moment the `AppTemplate` usage in both user management views is stubbed (both views use a shallow mount in their unit tests) and the `AppTemplate` doesn't have dedicated unit tests on its own. Both options are valid - testing the `AppTemplate` in a separate spec file would reduce boilerplate / test duplication.

packages/web-app-user-management/src/components/AppTemplate.vue
JanAckermann: Needs a little code base changes, but done

kulmann (approved):
As discussed, unit tests for app template will happen in a followup PR. Rest looks good to me! 💪

JanAckermann:
Tests: https://github.com/owncloud/web/pull/7461

PR #7467, opened 2022-08-16
Fetch and show quota limitations

JanAckermann:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/store/spaces.ts
+++ b/packages/web-runtime/src/store/spaces.ts
@@ -76,16 +76,14 @@ const actions = {
     context.commit('LOAD_SPACES', spaces)
   },
   async loadSpaceQuotas(context, { httpAuthenticatedClient }) {
-    // FIXME: Use graphClient
-    const response = await httpAuthenticatedClient.get(
+    // FIXME: Use graphClient after added to libre-graph-api specs https://github.com/owncloud/libre-graph-api
+    const { data } = await httpAuthenticatedClient.get(
       `${context.rootGetters.configuration.server}graph/v1.0/drivequotas`
     )
-    console.log(response)
-    // TODO: Make something with response
     context.commit('LOAD_SPACE_QUOTAS', {
-      maxPersonalQuota: 10 * Math.pow(10, 9),
-      defaultProjectQuota: 10 * Math.pow(10, 9),
-      maxProjectQuota: 100 * Math.pow(10, 9)
+      maxProjectQuota: data.max_project_quota || 0,
+      maxPersonalQuota: data.max_personal_quota || 0,
+      defaultProjectQuota: data.default_project_quota || 0
     })
   }
 }
--- a/packages/web-runtime/src/store/spaces.ts
+++ b/packages/web-runtime/src/store/spaces.ts
@@ -3,11 +3,13 @@ import Vue from 'vue'
 import { set, has } from 'lodash-es'
 
 const state = {
-  spaces: []
+  spaces: [],
+  spaceQuotas: {}
 }
 
 const getters = {
-  spaces: (state) => state.spaces
+  spaces: (state) => state.spaces,
+  spaceQuotas: (state) => state.spaceQuotas
 }
 
 const mutations = {
@@ -57,6 +59,9 @@ const mutations = {
   },
   CLEAR_SPACES(state) {
     state.spaces = []
+  },
+  LOAD_SPACE_QUOTAS(state, spaceQuotas) {
+    state.spaceQuotas = spaceQuotas
   }
 }
 
@@ -69,6 +74,13 @@ const actions = {
 
     const spaces = graphResponse.data.value.map((space) => buildSpace(space))
     context.commit('LOAD_SPACES', spaces)
+  },
+  async loadSpaceQuotas(context, { clientService }) {
+    await context.commit('LOAD_SPACE_QUOTAS', {
+      maxPersonalQuota: 10 * Math.pow(10, 9),
+      defaultProjectQuota: 10 * Math.pow(10, 9),
+      maxProjectQuota: 100 * Math.pow(10, 9)
+    })
   }
 }
 
--- a/packages/web-runtime/src/store/spaces.ts
+++ b/packages/web-runtime/src/store/spaces.ts
@@ -75,8 +75,14 @@ const actions = {
     const spaces = graphResponse.data.value.map((space) => buildSpace(space))
     context.commit('LOAD_SPACES', spaces)
   },
-  async loadSpaceQuotas(context, { clientService }) {
-    await context.commit('LOAD_SPACE_QUOTAS', {
+  async loadSpaceQuotas(context, { httpAuthenticatedClient }) {
+    // FIXME: Use graphClient
+    const response = await httpAuthenticatedClient.get(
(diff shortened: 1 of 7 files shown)


packages/web-runtime/src/store/spaces.ts
```diff
@@ -69,6 +74,13 @@ const actions = {
...
     const spaces = graphResponse.data.value.map((space) => buildSpace(space))
     context.commit('LOAD_SPACES', spaces)
+  },
+  async loadSpaceQuotas(context, { clientService }) {
+    await context.commit('LOAD_SPACE_QUOTAS', {
+      maxPersonalQuota: 10 * Math.pow(10, 9),
+      defaultProjectQuota: 10 * Math.pow(10, 9),
+      maxProjectQuota: 100 * Math.pow(10, 9)
+    })
   }
```
JanAckermann, line 84: Needs: 
https://github.com/owncloud/ocis/pull/4388

packages/web-runtime/src/store/spaces.ts
```diff
@@ -69,6 +74,17 @@ const actions = {
 
     const spaces = graphResponse.data.value.map((space) => buildSpace(space))
     context.commit('LOAD_SPACES', spaces)
+  },
+  async loadSpaceQuotas(context, { httpAuthenticatedClient }) {
+    // FIXME: Use graphClient after added to libre-graph-api specs https://github.com/owncloud/libre-graph-api
```
kulmann, line 79: Why not do that first and wait for that? Is this PR that urgent?

packages/web-runtime/src/store/spaces.ts
```diff
@@ -69,6 +74,17 @@ const actions = {
...
+    const { data } = await httpAuthenticatedClient.get(
+      `${context.rootGetters.configuration.server}graph/v1.0/drivequotas`
+    )
+    context.commit('LOAD_SPACE_QUOTAS', {
+      maxProjectQuota: data.max_project_quota || 0,
```
kulmann, line 84: I don't understand the API decision. Why is it not a map by drive type?

packages/web-runtime/src/store/spaces.ts
JanAckermann: I don’t think that it is planned right now.
Maybe @fschade knows more

kulmann: Is it available in the backend side graph api and just not in the spec? Because that would be a low hanging fruit. Or is the endpoint not implemented for graph API at all?

packages/web-runtime/src/store/spaces.ts
kulmann: We could even include a max quota in the individual drives, without the need for an additional endpoint.

packages/web-runtime/src/store/spaces.ts
JanAckermann: It’s not in the specs but implemented in the graph api on the experimental drive

packages/web-runtime/src/store/spaces.ts
JanAckermann: What do you mean with map by drive type?
project_quota and personal_quota are mapped by the naming already

JanAckermann: I also don’t know if this is future proof at all or just for the experimental branch, I think @butonic will have some objections against the endpoint as it is not defined in the ms graph

kulmann: I mean to **only** use the drive type. The endpoint is `drivequotas` and the state prop is also `quotas`. So IMO it would be sufficient to use the drive types, e.g. `personal`, `project` as keys. Already backend side, so that you wouldn't have to map the quota types manually into your own props.

kulmann: As I said, it could just be another prop in each drive.

kobergj: So I see there are two problems here:

1) The format of the response
I see you would prefer a map driveType to maxsize that makes sense. Allthough the endpoint could no longer report default quotas. But that is fine for me.

2) The location of the response
I agree that adding an endpoint that is not in the MS Graph spec doesn't make much sense. But the drive also has no prop for max quota, so that would be a custom change too. As far as I see there is no concept of a max quota in graph. Additionally, if we add it to the drive, a user with no drive will not be able to see this value.

JanAckermann: @kulmann @kobergj 
I would highly suggest using the capabilities provider for that, we already have a spaces property, where the (max) quota configuration could be integrated

kobergj: Unfortunately that is not possible. The capabilities live in reva. Graph is an ocis feature. We cannot get the config value from graph to the capabilities endpoint. The same value would have to be configured in both reva and graph. This is bad. The reva value will be outdated in 12mins.

So the only one who can answer question about max drive quota is the graph service

packages/web-runtime/src/store/spaces.ts
fschade: @kobergj does it make sense to add it to the graphapi? if yes, Is it a low hanging fruit?

packages/web-runtime/src/store/spaces.ts
fschade: we should think twice before we extend the graphapi with something that is not part of the spec...

packages/web-runtime/src/store/spaces.ts
kobergj: @fschade see discussion on other comment, unfortunately it is not that easy...

fschade: ok then we also should skip it, it seems to be not part of the spec anyway. @kulmann thoughts?

packages/web-runtime/src/store/spaces.ts
kobergj: Yes I agree. BUT the graph service enforces the max quotas. So it is the only one that can report the value for it.

fschade (approved):
please wait for the discussion & outcome above... then LGTM

packages/web-runtime/src/store/spaces.ts
fschade: oh no 😢 then i get it, thanks for clarifying

JanAckermann: @fschade @kobergj 
Maybe we can use this endpoint 
https://docs.microsoft.com/de-de/graph/api/usersettings-get?view=graph-rest-1.0

kobergj: That endpoint is supposed to return a `userSettings` object https://docs.microsoft.com/de-de/graph/api/resources/usersettings?view=graph-rest-1.0 We would also need to add a custom prop

JanAckermann:
@fschade good to merge or something we need to clarify?

Issue #7436, opened 2022-08-11
Space Viewer should not see the edit options on shares


micbar:
### Steps to reproduce
1. Create  Space "Mars"
2. Create `Neue Datei.xlsx`
3. Share `Neue Datei.xlsx` with Moss
4. Add Einstein to the space with the `Viewer` Role
5. Log in as Einstein
6. Open the share panel on `Neue Datei.xlsx`

### Expected behaviour
Einstein should see the share with Moss, but should not have the edit dialogue

### Actual behaviour
Einstein can try to edit the share and can submit the changes
1. The value of the role is changed in web, no error message
2. Reload, the edit didn't happen because not allowed by the backend

![1660212789693](https://user-images.githubusercontent.com/4378513/184112083-5fefc8cb-4909-485d-af1f-fcfd91d0c541.png)

### Environment general
**Operating system**:

**Backend (ownCloud Core or Infinite Scale)**:

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

### Environment ownCloud Infinite Scale
**Version: Infinite Scale 2.0.0-beta.5+93bf3356e Community**
**ownCloud Web UI 5.7.0-rc.9**

**Updated from an older Infinite Scale or fresh install:**

**Where did you install Infinite Scale from:**

### Environment ownCloud Core
**Web server:**

**Database:**

**PHP version:**

**Version:** (see ownCloud admin page)

**Updated from an older ownCloud or fresh install:**

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results into https://gist.github.com/ and put the link here.
```

**The content of config/config.php:**

```
Log in to the web-UI with an administrator account and click on
'admin' -> 'Generate Config Report' -> 'Download ownCloud config report'
This report includes the config.php settings, the list of activated apps
and other details in a well-sanitized form.

or 

If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

*ATTENTION:* Do not post your config.php file in public as is. Please use one of the above
methods whenever possible. Both, the generated reports from the web-ui and from occ config:list
consistently remove sensitive data. You still may want to review the report before sending.
If done manually then it is critical for your own privacy to diligently
remove *all* hostnames, passwords, usernames, salts and other credentials before posting.
[...]

PR #7470, opened 2022-08-17
Hide share actions for space viewers/editors

JammingBen:
## Description
We've fixed a bug where viewers and editors of a space could see the actions to edit and remove shares. We've also improved the error handling when something goes wrong while editing/removing shares.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7436

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
@@ -28,7 +28,7 @@
         <li v-for="collaborator in displayCollaborators" :key="collaborator.key">
           <collaborator-list-item
             :share="collaborator"
-            :modifiable="!collaborator.indirect"
+            :modifiable="shareIsModifiable(collaborator)"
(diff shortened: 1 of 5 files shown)


packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
```diff
@@ -28,7 +28,7 @@
         <li v-for="collaborator in displayCollaborators" :key="collaborator.key">
           <collaborator-list-item
             :share="collaborator"
-            :modifiable="!collaborator.indirect"
+            :modifiable="shareIsModifiable(collaborator)"
```
kulmann, line 31:
```suggestion
            :modifiable="isShareModifiable(collaborator)"
```
I personally like when methods with boolean return type have `is`, `are`, etc as prefix of the method name, not in between.

kulmann (approved):

PR #7485, opened 2022-08-19
Add resource name to the WebDAV properties

JammingBen:
## Description
We've added the resource name to the WebDAV properties. It is being used for building the resource (if given).

I also modified the extractFileExtension method to work with it.

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/helpers/resources.ts
+++ b/packages/web-app-files/src/helpers/resources.ts
@@ -49,13 +49,16 @@ export function buildResource(resource): Resource {
   }
 
   const id = resource.fileInfo[DavProperty.FileId]
+  const name = resource.fileInfo[DavProperty.Name]
+    ? resource.fileInfo[DavProperty.Name]
+    : path.basename(resource.name)
 
   return {
     id,
     fileId: resource.fileInfo[DavProperty.FileId],
     storageId: extractStorageId(resource.fileInfo[DavProperty.FileId]),
     mimeType: resource.fileInfo[DavProperty.MimeType],
-    name: path.basename(resource.name),
+    name,
     extension: isFolder ? '' : extension,
     path: resourcePath,
     webDavPath: resource.name,
--- a/packages/web-app-files/src/helpers/resources.ts
+++ b/packages/web-app-files/src/helpers/resources.ts
@@ -34,8 +34,9 @@ export function renameResource(resource, newName, newPath) {
 }
 
 export function buildResource(resource): Resource {
+  const name = resource.fileInfo[DavProperty.Name] || path.basename(resource.name)
   const isFolder = resource.type === 'dir' || resource.type === 'folder'
-  const extension = extractExtensionFromFile(resource)
+  const extension = extractExtensionFromFile({ ...resource, name })
   let resourcePath
 
   if (resource.name.startsWith('/files') || resource.name.startsWith('/space')) {
@@ -49,9 +50,6 @@ export function buildResource(resource): Resource {
   }
 
   const id = resource.fileInfo[DavProperty.FileId]
-  const name = resource.fileInfo[DavProperty.Name]
-    ? resource.fileInfo[DavProperty.Name]
-    : path.basename(resource.name)
 
   return {
     id,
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/helpers/resources.ts
```diff
@@ -49,17 +49,20 @@ export function buildResource(resource): Resource {
   }
 
   const id = resource.fileInfo[DavProperty.FileId]
+  const name = resource.fileInfo[DavProperty.Name]
```
kulmann, line 52: The three lines can be shortened to
```suggestion
  const name = resource.fileInfo[DavProperty.Name] || path.basename(resource.name)
```

packages/web-app-files/src/helpers/resource/resource.ts
```diff
@@ -16,9 +16,8 @@ export const extractNameWithoutExtension = (resource?: Resource): string => {
   return name.substring(0, extensionIndexInName)
 }
 
-export const extractExtensionFromFile = (resource: Resource): string => {
-  const name = resource.name
-  if (resource.type === 'dir' || resource.type === 'folder' || resource.isFolder) return ''
+export const extractExtensionFromFile = (name: string, type: string, isFolder: boolean): string => {
```
kulmann, line 19: Uff, to be honest I liked to only provide an object. What's your concern with it (genuinely interested)? If it's the fact that the `Resource` interface is too broad, you could introduce an interface `ExtractExtensionFromFileOptions` where you only define the attributes needed for the function and use that as type for the input param of the function. The `Resource` interface would then fulfill that options interface as well so that we could still pass in the resource objects.

kulmann (changes requested):

packages/web-app-files/src/helpers/resource/resource.ts
JammingBen: The issue I have is that the given `resource` is not always of the same type. In `buildResource`, the resource object is the one that comes from the SDK (before it's built). In `renameResource` e.g., it's the built resource object, which is completely different.

Now we mainly need the name of the resource object. That would be the `name` property of the built resource object. For the "unbuilt" object, it's something completely different (parsed path or the new DAV property).

We could introduce `ExtractExtensionFromFileOptions`. But it would have a few optional params because again, the resource object completely differs depending on where the method is called from.

kulmann: The nice thing about objects as params is that you can just omit the attributes that are not relevant for your context (as long as they are optional in the interface of course). So I'd still favour an object as param. More specific interface as discussed is definitely a good idea.

packages/web-app-files/src/helpers/resource/resource.ts
```diff
@@ -1,5 +1,13 @@
 import { Resource } from 'web-client'
 import fileExtensions from '../extensions/fileExtensions'
+import { DavProperty } from 'web-pkg/src/constants'
+
+interface ExtractExtensionFromFileOptions {
+  name: string
+  type: string
+  isFolder: boolean
+  fileInfo: object
```
kulmann, line 9: huh, why don't you pass `fileInfo[DavProperty.Name]` as `name` prop into the input object?

JammingBen: That's what I was trying to tell you earlier 😄  This method is called in `buildResource`, when the resource has not been built yet. Hence we don't have `fileInfo[DavProperty.Name]` as `name` yet - it just exists as `fileInfo` property at this point of time.

kulmann: but you can provide an object with `{ name: fileInfo[DavProperty.Name] }` instead of passing in the whole fileInfo struct.

JammingBen: Ah okay, now I get it. Well then we can basically keep `extractExtensionFromFile` untouched and just adjust the argument passed in `buildResource`.

kulmann (approved):
👍

PR #7491, opened 2022-08-19
Stuck after session expired

### LINKED ISSUE #7453: Stuck after session expired
### Steps to reproduce
1. Log in
2. Wait until session expires
3. Following screen appears:

Bildschirmfoto 2022-08-16 um 09 02 15

4. Click on "exit" as suggested

### Expected behaviour
Two options:
- Auto-forward to login screen without this message. I think this is the most common behaviour, I was surprised that the mentioned message showed up at all. 

But maybe this is a required step so second option:
- Show this message but make the "exit" link an actual link. Currently it seems it's just an anchor:
`<a id="exitAnchor">exit</a>`

### Actual behaviour
5. Nothing happens, "exit" is a link without target.

### Environment general
https://ocis.ocis-wopi.latest.owncloud.works

gitstart:
## Description

fixed exit link to redirect to login once session expires

## Related Issue

- [Fixes #7453 ](https://github.com/owncloud/web/issues/7453)
### **Demo**
https://www.loom.com/share/35c70267b1e743b38e468d65147880ff

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-runtime/src/pages/accessDenied.vue
```diff
@@ -12,7 +12,8 @@
         </h3>
         <h4 v-translate class="oc-mb-m">Your user session is invalid or has expired.</h4>
         <div v-translate class="oc-mb-m" @click="performLogout">
-          If you like to login with a different user please proceed to <a id="exitAnchor">exit</a>.
+          If you like to login with a different user please proceed to
+          <router-link id="exitAnchor" :to="{ name: 'login' }">exit</router-link>
```
kulmann, line 16: The click on the router-link now does two things at the same time, and both are needed: the click handler on the outer div fires and the router link is being followed. This is a somewhat confusing mix.

I'd like to suggest a solution as follows:
1. remove the click handler from the outer div and remove the method from the script block
2. introduce a new page in `packages/web-runtime/src/pages` called `logout`
3. define a route in `packages/web-runtime/src/router/index.ts` called `logout` which renders the page component from (2)
4. in the script block of `logout.vue` the `created()` hook should immediately call the `logoutUser` method on the `authService`.

IMO that would be cleaner because there would be an actual link to follow. Which is expected for an inline text element that is styled as a link.

The template of the page should never be visible, but you could add some simple message to it like `Please wait while you're being logged out`.

kulmann: While this solution would be cleaner regarding the flow in the business logic, it still doesn't cause a redirect to the `login` route. I think that this needs to be done in the `user unloaded` event handler in `authService.ts` of the web-runtime. When I tried that a while back it caused redirect loops like crazy when using the regular logout from the top bar user menu.

kulmann (changes requested):
Hey, thanks for your contribution! I have some suggestions, see comments. Happy to assist further of course!

packages/web-runtime/src/pages/accessDenied.vue
WajihaNiazi: Thank you for your nice feedback
let us walk on this.

packages/web-runtime/src/pages/logout.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <div>
+    <p>Please wait while you're being logged out.</p>
```
kulmann, line 3:
```suggestion
    <p v-translate>Please wait while you're being logged out.</p>
```

Please add the `v-translate` directive, so that the english string gets picked up by our extraction scripts for translation.

packages/web-runtime/src/pages/logout.vue
```diff
@@ -0,0 +1,15 @@
...
+  </div>
+</template>
+<script>
+import { authService } from '../services/auth'
+export default {
+  name: 'LogoutPage',
+  created() {
```
kulmann, line 10:
```suggestion
  async created() {
```

packages/web-runtime/src/pages/logout.vue
```diff
@@ -0,0 +1,15 @@
...
+    authService.logoutUser()
```
kulmann, line 11:
```suggestion
    await authService.logoutUser()
```

kulmann: The redirect to the login page should only ever happen after the logout has been finalized. Needs to be awaited.

changelog/unreleased/bugfix-stuck-after-session-expired
```diff
@@ -0,0 +1,9 @@
+Bugfix: Stuck After Session Expired
+
+We've fixed exit link to redirect to login once session expires
+
+We've removed the logout click handler and created a new logout component
+
+https://github.com/owncloud/web/issues/7453
+
```
kulmann, line 8: Please remove the blank line. The `calens` tool which we're using as changelog generator needs URLs at the end of the changelog file to follow each other without empty lines in between.

kulmann (changes requested):

kulmann (approved):
LGTM 💪 🥳

PR #7501, opened 2022-08-22
Right sidebar to views

dschmidt:
## Description
This PR converts the right sidebar of the files app to a child element of the respective view. No changelog, because there is no user- or (external-)dev-impact. The idea is to have the right sidebar as child of views, so that 

a: there is no magic access on shared state needed (still some refactoring to do for that) and 
b: views that are related to a space can pass in the already resolved space into the sidebar.

In other words: the right sidebar is/was the last component that was not a child of their view. With this PR being merged views are more or less in control of all their child components.

## Related Issue

- Works towards https://github.com/owncloud/web/issues/6648

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


kulmann:
I was kind of involved in this PR, maybe @JammingBen could take a look here?

packages/web-app-files/src/App.vue
```diff
@@ -1,60 +1,26 @@
 <template>
   <main id="files" class="oc-flex oc-height-1-1">
     <div v-if="dragareaEnabled" class="dragarea" />
-    <div ref="filesListWrapper" tabindex="-1" class="files-list-wrapper oc-width-expand">
```
kulmann, line 4: fyi: the wrapper div has been moved into a separate component, which is used inside all relevant views, because otherwise the right sidebar would have been a child of this file list wrapper here.

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -550,7 +550,6 @@ export default {
 #upload-info {
   background-color: var(--oc-color-background-secondary);
   width: 400px;
-  z-index: 4;
```
kulmann, line 553: fyi: z-index has been removed here because it's sufficient that the `snackbar` div in the `Application` layout has a high z-index. The notifications that get rendered into the sidebar have their own z-index again, which is not strictly necessary. Didn't want to make an ODS PR for that though.

packages/web-app-files/tests/unit/views/views.setup.js
```diff
@@ -207,20 +205,6 @@ export const getStore = function ({
               closed: sidebarClosed
             },
             namespaced: true
-          },
-          pagination: {
```
kulmann, line 211: fyi: there's no pagination store module in the files app for quite some time now. Removed the dead code here in the tests as well now.

JammingBen (approved):
Looks all good and reasonable to me. Also tested it and the sidebar works as expected in all views 👍

Issue #6996, opened 2022-05-17
Upload of folder with same name shows error


tbsbdr:
# Steps to reproduce
1. Login to https://ocis.ocis-wopi.released.owncloud.works/
2. Create a new folder "Folder" on the ocis instance and on your local drive.
3. upload your local "Folder" folder
4. upload does not start, shows error "Folder "Folder" already exists."
![image](https://user-images.githubusercontent.com/26610733/168844306-c51c1723-77e3-499c-9083-73b4d26ca93f.png)

# Expected behaviour
System should offer me alternative options to upload the folder.
The appropriate dialogues are consistently shown in case of conflicting resources, eg. in context of:
- [ ] Drag'n drop upload
- [ ] upload via upload-button
- [ ] move-dialog
- [ ] copy-dialog
- [ ] trashbin-restore
- etc... 

## General rule
**User gets asked what to do in case of conflicting resources**, as there is no common pattern across different services/platforms (OneDrive, Google Drive, Box, Windows, Mac, etc. - all different).

**Note** This dialog concept is also meant to be applied to https://github.com/owncloud/web/issues/1753

## Dialogs

![image](https://user-images.githubusercontent.com/26610733/168789953-c5d8d37a-5ceb-44ee-954f-0a72e310b476.png)

## Behaviour
1. **Replace** will replace the existing file (and creates a new version implicitly).
2. **Keep both** will create a new resource with an appended number "(xx)" - counting up, if ex. (2) already exists.
3. **Cancel** will skip the current resource, if applicable shows the next conflict.
4. **Merge** will combine folders with the same name; if they contain conflicting files, user gets asked what to do (dialog for files shows up).
5. **Do this for all XX conflicts** will apply the selected option to all XX upcoming conflicts. Is only shown, if there are at least 2 conflicts. 
6. **generally applicable:** The 

## Userflows
![image](https://user-images.githubusercontent.com/26610733/168798364-d4913c33-f1d7-4665-8a92-2c4b07106259.png)

____________
**solves:**
https://github.com/owncloud/web/issues/3751
https://github.com/owncloud/web/issues/5761
https://github.com/owncloud/web/issues/5106
https://github.com/owncloud/web/issues/6546

**groundwork for**
https://github.com/owncloud/web/issues/1753

kulmann:
[...]

PR #7504, opened 2022-08-22
[full-ci] Resolve upload existing folder conflict dialog

lookacat:
## Description
See https://github.com/owncloud/web/issues/6996

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6996

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
@@ -160,7 +160,8 @@ import {
   resolveFileExists,
   ResolveStrategy,
   ResolveConflict,
-  resolveFileNameDuplicate
+  resolveFileNameDuplicate,
+  FileExistsResolver
 } from '../../helpers/resource/copyMove'
 
 export default defineComponent({
@@ -800,7 +801,11 @@ export default defineComponent({
       this.$uppyService.uploadFiles(files)
     },
 
-    async displayOverwriteDialog(files: UppyResource[], conflicts, resolveFileExistsMethod) {
+    async displayOverwriteDialog(
+      files: UppyResource[],
+      conflicts,
+      resolveFileExistsMethod: FileExistsResolver
+    ) {
       let count = 0
       const allConflictsCount = conflicts.length
       const resolvedFileConflicts = []
@@ -839,7 +844,7 @@ export default defineComponent({
           false,
           isFolder
         )
-        count += 1
+        count++
         if (resolvedConflict.doForAllConflicts) {
           if (isFolder) {
             doForAllConflictsFolders = true
@@ -858,7 +863,20 @@ export default defineComponent({
       const filesToSkip = resolvedFileConflicts
         .filter((e) => e.strategy === ResolveStrategy.SKIP)
         .map((e) => e.name)
+      const foldersToSkip = resolvedFolderConflicts
+        .filter((e) => e.strategy === ResolveStrategy.SKIP)
+        .map((e) => e.name)
+
+      files = files.filter((e) => !filesToSkip.includes(e.name))
+      files = files.filter(
+        (file) =>
+          !foldersToSkip.some((folderName) => file.meta.relativeFolder.split('/')[1] === folderName)
+      )
+
       const filesToKeepBoth = resolvedFileConflicts
+        .filter((e) => e.strategy === ResolveStrategy.KEEP_BOTH)
+        .map((e) => e.name)
+      const foldersToKeepBoth = resolvedFolderConflicts
         .filter((e) => e.strategy === ResolveStrategy.KEEP_BOTH)
         .map((e) => e.name)
 
@@ -867,16 +885,6 @@ export default defineComponent({
         const extension = extractExtensionFromFile({ name: fileName } as Resource)
         file.name = resolveFileNameDuplicate(fileName, extension, this.files)
       }
-
-      files = files.filter((e) => !filesToSkip.includes(e.name))
-
-      const foldersToSkip = resolvedFolderConflicts
-        .filter((e) => e.strategy === ResolveStrategy.SKIP)
-        .map((e) => e.name)
-      const foldersToKeepBoth = resolvedFolderConflicts
-        .filter((e) => e.strategy === ResolveStrategy.KEEP_BOTH)
-        .map((e) => e.name)
-
       for (const folder of foldersToKeepBoth) {
         const filesInFolder = files.filter((e) => e.meta.relativeFolder.split('/')[1] === folder)
         for (const file of filesInFolder) {
@@ -895,10 +903,7 @@ export default defineComponent({
           file.meta.routeItem = `/${newFolderName}`
         }
       }
-      files = files.filter(
-        (file) =>
-          !foldersToSkip.some((folderName) => file.meta.relativeFolder.split('/')[1] === folderName)
-      )
+
       if (files.length === 0) return
       this.handleUppyFileUpload(files)
     }
--- a/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
@@ -162,6 +162,7 @@ import {
   ResolveConflict,
   resolveFileNameDuplicate
 } from '../../helpers/resource/copyMove'
+import { resolve } from 'dns'
 
 export default defineComponent({
   components: {
@@ -685,7 +686,8 @@ export default defineComponent({
     },
 
     async onFilesSelected(files: File[]) {
-      const conflicts = []
+      const fileConflicts = []
+      const folderConflicts = []
       const uppyResources: UppyResource[] = this.inputFilesToUppyFiles(files)
       const quotaExceeded = this.checkQuotaExceeded(uppyResources)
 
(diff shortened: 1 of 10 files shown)


ownclouders:
Results for <strong>e2e-tests oC10</strong> https://drone.owncloud.com/owncloud/web/27755/11/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27755/tracing/file-alice-2022-8-23-09-23-45.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27755/tracing/file-brian-2022-8-23-09-23-53.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27755/tracing/folder-alice-2022-8-23-09-22-13.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27755/tracing/folder-brian-2022-8-23-09-22-21.zip 


</p></details>

kulmann (changes requested):
The control flow in `onFilesSelected` is rather complex now and, if I understood correctly, still falls through to the old create version dialog. Can you simplify the control flow (e.g. early returns / continues where possible)? Also, the "create version" dialog should disappear entirely, as it's not needed anymore if we handle the `replace` use case in the different resolve strategies already. Please delete the code related to the create version dialog then.

packages/web-runtime/src/composables/upload/useUpload.ts
```diff
@@ -181,7 +181,9 @@ const createDirectoryTree = ({
             unref(publicLinkPassword)
           )
         } else {
-          await client.files.createFolder(`${file.meta.webDavBasePath}/${folderToCreate}`)
+          try {
+            await client.files.createFolder(`${file.meta.webDavBasePath}/${folderToCreate}`)
+          } catch (ex) {}
```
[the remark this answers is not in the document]
JanAckermann: Additionally, we either name the error variable `error` or `e`, we might want to proceed with this naming scheme. 
Usually, we also log the error with `console.error`, if its worth ignoring the error completely we should write a comment, about why we do this.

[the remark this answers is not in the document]
lookacat: I used it because it throws an error if the folder already exists, i'll try to avoid it by checking beforehand because this way every error gets ignored

kulmann (changes requested):

kulmann (changes requested):
Sorry if this appears to be harsh, but I don't get which concept you're following with this PR. I had the impression that the ticket describes pretty clear what's needed, but let me summarize it again here:
1. No matter if you try to upload a single file, a single folder, multiple files, multiple folders or any mix of that: we only handle conflicts on the "current folder"-level. Meaning if you already have `folder 1` in the current folder and try to upload a folder named `folder 1`, we show the conflict resolve dialog. We don't need to look into the folder and we only need to resolve the conflict once on the current folder level.
2. Files and folders are being handled the same (except for slightly different messages in the modal). I don't get why you have different handlers implemented for files and for folders. For folders you're invoking the conflict resolution directly, for files you only add them to the `conflicts` array and then ignore that array in the handler you're calling afterwards.
3. The technical process should be:
  a. Go through all `uppyResources` and check if they are conflicting with existing files. Resolve the conflict with a respective user choice. Don't do any backend requests here. Only memorize the user choices.
  b. Modify the `uppyResources` according to the user choices
  c. Push the upload queue into the uppy service

What's not clear to me is how we should handle the `Overwrite` case for folders, as the backend can only create a new version for files, not for folders. Sounds like a `delete & upload`-thing to me, which would then need to be reverted if an error happens during the upload. Ideally the backend supports an overwrite flag for folders, but not sure if that exists.

By the way, trying this locally, I immediately ran into situations where folders are attempted to be created and fail because they already exist.

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -797,51 +801,93 @@ export default defineComponent({
...
+        .filter((e) => e.strategy === ResolveStrategy.KEEP_BOTH)
+        .map((e) => e.name)
+
+      for (const fileName of filesToKeepBoth) {
+        const file = files.find((e) => e.name === fileName && !e.meta.relativeFolder)
+        const extension = extractExtensionFromFile({ name: fileName } as Resource)
+        file.name = resolveFileNameDuplicate(fileName, extension, this.files)
+      }
+
+      files = files.filter((e) => !filesToSkip.includes(e.name))
```
JammingBen, line 858: I think we should include `!e.meta.relativeFolder` in the condition to make sure we only deal with root files here. Should be fine without, but better be safe.

packages/web-runtime/src/composables/upload/useUpload.ts
```diff
@@ -181,7 +181,11 @@ const createDirectoryTree = ({
             unref(publicLinkPassword)
           )
         } else {
-          await client.files.createFolder(`${file.meta.webDavBasePath}/${folderToCreate}`)
+          try {
+            await client.files.createFolder(`${file.meta.webDavBasePath}/${folderToCreate}`)
+          } catch (error) {
+            console.log(error)
```
JammingBen, line 187:
```suggestion
            console.error(error)
```

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -797,51 +801,93 @@ export default defineComponent({
...
+
+      const foldersToSkip = resolvedFolderConflicts
+        .filter((e) => e.strategy === ResolveStrategy.SKIP)
+        .map((e) => e.name)
+      const foldersToKeepBoth = resolvedFolderConflicts
+        .filter((e) => e.strategy === ResolveStrategy.KEEP_BOTH)
+        .map((e) => e.name)
+
+      for (const folder of foldersToKeepBoth) {
+        const filesInFolder = files.filter((e) => e.meta.relativeFolder.startsWith(`/${folder}`))
```
JammingBen, line 868: I found a scenario where this is an issue:

* Upload folder `abc`
* Then upload the folders `abc` and `abcd`
* Click "Keep both" in the overwrite modal -> The folder `abc` gets a suffix (correct) as well as the folder `abcd` (wrong).

Splitting `e.meta.relativeFolder` via `/` and using the part between the first two `/` should solve this.

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -797,51 +801,93 @@ export default defineComponent({
...
+          const data = file.data as any
+          data.relativePath = data.relativePath.replace(`/${folder}/`, `/${newFolderName}/`)
+          file.meta.routeItem = `/${newFolderName}`
         }
       }
-
-      this.createModal(modal)
+      files = files.filter(
+        (file) =>
+          !foldersToSkip.some((folderName) => file.meta.relativeFolder.startsWith(`/${folderName}`))
```
JammingBen, line 887: Same as above, we need to split `e.meta.relativeFolder` here.

JammingBen (dismissed):

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
lookacat: cant do that, or we loose all subfolder files

JammingBen: But we don't care about these here, right? Those are being handled via the folder conflicts I thought 🤔

JammingBen: Nevermind, works as expected as we filter out nested files beforehand.

kulmann (changes requested):
Added some comments regarding code style. Checked the branch locally, works like a charm. 🚀

packages/web-app-files/src/helpers/resource/copyMove.ts
```diff
@@ -19,6 +19,19 @@ interface FileConflict {
...
+  (
+    createModal: void,
+    hideModal: void,
+    resource: Resource,
+    conflictCount: number,
+    $gettext: void,
+    $gettextInterpolate: void,
+    isSingleConflict: boolean,
+    suggestMerge: boolean
+  )
```
kulmann, line 32: The return type is missing 🤔

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -858,25 +863,28 @@ export default defineComponent({
...
-      for (const fileName of filesToKeepBoth) {
-        const file = files.find((e) => e.name === fileName && !e.meta.relativeFolder)
-        const extension = extractExtensionFromFile({ name: fileName } as Resource)
-        file.name = resolveFileNameDuplicate(fileName, extension, this.files)
-      }
-
       files = files.filter((e) => !filesToSkip.includes(e.name))
+      files = files.filter(
+        (file) =>
+          !foldersToSkip.some((folderName) => file.meta.relativeFolder.split('/')[1] === folderName)
```
kulmann, line 873: Please include the condition in line 870. Doesn't make much sense iterating twice for filtering, when you can do all the filtering in one iteration.

Filter should look like this:
```js
files = files.filter(
        (file) =>
          !filesToSkip.includes(file.name) &&
          !foldersToSkip.some((folderName) => file.meta.relativeFolder.split('/')[1] === folderName)
      )
```

kulmann (approved):
I have two recommendations (see comments) - but they can happen in a subsequent PR if you want to get this one merged (squash merge please!).

Issue #7498, opened 2022-08-22
Sidebar content floats out of the screen


hurradieweltgehtunter:
### Steps to reproduce
1. Open sidebar in file list

### Expected behaviour
The Sidebar should always fully display all elements despite of the screen size.

### Actual behaviour
Sidebar exceeds right window border on different windows widths
see red box in the bottom right corner in the video for dimensions when it happens

https://user-images.githubusercontent.com/16665512/185870242-68c83d81-b7c9-46bf-a0e6-6b91b86187bd.mp4

### Environment general
https://ocis.ocis-wopi.latest.owncloud.works

JammingBen:
Ahh okay, it happens when selecting German as language, because the words are generally longer. No problem in English though. Thx for noticing!

I adjusted the title and the expected behavior a bit as the sidebar is supposed to have a fixed width and shouldn't resize.

tbsbdr:
Proposal:

**Some space left** --> show Icons only

![](https://user-images.githubusercontent.com/26610733/185883052-ad781f18-71aa-4002-aca9-f803bb535de6.png)

**Almost no space left** --> move quickactions to 3-dots menu

PR #7508, opened 2022-08-23
[full-ci] Fix right sidebar content on small screens

JammingBen:
## Description
We've fixed the right sidebar content on small screens because some screen sizes caused the content to flow out of the screen. Things that have been done to achieve this:

* Selection info has been removed.
* Labels of the batch actions will hide on screens <1280px if the sidebar is open.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7498

## Types of changes

- [ ] Technical debt
- [ ] Tests


JammingBen:
@kulmann How "bad" is it that the batch actions always show a tooltip if the right sidebar is open? If that's an issue, I probably need to solve this in JS and use a resize observer... which I _really_ don't want to 😅

ownclouders:
Results for <strong>e2e-tests oC10</strong> https://drone.owncloud.com/owncloud/web/27765/11/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27765/tracing/file-alice-2022-8-23-01-56-26.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27765/tracing/file-brian-2022-8-23-01-56-32.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27765/tracing/folder-alice-2022-8-23-01-54-57.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/27765/tracing/folder-brian-2022-8-23-01-55-04.zip 


</p></details>

packages/web-app-files/src/components/AppBar/SelectedResources/BatchActions.vue
```diff
@@ -84,5 +87,14 @@ export default {
     display: flex;
     gap: var(--oc-space-small);
   }
+
+  &-squashed {
```
kulmann, line 91: Why only for squashed state? IMO we could do the same for when the sidebar is closed.

packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -32,13 +32,18 @@
         </div>
       </div>
       <div class="files-app-bar-actions">
-        <div class="oc-flex-1 oc-flex oc-flex-start" style="gap: 15px">
+        <div class="oc-flex-1 oc-flex oc-flex-start">
           <slot v-if="showActionsOnSelection || selectedFiles.length === 0" name="actions" />
-          <size-info
+          <oc-button
```
kulmann, line 37: We could easily implement the `Clear selection` action as a file action (see actions mixins). That would remove the need for special code here. What do you think?

JammingBen: Good idea, I'll do it that way 👍

kulmann:
I have one more: with an active file clipboard we'd need to apply the same logic to buttons without selection. You can reach this state by e.g. copying or cutting files and then navigating into a different folder. See screenshot below, on small screens the right sidebar gets cropped again. For the `Paste here` button I suspect that we'll need a divider between the `Paste` and `Cancel` icons in the button group...

Screenshot 2022-08-24 at 14 47 01

JammingBen:
Oof. Yeah that's right, haven't thought of that. The same applies to the "New" and the "Upload" button as well, but only in the "active clipboard" + sidebar open state, correct?

kulmann:
> Oof. Yeah that's right, haven't thought of that. The same applies to the "New" and the "Upload" button as well, but only in the "active clipboard" + sidebar open state, correct?

Thought about it with @tbsbdr - could we hide the labels of `New` and `Upload`, but keep the label of `Paste here`? The first two buttons are easier to understand than the paste button.

JammingBen:
Sure. One question remains though: How do we deal with tooltips? The batch actions now always have a tooltip - the same would apply here. The alternative would be to mess around with a resize observer and toggle the tooltips conditionally depending on the screen size.

tbsbdr:
would be good to hide the tooltips if the labels are visible

packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -79,6 +76,12 @@ export default {
     hasViewOptions: { type: Boolean, default: true },
     showActionsOnSelection: { type: Boolean, default: false }
   },
+  data: function () {
+    return {
+      resizeObserver: new ResizeObserver(this.onResize),
+      limitedScreenSpace: false
```
JammingBen, line 82: Bit of an ugly name. I had set it to "showTooltips" originally, but the TopBar component should not determine wether to show or hide tooltips - that's up to the component itself. Therefore the TopBar only provides the info that limited screen space is available.

kulmann: Isn't the same valid for the `BatchActions` component then? That one still has the `showTooltips` prop.

kulmann (approved):

packages/web-app-files/src/components/AppBar/AppBar.vue
JammingBen: For `BatchActions`, there is no other condition to show tooltips. Hence I figured this component can have a `showTooltips` prop.

Issue #7540, opened 2022-08-29
Inconsistent hover effect (left sidebar)


hurradieweltgehtunter:
### Steps to reproduce
1. Hover over elements of left sidebar

### Expected behaviour
Hover effect should be the same as in context menu, file list or right sidebar (grey background)

### Actual behaviour
No hover effect is visible

### Environment general
**Operating system**: https://ocis.ocis-wopi.latest.owncloud.works

PR #7567, opened 2022-09-01
Add hover effect for left sidebar

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/7540

## Screenshots:

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/components/SidebarNav/SidebarNav.vue
+++ b/packages/web-runtime/src/components/SidebarNav/SidebarNav.vue
@@ -141,6 +141,22 @@ export default {
   .toggle-sidebar-button-expanded {
     justify-content: flex-end !important;
   }
+
+  .oc-sidebar-nav li a {
+    &:hover,
+    &:focus {
+      text-decoration: none !important;
+      background-color: var(--oc-color-background-hover);
+      color: var(--oc-color-swatch-passive-default);
+    }
+  }
+
+  .oc-sidebar-nav li a.active {
+    &:focus,
+    &:hover {
+      color: var(--oc-color-swatch-inverse-default);
+    }
+  }
 }
 .oc-app-navigation-expanded {
   min-width: 230px !important;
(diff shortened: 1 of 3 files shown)


JanAckermann (approved):

packages/web-runtime/src/components/SidebarNav/SidebarNav.vue
```diff
@@ -141,6 +141,22 @@ export default {
   .toggle-sidebar-button-expanded {
     justify-content: flex-end !important;
   }
+
+  .oc-sidebar-nav li a {
+    &:hover,
+    &:focus {
+      text-decoration: none !important;
+      background-color: var(--oc-color-background-hover);
+      color: var(--oc-color-swatch-passive-default);
```
kulmann, line 150: can you reproduce the following? in dark mode there is a slight delay until the text is rendered in the brighter color. This is not observable in light mode because there the text color is the same as without highlighting.

JammingBen: Indeed. I've removed the transition (see https://github.com/owncloud/web/pull/7567/commits/8f1d848dcf797ee22b8fd15ecf134871224efaf4), please check again. I didn't notice any weird behavior after removing this.

kulmann: 👍

kulmann (approved):

diocas:
This PR is causing a strange visual effect. When I click in the menu, the background of the entry changes immediately. But then the effect that was there before still happens, so you see 2 blue boxes merging.

In the picture you can see a switch from the previously selected "Personal" to "Deleted files"
Screenshot 2022-09-01 at 17 29 56

kulmann:
> This PR is causing a strange visual effect. When I click in the menu, the background of the entry changes immediately. But then the effect that was there before still happens, so you see 2 blue boxes merging.
> 
> In the picture you can see what happens when switching from (the previously selected) "Personal" to "Deleted files" Screenshot 2022-09-01 at 17 29 56

True... thank you! Fixed in https://github.com/owncloud/web/pull/7575

Issue #7506, opened 2022-08-22
(incorrectly) limited link roles in personal space


kulmann:
### Steps to reproduce
1. Log in as einstein
2. Create a folder "from-einstein"
3. Share folder "from-einstein" with marie
4. Log in as marie
5. Visit "Shared with me" page
6. Accept "from-einstein" share
7. Create a quick link with viewer permissions (= default) on "from-einstein" (e.g. via right-click context menu)
8. Navigate to "Personal" space (still as marie)
9. Create a folder "my-stuff"
10. Create quick link on "my-stuff"
11. Try to edit the role of the quicklink on "my-stuff"

### Expected behaviour
As "my-stuff" belongs to marie, all possible roles need to be available

### Actual behaviour
Only the "Viewer" role is available

### Bug analysis
`FileLinks.vue` has a computed prop `share` which only ever picks the first incoming share it finds in the root of the shares tree. This is wrong because the root contains more share entities than just the ones related to the  given folder "my-stuff", in this case also the reshared folder "from-einstein". Because of that the link from (10) is identified as a reshare, which is wrong.

PR #7580, opened 2022-09-02
[full-ci] Fix sharesTree loading

JammingBen:
## Description
We've improved loading of the shares tree:

* It now happens more globally in the sidebar component instead of in each sidebar panel.
* Shares won't be loaded for resources without a path anymore.

These changes massively improve the sidebar performance and fix several issues with (re-)share permissions.

Note: Needs ocis with the latest Reva update to include the `shareId`.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7506
- Fixes https://github.com/owncloud/web/issues/7593
- Fixes https://github.com/owncloud/web/issues/7592

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-shares-tree-loading
+++ b/changelog/unreleased/bugfix-shares-tree-loading
@@ -1,6 +1,11 @@
-Bugfix: Hide share actions for space viewers/editors
+Bugfix: Shares tree loading
 
-We now make sure that an empty string will never be used as key for the shares tree. This fixes several issues with wrong share permissions being loaded and also improves performance in folders.
+We've improved loading of the shares tree:
+
+* It now happens more globally in the sidebar component instead of in each sidebar panel.
+* Shares won't be loaded for resources without a path anymore.
+
+These changes massively improve the sidebar performance and fix several issues with (re-)share permissions.
 
 https://github.com/owncloud/web/issues/7506
 https://github.com/owncloud/web/issues/7593
--- a/changelog/unreleased/bugfix-shares-tree-loading
+++ b/changelog/unreleased/bugfix-shares-tree-loading
@@ -9,4 +9,5 @@ These changes massively improve the sidebar performance and fix several issues w
 
 https://github.com/owncloud/web/issues/7506
 https://github.com/owncloud/web/issues/7593
+https://github.com/owncloud/web/issues/7592
 https://github.com/owncloud/web/pull/7580
--- a/changelog/unreleased/bugfix-shares-tree-loading
+++ b/changelog/unreleased/bugfix-shares-tree-loading
@@ -1,0 +1,7 @@
+Bugfix: Hide share actions for space viewers/editors
+
+We now make sure that an empty string will never be used as key for the shares tree. This fixes several issues with wrong share permissions being loaded and also improves performance in folders.
+
+https://github.com/owncloud/web/issues/7506
+https://github.com/owncloud/web/issues/7593
+https://github.com/owncloud/web/pull/7580
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -352,39 +352,40 @@ export default defineComponent({
   watch: {
     file() {
       this.loadData()
-      this.refreshShareDetailsTree()
     },
-    sharesTree() {
-      // missing early return
-      this.sharedItem = null
-      this.shareIndicators = getIndicators(this.file, this.sharesTree)
-      const sharePathParentOrCurrent = this.getParentSharePath(this.file.path, this.sharesTree)
-      if (sharePathParentOrCurrent === null) {
-        return
-      }
-      const shares = this.sharesTree[sharePathParentOrCurrent]?.filter((s) =>
-        ShareTypes.containsAnyValue(
-          [...ShareTypes.individuals, ...ShareTypes.unauthenticated],
-          [s.shareType]
+    sharesTree: {
+      handler() {
+        // missing early return
+        this.sharedItem = null
+        this.shareIndicators = getIndicators(this.file, this.sharesTree)
+        const sharePathParentOrCurrent = this.getParentSharePath(this.file.path, this.sharesTree)
+        if (sharePathParentOrCurrent === null) {
+          return
+        }
+        const shares = this.sharesTree[sharePathParentOrCurrent]?.filter((s) =>
+          ShareTypes.containsAnyValue(
+            [...ShareTypes.individuals, ...ShareTypes.unauthenticated],
+            [s.shareType]
+          )
         )
-      )
-      if (shares.length === 0) {
-        return
-      }
+        if (shares.length === 0) {
+          return
+        }
 
-      this.sharedItem = shares[0]
-      this.sharedByName = this.sharedItem.owner?.name
-      this.sharedByDisplayName = this.sharedItem.owner?.displayName
-      if (this.sharedItem.owner?.additionalInfo) {
-        this.sharedByDisplayName += ' (' + this.sharedItem.owner.additionalInfo + ')'
-      }
-      this.sharedTime = this.sharedItem.stime
-      this.sharedParentDir = sharePathParentOrCurrent
+        this.sharedItem = shares[0]
+        this.sharedByName = this.sharedItem.owner?.name
+        this.sharedByDisplayName = this.sharedItem.owner?.displayName
+        if (this.sharedItem.owner?.additionalInfo) {
+          this.sharedByDisplayName += ' (' + this.sharedItem.owner.additionalInfo + ')'
+        }
+        this.sharedTime = this.sharedItem.stime
+        this.sharedParentDir = sharePathParentOrCurrent
+      },
+      immediate: true
     }
   },
   mounted() {
     this.loadData()
-    this.refreshShareDetailsTree()
   },
   asyncComputed: {
     preview: {
@@ -405,16 +406,8 @@ export default defineComponent({
     }
   },
   methods: {
-    ...mapActions('Files', ['loadPreview', 'loadVersions', 'loadSharesTree']),
-    async refreshShareDetailsTree() {
-      await this.loadSharesTree({
-        client: this.$client,
-        path: this.file.path,
-        $gettext: this.$gettext,
-        storageId: this.file.fileId
-      })
-      this.shareIndicators = getIndicators(this.file, this.sharesTree)
-    },
+    ...mapActions('Files', ['loadPreview', 'loadVersions']),
+
     getParentSharePath(childPath, shares) {
       let currentPath = childPath
       if (!currentPath) {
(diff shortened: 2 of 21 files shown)


JammingBen:
Shares loading via empty string `''` leads to ALL shares being loaded, which is a bad thing when loading shares for your personal root for example (and in general IMO). The only reason for this is to retrieve the parent share in the shares jail, so the parent permissions won't be exceeded.

Maybe we shouldn't go the shares tree route here. Instead we could retrieve the parent share via the shareId, which we got in the resource now. The only issue here is that we somehow need to store the parent share somewhere? Or always load it when the shares panel gets opened. @kulmann What do you think?

packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
```diff
@@ -183,7 +183,7 @@ export default defineComponent({
     },
     share() {
       // the root share has an empty key in the shares tree. That's the reason why we retrieve the share by an empty key here
```
kulmann, line 185: please adjust comment as well

packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -184,7 +184,7 @@ export default defineComponent({
 
     share() {
       // the root share has an empty key in the shares tree. That's the reason why we retrieve the share by an empty key here
```
kulmann, line 186: please adjust comment as well

ownclouders:
Results for <strong>oC10SharingExternalRoot</strong> https://drone.owncloud.com/owncloud/web/28114/43/1
:boom: The oC10SharingExternalRoot tests pipeline failed. The build has been cancelled.

packages/web-app-files/src/helpers/path.js
```diff
@@ -27,6 +27,10 @@ export function getParentPaths(path = '', includeCurrent = false) {
 
   sections.pop()
   while (sections.length > 0) {
+    if (!sections.join('/')) {
```
kulmann, line 30: What about changing line 22 to
```js
const sections = s.split('/').filter(Boolean)
```
instead to remove empty strings from the array?

packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
```diff
@@ -223,7 +223,8 @@ export default defineComponent({
     window.removeEventListener('keydown', this.cycleRoles)
   },
 
-  mounted() {
+  async mounted() {
+    await this.loadIncomingParentShare.perform(this.resource)
```
kulmann, line 227: doesn't this try to load the incoming parent shares for each single `RoleDropdown` component instance? Can we load the incoming parent shares once in the `SideBar.vue` or in the `SharesPanel.vue` and hand it down to the child components via provide/inject (if from SideBar.vue) or via component properties (if coming from SharesPanel.vue)?

kulmann (changes requested):

packages/web-app-files/src/helpers/path.js
JammingBen: That would actually result in a different array at the end. When `''` is removed beforehand, the leading slash would be missing.

packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
```diff
@@ -258,11 +251,6 @@ export default defineComponent({
       return this.sharedParentRoute?.params?.item.split('/').pop()
     },
 
-    shareDetailsHelperContent() {
```
JammingBen, line 261: Fyi unused property.

packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
```diff
@@ -188,38 +188,6 @@ export default {
       return this.collaborators.length > 0
     },
 
-    /**
```
JammingBen, line 191: Fyi unused property.

packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
```diff
@@ -294,20 +259,6 @@ export default {
       const translatedFolder = this.$gettext("You don't have permission to share this folder.")
       return this.highlightedFile.type === 'file' ? translatedFile : translatedFolder
     },
-
```
JammingBen, line 297: Fyi unused property.

packages/web-app-files/src/composables/parentShare/useIncomingParentShare.ts
```diff
@@ -0,0 +1,34 @@
+import { buildShare } from '../../helpers/resources'
+import { useStore } from 'web-pkg/src/composables'
+import { computed, ref, unref } from '@vue/composition-api'
+import { useTask } from 'vue-concurrency'
+import { clientService } from 'web-pkg/src/services'
+
+export function useIncomingParentShare() {
+  const store = useStore()
+  const incomingParentShare = ref(null)
+  const sharesTree = computed(() => store.state.Files.sharesTree)
```
kulmann, line 10: Please use the getter for the sharesTree: `store.getters['Files/sharesTree']`.
Reason: In my experience that's a good idea for building backwards compatibility. If something from the state gets moved or removed you can write some logic in the getter to keep it compatible for a while and deprecate it. With state that's not possible. That being said I'd even recommend to never access vuex state directly (outside of the store module itself).

packages/web-app-files/src/composables/parentShare/useIncomingParentShare.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const loadIncomingParentShare = useTask(function* (signal, resource) {
+    let parentShare
+    for (const shares of Object.values(unref(sharesTree)) as any) {
```
kulmann, line 14: Please walk the sharesTree from the leaf up to the root in sorted manner to get the closest match in case of parent shares on multiple levels.

kulmann: Can we ensure that the `sharesTree` is already loaded?

packages/web-app-files/src/composables/parentShare/useIncomingParentShare.ts
```diff
@@ -0,0 +1,34 @@
...
+      parentShare = shares.find((s) => s.incoming)
+      if (parentShare) {
+        incomingParentShare.value = parentShare
+        return
+      }
+    }
+
+    if (resource.shareId) {
+      const parentShare = yield clientService.owncloudSdk.shares.getShare(resource.shareId)
```
kulmann, line 23: the `parentShare` variable is already defined above, so you should drop the `const` here, no?

kulmann (changes requested):

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -352,39 +352,41 @@ export default defineComponent({
...
 
-      this.sharedItem = shares[0]
-      this.sharedByName = this.sharedItem.owner?.name
-      this.sharedByDisplayName = this.sharedItem.owner?.displayName
-      if (this.sharedItem.owner?.additionalInfo) {
-        this.sharedByDisplayName += ' (' + this.sharedItem.owner.additionalInfo + ')'
-      }
-      this.sharedTime = this.sharedItem.stime
-      this.sharedParentDir = sharePathParentOrCurrent
+        this.sharedItem = shares[0]
```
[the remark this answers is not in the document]
JammingBen: Sounds worth trying, although I would like to do that in a separate PR.

packages/web-app-files/src/composables/parentShare/useIncomingParentShare.ts
JammingBen: > Please walk the sharesTree from the leaf up to the root in sorted manner to get the closest match in case of parent shares on multiple levels.

It seems that this is already the case. Honestly, I don't know why, maybe the combination of `Object.values()` and `unref()` leads to this?

> Can we ensure that the sharesTree is already loaded?

Yes. `loadIncomingParentShare()` is being called in a `sharesTree` watcher in the `SharesPanel` component (= after the `sharesTree` has been loaded.

tests/e2e/support/objects/app-files/share/actions.ts
```diff
@@ -81,13 +81,14 @@ export const inviteMembers = async (args: inviteMembersArgs): Promise<void> => {
...
     const shareInputLocator = page.locator(invitationInput)
+    await shareInputLocator.click()
     await Promise.all([
       page.waitForResponse((resp) => resp.url().includes('sharees') && resp.status() === 200),
       shareInputLocator.fill(recipient.id)
     ])
     await shareInputLocator.focus()
     await page.waitForSelector('.vs--open')
-    await page.locator(invitationInput).press('Enter')
+    await page.locator('.vs__dropdown-option').click()
```
JammingBen, line 91: I don't know why these changes are necessary, but without them, e2e tests won't succeed. I assume this happens because the panel is now loading way faster than before... but that's just a guess.

changelog/unreleased/bugfix-shares-tree-loading
```diff
@@ -0,0 +1,12 @@
...
+We've improved loading of the shares tree:
+
+* It now happens more globally in the sidebar component instead of in each sidebar panel.
+* Shares won't be loaded for resources without a path anymore.
+
+These changes massively improve the sidebar performance and fix several issues with (re-)share permissions.
+
+https://github.com/owncloud/web/issues/7506
+https://github.com/owncloud/web/issues/7593
+https://github.com/owncloud/web/pull/7580
```
kulmann, line 12: I added https://github.com/owncloud/web/issues/7592 to the PR description. Could you add it to the changelog as well?

packages/web-app-files/src/components/SideBar/SideBar.vue
```diff
@@ -53,6 +53,8 @@ import {
 } from 'web-pkg/src/composables'
 import { bus } from 'web-pkg/src/instance'
 import { SideBarEventTopics } from '../../composables/sideBar'
+import _ from 'lodash'
```
kulmann, line 56: Please only import lodash functions via `lodash-es/<functionName>`

packages/web-app-files/src/components/SideBar/SideBar.vue
```diff
@@ -197,6 +201,10 @@ export default defineComponent({
         return
       }
 
+      if (oldFile && _.isEqual(newFile, oldFile)) {
```
kulmann, line 204: Please import `lodash-es/isEqual` instead (see comment on import)

packages/web-app-files/src/composables/parentShare/useIncomingParentShare.ts
kulmann: 👍

kulmann (changes requested):
One last tiny nitpick (about lodash imports, see comment). Otherwise superb, love it!

kulmann (approved):
LGTM now 🚀

PR #7596, opened 2022-09-07
[full-ci]  Performance improvements

diocas:
Attempt to improve the performance number 2. As opposed to the previous PR (done a couple of months ago), this one touches less things (doesn't improve as much, but also doesn't make assumptions about execution dependencies that might change in the future).

But again, the objective here is to parallelise calls to the server... Currently, web does a lot of sequential calls. When these take long to reply (the case in CERNBox, maybe not as visible in a clean and small ocis deployment), these add up.

I've made the OCS calls for the sharing indicators in parallel with the PROPFIND request. This might be good enough for us while we wait for the async/vue3 table updates (even in that case, I still believe it makes sense to do all of them in parallel...).
The change here is that I do not wait for the PROPFIND request to retrieve a resource path, since this might've been passed as parameter.
~~This change is only in the legacy loader.~~

It would be beneficial to do some research to see what other places would benefit for parallelisation.

### DIFF
--- a/packages/web-runtime/src/index.ts
+++ b/packages/web-runtime/src/index.ts
@@ -47,7 +47,7 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> => 
   const store = await announceStore({ runtimeConfiguration })
   announcePermissionManager({ app, store })
 
-  await initializeApplications({
+  const applicationsPromise = await initializeApplications({
     runtimeConfiguration,
     configurationManager,
     store,
@@ -55,13 +55,14 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> => 
     router,
     translations
   })
+  const themePromise = announceTheme({ store, app, designSystem, runtimeConfiguration })
+  await Promise.all([applicationsPromise, themePromise])
 
   announceTranslations({ app, availableLanguages: supportedLanguages, translations })
   announceClientService({ app, runtimeConfiguration })
   announceUppyService({ app })
   await announceClient(runtimeConfiguration)
   await announceAuthService({ app, configurationManager, store, router })
-  await announceTheme({ store, app, designSystem, runtimeConfiguration })
   announceDefaults({ store, router })
 
   app.use(router)
--- a/packages/web-runtime/src/services/auth/userManager.ts
+++ b/packages/web-runtime/src/services/auth/userManager.ts
@@ -157,20 +157,25 @@ export class UserManager extends OidcUserManager {
   }
 
   private async fetchUserInfo(accessToken): Promise<void> {
-    const login = await this.clientService.owncloudSdk.getCurrentUser()
-    await this.fetchCapabilities({ accessToken })
+    const [login] = await Promise.all([
+      this.clientService.owncloudSdk.getCurrentUser(),
+      this.fetchCapabilities({ accessToken })
+    ])
 
     let role, graphUser, userGroups
 
-    const user = await this.clientService.owncloudSdk.users.getUser(login.id)
+    // start the network request here and let it run in parallel to the next calls,
+    // until awaiting the promise execution further below
+    const userPromise = this.clientService.owncloudSdk.users.getUser(login.id)
 
     if (this.store.getters.capabilities.spaces?.enabled) {
       const graphClient = this.clientService.graphAuthenticated(
         this.configurationManager.serverUrl,
         accessToken
       )
-      graphUser = await graphClient.users.getMe()
-      const [roles, settings] = await Promise.all([
+      let roles, settings
+      ;[graphUser, roles, settings] = await Promise.all([
+        graphClient.users.getMe(),
         this.fetchRoles({ accessToken }),
         this.fetchSettings()
       ])
@@ -181,6 +186,9 @@ export class UserManager extends OidcUserManager {
     } else {
       userGroups = await this.clientService.owncloudSdk.users.getUserGroups(login.id)
     }
+
+    const user = await userPromise
+
     this.store.commit('SET_USER', {
       id: login.id,
       uuid: graphUser?.data?.id || '',


ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/28138/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/28138/tracing/participant-management-alice-2022-9-7-10-53-52.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/28138/tracing/participant-management-brian-2022-9-7-10-53-57.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/28138/tracing/public-link-for-space-alice-2022-9-7-10-56-14.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/28138/tracing/unstructured-collection-of-testable-space-interactions-alice-2022-9-7-10-55-05.zip 


</p></details>

packages/web-runtime/src/index.ts
```diff
@@ -32,20 +32,21 @@ export const bootstrap = async (configurationPath: string): Promise<void> => {
...
     store,
     supportedLanguages,
     router,
     translations
   })
+  const theme = announceTheme({ store, vue: Vue, designSystem, runtimeConfiguration })
   announceClientService({ vue: Vue, runtimeConfiguration })
+  await announceClient(runtimeConfiguration)
+  const authService = announceAuthService({ vue: Vue, configurationManager, store, router }) // Requires client
+  await Promise.allSettled([applications, authService, theme])
```
kulmann, line 46: Please use `Promise.all`, as `allSettled` would swallow any errors. But we want to show the error page If any of the three promises fails, so needs to be `Promise.all`.

kulmann (commented):
Hey, parallelizing requests looks good in this PR. However, there are some bigger changes going on in https://github.com/owncloud/web/pull/7430 - would it be ok if we let this PR sit on hold for a few more days and then I bring it over the finish line when #7430 is merged? I guess you have the changes in your fork anyway, right?

diocas:
Ok, no problem. I'll wait and rebase + do the changes once that one is done.

diocas:
@kulmann Done, rebased and fixed the issue you pointed. But, ofc, things changed slightly from the first iteration.

packages/web-app-files/src/services/folder/loaderSpace.ts
```diff
@@ -53,9 +62,17 @@ export class FolderLoaderSpace implements FolderLoader {
       try {
         store.commit('Files/CLEAR_CURRENT_FILES_LIST')
 
-        const resources = yield webdav.listFiles(space, { path, fileId })
-        let currentFolder = resources.shift()
-        replaceInvalidFileRoute({ space, resource: currentFolder, path, fileId })
+        const resourcesPromise = webdav.listFiles(space, { path, fileId })
```
kulmann, line 65: This doesn't work - the `listFiles` call has an internal path correction mechanism. Works as follows:
1. you reach a URL with path and fileId (please note: CERNbox is not supposed to use the fileIds in URLs. We build a disable switch just for you, see documentation for `options.routing.idBased` in https://owncloud.dev/clients/web/getting-started/#options
2. both the path and the fileId get passed into `listFiles` here
3. the request is done by path (webdav works path based, so that's only natural)
4. the fileId from the webdav response of (3) is compared with the fileId provided in the params
5. if the fileIds don't match, we request "the new path" by fileId, using the `meta` endpoint. This can happen if you open a bookmarked URL and the relative item path has changed since you bookmarked the URL (e.g. a folder has been renamed or moved to a different location)
6. the `listFiles` request happens again, but with the corrected path

IMPORTANT: if the fileId changed (due to e.g. rename) it is possible that a new file/folder has been created in the old, bookmarked location / path. You would request shares for a completely different file/folder in that case.

This is the reason why we have to await the propfind before we request the shares. Not possible to parallelize this!

I see one option though: we could
1. await/yield a `getFileInfo` request, which is a propfind with depth 0 internally. This would solve our path correction issue
2. we use your approach to parallelize a `listFiles` (= propfind with depth 1) and the OCS getShares call.

I can't tell if it's worth the effort. My assumption is, that a propfind with depth 0 is reasonably faster than a propfind with depth 1 for CERNbox, but you need to tell me if that assumption is valid.
If you want to I can provide a pull request solving the stuff above - or if your time permits you could solve this as well of course.
In any case I'd be very happy to see (and merge) a pull request with your other changes!! Really good that you started parallelizing requests. 👍

kulmann (changes requested):

packages/web-app-files/src/services/folder/loaderSpace.ts
diocas: What if I use the switch `routing.idBased` in here? If this is false, then we can safely assume that the provided id is the expected one, right? Or at least more or less (see last paragraph)...

But actually, I think that this happens already (did this some time ago so I'm not sure anymore).. Basically, if I land on a page the `fileId` will be undefined (because we have `idBased = false`) so the url doesn't contain it. So we will retrieve it just like the previous behaviour, i.e it will be the first thing we do. The id is only defined if this method is reached having before PROPFINDING the current folder. 
But, ofc, if `idBased = true`, we will have always an `id` to pass to this method (unless we distinguish between an id given by the route or by the current folder info). So, if we separated the id the was provided by the route from the one provided by the current folder PROPFIND, we would be able to force the re-check only in the first case.

The only issue I see is if something changes in the background (some other open tab, for example), so our current folder info is no longer up to date. Not sure we would want to accept this corner case inconsistency.

diocas: @kulmann ^

diocas: Just checked, and since we eventually `replaceInvalidFileRoute` and redirect the user to the new path, what happens is that what is presented is the correct information. The worst that happens is that there is an extra call to the shares for the wrong folder that is not the current one.
isn't it better to do a bad, unnecessary, request on the (exceptional) possibility that a path has changed, if by doing so we are improving the more common case that the path is correct?

diocas:
With the removal of the OCS calls, this PR became very small... But maybe still worth to merge, @kulmann ?

kulmann (approved):
Nice, thank you 👍

Issue #7628, opened 2022-09-09
Upload via drag and drop doesn't preserve time of file


michaelstingl:
### Steps to reproduce

1. Upload old file via drag and drop to **demo.owncloud.com** ==> "Modified" column shows **original time** from file ✅ 
2. Upload old file via drag and drop to **web.owncloud.com** ==> "Modified" column shows **upload tim**e from file ⚠️ 
3. Upload old file via drag and drop to **ocis.owncloud.com** ==> "Modified" column shows **upload time** from file ⚠️ 

### Expected behaviour

Upload via ownCloud web should preserve the file time, same as oC10 does.

### Actual behaviour

![CleanShot 2022-09-09 at 14 44 40@2x](https://user-images.githubusercontent.com/214010/189353133-9dc3a6f2-2044-4064-865c-f1f7f454a8e4.png) | ![CleanShot 2022-09-09 at 14 44 46@2x](https://user-images.githubusercontent.com/214010/189353151-94e15fbb-5f58-4fee-a04b-9e1610f0615d.png) | ![CleanShot 2022-09-09 at 14 44 56@2x](https://user-images.githubusercontent.com/214010/189353157-1a86173c-76dc-45cb-8ba7-6aadea9b4933.png)
-- | -- | --
demo.owncloud.com | web.owncloud.com | ocis.owncloud.com

### Client configuration
**Browser:**
Safari 15.6.1 (17613.3.9.1.16)

**Operating system:**
macOS 12.5.1 (21G83) Monterey

JammingBen:
This has been fixed for Web, however, there is still a bug in the backend tus implementation. See https://github.com/owncloud/ocis/issues/4559

PR #7630, opened 2022-09-12
Include `x-oc-mtime` header in upload requests

JammingBen:
## Description
We've included the `x-oc-mtime` header in upload requests to tell the backend the proper modify date of uploaded resources.

Fyi: This is not yet being supported for tus by ocis (https://github.com/owncloud/ocis/issues/4559)

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7628

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/composables/upload/useUpload.ts
+++ b/packages/web-runtime/src/composables/upload/useUpload.ts
@@ -10,7 +10,7 @@ import {
   useStore
 } from 'web-pkg/src/composables'
 import { computed, Ref, unref, watch } from '@vue/composition-api'
-import { UppyService } from '../../services/uppyService'
+import { UppyService, uppyHeaders } from '../../services/uppyService'
 import * as uuid from 'uuid'
 
 export interface UppyResource {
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/composables/upload/useUpload.ts
```diff
@@ -79,7 +79,18 @@ export function useUpload(options: UploadOptions): UploadResult {
...
       },
-      headers: () => unref(headers),
+      headers: (file) => {
+        const reqHeaders = {
+          'x-oc-mtime': file.data.lastModified / 1000,
+          ...unref(headers)
+        } as uppyHeaders
+
+        if (unref(isTusSupported)) {
+          // Tus includes auth headers before each request
```
kulmann, line 89: wouldn't the TUS handling just overwrite the auth headers then? In that case I'd prefer not manipulating them here to keep the code clean(er).

JammingBen: > wouldn't the TUS handling just overwrite the auth headers then?

Unfortunately not. The headers here overwrite the modifications in `beforeEachRequest()`.

kulmann: ah, ok :-)

kulmann (approved):
LGTM 👍

Issue #1753, opened 2019-08-09
restoring a file from "Deleted files" (trashbin) is not possible if the original folder does not exist any-more


individual-it:
### Steps to reproduce
1. create a file in a folder
2. delete the file
3. rename the folder
4. try to restore the file

### Expected behaviour
file should be restored

### Actual behaviour
file is not restored. Error: ` Restoration of file failed The destination node is not found`
The user has no idea from where the file came from.  see #1525
So its very hard to restore the file

PVince81:
I'd expect the trashbin API to know where the original location is as it's stored in the database.

The behavior in OC 10 was that if the target folder does not exist, we fall back to putting the restored item into the user's home root.

PVince81:
Maybe in Phoenix we can detect this situation, this is likely a 409 if the folder does not exist, so after that fall back to a MOVE to the root

PVince81:
in case we rely on the error code, need to make sure it's not happening also if the target destination file already exists like in https://github.com/owncloud/phoenix/issues/1730

individual-it:
assigned @jasson99 and put into sprint for creating tests

HariBhandari07:
@individual-it since tests have been added should we mark this issue as done and remove from current sprint?

individual-it:
@HariBhandari07 the issue is not fixed, only tests were added, so I took off @jasson99 assignment, but it should stay on the sprint

pascalwengerter:
Can confirm this still happens on https://ocis.ocis-web.latest.owncloud.works

kulmann:
The trash (at least for oc10) has no parent file id, only an original location of the deleted file. We have no way of finding out where to restore the file/folder to if the original location doesn't exist anymore.

- For a renamed original location, I'd expect restoring to the now renamed original location. We cannot do that as the information doesn't exist.
- For a deleted original location, I'd expect restoring to fail with a fallback to be chosen by the user.

@tbsbdr what do you think about only informing the user that the original location doesn't exist anymore and then
- a) restore to the users personal home (could require a confirmation upfront) as described by @PVince81 in https://github.com/owncloud/web/issues/1753#issuecomment-535862744
- b) utilize the `LocationPicker` that we also have for copy and move operations for letting the user choose an alternative location.

If we choose option b, then we could also introduce an additional action `restore to` where the user can immediately choose a different restore location upfront.

tbsbdr:
[...]

PR #7635, opened 2022-09-12
[full-ci] Restore files: Implement conflict dialog

lookacat:
## Description
See https://github.com/owncloud/web/issues/1753

## Related Issue
- Fixes https://github.com/owncloud/web/issues/1753
- Fixes https://github.com/owncloud/enterprise/issues/5325

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -47,7 +57,89 @@ export default {
...
+      const resolvedResources = []
+      for (const resource of resources) {
+        const webDavParentPath = this.getWebdavParentFolderFromResource(resource)
+
+        // ? check if parent folder has already been requested
+        let parentResources = []
+        if (webDavParentPath in parentFolders) {
+          parentResources = parentFolders[webDavParentPath]
+        } else {
+          const listResponse = await this.$client.files.list(
```
JammingBen, line 80: This returns the root directory as well. I think you need to respect that in your check down below: `parentResources.some((e) => e.name === resource.name)`. Otherwise you could not restore a folder `aaa` inside folder `aaa`.

Not 100% sure though as I can't try it myself.

JammingBen (commented):
I'd like to try it, but I always get an error on restoring files: `TypeError: Cannot read properties of undefined (reading 'includes')` in this line: `const overwrite = filesToOverwrite.includes(resource)`.

lookacat:
sorry that it didnt worked, im confused too as it did work before :O will look into it

packages/web-app-files/src/mixins/actions/restore.ts
lookacat: yes you are right indeed

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -47,7 +57,88 @@ export default {
...
+            1,
+            DavProperties.Default
+          )
+          parentResources = listResponse.map((i) => buildResource(i))
+          parentFolders[webDavParentPath] = parentResources
+        }
+        // ? Check for naming conflict in parent folder and between resources batch
+        const hasConflict =
+          parentResources.some((e) => e.name === resource.name) ||
+          resources.filter((e) => e.id !== resource.id).some((e) => e.name === resource.name)
```
JammingBen, line 90: I don't quite get this check. Imagine I have the following 2 files:

* `/someFile`
* `/folder/someFile`

I delete both `someFile`, now I want to restore both at the same time -> conflict dialog shows. Isn't the parent path check enough?

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -43,7 +53,129 @@ export default {
...
+      for (const resource of resources) {
+        const webDavParentPath = this.getWebdavParentFolderFromResource(resource)
+
+        // ? check if parent folder has already been requested
+        let parentResources = []
+        if (webDavParentPath in parentFolders) {
+          parentResources = parentFolders[webDavParentPath]
+        } else {
+          try {
+            const listResponse = await this.$client.files.list(
```
dschmidt, line 76: Could you please use the webdav client from clientService for new code?

lookacat: yes

JammingBen:
The `Replace` strategy does not work for me. The file disappears in the trashbin after clicking "Replace", but it re-appears after a page reload. Also, no new version gets created.

https://github.com/owncloud/web/issues/1753 is not fixed by this, right? It still gives me an error.

lookacat:
@JammingBen thats weird, i will look into it. But yes the bug ticket will be fixed by this PR.

tests/acceptance/features/webUITrashbinRestore/trashbinRestore.feature
```diff
@@ -134,73 +134,73 @@ Feature: Restore deleted files/folders
...
+
+
+
+
+
+
+
+
+
+
```
dschmidt, line 203: Why all these empty lines?

packages/web-app-files/tests/unit/mixins/actions/restore.spec.js
```diff
@@ -79,8 +102,21 @@ function getWrapper({ invalidLocation = false, resolveClearTrashBin: resolveRest
...
       $gettext: jest.fn(),
       $gettextInterpolate: jest.fn(),
+      createModal: jest.fn(),
+      $clientService: {
+        webdav: {
+          listFiles: jest.fn().mockImplementation(() => {
+            return []
+          })
+        }
+      },
```
dschmidt, line 112: I understand if you don't want to port everything to the defaultComponentMocks, but you could mock the whole clientService like this:

https://github.com/owncloud/web/blob/3808ee02ee77549bb6133151f4fe15ce80fd2d95/tests/unit/mocks/defaultComponentMocks.ts#L20

packages/web-app-files/tests/unit/mixins/actions/restore.spec.js
```diff
@@ -79,8 +102,21 @@ function getWrapper({ invalidLocation = false, resolveClearTrashBin: resolveRest
...
       $client: {
         ...sdkMock,
+        files: {
+          list: jest.fn().mockImplementation(() => {
+            return Promise.resolve([])
+          })
+        },
```
dschmidt, line 119: This is probably not needed anymore after you ported your code to `webdav.listFiles`

dschmidt (commented):
Just a few more or less nitpicky comments

tests/acceptance/features/webUITrashbinRestore/trashbinRestore.feature
kulmann: You never just delete lines in feature files if you can avoid it. Replacing with empty lines is the way to go, because this way the line numbers of other test scenarios are not modified. Why is this good? Because we have `expected to fail`-files which define expected failures by line numbers.

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -120,6 +247,42 @@ export default {
         const user = await this.$client.users.getUser(this.user.id)
         this.SET_QUOTA(user.quota)
       }
+    },
+    async $_restore_trigger({ resources }) {
+      //! collect and request existing files in associated parent folders of each resource
```
kulmann, line 252: Why the `//!`? Just curious, as I don't know this in JS/TS context... would prefer `//` without the `!` if this is not some convention I simply don't know about. Please enlighten me. :-) (Same for the `//?` above)

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -43,7 +51,126 @@ export default {
     ...mapMutations('runtime/spaces', ['UPDATE_SPACE_FIELD']),
     ...mapMutations(['SET_QUOTA']),
 
-    async $_restore_trigger({ resources }) {
+    getParentFolderFromResource(resource) {
+      return resource.path.slice(0, resource.path.lastIndexOf(resource.name))
```
kulmann, line 55: This is the same is `dirname`, isn't it? See https://nodejs.org/docs/latest/api/path.html#pathdirnamepath
You can import it via `import { dirname } from 'path'`

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -43,7 +51,126 @@ export default {
...
+    },
+    getWebdavParentFolderFromResource(resource) {
+      const parentPath = this.getParentFolderFromResource(resource)
+      return buildWebDavFilesPath(this.user.id, parentPath)
```
kulmann, line 59: Erm... why the "old" webdav v2 in a hardcoded way? All our space objects from the runtime are capable of giving you a webDavPath (if needed at all), so that you don't need to know how to build a webDav path. This can't work for any space trashbin, only for personal spaces.

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -43,7 +51,126 @@ export default {
...
+    },
+    async collectRestoreConflicts(resources) {
+      const parentFolders = {}
+      const conflicts = []
+      const resolvedResources = []
+      for (const resource of resources) {
+        const webDavParentPath = this.getParentFolderFromResource(resource)
```
kulmann, line 66: Name is confusing, this is not a webDav path but just simply a relative path.

packages/web-app-files/src/mixins/actions/restore.ts
kulmann: I'd vote for removing the function and just using `dirname(resource.path)` in all the locations where you used this function before.

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -56,11 +183,12 @@ export default {
         const restorePath = hasShareJail
           ? buildWebDavSpacesPath(this.space.id, resource.path)
           : buildWebDavFilesPath(this.user.id, resource.path)
+        const overwrite = filesToOverwrite.includes(resource)
 
         restorePromises.push(
           restoreQueue.add(async () => {
             try {
-              await this.$client.fileTrash.restore(path, resource.id, restorePath)
+              await this.$client.fileTrash.restore(path, resource.id, restorePath, overwrite)
```
kulmann, line 191: Let's go the extra mile and introduce `restoreDeletedFile` (or similar wording) in `web-client`. We can do that together.

kulmann (changes requested):

packages/web-app-files/tests/unit/mixins/actions/restore.spec.js
lookacat: doesnt work 😬

packages/web-app-files/src/mixins/actions/restore.ts
lookacat: done

packages/web-app-files/tests/unit/mixins/actions/restore.spec.js
kulmann: I doubt that it just `doesn't work` 👀  any specific reason why it `doesn't work`?

kulmann (approved):

Issue #7655, opened 2022-09-19
Indirect shares can't be loaded for nested folders in spaces


JammingBen:
### Steps to reproduce
1. Create the following folder structure in a space: `level1/level2/somefile.txt`
2. Share `level1` with some user
3. Navigate into `level2` -> Everything is fine: `somefile.txt` shows a share indicator, the sidebar reveals the indirect share
4. Reload the page -> Share indicator is missing, console shows errors when loading the shares tree.

As described in https://github.com/owncloud/web/pull/7638#discussion_r970761149: This happens because we only pass the `storageId` of the current resource to `loadSharesTree`. However, we need the `storageId` of each parent resource.

Edit: The same error exists in the share jail.

PR #7638, opened 2022-09-13
[full-ci] Optimize share loading part II

JammingBen:
## Description
Continuation of https://github.com/owncloud/web/pull/7580

* Remove `loadCurrentFileOutgoingShares` and `loadIncomingShares` mutations. Instead, incoming and outgoing shares are now being loaded via `loadSharesTree()`. This avoids `getShare()` requests from being executed multiple times.
* Decouple shares member loading from shares loading in store. This reduces the fetching of space members to a minimum and improves the structure of the code.
* Fix reactive loading of share indicators in sidebar details panel.
* Fix reactive loading of space member count in the spaces overview.
* Fix loading of indirect shares within spaces.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

PR #7656, opened 2022-09-19
Disable shares tree cache for flat file lists

JammingBen:
## Description
Shares tree caching is a problem in flat file lists that gather resources from various places (like "Shared with others", "Shared via link", ...). This is an issue because the caching currently happens purely based on the filename. However, these lists can contain multiple resources with the same name.

In the future, we should think about something else that also includes the storage ID. This is necessary anyway because of https://github.com/owncloud/web/issues/7655.

However, we don't have any performance impact by disabling the cache here as we can't navigate into folders anyway.

## Motivation and Context
Followup for https://github.com/owncloud/web/pull/7638

## Types of changes

- [ ] Technical debt
- [ ] Tests


kulmann (approved):
Very good fix, nice! 💪

Issue #7666, opened 2022-09-21
webUISharingFilePermissionMultipleUsers/shareFileWithMultipleUsers.feature:51 in OC10 failed in nightly and needs further investigation 


amrita-shrestha:
Build link: https://drone.owncloud.com/owncloud/web/28470/24/19

Failed Scenario:
```
 Scenario: share a file with multiple users with different roles and permissions # features/webUISharingFilePermissionMultipleUsers/shareFileWithMultipleUsers.feature:51
   ✔ Before # setup.js:38
   ✔ Before # setup.js:42
   ✔ Before # setup.js:53
   ✔ Before # setup.js:60
   ✔ Before # setup.js:71
   ✔ Before # stepDefinitions/filesContext.js:17
   ✔ Before # stepDefinitions/generalContext.js:206
   ✔ Before # stepDefinitions/generalContext.js:244
   ✔ Before # stepDefinitions/middlewareContext.js:46
   ✔ Given the setting "shareapi_auto_accept_share" of app "core" has been set to "no" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And the administrator has set the default folder for received shares to "Shares" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And these users have been created with default attributes and without skeleton files in the server: # stepDefinitions/middlewareContext.js:76
       | username |
       | Alice    |
       | Brian    |
       | Carol    |
   ✔ Given user "Alice" has created file "lorem.txt" in the server # stepDefinitions/middlewareContext.js:66
   ✔ And user "Alice" has logged in using the webUI # stepDefinitions/loginContext.js:68
   ✔ When the user opens the share dialog for file "lorem.txt" using the webUI # stepDefinitions/sharingContext.js:588
   ✔ And the user selects the following collaborators for the share as "Custom permissions" with "update" permissions: # stepDefinitions/sharingContext.js:371
       | collaborator | type |
       | Brian Murphy | user |
       | Carol King   | user |
   ✔ And the user shares with the selected collaborators # stepDefinitions/sharingContext.js:420
   ✔ And user "Brian" accepts the share "Shares/lorem.txt" offered by user "Alice" using the sharing API in the server # stepDefinitions/middlewareContext.js:66
   ✔ And user "Carol" accepts the share "Shares/lorem.txt" offered by user "Alice" using the sharing API in the server # stepDefinitions/middlewareContext.js:66
[...]

PR #7671, opened 2022-09-22
[full-ci][tests-only] Wait more for element collaboratorAccessDetailsDrop not to be present

amrita-shrestha:
## Description

From investigation, it was found that sometimes the `collaboratorAccessDetailsDrop` doesn't resolve up to the time limit 300ms used by this
``` 
await this.waitForElementNotPresent(
          '@collaboratorAccessDetailsDrop',
          this.api.globals.waitForNegativeConditionTimeout
        )
```
that result `.share-access-details-drop` element to be visible. So replace  waitForNegativeConditionTimeout with waitForConditionTimeout.
For detail visit https://github.com/owncloud/web/pull/7671#discussion_r993341501

## Related Issue

- Fixes [webUISharingFilePermissionMultipleUsers/shareFileWithMultipleUsers.feature:51 in OC10 failed in nightly and needs further investigation](https://github.com/owncloud/web/issues/7666)

## How Has This Been Tested?

- locally.

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
+++ b/tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
@@ -199,7 +199,7 @@ module.exports = {
         await this.api.elementIdClick(collaboratorAvatarElId)
         await this.waitForElementNotPresent(
           '@collaboratorAccessDetailsDrop',
-          this.api.globals.waitForNegativeConditionTimeout
+          this.api.globals.waitForConditionTimeout
         )
       }
 


tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
```diff
@@ -196,7 +196,7 @@ module.exports = {
           }
         )
         // to hide the access details dialog from the screen, click on the collaborator avatar
-        await this.api.elementIdClick(collaboratorAvatarElId)
+        this.api.elementIdClick(collaboratorAvatarElId)
```
saw-jan, line 199:
```suggestion
        await this.api.elementIdClick(collaboratorAvatarElId)
```

saw-jan (changes requested):
I am not sure about this.
I strongly suggest to use `await` for nightwatch commands

tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
saw-jan: Could elaborate what happens and why removing `await` works

tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
```diff
@@ -196,7 +196,7 @@ module.exports = {
           }
         )
         // to hide the access details dialog from the screen, click on the collaborator avatar
-        await this.api.elementIdClick(collaboratorAvatarElId)
+        await this.api.elementIdClick(collaboratorAvatarElId).pause(500)
```
saw-jan, line 199: :cry: the `waitForElementNotPresent` is exactly for such waits.
```suggestion
        await this.api.elementIdClick(collaboratorAvatarElId)
```

amrita-shrestha: yes, that's why I have to use explicit wait. Time waiting for **waitForElementNotPresent** is not enough so it fails intermittently.

SagarGi: if this function `waitForElementNotPresent` did not wait enough you might need to increase the timeout for that function.
looks like the function `waitForElementNotPresent` below has timeout `waitForNegativeConditionTimeout` which is `300 ms` as per the config. https://github.com/owncloud/web/blob/a4bfc0090776ad384148259d96fedbe5858607b4/tests/acceptance/nightwatch.conf.js#L47-L48


You can use `waitForConditionTimeout` instead of `waitForNegativeConditionTimeout` for function `waitForElementNotPresent` below.(No need to use explicit pause)

And check if it affects other tests runtime. CC @saw-jan ?

saw-jan: > Regarding increasing waitForNegativeConditionTimeout can we afford increasing time for this variable 

I am afraid we should increase the time for `waitForNegativeConditionTimeout`.

>That's a nice fix but what is the use of naming two variables waitForNegativeConditionTimeout and waitForConditionTimeout if we are going to use it in a zigzag pattern like using waitForConditionTimeout in a negative condition like waitForElementNotPresent?

The purpose of `waitForNegativeConditionTimeout` is not to use in commands like `waitForElementNotPresent` or `waitForElementNotVisible` (You got confused here).

The main purpose is to use `waitForNegativeConditionTimeout` in commands like `isVisible` (to determine the visibility) if we expect that the element should not be visible (**negative**).

amrita-shrestha: I found that people are using these two variables according to the time needed. So i think it ok to use

saw-jan: For example here: https://github.com/owncloud/web/blob/c73b8d03264e4a0e1456a512831eba008ac9579a/tests/acceptance/pageObjects/personalPage.js#L301-L317

tests/acceptance/pageObjects/FilesPageElement/SharingDialog/collaboratorsDialog.js
```diff
@@ -199,7 +199,7 @@ module.exports = {
         await this.api.elementIdClick(collaboratorAvatarElId)
         await this.waitForElementNotPresent(
           '@collaboratorAccessDetailsDrop',
-          this.api.globals.waitForNegativeConditionTimeout
+          this.api.globals.waitForConditionTimeout
```
saw-jan, line 202: Indeed, sometimes the element takes more than `300ms` (`waitForNegativeConditionTimeout`) to be not present.
Screenshot from 2022-10-12 17-08-51

LGTM :+1:

saw-jan: @amrita-shrestha, can PR title be more meaningful? (related to what your changes are)

saw-jan (commented):
Looks good :+1:

individual-it (approved):

saw-jan (approved):

Issue #7617, opened 2022-09-09
Search icon is misplaced on small screens


tbsbdr:
### Problem description

Search icon looks like this on viewports smaller than 960 px:  

![Screenshot 000494@2x](https://user-images.githubusercontent.com/26610733/189308105-6bed0a4c-db30-4add-a06a-199928670795.png)

### Current and expected behaviour

![image](https://user-images.githubusercontent.com/26610733/189614688-17f4cadb-e1e2-439c-80d0-ed1c564a09e1.png)

![image](https://user-images.githubusercontent.com/26610733/189615972-86c59a78-db2c-419b-b75c-9337ef484a0b.png)

![image](https://user-images.githubusercontent.com/26610733/189616029-0c5d5ba8-9ccb-4598-9af6-5ef60aba4cdd.png)

*   [x] remove the label "dark/light mode" so that we can place additional icons there; it would also balance the left and right areas cc @JammingBen

PR #7675, opened 2022-09-22
Fix search bar display on small screens

JammingBen:
## Description
We've fixed the display of the search bar on small screens.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7617
- 
## Screenshots:

See issue.

## Types of changes

- [ ] Technical debt
- [ ] Tests


ownclouders:
Results for <strong>oC10iPhone1</strong> https://drone.owncloud.com/owncloud/web/28514/49/1
:boom: The oC10iPhone1 tests pipeline failed. The build has been cancelled.

JammingBen:
@individual-it One acceptance test keeps failing here: https://drone.owncloud.com/owncloud/web/28569/57/14. However, on my local machine, the test succeeds. I have no idea what's the issue here... Maybe your team could have a look if time permits?

amrita-shrestha:
> @individual-it One acceptance test keeps failing here: https://drone.owncloud.com/owncloud/web/28569/57/14. However, on my local machine, the test succeeds. I have no idea what's the issue here... Maybe your team could have a look if time permits?

@JammingBen, I was working separately to solve acceptance failure in https://github.com/owncloud/web/pull/7687 PR. But mistakenly I pushed my commit to this PR in place of https://github.com/owncloud/web/pull/7687 PR. Sorry for inconvience. You can drop my commit

JammingBen:
@amrita-shrestha 

> I found something odd in the search feature. In the tests run, I noticed that after renaming resources or uploading resources Personal page contained renamed/uploaded resources. But if we search currently renamed/uploaded, then renamed/uploaded resources can't be found.

That's strange though, as this PR only touches visuals. Also, the failing test suite is for iPhone only (meaning on small screens).

Are you sure that the search bar is being accessed the right way during the test? This PR changed the behavior on mobile devices so the search icon has to be clicked first, then the search input appears. Maybe the time between the click and the appearing search input is too small and we need to wait for the search input to be fully visible and usable?

amrita-shrestha:
@JammingBen **the indexing of uploaded files for search done async** could be the culprit behind `webUIFilesSearch/search.feature:125 scenario ` failing. I will look for a way to verify upload has been completed before the search

individual-it:
I believe that it worked before because the item `'#files-open-search-btn'` did not exist in the DOM, so the test would look for it till an timeout was reached and that gave the backend enough time to index the files.
Because @JammingBen put a valid selector for `openSearchButton` the step runs faster and sometimes there is not enough time to index the file

JammingBen:
@individual-it I added a pause statement as discussed in the daily yesterday: https://github.com/owncloud/web/pull/7675/commits/f77d04015c66770718637f4668ce293166cb6874

Is that okay from your side (at least for now)?

individual-it:
@JammingBen yes lets do it that way. I cannot think of any other way to do it :-(

packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -281,18 +309,43 @@ export default {
...
+      document.getElementsByClassName('oc-search-input')[0].focus()
+      this.showCancelButton = true
+    },
+    cancelSearch() {
+      document.getElementById('files-global-search-bar').style.visibility = 'hidden'
+      this.showCancelButton = false
+    },
+    onResize() {
+      const searchBarEl = document.getElementById('files-global-search-bar')
+      if (window.innerWidth > 639) {
```
lookacat, line 324: shouldnt this be possible via css? I mean its fine with me just weird to see this number in code rather than css

JammingBen: We need it unfortunately to set the `visible` state back again if you search on a small screen and resize your screen to full again. Without, the search would stay hidden because `showSearchBar()` actively sets `style.visibility = 'visible'`.

JanAckermann (approved):

PR #7689, opened 2022-09-26
add option to inject css and js files

fschade:
This PR introduces the ability to configure scripts and style resources that will be injected into the SPA application.
This enables for example injection of small self containing scripts or style adjustments without rebuilding web or build a custom application.

needs https://github.com/owncloud/ocis/pull/4664


JanAckermann (changes requested):
I don't think we need an async function here

JanAckermann (approved):

Issue #7713, opened 2022-09-29
Resource name in sidebar does not update reactively


JammingBen:
### Steps to reproduce
1. Open the sidebar for a resource
2. Rename the resource in the files list
3. Check the name in the sidebar

### Expected behaviour
The name should be updated.

### Actual behaviour
The name is not being updated.

PR #7734, opened 2022-10-04
Fix file name reactivity in sidebar

JammingBen:
## Description
We've fixed a bug where the file name would not update reactively in the sidebar after changing it.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7713

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/SideBar/FileInfo.vue
+++ b/packages/web-app-files/src/components/SideBar/FileInfo.vue
@@ -34,7 +34,6 @@ export default {
   components: {
     PrivateLinkItem
   },
-  inject: ['displayedItem'],
   props: {
     isSubPanelActive: {
       type: Boolean,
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/SideBar/FileInfo.vue
```diff
@@ -34,7 +34,6 @@ export default {
   components: {
     PrivateLinkItem
   },
-  inject: ['displayedItem'],
```
kulmann, line 37: What exactly is/was the issue? Is the displayedItem not updating properly, or is the usage within this component not reactive?

JammingBen: Because `displayedItem` is only a copy of the highlighted file: https://github.com/owncloud/web/blob/c8f8589bf6d66353dd004807872ab3cfb66c7a80/packages/web-app-files/src/components/SideBar/SideBar.vue#L244

kulmann (approved):

Issue #7347, opened 2022-07-26
Shared via link > "parent folder" shows "Resource not found" on click


tbsbdr:
### Steps to reproduce
1. receive a share
2. go to "shares" > "shared with me" > create a link for the with-you-shared file
3. go to "shares" > "shared via link" > click on the parent folder of the shared resource
4. View shows "Resource not found"

https://user-images.githubusercontent.com/26610733/181061204-6d205bdf-f99c-486a-b65b-cc2e813cad9a.mp4

### Expected behaviour
show "Shares" > "Shared with me" > "highlight resource"

### Actual behaviour
View shows "Resource not found"

JammingBen:
It also breaks when sharing resources within the share via link. This is a similar challenge as in https://github.com/owncloud/web/issues/7043: We somehow need to determine if the resource is a share (that is easy) OR part of a share (that's not easy). Then we need to somehow retrieve the id of the underlying share to build the correct route and to show the proper parent name.

PR #7771, opened 2022-10-12
Fix routing for re-shares

JammingBen:
## Description
We've fixed a bug where routing into re-shares and their parent folders from the "Shared with others/via link" page was broken.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7347

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-routing-for-re-shares
+++ b/changelog/unreleased/bugfix-routing-for-re-shares
@@ -1,0 +1,6 @@
+Bugfix: Routing for re-shares
+
+We've fixed a bug where routing into re-shares and their parent folders from the "Shared with others/via link" page was broken.
+
+https://github.com/owncloud/web/pull/7771
+https://github.com/owncloud/web/issues/7347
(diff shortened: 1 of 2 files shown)


kulmann (commented):
Same solution is needed for the search `Preview` component and for the upload widget, could you apply it there as well?
I have the feeling that we need to clean up this link creation mess throughout the different components...

JammingBen:
> Same solution is needed for the search Preview component and for the upload widget, could you apply it there as well?
I have the feeling that we need to clean up this link creation mess throughout the different components...

Could you give me a hint in which scenarios exactly? In the share jail itself these seems to work..?

kulmann:
> > Same solution is needed for the search Preview component and for the upload widget, could you apply it there as well?
> > I have the feeling that we need to clean up this link creation mess throughout the different components...
> 
> Could you give me a hint in which scenarios exactly? In the share jail itself these seem to work..?

Never mind, brainfart 😅

kulmann (approved):

Issue #7785, opened 2022-10-13
"Open folder"-action on a reshare does not resolve properly


JammingBen:
### Steps to reproduce
1. Create a public link for a received folder share
2. Go to the "Shared via link" page
3. Open the context menu for this resource and click "Open folder"

### Expected behaviour
The link should resolve correctly into the share jail (oCIS).

### Actual behaviour
The link is being resolved incorrectly.

For now, this has to be solved via a private link. See https://github.com/owncloud/web/pull/7771 where this has been done already.

PR #7790, opened 2022-10-13
Fix 'Open folder'-action for re-shares

JammingBen:
## Related Issue
- Fixes https://github.com/owncloud/web/issues/7785

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/mixins/actions/navigate.ts
+++ b/packages/web-app-files/src/mixins/actions/navigate.ts
@@ -9,7 +9,7 @@ import {
 } from '../../router'
 import { ShareStatus } from 'web-client/src/helpers/share'
 import merge from 'lodash-es/merge'
-import { buildShareSpaceResource } from 'web-client/src/helpers'
+import { buildShareSpaceResource, isShareSpaceResource } from 'web-client/src/helpers'
 import { configurationManager } from 'web-pkg/src/configuration'
 import { createFileRouteOptions } from 'web-pkg/src/helpers/router'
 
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/mixins/actions/navigate.ts
```diff
@@ -52,6 +52,15 @@ export default {
           canBeDefault: true,
           componentType: 'router-link',
           route: ({ resources }) => {
+            if (
+              isShareSpaceResource(this.space) &&
```
kulmann, line 56: I think you need to use `this.$_navigate_getSpace` here instead of `this.space`.

JammingBen: `this.space` is actually always set in this scenario, at least on those 2 relevant views.

ownclouders:
Results for <strong>oC10XGAPortrait1</strong> https://drone.owncloud.com/owncloud/web/29083/45/1
:boom: The oC10XGAPortrait1 tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>oC10Resharing</strong> https://drone.owncloud.com/owncloud/web/29083/39/1
:boom: The oC10Resharing tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>oC10SharingExternal</strong> https://drone.owncloud.com/owncloud/web/29083/42/1
:boom: The oC10SharingExternal tests pipeline failed. The build has been cancelled.

packages/web-app-files/src/mixins/actions/navigate.ts
kulmann: But how? The `shared with others` and `shared via link` page don't have one common parent space.

kulmann: But it's not available as `this.space`.  The linked function is used for binding the space into the context menu: https://github.com/owncloud/web/blob/d44c2bfa926b2195fb37540aecb071dd3d7ee7cd/packages/web-app-files/src/views/shares/SharedViaLink.vue#L38

The `$_navigate_getSpace` function resembles the same functionality.

JammingBen: The space is given as property to `ContextActions`, which has all the actions as mixin. Isn't `space` available for the mixin then via `this`? Otherwise I don't now where that comes from 🤔

kulmann: argh. yes, you're right. 🙈

kulmann (approved):

PR #7816, opened 2022-10-17
[full-ci] Make dependencies between modules more explicit and resolve some of them

dschmidt:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-client/src/helpers/resource/functions.ts
```diff
@@ -1,4 +1,11 @@
-import { urlJoin } from 'web-pkg/src/utils'
+import { basename } from 'path'
+import { urlJoin } from '../../utils'
+import { DavPermission, DavProperty } from '../../webdav/constants'
+import { Resource } from './types'
+
+const fileExtensions = {
```
kulmann, line 6: Why did you move `fileExtensions` to `functions` instead of `types`? 🤔

dschmidt: Probably because I moved the functions going along with it here, it's not exported after all.

Have no strong preference. If you prefer it in `types` I can move it there, but it's not really a "type", is it? (I mean technically yes, but nothing you'd use in an interface or something like that)

kulmann: Hmm true, not exported. Then it's fine

kulmann (approved):

dschmidt:
Yay 🤓

Issue #7784, opened 2022-10-13
"Copy" and "Cut" actions don't work for the current folder


JammingBen:
### Steps to reproduce

1.  Create a folder in your personal space
2.  Navigate into the folder
3.  Select an action for the current folder via the breadcrumb action menu: "Copy" or "Cut"

### Expected behaviour

The resource should either be copied/cut, ~or these actions should be hidden.~

### Actual behaviour

The actions can be clicked, but nothing is being copied/cut.

I am not sure about the expected behavior, what do you think @tbsbdr @kulmann ?

tbsbdr:
the resource should be copied / cut.

updated initial post

PR #7830, opened 2022-10-19
Fix Cut and Copy actions for current folder

JammingBen:
## Description
The "Cut" and "Copy" actions for the current folder have been fixed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7784

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/FilesList/KeyboardActions.vue
+++ b/packages/web-app-files/src/components/FilesList/KeyboardActions.vue
@@ -35,7 +35,7 @@ export default defineComponent({
 
   computed: {
     ...mapState('Files', ['latestSelectedId']),
-    ...mapGetters('Files', ['files', 'selectedFiles'])
+    ...mapGetters('Files', ['selectedFiles'])
   },
 
   mounted() {
--- a/packages/web-app-files/src/components/FilesList/KeyboardActions.vue
+++ b/packages/web-app-files/src/components/FilesList/KeyboardActions.vue
@@ -133,7 +133,6 @@ export default defineComponent({
       }
 
       if (key === keycode('c') && ctrl) {
-        console.log('selectedFiles', this.selectedFiles)
         return this.copySelectedFiles({ space: this.space, resources: this.selectedFiles })
       }
       if (key === keycode('v') && ctrl) {
--- a/packages/web-app-files/src/components/FilesList/KeyboardActions.vue
+++ b/packages/web-app-files/src/components/FilesList/KeyboardActions.vue
@@ -5,7 +5,7 @@
 <script lang="ts">
 import keycode from 'keycode'
 import { eventBus } from 'web-pkg/src/services/eventBus'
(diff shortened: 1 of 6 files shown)


packages/web-app-files/src/mixins/actions/copy.js
```diff
@@ -53,9 +56,19 @@ export default {
...
-
-    $_copy_trigger() {
+    ...mapMutations('Files', ['SET_FILE_SELECTION']),
+    $_copy_trigger({ resources }) {
+      this.SET_FILE_SELECTION(resources)
       this.copySelectedFiles({ space: this.space })
+      const currentFolderSelected = resources.some((r) => r.id === this.currentFolder?.id)
+      if (currentFolderSelected) {
+        return this.$router.push(
+          createFileRouteOptions(this.space, {
```
kulmann, line 66: Are you navigating to the parent because otherwise the selection doesn't make sense? Or what's the reason?

JammingBen: Yes. No matter if you copy or cut, you can't perform the action within the current folder. I admit, it might be a bit confusing for the user though. Do you have an alternative solution?

kulmann: Accessibility wise (and probably also in general UX wise) it's an issue to navigate to another page without communicating clearly. As far as I can tell we need to stay in the current folder. I'd be fine with the notification.

packages/web-app-files/src/mixins/actions/copy.js
```diff
@@ -53,9 +56,19 @@ export default {
   },
   methods: {
     ...mapActions('Files', ['copySelectedFiles']),
-
-    $_copy_trigger() {
+    ...mapMutations('Files', ['SET_FILE_SELECTION']),
+    $_copy_trigger({ resources }) {
+      this.SET_FILE_SELECTION(resources)
```
kulmann, line 61: Can we change the `copySelectedFiles` action in the store instead? You can add an optional param for the resources to be copied/moved. Feels weird needing to set the selection in the store first (which could have other side effects).

packages/web-app-files/src/components/FilesList/KeyboardActions.vue
```diff
@@ -34,7 +34,8 @@ export default defineComponent({
   },
 
   computed: {
-    ...mapState('Files', ['latestSelectedId'])
+    ...mapState('Files', ['latestSelectedId']),
+    ...mapGetters('Files', ['files', 'selectedFiles'])
```
kulmann, line 38: Why did you add `files` as well? I only see `selectedFiles` being used in the diff.

kulmann (approved):

packages/web-app-files/src/components/FilesList/KeyboardActions.vue
JammingBen: Ah damn, left-over from dev

PR #7833, opened 2022-10-20
[full-ci] Clean up packages and dependencies some more

dschmidt:
## Description
This is a follow up for https://github.com/owncloud/web/pull/7816.

The first PR made it possible to build oC Web with rollup after installing the deps with `pnpm import && pnpm install`.
Running the tests wasn't possible yet, this PR addresses that.

To achieve that:
- more explicit dependencies are added to packages
- superfluous/obsolete dependencies are removed
- mocks are moved into a new `web-test-helpers` package (which in turn can import types from e.g. `vue-router`, so not all packages using a mock are required to add `vue-router` as a direct dep)

Moreover a bunch of other nitpick fixes are included:
- `@files` import alias is gone for good now
- composition-api import fixes

To test this PR with `pnpm`:
```bash
yarn install
pnpm import
rm -rf **/node_modules/
pnpm install
pnpm build
pnpm test:unit --coverage
```

## Related Issue

- Fixes <issue_link>

## Motivation and Context

We (at least @fschade and I) would like to switch to `pnpm`, as it obviously enforces way better declaration of package dependencies and makes it harder to add imports that would break when publishing packages separately to npm.
For instance when we publish `web-client` or `web-pkg`, those packages should never import anything from `web-runtime` or `web-app-files`. `pnpm` makes it impossible to do that without adding those as dependencies in `package.json` which is way easier to spot in a review than just a broken import.

Additionally it has some nice features for publishing multiple packages from a monorepo.

## How Has This Been Tested?
- 

Screenshot_20221020_112540
:trollface: 

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Followup tasks:
- Switch to pnpm


kulmann (approved):
LGTM 💪

Issue #7657, opened 2022-09-19
Wrong share permissions for re-shares from "Shared with others/via link" page


JammingBen:
### Steps to reproduce
1. Create a link for a resource WITHIN a received share that has only "View"-permissions
2. Head the to "Shared via link" page
3. Open the sidebar for this resource and check the permission dropdown menus in the share panel

### Expected behaviour
The "View"-permissions can't be exceeded.

### Actual behaviour
All roles can be selected.

~We need the `shareId` in the OCS share response to retrieve the parent share and its permissions.~ -> not possible. Hence we need another solution.

Edit: This could work if we had the storage ID for each of the parent resources. Then we could retrieve the parent shares via the `sharesTree`.

PR #7844, opened 2022-10-21
Fix selectable share permissions for re-shares

JammingBen:
## Description
We've fixed a bug where the selectable roles on a re-share could exceed the parent share's permissions in certain scenarios.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7657

## Types of changes

- [ ] Technical debt
- [ ] Tests


kulmann (changes requested):
This introduces quite some code duplication with the private link resolving 🤔 Does it make sense to port the composable to `web-pkg`?

JammingBen:
> This introduces quite some code duplication with the private link resolving 🤔 Does it make sense to port the composable to `web-pkg`?

I had the same idea but some parts differ a bit, and either way, it's only temporary solution. Hence I figured to leave it. What do you think?

kulmann (approved):

Issue #7603, opened 2022-09-08
Searching in public link takes you to the personal of the creator of the public link


SwikritiT:
### Steps to reproduce
1. As user `Alice` create folders `test`, `test/test-folder`, `test/test-test`
2. create a public link for folder `test`
3. navigate to the public link
4. In global search, search for `test`

### Expected behaviour
The search list should display the resources with `test` keyword in them

### Actual behaviour
the search list doesn't display the result of the search query but if I `search all files`  it navigates to the `Personal` of user `Alice` 

#### OCIS
https://user-images.githubusercontent.com/41103328/189056363-8edd6870-0038-4cb0-9dcf-380cce0e6f6a.mp4

#### OC10
https://user-images.githubusercontent.com/41103328/189056656-089499ce-a16b-4888-9b86-73f4cb8c167c.mp4

If I open the link in incognito mode the search bar is not displayed

Both ocis and oc10 are the latest master

individual-it:
with the web in the current ocis the search does not show anything but throws errors
![image](https://user-images.githubusercontent.com/2425577/195329050-57552751-36c7-4d69-aca0-1d96a8f83cee.png)

PR #7849, opened 2022-10-24
Hide search bar in public link context

JammingBen:
## Description
The search bar in a public link context has been hidden because it requires authentication.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7603

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-search/src/portals/SearchBar.vue
+++ b/packages/web-app-search/src/portals/SearchBar.vue
@@ -99,12 +99,18 @@ import { providerStore } from '../service'
 import { createLocationCommon } from 'web-app-files/src/router'
 import Mark from 'mark.js'
 import debounce from 'lodash-es/debounce'
+import { useStore, useUserContext } from 'web-pkg/src/composables'
 import { eventBus } from 'web-pkg/src/services/eventBus'
 import { defineComponent } from '@vue/composition-api'
-import {mapGetters} from "vuex";
 
 export default defineComponent({
   name: 'SearchBar',
+  setup() {
+   const store = useStore()
+    return {
+      isUserContext: useUserContext({ store })
+    }
+  },
 
   data() {
     return {
@@ -122,8 +128,6 @@ export default defineComponent({
     }
   },
   computed: {
-    ...mapGetters(['user']),
-
     showNoResults() {
       return this.searchResults.every(({ result }) => !result.values.length)
     },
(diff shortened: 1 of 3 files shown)


packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -101,6 +101,7 @@ import Mark from 'mark.js'
 import debounce from 'lodash-es/debounce'
 import { eventBus } from 'web-pkg/src/services/eventBus'
 import { defineComponent } from '@vue/composition-api'
+import {mapGetters} from "vuex";
```
kulmann, line 104: Your autolinting looks broken 🤔

packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -121,12 +122,17 @@ export default defineComponent({
...
+    ...mapGetters(['user']),
+
     showNoResults() {
       return this.searchResults.every(({ result }) => !result.values.length)
     },
     availableProviders() {
       return this.providerStore.availableProviders
     },
+    isSearchBarEnabled() {
+      return this.availableProviders.length && this.user?.id
```
kulmann, line 134: Could you please add a setup function with
```js
return {
  isUserContext: useUserContext({ store })
}
```
and use `this.isUserContext` instead of `this.user?.id`?

kulmann (changes requested):

packages/web-app-search/src/portals/SearchBar.vue
JammingBen: Only this one file which does not lint correctly... what was the issue here again?

kulmann: Previously it was `script type="ts"` instead of `script lang="ts"`. But that's not the issue here. 🤔

kulmann (approved):
👍

packages/web-app-search/src/portals/SearchBar.vue
JammingBen: It seems that my IDE doesn't like the template structure of this file, hence it won't lint at all. `<li><oc-list><li>...</li></oc-list></li>`

image

kulmann: Ah, I have the same (in WebStorm).

PR #7861, opened 2022-10-25
continue on non 2xx responses

fschade:
## Description
Axios stopped execution if the response code was not 2xx, in those cases the app error message was never shown as expected. This pr allows axios to continue and does the error handling on its own.


kulmann (approved):
👍

PR #7865, opened 2022-10-25
Disable share renaming

JammingBen:
## Description
Renaming of shares has been disabled temporarily until it works as expected. Also see https://github.com/owncloud/ocis/issues/4866

**Note:** We need to merge this before GA if the issue has not been solved until then.

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/mixins/actions/rename.ts
+++ b/packages/web-app-files/src/mixins/actions/rename.ts
@@ -41,6 +41,12 @@ export default {
               return false
             }
 
+            // FIXME: Remove this check as soon as renaming shares works as expected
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/mixins/actions/rename.ts
```diff
@@ -41,6 +41,12 @@ export default {
               return false
             }
 
+            // FIXME: Remove this check as soon as renaming shares works as expected
```
kulmann, line 44: Could you add the ocis issue link here? Otherwise it's really hard to find out when it's supposed to work again.

kulmann (approved):
👍

Issue #7863, opened 2022-10-25
Routes auth config not honoured


diocas:
I might want to create an extension that exposes a route publicly (no user context).
Before I could set `meta.auth: false` and everything would work, but this configuration is no longer honoured.

The auth logic (auth guard) eventually enters in the user context 'if' (`isUserContext === true` because `contextRoute === null`), where it checks that `isUserContextReady === false`, thus redirecting the user back to login.

diocas:
I tried replacing (from router/helpers.ts > isUserContext)
```js
  const contextRoute = getContextRoute(router, to)
  return !contextRoute || contextRoute.meta?.auth !== false
```
with
```js
  const contextRoute = getContextRoute(router, to)
  if (contextRoute) {
    return contextRoute.meta?.auth !== false
  }
  return false
```
Which prevents the login (does `next()`), but then something gets stuck somewhere else...

diocas:
Think I found it.. I can make it simpler.. Will run the tests and do a PR.

kulmann:
There was this PR https://github.com/owncloud/web/pull/7437 approx. 2.5 months ago, where we changed the behaviour of the authGuard to interpret the absence of a contextRoute as requiring a login. This still makes sense to me. Defining `meta.auth: false` on the route itself always takes precedence over context route handling. That being said I don't understand the exact details of how you reach the faulty state. Can you disclose more details about your use case?

diocas:
The problem is that the target route auth is not absent and I'm still redirected to login if I'm not logged in.
I would expect non oC extensions (especially if not opened via files app), to not have the contextRoute and still be able to expose public routes.

I just created an extension that sets `auth: false` and observe this. This extension, in particular, is to redirect old CERNBox URLs to the new ones, including the public links. This could be handled on our nginx or on the gateway (like upstream oCIS does), but this way I kept all logic in a single place. Users should be able to reach `[appId]/s/:token` (with `appId = index.php`), which was supposed to be available worldwide but is not.
When we pushed this to all CERN on Monday without realising the issue, we had to create a workaround.

You can see here the definition:
https://github.com/cernbox/web-extensions/blob/main/old-web-redirector/src/index.js#L14

kulmann:
> The problem is that the target route auth is not absent and I'm still redirected to login if I'm not logged in. I would expect non oC extensions (especially if not opened via files app), to not have the contextRoute and still be able to expose public routes.
> 
[...]

Issue #7234, opened 2022-07-07
route.meta.auth should be deprecated and renamed to something more meaningful


dschmidt:
This is a follow up ticket for #7072.

`meta.auth` is used in `isUserContext`. If it's `false`, we don't necessarily need a user context - it could be a public link context as well.
The name doesn't reflect what it's actually good for. We need a better name and a sane default.

First ideas included `userContextRequired: false`, `hybridContext: true`, …

Issue #7338, opened 2022-07-26
Deprecations


kulmann:
The purpose of this issue is to keep track of deprecations and dates of code removal. Please subscribe to this issue if you are a developer and want to stay informed about deprecations / code deletion.

# web runtime

## `getToken` getter in the `user` store module (global namespace)
- deprecated on 2022-07-07
- to be removed in major version 6
- migration: use the `runtime/auth/accessToken` getter
- reasoning: `user` state is not supposed to deal with authentication details. We introduced the namespaced `auth` module for `user` and `public-link` related authentication contexts.

LukasHirt:
This has been replaced by a new section in dev docs — https://owncloud.dev/clients/web/releasing/deprecations/

PR #7875, opened 2022-10-26
Introduce meta authContext

kulmann:
## Description
This PR introduces a `meta.authContext` prop on routes so that we give fine grained control to extension developers which kind of auth context they want to require from ownCloud Web.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/7234
- Fixes https://github.com/owncloud/web/issues/7863

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-runtime/src/App.vue
+++ b/packages/web-runtime/src/App.vue
@@ -49,7 +49,7 @@ import LayoutLoading from './layouts/Loading.vue'
 import LayoutPlain from './layouts/Plain.vue'
 import { getBackendVersion, getWebVersion } from './container/versions'
 import { defineComponent } from '@vue/composition-api'
-import { isPublicLinkContext, isUserContext, isAuthenticationRequired } from './router'
+import { isPublicLinkContext, isUserContext, isAnonymousContext } from './router'
 import { additionalTranslations } from './helpers/additionalTranslations' // eslint-disable-line
 import { eventBus } from 'web-pkg/src/services'
 
@@ -70,7 +70,7 @@ export default defineComponent({
     ...mapGetters(['configuration', 'capabilities', 'getSettingsValue']),
     ...mapGetters('runtime/auth', ['isUserContextReady', 'isPublicLinkContextReady']),
     layout() {
-      if (!this.$route.name || !isAuthenticationRequired(this.$router, this.$route)) {
+      if (!this.$route.name || isAnonymousContext(this.$router, this.$route)) {
         return LayoutPlain
       }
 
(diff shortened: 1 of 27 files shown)


dschmidt (approved):
Haven't tested and would be better with https://dev.to/ayc0/typescript-49-satisfies-operator-1e4i but alas ... better than before!

I especially like the very useful runtime/deprecation warnings!

dschmidt:
When this is merged, please add an entry here: https://github.com/owncloud/web/issues/7338

Might be worth pinning that issue...

kulmann:
> When this is merged, please add an entry here: #7338
> 
> Might be worth pinning that issue...

It was pinned when I created it in the first place (see screenshot)... thanks for re-pinning. And yes, of course, the deprecation info needs to go into the ticket. :-)

Screenshot 2022-10-27 at 08 59 45

diocas (commented):
Does what is really necessary for now (even if the presentation is inconsistent: see my comment in the original issue).

packages/web-runtime/src/App.vue
```diff
@@ -70,18 +70,24 @@ export default defineComponent({
     ...mapGetters(['configuration', 'capabilities', 'getSettingsValue']),
     ...mapGetters('runtime/auth', ['isUserContextReady', 'isPublicLinkContextReady']),
     layout() {
-      if (!this.$route.name || !isAuthenticationRequired(this.$router, this.$route)) {
-        return LayoutPlain
+      const pageLayouts = {
```
dschmidt, line 73:
```suggestion
      const pageLayouts: Record<PageLayout, VueComponentSomething> = { 
```

.... again satisfies would be very nice here - then the object could simply satisfy `Record<PageLayout, any>` without narrowing or widening the type

dschmidt (approved):
Codewise the PR seems completely fine to me, I'm just not so enthusiastic about exposing the `pageLayout` setting. Especially the loading layout seems a bit weird, but then again it's probably not harmful

kulmann:
> Codewise the PR seems completely fine to me, I'm just not so enthusiastic about exposing the `pageLayout` setting. Especially the loading layout seems a bit weird, but then again it's probably not harmful

You're right, I didn't remember our decision to force all apps into the `application layout`. The more I think about it, the less I want to expose the pageLayout to developers. Instead I'd rather use the `application` layout as fallback for anonymous routes (which solves @diocas use case as well). Will discuss with @tbsbdr if we want to drop the `plain` layout for our own internal anonymous routes as well (like accessDenied page, oidcCallback page, etc)... converting back to draft.

kulmann:
Updated the PR to not expose/use the `pageLayout` prop anymore. As described earlier today, all app routes, including the ones requesting an `anonymous` context, will render in the `Application` layout. I kept the plain layout for login, logout, oidcCallback, oidcSilentRedirect, resolvePublicLink and accessDenied. Those were our implicitly anonymous page before. Now they are our implicitly plain-layout pages (they all received an `authContext: "anonymous"` meta prop in this PR).

PR #7880, opened 2022-10-26
[full-ci] Refactor webdav listFiles

lookacat:
## Description
Refactor webDav listFiles so the result isn't only an array but rather an object containing the requested node and its children.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests updated
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/mixins/actions/restore.ts
+++ b/packages/web-app-files/src/mixins/actions/restore.ts
@@ -68,10 +68,9 @@ export default {
           existingResources = existingResourcesCache[parentPath]
         } else {
           try {
-            existingResources = await this.$clientService.webdav.listFiles(this.space, {
+            existingResources = (await this.$clientService.webdav.listFiles(this.space, {
               path: parentPath
-            })
-            existingResources = existingResources.slice(1)
+            })).children
           } catch (error) {
             missingFolderPaths.push(parentPath)
           }
--- a/packages/web-app-files/src/mixins/actions/restore.ts
+++ b/packages/web-app-files/src/mixins/actions/restore.ts
@@ -68,9 +68,11 @@ export default {
           existingResources = existingResourcesCache[parentPath]
         } else {
           try {
-            existingResources = (await this.$clientService.webdav.listFiles(this.space, {
-              path: parentPath
-            })).children
+            existingResources = (
+              await this.$clientService.webdav.listFiles(this.space, {
+                path: parentPath
+              })
+            ).children
           } catch (error) {
             missingFolderPaths.push(parentPath)
           }
--- a/packages/web-app-files/src/services/folder/loaderSpace.ts
+++ b/packages/web-app-files/src/services/folder/loaderSpace.ts
@@ -55,7 +55,7 @@ export class FolderLoaderSpace implements FolderLoader {
 
         const listFilesResult = yield webdav.listFiles(space, { path, fileId })
         const resources = listFilesResult.children
-        let currentFolder = listFilesResult.folder
+        let currentFolder = listFilesResult.node
         replaceInvalidFileRoute({ space, resource: currentFolder, path, fileId })
 
         if (path === '/') {
--- a/packages/web-app-files/src/services/folder/loaderSpace.ts
+++ b/packages/web-app-files/src/services/folder/loaderSpace.ts
@@ -53,9 +53,9 @@ export class FolderLoaderSpace implements FolderLoader {
       try {
         store.commit('Files/CLEAR_CURRENT_FILES_LIST')
 
-        const propfindResult = yield webdav.listFiles(space, { path, fileId })
-        const resources = propfindResult.children
-        let currentFolder = propfindResult.resource
+        const listFilesResult = yield webdav.listFiles(space, { path, fileId })
+        const resources = listFilesResult.children
+        let currentFolder = listFilesResult.resource
         replaceInvalidFileRoute({ space, resource: currentFolder, path, fileId })
 
         if (path === '/') {
--- a/packages/web-app-files/src/services/folder/loaderSpace.ts
+++ b/packages/web-app-files/src/services/folder/loaderSpace.ts
@@ -53,8 +53,9 @@ export class FolderLoaderSpace implements FolderLoader {
       try {
         store.commit('Files/CLEAR_CURRENT_FILES_LIST')
 
-        const resources = yield webdav.listFiles(space, { path, fileId })
-        let currentFolder = resources.shift()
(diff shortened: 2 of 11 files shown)


packages/web-client/src/webdav/listFiles.ts
```diff
@@ -20,7 +20,7 @@ export const ListFilesFactory = ({ sdk }: WebDavOptions) => {
       space: SpaceResource,
       { path, fileId }: { path?: string; fileId?: string | number } = {},
       { depth = 1, davProperties }: ListFilesOptions = {}
-    ): Promise<Resource[]> {
+    ): Promise<{ folder: Resource; children: Resource[] }> {
```
kulmann, line 23: - Please define an interface for the return type
- `children` is optional, so you need to declare it as `children?: Resource[]`
- `folder` is not such a good fit, naming wise, since it can also be a space. Maybe `node` and `children` works better? @dschmidt what's your opinion on this?

dschmidt: Naming is hard... 

folder is very bad, I agree. I'd say even file would be better than folder, as every folder is a file but not every file a folder (in POSIX terms). node is a bit weird imho as we don't use that term anywhere afaict. 

I know it's semantically weak, but what about resource? main? mainResource? result? 
At least it's not more specific than the actual result - which might be misleading (eg when the result is a space node seems a bit off)

und jetzt weiß ich auch nicht 😅

packages/web-app-files/src/services/folder/loaderSpace.ts
```diff
@@ -53,8 +53,9 @@ export class FolderLoaderSpace implements FolderLoader {
       try {
         store.commit('Files/CLEAR_CURRENT_FILES_LIST')
 
-        const resources = yield webdav.listFiles(space, { path, fileId })
-        let currentFolder = resources.shift()
+        const listFilesResult = yield webdav.listFiles(space, { path, fileId })
+        const resources = listFilesResult.children
+        let currentFolder = listFilesResult.folder
```
dschmidt, line 58: Looking at this code, maybe `node` is still our best bet. But maybe keeping listFiles as a name wasn't too smart either. 
propfindResult.resource or similar seems okayish? Dunno... 

The result could be destructured directly

packages/web-app-files/src/mixins/actions/restore.ts
```diff
@@ -68,10 +68,11 @@ export default {
...
           try {
-            existingResources = await this.$clientService.webdav.listFiles(this.space, {
-              path: parentPath
-            })
-            existingResources = existingResources.slice(1)
+            existingResources = (
+              await this.$clientService.webdav.listFiles(this.space, {
+                path: parentPath
+              })
+            ).children
```
dschmidt, line 75: This is some semantically telling code, me gusta

packages/web-client/src/webdav/listFiles.ts
lookacat: i also think node is a better fit 👍🏼

packages/web-app-files/src/services/folder/loaderSpace.ts
lookacat: yes right, but keeping listFiles as the method name?

kulmann: I would like to avoid the word `propfind`. 😉 `listFiles` as method name is ok IMO.

packages/web-client/src/webdav/listFiles.ts
lookacat: alright

lookacat: fixed 👍🏼

packages/web-app-files/src/services/folder/loaderSpace.ts
kulmann: Stop stop stop. Why not `listFiles` and `listFilesResult`?!

kulmann: The word `propfind` outside of the web-client feels wrong.

ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/29449/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/29449/tracing/file-alice-2022-10-27-08-41-00.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/29449/tracing/file-brian-2022-10-27-08-41-21.zip 


</p></details>

packages/web-client/src/helpers/space/types.ts
```diff
@@ -75,3 +75,8 @@ export interface PublicSpaceResource extends SpaceResource {
 export const isPublicSpaceResource = (resource: Resource): resource is PublicSpaceResource => {
   return resource.driveType === 'public'
 }
+
+export interface PropfindResult {
```
kulmann, line 79: Again, the term `propfind` is something API specific, in this case webdav. We want to avoid leaking API specifics as much as possible. Please use a name like `ListFilesResult`.

Apart from that, the code location of the interface is not good. This is no space type. Please export the type from `web-client/src/webdav/listFiles.ts`.

kulmann (changes requested):

packages/web-client/src/helpers/space/types.ts
kulmann: Need to un-resolve this, sorry. Please see second part of the comment above - interface needs to be moved to the `listFiles` file.

Also, again, please don't use `propfind` in naming. There are still locations in the code left. A simple `git grep "propfind"` will show you the locations.

packages/web-app-files/tests/unit/helpers/resource/resourcesTransfer.spec.ts
```diff
@@ -74,8 +75,12 @@ describe('resourcesTransfer', () => {
     it.each([TransferType.COPY, TransferType.MOVE])(
       'should copy / move files without renaming them if no conflicts exist',
       async (action: TransferType) => {
+        const listFilesResult = {
+          resource: {} as Resource,
+          children: [] as Resource[]
+        } as ListFilesResult
```
dschmidt, line 81: Why not?
```suggestion
        const listFilesResult: ListFilesResult = {
          resource: {},
          children: []
        }
```

lookacat: resource braucht as a Resource sonst meckert typescript ^^

kulmann (approved):

Issue #7866, opened 2022-10-25
Fresh `pnpm i` fails


pascalwengerter:
### Steps to reproduce
1. Clone this repo and `cd` into it
2. Run `pnpm i`
3. See this
<img width="717" alt="Screenshot 2022-10-25 at 23 11 34" src="https://user-images.githubusercontent.com/16822008/197882275-bfee7e09-065a-4af3-ba6f-1db390d059c5.png">

### Expected behaviour
Things work

### Actual behaviour
Things don't work. Not sure if that's related to my local `pnpm` cache maybe? I was trying pretty hard getting a local ODS linked via `pnpm`

dschmidt:
@saw-jan had the same issue

Do you remember, what solved the problem for you? I think it was missing the `.npmrc` file, but it started working, when I added it to the repo iirc.

```
11:27
yeah, I am trying to build web repo locally
I did pnpm upgrade to upgrade the packages
And now the install command works
but I don't think we need to do upgrade
```

Agreed, upgrade installs everything in the valid version ranges, which probably gives you too new versions of a bunch of our deps.
Maybe it also fails now because we pinned dependency versions via renovatebot?

pascalwengerter:
The culprint seems to be `vue-jest` having been set to `v3.0.7` by renovate-bot, fixing it at `v3.0.0` seems to resolve it for now 🥳

dschmidt:
I don't get that, if you look at  https://github.com/owncloud/web/pull/7842/files :

Before `^3.0.7` was pinned, now `3.0.7` is pinned - in both `package.json` and `pnpm-lock.yaml`. If at all the release should be older than before, but `3.0.7` is already the latest (stable?) release...

Maybe you can paste your diff or send a PR with your changes and the updated lock file?

saw-jan:
I found out something.
I use `nvm` to install node versions and `npm` to install `pnpm` and works just fine.
1. :heavy_check_mark: Working
i. Install node 16: `nvm install 16`
ii. Install pnpm: `npm i -g pnpm` (path: `~/.nvm/versions/node/v16.18.0/bin/pnpm`)
iii. `pnpm i` works :heavy_check_mark: 

2. :x: `spawn node-gyp` Error
i. Install node 16: `nvm install 16`
ii. Install pnpm: `curl -fsSL https://get.pnpm.io/install.sh | sh -` (path: `~/.local/share/pnpm/pnpm`)
iii. `pnpm i` fails :x:

[...]

PR #7898, opened 2022-11-02
Update vue-jest

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/7866

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "presets": ["@ownclouders/babel-preset"]
-}
--- a/.npmrc
+++ b/.npmrc
@@ -20,9 +20,6 @@ hoist-pattern[]=vuex
 # eslint (public so they are picked up by IDEs)
 public-hoist-pattern[]=*eslint*
 
-# vue-jest
-hoist-pattern[]=convert-source-map
-
 # owncloud-design-system
 hoist-pattern[]=*babel*
 hoist-pattern[]=deepmerge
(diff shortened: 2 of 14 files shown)


dschmidt:
Hehe, this is how far I got, when I tested locally :-) 

Somehow JS support is broken and I couldn't figure out why. For some reason javascript files (and non ts .vue files) don't seem to be piped through babel, that's why module syntax is not supported. I started making a ton of files typescript files but that had other issues: for some reason jest and rollup seem to handle a bunch of imports in typescript differently (we worked around that in a bunch of places, but I didn't feel like doing that this time).

We should probably figure out why
1) babel isn't used anymore (for this PR)
2) typescript imports behave differently in different environments (in general/the long run)

dschmidt:
btw `@vue/vue2-jest` has `convert-source-map` in its dependencies, so we can remove that `hoist-pattern` from `.npmrc`

... and thanks for looking into this!

dschmidt:
... and another btw: I know the README of `vue-jest` says one should use version `28`, but that info is outdated. We can use the latest version, which currently is `29.1.1`

Issue #7900, opened 2022-11-02
Upload of big files is failing in CERNBox: Uppy error "Upload stalled for 30 seconds, aborting."


elizavetaRa:
### Steps to reproduce
1. Upload a large file (6-7 GB)

### Expected behaviour
Upload works

### Actual behaviour
At the very end of the upload it always fails for any user, in console we get 
`[Uppy]  Failed to upload xxx. Upload stalled for 30 seconds, aborting.`

For smaller files (tested 2 and 3 GB) everything works.
With curl we are able to upload the same big files, so it must be the frontend issue.

We tried to set the timeout in uppyOptions to 1h and to 0 (deactivates check) but still get the same error.

For smaller files (tested 2 and 3 GB) everything works.

The nginx upload limit is 8GB.

kulmann:
This is without TUS, right? So no chunking in place, only one big XHR upload, correct?

elizavetaRa:
Xhr upload, yes

JammingBen:
> We tried to set the timeout in uppyOptions to 1h and to 0 (deactivates check) but still get the same error.

@elizavetaRa Where exactly did you set the timeout? The proper place would be where the XHR plugin is being initialized: https://github.com/owncloud/web/blob/54e26b5010bb0ca99626808d5a67579466731da8/packages/web-runtime/src/services/uppyService.ts#L78

I did get the same error, however, after increasing the timeout in the XHR options, everything works.

elizavetaRa:
@JammingBen I was setting in `packages/web-runtime/src/composables/upload/useUpload.ts`  in `uppyOptions`. Will try your way, thanks!

elizavetaRa:
@JammingBen it worked by us as well, maybe you could add the fix to web

JammingBen:
https://github.com/owncloud/web/pull/7912 increases the default timeout and also makes it configurable.

PR #7912, opened 2022-11-03
Increase default XHR timeout and make it configurable

JammingBen:
## Description
The default timeout for XHR uploads has been increased from 30 to 60 seconds. Also, it can now be configured via the `config.json` file (in ms).

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7900

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/composables/upload/useUpload.ts
+++ b/packages/web-runtime/src/composables/upload/useUpload.ts
@@ -104,7 +104,8 @@ export function useUpload(options: UploadOptions): UploadResult {
       ...(isTusSupported && {
         tusMaxChunkSize: unref(tusMaxChunkSize),
         tusHttpMethodOverride: unref(tusHttpMethodOverride),
-        tusExtension: unref(tusExtension)
+        tusExtension: unref(tusExtension),
+        xhrTimeout: store.getters.configuration?.options?.upload?.xhr?.timeout || 60000
       })
     }
   })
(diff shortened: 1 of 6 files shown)


packages/web-runtime/src/composables/upload/useUpload.ts
```diff
@@ -104,7 +104,8 @@ export function useUpload(options: UploadOptions): UploadResult {
       ...(isTusSupported && {
         tusMaxChunkSize: unref(tusMaxChunkSize),
         tusHttpMethodOverride: unref(tusHttpMethodOverride),
-        tusExtension: unref(tusExtension)
+        tusExtension: unref(tusExtension),
+        xhrTimeout: store.getters.configuration?.options?.upload?.xhr?.timeout || 60000
```
kulmann, line 108: huh, why do you set the xhr timeout if tus is supported?

kulmann (changes requested):
Please add a short description of the new option to the `getting started` page in the `docs` folder (there is a section for options).

packages/web-runtime/src/composables/upload/useUpload.ts
JammingBen: Arrrghs 😅

kulmann (approved):
Nice! 👍

Issue #7608, opened 2022-09-08
Right-click on spaces does not open context menu


hurradieweltgehtunter:
### Steps to reproduce
1. Create Space
2. Right click on it

### Expected behaviour
Context menu opens

### Actual behaviour
Context menu does not open. Need to click on the three dots icon

PR #7917, opened 2022-11-04
open context menu on right click

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/7608

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/views/spaces/Projects.vue
+++ b/packages/web-app-files/src/views/spaces/Projects.vue
@@ -35,7 +35,7 @@
                 class="spaces-list-card oc-card oc-card-default oc-rounded"
                 :data-space-id="space.id"
                 :class="getSpaceCardAdditionalClass(space)"
-                @contextmenu="(event) => contextMenuClicked(event, space)"
+                @contextmenu="(event) => showSpaceContextDrop(event, space)"
               >
                 <div class="oc-card-media-top oc-border-b">
                   <component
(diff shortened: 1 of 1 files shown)


kulmann (changes requested):
In the file list we open the context menu on right click at the current position of the mouse cursor. Could you check if it's feasible to do the same here?

JanAckermann:
sure, done

kulmann (changes requested):
Cool, thanks for the improvement! Another nitpick about the new state: the context menu seems to open to the left be default. It should open to the right by default.

JanAckermann:
@kulmann I can't tell that there is a difference between this pr and master

image
image
image

packages/web-app-files/src/views/spaces/Projects.vue
```diff
@@ -265,6 +267,26 @@ export default defineComponent({
...
+        return
+      }
+
+      drop.tippy.setProps({
+        getReferenceClientRect: () => ({
+          width: 0,
+          height: 0,
+          top: event.clientY,
+          bottom: event.clientY,
+          left: event.clientX,
```
kulmann, line 284: Like for the `ResourceTable` the `left` and `right` prop needs to be set to bounding client rect `x` value of the context menu trigger button. Otherwise, using `Enter` with the keyboard opens the context menu in weird places.

kulmann (changes requested):

kulmann:
> @kulmann I can't tell that there is a difference between this pr and master
> 
> image image image

Doesn't mean that it should stay like that :-P but probably just out of scope for this PR then, which is fine 👍

kulmann (approved):

Issue #7924, opened 2022-11-04
Spaces not being counted in files lists


JammingBen:
### Steps to reproduce
1. Create a public link for a folder, a file and a space
2. Head to the "Shared via link"-page

### Expected behaviour
At the bottom of the files list, it should say `3 items in total (1 file, 1 folder, 1 space)`

### Actual behaviour
The space is being ignored:

![image](https://user-images.githubusercontent.com/50302941/199970327-9ceb79b1-7bb2-40d6-9828-6142c7c18e7c.png)

PR #7926, opened 2022-11-04
Include spaces in the list info

JammingBen:
## Description
Spaces have been included in the list info below file lists that support displaying spaces.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7924

## Screenshot:
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/FilesList/ListInfo.vue
+++ b/packages/web-app-files/src/components/FilesList/ListInfo.vue
@@ -5,6 +5,7 @@
       :data-test-items="items"
       :data-test-files="files"
       :data-test-folders="folders"
+      :data-test-spaces="spaces"
       :data-test-size="size"
       class="oc-text-muted"
     >
@@ -26,6 +27,16 @@ export default {
       type: Number,
       required: true
     },
+    spaces: {
+      type: Number,
+      default: 0,
+      required: false
+    },
+    showSpaces: {
+      type: Boolean,
+      default: false,
+      required: false
+    },
     /**
      * Total size in bytes. Unformatted strings and integers are allowed.
      */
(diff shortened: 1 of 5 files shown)


kulmann (changes requested):
As discussed, this PR can't go into the stable branch because of changes in or additional translation strings. Up to you if you want to make two different PRs for stable and master, or only point this PR to master.

packages/web-app-files/src/components/FilesList/ListInfo.vue
```diff
@@ -37,7 +48,8 @@ export default {
   },
   computed: {
     items() {
-      return this.files + this.folders
+      const filesAndFolderCount = this.files + this.folders
+      return this.showSpaces ? filesAndFolderCount + this.spaces : filesAndFolderCount
```
kulmann, line 52: Why not single line as follows?
```js
return this.files + this.folders + (this.showSpaces ? this.spaces : 0)
```

kulmann: And more radical: isn't `this.spaces` always `0` if `this.showSpaces` evaluates to `false`? In that case you could just do

```js
return this.files + this.folders + this.spaces
```

packages/web-app-files/src/store/getters.ts
```diff
@@ -32,9 +32,11 @@ export default {
   totalFilesCount: (state, getters) => {
     const fileCount = getters.filesAll.filter((file) => file.type === 'file').length
     const folderCount = getters.filesAll.filter((file) => file.type === 'folder').length
+    const spaceCount = getters.filesAll.filter((file) => file.type === 'space').length
```
kulmann, line 35: Is this sufficient? As far as I can tell we don't want to count personal spaces for example. Which would mean that you need to extend the filter callback to also check the driveType.

kulmann (changes requested):
Rebase please :-) Found some small issues, please see comments.

packages/web-app-files/src/store/getters.ts
JammingBen: Good catch, t hx!

Issue #7933, opened 2022-11-07
Add origin check to events


C0rby:
This event listener should check the event's origin.
https://github.com/owncloud/web/blob/37e83b008f5baa762c6544e126e3836d44c11c10/packages/web-app-draw-io/src/App.vue#L75-L93
And these lines should contain the target origin:
https://github.com/owncloud/web/blob/37e83b008f5baa762c6544e126e3836d44c11c10/packages/web-app-draw-io/src/App.vue#L118
https://github.com/owncloud/web/blob/37e83b008f5baa762c6544e126e3836d44c11c10/packages/web-app-draw-io/src/App.vue#L145
https://github.com/owncloud/web/blob/37e83b008f5baa762c6544e126e3836d44c11c10/packages/web-app-draw-io/src/App.vue#L186
https://github.com/owncloud/web/blob/37e83b008f5baa762c6544e126e3836d44c11c10/packages/web-app-draw-io/src/App.vue#L211

See this for more information: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#security_concerns
If you want a demo of a poc exploit, just ping me. :)

JammingBen:
This has been fixed via https://github.com/owncloud/web/pull/7941 in the `stable-6.0` branch. Will come to master soon.

PR #7941, opened 2022-11-09
Add origin checks to all Draw.io events

JammingBen:
## Description
Origin checks have been added to all Draw.io events due to security reasons.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7933

## Types of changes

- [ ] Technical debt
- [x] Security
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-origin-event-check
+++ b/changelog/unreleased/enhancement-origin-event-check
@@ -1,0 +1,6 @@
+Enhancement: Add origin check to Draw.io events
+
+Origin checks have been added to all Draw.io events due to security reasons.
+
+https://github.com/owncloud/web/pull/7941
+https://github.com/owncloud/web/issues/7933
--- a/packages/web-app-draw-io/src/App.vue
+++ b/packages/web-app-draw-io/src/App.vue
@@ -64,7 +64,7 @@ export default defineComponent({
   watch: {
     currentFileContext: {
       handler: function () {
-        this.load()
+        this.checkPermissions()
       },
       immediate: true
     }
@@ -74,6 +74,9 @@ export default defineComponent({
     this.fileExtension = this.filePath.split('.').pop()
     window.addEventListener('message', (event) => {
       if (event.data.length > 0) {
+        if (event.origin !== this.config.url) {
+          return
+        }
         const payload = JSON.parse(event.data)
         switch (payload.event) {
           case 'init':
@@ -115,7 +118,7 @@ export default defineComponent({
           message: error,
           modified: false
         }),
-        '*'
+        this.config.url
       )
     },
     async checkPermissions() {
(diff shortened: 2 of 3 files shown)


packages/web-app-draw-io/src/App.vue
```diff
@@ -64,7 +64,7 @@ export default defineComponent({
   watch: {
     currentFileContext: {
       handler: function () {
-        this.load()
+        this.checkPermissions()
```
JammingBen, line 67: No need to call `load()` here, it will be called later anyways. `checkPermissions()` is sufficient.

packages/web-app-draw-io/src/App.vue
```diff
@@ -74,6 +74,9 @@ export default defineComponent({
     this.fileExtension = this.filePath.split('.').pop()
     window.addEventListener('message', (event) => {
       if (event.data.length > 0) {
+        if (event.origin !== this.config.url) {
```
dschmidt, line 77: I don't know much about draw io, but can it be hosted in a subfolder? Do we maybe need to extract the host from the url?

JammingBen: The URL is currently hard-coded to https://embed.diagrams.net.

dschmidt: "Returns the scheme, hostname and port of the document that caused the [onmessage](http://help.dottoro.com/ljjqtjsj.php) event."

packages/web-app-draw-io/src/App.vue
```diff
@@ -47,6 +47,11 @@ export default defineComponent({
       } = this.applicationConfig
       return { url, theme, autosave: autosave ? 1 : 0 }
     },
+    urlHost() {
+      const url = new URL(this.config.url)
+      const urlHost = `${url.protocol}//${url.hostname}`
```
dschmidt, line 52: Does protocol contain a `:`? 
Looks like it's missing here

JammingBen: Yep, see https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol

dschmidt: Surprising, alright then!

dschmidt (approved):

changelog/unreleased/enhancement-origin-event-check
```diff
@@ -0,0 +1,6 @@
+Enhancement: Add origin check to Draw.io events
```
kulmann, line 1: Please change `Enhancement` to `Bugfix`. We handle security patches as bugfixes. :-) (filename respectively)

kulmann (changes requested):

kulmann (approved):

PR #7965, opened 2022-11-14
Up roles defaults

diocas:
Follow up to what we discussed in our last meeting (some weeks ago).
This is not final yet (the tests are failing, and I need to check why, as the default behaviour should stay the same...) and I would appreciate feedback.

The Contributor role is now optional, since we do not support it in CERBox.

It also makes the re-sharing permission optional, as allowing re-sharing doesn't mean we want to make it the default.
We discussed this and you were ok with the change, but now I see that oCIS is not honouring the change... Example: I have re-sharing default to false, in the share dialog I set the Viewer role, the request sends the proper permission code (1), together with the role name, but oCIS replies with the sharing permission (1 + 16). Is it using the role name to define the permissions?
Also, the implementation might need some suggestions...

### DIFF
--- a/packages/web-client/src/helpers/share/role.ts
+++ b/packages/web-client/src/helpers/share/role.ts
@@ -356,30 +356,26 @@ export abstract class PeopleShareRoles {
 }
 
 export abstract class LinkShareRoles {
-  static readonly all = [
-    linkRoleViewerFile,
-    linkRoleViewerFolder,
-    linkRoleContributorFolder,
-    linkRoleEditorFolder,
-    linkRoleUploaderFolder
-  ]
-
   static list(
     isFolder: boolean,
     canEditFile = false,
+    canContribute = false,
     hasAliasLinks = false,
     hasPassword = false
   ): ShareRole[] {
     return [
       ...(hasAliasLinks && !hasPassword ? [linkRoleInternalFile, linkRoleInternalFolder] : []),
-      ...this.all,
+      linkRoleViewerFile,
+      linkRoleViewerFolder,
+      ...(canContribute ? [linkRoleContributorFolder] : []),
+      linkRoleEditorFolder,
+      linkRoleUploaderFolder,
       ...(canEditFile ? [linkRoleEditorFile] : [])
     ].filter((r) => r.folder === isFolder)
   }
 
   static getByBitmask(bitmask: number, isFolder: boolean): ShareRole {
-    return [...this.all, linkRoleEditorFile, linkRoleInternalFile, linkRoleInternalFolder] // Always return all roles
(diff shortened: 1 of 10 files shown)


diocas:
@kulmann I think I solved the tests' issues... Ready for review.

packages/web-client/src/helpers/share/role.ts
```diff
@@ -359,26 +359,33 @@ export abstract class LinkShareRoles {
...
+    canContribute = false,
     hasAliasLinks = false,
     hasPassword = false
   ): ShareRole[] {
     return [
       ...(hasAliasLinks && !hasPassword ? [linkRoleInternalFile, linkRoleInternalFolder] : []),
       ...this.all,
-      ...(canEditFile ? [linkRoleEditorFile] : [])
+      ...(canEditFile ? [linkRoleEditorFile] : []),
+      ...(canContribute ? [linkRoleContributorFolder] : [])
```
kulmann, line 377: With this change the `Contributor` role is at the end of the roles list. Should stay at the same position as before please.
I'd say at this point it doesn't make much sense to keep the `all` array at the top of the class. I'd like to propose that you inline all roles in the `list` function here to keep the correct order. And then in the `getByBitmask` function  just call `list(isFolder, true, true, true, false)` instead of building another array again.

kulmann (changes requested):
One issue about the order of link share roles in the dropdown (see comment). Other than that everything fine, so if you get that fixed I'll merge.

diocas:
Made some changes in the tests. Let me know if it's not ok.

packages/web-client/src/helpers/share/role.ts
```diff
@@ -356,54 +356,50 @@ export abstract class PeopleShareRoles {
...
+      linkRoleEditorFolder,
+      linkRoleUploaderFolder,
       ...(canEditFile ? [linkRoleEditorFile] : [])
     ].filter((r) => r.folder === isFolder)
   }
 
   static getByBitmask(bitmask: number, isFolder: boolean): ShareRole {
-    return [...this.all, linkRoleEditorFile, linkRoleInternalFile, linkRoleInternalFolder] // Always return all roles
-      .find((r) => r.folder === isFolder && r.bitmask(false) === bitmask)
+    return this.list(isFolder, true, true, true, false).find((r) => r.bitmask(false) === bitmask)
```
diocas, line 378: This means 2 finds, but the list is small.. Let me know if you prefer this more optimized.

kulmann: that's fine. I prefer the readability over the performance in this case (like you said, only few items in the list).

kulmann (approved):
👍

PR #7966, opened 2022-11-14
UI improvements

diocas:
This PR fixes a bug: if `hoverableQuickActions: true`, the context menus disappear if the mouse exits them or the table row area.

...and improves the sorting in the shared with view.
In the current behaviour it only sorts by the first sharee name, which means we might have rows with multiple sharees (user + groups) appearing in the middle of other rows that only have the user in the sharees. 
This was particularly visible and bad because we even allow multiple groups to appear in this list (as opposed to ocis which seems to be showing only the first group?).

(like in all PRs, I can add the changelog if you think this is ok to be merged)


kulmann (approved):
Please rebase and run `pnpm lint --fix`

PR #7967, opened 2022-11-14
Text-editor extensions

diocas:
Sets `canBeDefault: true` to ensure that files are not downloaded first.

Adds a new configuration parameter that allows us to open a configurable number of other extensions.


kulmann:
> It looks like `canBeDefault: true` is preventing us from opening .md on codimd by default... @elizavetaRa can you check please? Maybe instead of using that, what we need to do is ensuring that the download action is always the last (so any app takes precedence)

Yes, at the moment an app that handles a specific file extension explicitly (like the `text-editor` that explicitly handles `.txt` and `.md`) would have precedence over the `external` app. The perfect solution is probably to let the user choose which app they want to use for a specific file type.

diocas:
That would be the best, but I would be surprised if we could have that in the short term, or am I wrong? :)

I've asked @elizavetaRa to have a look and suggested we order the actions like following: 1st `canBeDefault`, 2nd all remaining apps, 3rd system actions (like download). Then we can always execute the first one on the list.

To me it makes more sense that we use an app if one is registered. What do you think, @kulmann ?

kulmann:
> That would be the best, but I would be surprised if we could have that in the short term, or am I wrong? :)

Yes, nothing for short term. :trollface: 
One could store the user choice in the browsers localStorage, like we already do for other user choices (e.g. showing/hiding hidden files, showing/hiding file extensions in file names). My team can't work on that at the moment.


> I've asked @elizavetaRa to have a look and suggested we order the actions like following: 1st `canBeDefault`, 2nd all remaining apps, 3rd system actions (like download). Then we can always execute the first one on the list.
> 
> To me it makes more sense that we use an app if one is registered. What do you think, @kulmann ?

The name `canBeDefault` would be pretty misleading then, as it would just be a boolean way of prioritizing certain apps over other apps. But I agree with the approach. If we have an app for a mime type, it doesn't make sense to ignore it when trying to open a file.

kulmann:
> > The name canBeDefault would be pretty misleading then, as it would just be a boolean way of prioritizing certain apps over other apps.
> 
> Isn't this what we're doing already in a way? Should we rename it?

Yes and yes 😄 Do you have a proposal for a better var name?

diocas:
`priority`? Or too generic?

kulmann:
works for me, `priority` or `hasPriority`? (I prefer `hasPriority` but don't care too much).

So as a result: we want to sort all the apps and external-apps in one pool by following criteria:
- `hasPriority=true` first, `hasPriority=false / absent` last
- "dedicated" apps first, external-apps last
- by name

Correct?

elizavetaRa:
> works for me, `priority` or `hasPriority`? (I prefer `hasPriority` but don't care too much).
> 
> So as a result: we want to sort all the apps and external-apps in one pool by following criteria:
> 
> * `hasPriority=true` first, `hasPriority=false / absent` last
> * "dedicated" apps first, external-apps last
> * by name
> 
> Correct?

sounds good! The priority sort should apply after the "external apps last" sort

ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/34152/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/34152/tracing/unstructured-collection-of-testable-space-interactions-alice-2023-3-24-09-49-18.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/34152/tracing/user-login-can-be-managed-in-the-admin-settings-admin-2023-3-24-09-39-05.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/34152/tracing/user-login-can-be-managed-in-the-admin-settings-admin-2023-3-24-09-39-15.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/34152/tracing/user-login-can-be-managed-in-the-admin-settings-alice-2023-3-24-09-39-13.zip 


</p></details>

diocas:
@kulmann please let us know if you prefer to split the PR into text-editor stuff and the more general default extensions handling.

elizavetaRa:
@kulmann and please let us know if you see the use case to implement `hasPriority` for internal editors as well. Right now there are no internal editors that are in conflict with each other to open the same extension, so the internal editor will be taken first unless there is a `hasPriority` for external app. But maybe something is planned for the future?

packages/web-pkg/src/composables/actions/types.ts
```diff
@@ -12,6 +12,7 @@ export interface Action<T = ActionOptions> {
   componentType: 'button' | 'router-link'
   class: string
   canBeDefault?: boolean
```
kulmann, line 14: Hey @diocas and @elizavetaRa 

my intend (see discussion, I think yours as well at some point) was to rephrase `canBeDefault` to `hasPriority`. Meaning, having both in parallel now makes things unnecessarily complex. 🙈 

On top of that: I think the approach with the app provider - define via config which app has priority - is the way to go. An extension developer should never be the one who decides which application has priority over another one. The admin has to decide that, so that's configuration only, not code.

Could you look into realizing the following?
- get rid of `canBeDefault` entirely
- introduce `hasPriority` but only through configuration, nothing hardcoded. For the app provider apps it's already there with the `default_application` attribute on mime types.
- determine the app that gets opened on click in file list (= "default app") by throwing all standalone apps and all app provider apps into one list, sort them by priority, then by internal vs app provider, and pick the first.

What do you think? Does that make sense?

pascalwengerter: @diocas @elizavetaRa anything we can do to support you here? Do you have feedback or further need for clarification?

diocas: @elizavetaRa has been a bit busy. I'll check with her tomorrow when she can fit this in her tasks.

pascalwengerter: @diocas if you're generally fine with what Benedikt suggested I could also take over?

elizavetaRa: @pascalwengerter @kulmann I'll take this issue, will update the pr these days, thanks for waiting!

elizavetaRa: @kulmann  2 questions:
1) what is the purpose of "canBeDefault" in terms priority in the composables/actions files like `web-app-files/src/composables/actions/files/useFileActionsCreateNewFile.ts`?
2) "introduce `hasPriority` but only through configuration, nothing hardcoded."  You mean move setting` hasPriority` of all the extensions, also editors to config.json instead of index.ts of each extension?

kulmann: hey @elizavetaRa - cool that you are working on this, thank you! 🤩 
1. in the case of file action composables, `canBeDefault` doesn't belong there. Could be removed entirely, doesn't make sense.
2. yes, that's exactly what I meant. :-) Please note that providing config for an app is not possible in the `apps` array of the config.json, but only for the `external_apps` array. That one is a bit misleading, name wise, but that's fine for now. Only means that if an admin wants to give an app priority over another one for the same file type they need to define the app in `external_apps`. We need to clean that up (big time), but I wouldn't do that in the scope of this PR. So please just keep the assumption that providing config for an app is only possible in the `external_apps` array of the config.json. That's fine.

tbsbdr:
@elizavetaRa you will take care of this, right?

packages/web-pkg/src/composables/actions/types.ts
elizavetaRa: So proposal 2 for config.json:

  "external_apps": [
    ...,
    {
      "id": "A",
      "path": "/aaa",
      "config": {
        "priorityExtensions": ["txt"]
        ]
      }
    }
  ],

elizavetaRa:
@tbsbdr on it, yes

pascalwengerter:
@elizavetaRa could you add a changelog item? :)

kulmann (approved):
nice, thank you for taking care! 💪

tbsbdr:
relates to https://github.com/owncloud/web/pull/9159/commits/323030a42bc6184850b8f20f3106e827b6443e63

Issue #7956, opened 2022-11-14
upload of big file 10gb takes time to be seen in web after upload completed


bishwasojha:
# Description
When a big file 10gb is uploaded, after the upload is completed, the file is not seen in web. And after some minute then file is seen in web.

### Steps to reproduce
1. upload big file 10gb in web
2. wait for it to upload 
3. After upload is completed, uploaded file isnot seen in web

### Expected behaviour
File should be seen in web after the upload is complete.

### Actual behaviour
File isnot seen in web after the upload is complete. After some minute file is seen in web.

### Environment ownCloud Infinite Scale
**Version:**  owncloud/ocis:2.0.0-rc.1
**Updated from an older Infinite Scale or fresh install:** fresh install
**issue is reproduced with oCIS** : in localstorage, docker container , s3 backend(Minio), nfs

### Client configuration
**Browser:** : Google Chrome Version 106.0.5249.61 (Official Build) (64-bit)
**Operating system:** : Ubuntu 22.04 LTS

# Additional Information
[Screencast from 14-11-22 11:36:25 पूर्वाह्न +0545.webm](https://user-images.githubusercontent.com/33860671/201587965-e80c8db2-29ec-4927-a224-24136bf2b613.webm)

bishwasojha:
- with nfs, big file isnot seen as soon as upload is completed. But seen after a few second.

JammingBen:
I don't quite get the video. You clicked on "pause" just before the upload finished. What happened next? Did the file suddenly occur a minute later?

Is it the same behaviour without clicking "pause"?

bishwasojha:
- This time i clicked `pause`, after progress bar is completed, `Upload complete` notification comes. But file doesnot appear in web.

[Screencast from 15-11-22 03:19:37 अपराह्न +0545.webm](https://user-images.githubusercontent.com/33860671/201884278-234b5688-476c-428a-b4c5-3c87a41d48fb.webm)

JammingBen:
Okay so I can reproduce it when clicking pause right before the upload is finished. This actually makes sense:

Once the file transfer to the backend is completed, the backend still needs to write all data to the storage. During this time window, the progress bar is at 100%, but the overlay still says "1 item uploading". Which is correct, because the upload-process is technically not yet finished.

Pausing or cancelling during this time window will pause/cancel the data transfer, which already finished though. So it just leads to the request being paused/cancelled. The backend still proceeds to write the data to the storage. Once this is done, the file appears in the frontend.

[...]

PR #7974, opened 2022-11-15
Introduce 'upload finished'-state in upload overlay

JammingBen:
## Description
The "upload finished"-state has been introduced to the upload overlay. This state is relevant during the time window when all data has been transferred to the server (= progress bar is at 100%), but the server still needs to write all data to the storage.

The "cancel"- and "pause"-actions are disabled during the "upload finished"-state as the data transfer is technically finished. Previously, when pausing and resuming when being in this state, the upload would be marked as successful instantly, despite the server still writing to the storage.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7956

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/components/UploadInfo.vue
+++ b/packages/web-runtime/src/components/UploadInfo.vue
@@ -53,7 +53,7 @@
           <oc-icon name="restart" fill-type="line" />
         </oc-button>
         <oc-button
-          v-if="runningUploads && uploadsPausable && !inPreparation"
+          v-if="runningUploads && uploadsPausable && !inPreparation && !uploadsWritingToStorage"
           id="pause-upload-info-btn"
           v-oc-tooltip="uploadsPaused ? $gettext('Resume upload') : $gettext('Pause upload')"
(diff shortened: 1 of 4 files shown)


packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -162,6 +162,7 @@ export default defineComponent({
     totalProgress: 0, // current uploads progress (0-100)
     uploadsPaused: false, // all uploads paused?
     uploadsCancelled: false, // all uploads cancelled?
+    uploadsWritingToStorage: false, // uploads transferred but still need to be written to storage
```
kulmann, line 165: I'd prefer a more generic variable name that doesn't imply backend knowledge. What do you think about `inFinalization`?

```suggestion
    inFinalization: false, // uploads transferred but still need to be written to storage
```

kulmann (approved):
awesome 🤩

JanAckermann (approved):

Issue #7980, opened 2022-11-16
version of file cannot be restored by shared user


bishwasojha:
### Descriptiop of bug
A file is shared to a user with editor role. The old version of file cannot be restored and download by a shared user.

### Steps to reproduce
1. create a file and write content
2. again add content to a file and make new version 
3. share file to a user with editor role
4. user accepts a file and check version of file
5. click on restore  
6. old version cannot be restored. 
7. click on download; it throws error 

**Note: This issue isnot reproducable if file is within a folder**

### Expected behaviour 
Old version of file can be restore

## Environment
- server : owncloud/ocis:2.0.0-rc.1

## Client configuration
- Browser: : Google Chrome Version 106.0.5249.61 (Official Build) (64-bit)
- Operating system: : Ubuntu 22.04 LTS

## Additional Information

[Screencast from 16-11-22 03:47:55 अपराह्न +0545.webm](https://user-images.githubusercontent.com/33860671/202164143-7925bb6e-0910-40f4-a448-4ec020d2a809.webm)

kulmann:
@bishwasojha does it work with oc10 for single file shares? 🤔

bishwasojha:
@kulmann it doesnot work in oc10 with new design. But it works fine with classic web of oc10. Seems like a problem with new design of web.

# in classic web

[Screencast from 17-11-22 10:01:42 पूर्वाह्न +0545.webm](https://user-images.githubusercontent.com/33860671/202354071-157bbbda-1dc8-4f22-8e9d-4b37efb2e67c.webm)

# in new web

[Screencast from 17-11-22 10:04:23 पूर्वाह्न +0545.webm](https://user-images.githubusercontent.com/33860671/202354448-9f57c3b9-a449-4992-ab9e-08e513339618.webm)

JammingBen:
https://github.com/owncloud/web/pull/7989 fixes downloading and reverting versions on the "Shared with me"-page, which was the issue here. However, there are still issues with oCIS: https://github.com/owncloud/ocis/issues/3868. These seem to be on the sever side as far as I can see.

JammingBen:
Closing as the issue has been solved in Web and oCIS 👍

PR #7989, opened 2022-11-17
Fix file versions on 'Shared with me'-page

JammingBen:
## Description
Downloading and reverting versions on the "Shared with me"-page has been fixed.

## Related Issue

- https://github.com/owncloud/web/issues/7980

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
+++ b/packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
@@ -93,7 +93,7 @@ export default defineComponent({
       this.loading = false
     },
     async revertVersion(file) {
-      const { fileId, path } = this.highlightedFile
+      const { fileId, id, path } = this.highlightedFile
       await this.$client.fileVersions.restoreFileVersion(fileId, this.currentVersionId(file), path)
       const resource = await (this.$clientService.webdav as WebDAV).getFileInfo(
         this.space,
@@ -103,7 +103,7 @@ export default defineComponent({
       const fieldsToUpdate = ['size', 'mdate']
       for (const field of fieldsToUpdate) {
         if (this.highlightedFile[field]) {
-          this.UPDATE_RESOURCE_FIELD({ id: fileId, field, value: resource[field] })
+          this.UPDATE_RESOURCE_FIELD({ id, field, value: resource[field] })
         }
       }
 
--- a/packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
+++ b/packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
@@ -52,10 +52,10 @@ import { mapActions, mapGetters, mapMutations } from 'vuex'
 import { DavProperty } from 'web-client/src/webdav/constants'
 import { formatRelativeDateFromHTTP, formatFileSize } from 'web-pkg/src/helpers'
 import { WebDAV } from 'web-client/src/webdav'
-import { inject } from '@vue/composition-api'
+import { defineComponent, inject } from '@vue/composition-api'
 import { SpaceResource } from 'web-client/src/helpers'
 
-export default {
+export default defineComponent({
   name: 'FileVersions',
   setup() {
     return {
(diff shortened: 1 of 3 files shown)


JammingBen:
Fyi: I don't want to go crazy on the unit tests here because they changed quite heavily on master.

butonic:
it seems the web ui is making a request that uses `:` as the seperator for reference properties:
```
https://cloud.ocis.test/remote.php/dav/meta/1284d238-aa92-42ce-bdc4-0b0000009157%3Af7fbf8c8-139b-4376-b307-cf0a8c2d0d9c%3A12ac22bf-e11c-4a35-ae7d-179f7d9638d9/v/2f7a4e7b-3a93-4690-b66d-a172db22311f.REV.2022-11-18T10%3A42%3A14.151912819Z
```

that url should be taken literally from the href:
```
<d:href>/remote.php/dav/meta/1284d238-aa92-42ce-bdc4-0b0000009157$f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c%212f7a4e7b-3a93-4690-b66d-a172db22311f/v/2f7a4e7b-3a93-4690-b66d-a172db22311f.REV.2022-11-18T10:42:16.061925767Z</d:href>
```
that uses `$` and `%21` (= `!`)

downloading revisions also requires https://github.com/cs3org/reva/pull/3473

JammingBen:
`it seems the web ui is making a request that uses : as the seperator for reference properties:`

This should be fixed with this PR. I'll test it when https://github.com/cs3org/reva/pull/3473 is merged.

packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
```diff
@@ -92,21 +92,22 @@ export default {
...
+      await this.$client.fileVersions.restoreFileVersion(fileId, this.currentVersionId(file), path)
+      const resource = await (this.$clientService.webdav as WebDAV).getFileInfo(
+        this.space,
+        this.highlightedFile
+      )
+
+      const fieldsToUpdate = ['size', 'mdate']
+      for (const field of fieldsToUpdate) {
+        if (this.highlightedFile[field]) {
+          this.UPDATE_RESOURCE_FIELD({ id: fileId, field, value: resource[field] })
```
kulmann, line 106: When being on e.g. the `Shared with me` page the store holds `sharedResource` entities. Since the `UPDATE_SOURCE_FIELD` mutation identifies entities by the `id` prop it looks wrong to me that you're using the `fileId` here. How does it actually work at all if you're not in a personal- or project-space (where id=fileId)? 😅

kulmann (changes requested):

packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
JammingBen: Good catch, the need the `id` here indeed. I didn't notice it because the share tables don't have the columns "size" and "mdate" anyway.

packages/web-runtime/src/plugins/web.js
```diff
@@ -29,7 +29,7 @@ export default {
             if (version === null) {
               url = `${this.$client.helpers._davPath}${file.webDavPath}`
             } else {
-              url = this.$client.fileVersions.getFileVersionUrl(file.id, version)
+              url = this.$client.fileVersions.getFileVersionUrl(file.fileId, version)
```
dschmidt, line 32: 🙈

packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
```diff
@@ -52,10 +52,10 @@ import { mapActions, mapGetters, mapMutations } from 'vuex'
 import { DavProperty } from 'web-client/src/webdav/constants'
 import { formatRelativeDateFromHTTP, formatFileSize } from 'web-pkg/src/helpers'
 import { WebDAV } from 'web-client/src/webdav'
-import { inject } from '@vue/composition-api'
+import { defineComponent, inject } from '@vue/composition-api'
```
dschmidt, line 55: We need to be careful when merging to master later

dschmidt (approved):

packages/web-runtime/src/plugins/web.js
kulmann: 🙈

kulmann (approved):

PR #8000, opened 2022-11-18
Update README.md

mmattel:
## Description
Run the `pnmp install` command to initialize or update dependencies.

## Motivation and Context
For beginners 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ We currently publish a new release every couple of weeks, strictly following [se
 ## Documentation
 The documentation is an important part of this project and can be found in the [docs](https://github.com/owncloud/web/tree/master/docs) folder, or in its deployed [live version](https://owncloud.dev/clients/web).
 
+## Dependencies
+
+To initialize or update dependencies, run `pnpm install`.
+
 ## Contribution
 Contribution in the form of bug reports, user feedback or actual code is always welcome! We do have a [contribution guide](https://github.com/owncloud/web/blob/master/.github/CONTRIBUTING.md), actively use the [good-first-issue](https://github.com/owncloud/web/issues?q=is%3Aopen+is%3Aissue+label%3ATopic%3Agood-first-issue) label and try to feedback on issues and pull requests in a timely manner. There is also a [setup guide](https://owncloud.dev/clients/web/getting-started/) for building and running `web` locally.
 


README.md
```diff
@@ -77,6 +77,10 @@ We currently publish a new release every couple of weeks, strictly following [se
 ## Documentation
 The documentation is an important part of this project and can be found in the [docs](https://github.com/owncloud/web/tree/master/docs) folder, or in its deployed [live version](https://owncloud.dev/clients/web).
 
+## Dependencies
+
+To initialize or update dependencies, run `pnpm install`.
+
 ## Contribution
 Contribution in the form of bug reports, user feedback or actual code is always welcome! We do have a [contribution guide](https://github.com/owncloud/web/blob/master/.github/CONTRIBUTING.md), actively use the [good-first-issue](https://github.com/owncloud/web/issues?q=is%3Aopen+is%3Aissue+label%3ATopic%3Agood-first-issue) label and try to feedback on issues and pull requests in a timely manner. There is also a [setup guide](https://owncloud.dev/clients/web/getting-started/) for building and running `web` locally.
```
pascalwengerter, line 85: @mmattel that information is already available in the linked setup guide (albeit I don't know how up-do-date it is after the switch to `pnpm`)

dschmidt: Yes, there is https://owncloud.dev/clients/web/getting-started/#docker and https://owncloud.dev/clients/web/building/ which partially overlap and which could certainly be improved, but they are up to date and are much more complete than this new isolated info. Alright, I've got dependencies installed - now what? 
Also I don't like this additional section  as the full documentation is linked from another, that's pretty confusing to me.

mmattel: Hey, instead of complaining about this PR, you could think about _why_ this PR was made. I just wanted to improve some text strings (other PR's) but the linter complained unexpectedly. To fix this, I needed to run a local command `pnpm lint --fix` but this is only possible when installing dependencies. I do not to build web. Now, I see a link named `Documentation` pointing to a folder contining a bunch of other documents. First of all they are not directly linked from readme.md and second you have to parse the documents for finding the right info. IMHO, a second readme for preparing the environment would solve the problem, makes this PR obsolete and clearly highlights what is necessary to prepare the environment and gets that info into one point avoiding redundancies. Just my 2c.

dschmidt: Maybe we can agree first, that reviewing is not complaining - otherwise I don't see any basis for discussion.

tbsbdr: > Yes, there is https://owncloud.dev/clients/web/getting-started/#docker and https://owncloud.dev/clients/web/building/ which partially overlap and which could certainly be improved, but they are up to date and are much more complete than this new isolated info. Alright, I've got dependencies installed - now what?
> Also I don't like this additional section as the full documentation is linked from another, that's pretty confusing to me.

- agree
- lets go with #8014

tbsbdr:
won't change
lets go with #8014

PR #8002, opened 2022-11-18
[tests-only] Add unit tests for the text editor

JammingBen:
This also includes a mock for the `useAppDefaults` composable.

## Types of changes

- [ ] Technical debt
- [x] Tests

### DIFF
--- a/packages/web-app-text-editor/tests/unit/app.spec.ts
+++ b/packages/web-app-text-editor/tests/unit/app.spec.ts
@@ -30,10 +30,19 @@ describe('Text editor app', () => {
       expect(wrapper.find('oc-textarea-stub').exists()).toBeTruthy()
     })
   })
-  it('shows the preview', async () => {
-    const { wrapper } = getWrapper({ fileName: 'markdown.md' })
-    await wrapper.vm.loadFileTask.last
-    expect(wrapper.find('#text-editor-preview').exists()).toBeTruthy()
+  describe('preview', () => {
+    it.each([
+      { fileExtension: 'txt', showPreview: false },
+      { fileExtension: 'js', showPreview: false },
+      { fileExtension: 'php', showPreview: false },
+      { fileExtension: 'json', showPreview: false },
+      { fileExtension: 'xml', showPreview: false },
+      { fileExtension: 'md', showPreview: true }
+    ])('shows only for supported file types', async (data) => {
+      const { wrapper } = getWrapper({ fileName: `file.${data.fileExtension}` })
+      await wrapper.vm.loadFileTask.last
+      expect(wrapper.find('#text-editor-preview').exists()).toBe(data.showPreview)
+    })
   })
 })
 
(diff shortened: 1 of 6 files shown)


JammingBen:
SonarCloud fails because of 0.0% condition coverage... makes sense 🤡 
image

dschmidt (approved):
I like, can we add this to stable instead?

packages/web-app-text-editor/tests/unit/app.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+function getWrapper({ fileName = 'someFile.txt' }: { fileName?: string } = {}) {
+  jest.mocked(useAppDefaults).mockImplementation(() =>
+    useAppDefaultsMock({
+      currentFileContext: ref(mockDeep<FileContext>({ path: fileName }))
+    })
+  )
+  const defaultMocks = { ...defaultComponentMocks() }
+  const storeOptions = { ...defaultStoreMockOptions }
+  const localVue = defaultLocalVue()
+  const store = createStore(Vuex.Store, storeOptions)
```
dschmidt, line 49: Out of scope for this PR, but:

I don't like that you use options from test helpers but have to add `vuex-extensions` to the package dependencies, I would like to see this handled in `web-test-helpers` completely.
We should really throw some time at unit tests at some point, unify the setup and clean up stuff like this.

packages/web-test-helpers/src/mocks/useAppDefaultsMock.ts
```diff
@@ -0,0 +1,29 @@
+import { mockDeep } from 'jest-mock-extended'
+import { ref } from '@vue/composition-api'
+import { Resource } from 'web-client/src'
+import { FileResource } from 'web-client/src/helpers'
+import { GetFileContentsResponse } from 'web-client/src/webdav/getFileContents'
+import { AppConfigObject, FileContext, useAppDefaults } from 'web-pkg/src'
+
+export const useAppDefaultsMock = (
```
kulmann, line 8: Still amazed by the mocks with the partial type for options 😎

dschmidt (approved):

packages/web-app-text-editor/tests/unit/app.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+      expect(wrapper.find('oc-spinner-stub').exists()).toBeTruthy()
+    })
+    it('shows the editor after loading', async () => {
+      const { wrapper } = getWrapper()
+      await wrapper.vm.loadFileTask.last
+      expect(wrapper.find('oc-spinner-stub').exists()).toBeFalsy()
+      expect(wrapper.find('oc-textarea-stub').exists()).toBeTruthy()
+    })
+  })
+  it('shows the preview', async () => {
```
kulmann, line 33: Could you make this a `describe` block about previews and then check that different file types show/don't show the preview? The preview is only there for markdown. E.g. a `.txt` file will not have a preview.

kulmann (changes requested):

kulmann (approved):

Issue #8003, opened 2022-11-19
Public link of folder with `Uploader`role, public is able to create `(file | folder)` in that folder


bishwasojha:
### Description of bug
when the public link of folder with `Editor` role is changed to `Uploader` , public is able to create file | folder in that folder.

### Steps to reproduce
1. Create a Folder
2. Add Public Link with role `editor`
3. Open the Link in incognito 
4. Change the role `editor` to `Uploader`
5. Refresh the link

Then, in the shared folder, a new file | folder can be created. 

### Expected behaviour
with `Uploader` role , the UI mustnot have option to create file | folder.

### Actual behaviour
Public can create file|folder.
### Environment general
**Operating system**:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
```

**Backend (ownCloud Core or Infinite Scale)**:oCIS

### Environment ownCloud Infinite Scale
**Version:**  owncloud/ocis: 2.0.0-rc.1 

### Client configuration
**Browser:**  Google Chrome Version 106.0.5249.61 (Official Build) (64-bit)

**Operating system:** Ubuntu 22.04 LTS

### Additional Information
[Screencast from 21-11-22 09:52:17 पूर्वाह्न +0545.webm](https://user-images.githubusercontent.com/33860671/202963730-9a641d78-186a-495e-9352-042bb91b6576.webm)

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8006.

PR #8006, opened 2022-11-21
[full-ci] Fix public link loading on role change

JammingBen:
## Description
The loading and resolving of public links when their respective roles have changed has been fixed.
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8003

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/views/spaces/DriveResolver.vue
+++ b/packages/web-app-files/src/views/spaces/DriveResolver.vue
@@ -15,10 +15,19 @@ import DriveRedirect from './DriveRedirect.vue'
 import GenericSpace from './GenericSpace.vue'
 import GenericTrash from './GenericTrash.vue'
 
-import { defineComponent } from '@vue/composition-api'
-import { useDriveResolver, useRouteParam, useStore } from 'web-pkg/src/composables'
+import { defineComponent, onMounted, unref } from '@vue/composition-api'
+import {
+  useClientService,
+  useDriveResolver,
+  useRouteParam,
+  useRouter,
+  useStore
+} from 'web-pkg/src/composables'
 import { useActiveLocation } from '../../composables'
 import { isLocationSpacesActive, isLocationTrashActive } from '../../router'
+import { isPublicSpaceResource, PublicSpaceResource, SpaceResource } from 'web-client/src/helpers'
+import { locationPublicUpload } from 'web-app-files/src/router/public'
+import { linkRoleUploaderFolder } from 'web-client/src/helpers/share'
 
 export default defineComponent({
   components: {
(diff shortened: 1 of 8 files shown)


ownclouders:
Results for <strong>oCISSharingPublic2</strong> https://drone.owncloud.com/owncloud/web/30019/68/1

<p>


### webUISharingPublicManagement-sharingPublicSession_feature-L35.png
  

### webUISharingPublicManagement-sharingPublicSession_feature-L48.png
  


</p></details>

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -321,7 +323,23 @@ export default defineComponent({
...
+        publicSpace = this.space
+      }
+
+      if (linkRoleUploaderFolder.bitmask(false) === publicSpace.publicLinkPermission) {
+        return this.$router.push({
+          name: locationPublicUpload.name,
+          params: { token: this.space.id }
+        })
+      }
+    }
```
JammingBen, line 341: Not the biggest fan of putting such logic in the view directly... But the public link is already resolved here, so I don't see another good place to put it.

kulmann: What about putting it in the `DriveResolver.vue`? 🤔

JammingBen: This would work as well I think, but do you think it's a better fit though? It's "just another view". Also, the logic only applies to the `GenericSpace` view, not to `GenericTrash` and `DriveRedirect`.

JammingBen: I moved it to the DriveResolver. I agree it's a better fit than the already resolved space.

packages/web-pkg/src/composables/driveResolver/useDriveResolver.ts
```diff
@@ -16,7 +16,13 @@ interface DriveResolverOptions {
   driveAliasAndItem?: Ref<string>
 }
 
-export const useDriveResolver = (options: DriveResolverOptions = {}) => {
+interface ResolvedDrive {
```
kulmann, line 19: I'd prefer
```suggestion
interface DriveResolverResult {
```
as we've already used that naming scheme a lot (`...Options` and `...Result`).

packages/web-app-files/src/views/spaces/DriveResolver.vue
```diff
@@ -28,11 +37,39 @@ export default defineComponent({
...
+        return publicSpace
+      } catch (e) {
+        return space
+      }
+    }
+
+    onMounted(async () => {
+      const space = unref(resolvedDrive.space)
+      if (space && isPublicSpaceResource(space)) {
+        let publicSpace = (await getSpaceResource()) as PublicSpaceResource
```
dschmidt, line 60: What happens if this request fails for whatever reason?

packages/web-app-files/tests/unit/views/spaces/DriveResolver.spec.ts
```diff
@@ -28,6 +33,24 @@ describe('DriveResolver view', () => {
...
+    clientService.webdav.getFileInfo.mockResolvedValue(
+      mockDeep<PublicSpaceResource>({ publicLinkPermission: SharePermissionBit.Create })
+    )
+    const { wrapper, mocks } = getMountedWrapper({
+      space,
+      mocks: { $clientService: clientService }
+    })
+
+    await wrapper.vm.$nextTick()
+    await wrapper.vm.$nextTick()
```
dschmidt, line 48: I'm afraid we need this? but it's ugly :(

dschmidt (approved):

packages/web-app-files/src/views/spaces/DriveResolver.vue
JammingBen: The space from `useDriveResolver()` will be used as default. It won't have `publicLinkPermission` though, so the following if-condition will never resolve to `true`.

packages/web-app-files/tests/unit/views/spaces/DriveResolver.spec.ts
JammingBen: Yes, I didn't find a way around it unfortunately :/

packages/web-app-files/src/views/spaces/DriveResolver.vue
dschmidt: Yeah, my point is: should we show some kind of error?

JammingBen: Hmm IMO we don't need a user facing error here, as a user would most likely not know what to do with it anyway. But a console error would be good indeed. I'll wait for the review from Benedikt and add it after.

kulmann: Hm... I think it would make more sense to do the `getFileInfo` before inserting the `PublicSpaceResource` into the store, i.e. inside the `store.watch` of the `web-runtime/src/index.ts`.

kulmann (changes requested):

kulmann (approved):

packages/web-app-files/src/views/spaces/DriveResolver.vue
kulmann: for the record: this was a can of worms. 😅

Issue #7670, opened 2022-09-21
Public link does not require a password after adding it 


ScharfViktor:
ownCloud Web UI 5.7.0 

Steps:

- admin creates public link to the folder (upload permissions)
- public open link
- admin adds password
- public refresh the opened link

Expected: a window appears in which the public must enter the password

Actual: no requiring pass

if public tries to upload files-> 401 (correct)

https://user-images.githubusercontent.com/84779829/191605907-a8af9ae4-a9bb-4354-876a-3c890f8a6ce1.mov

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8007

PR #8007, opened 2022-11-21
Fix missing password form on public drop page

JammingBen:
## Description
We've fixed a bug where the password form on a public drop page would not show after setting a required password.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7670

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/views/FilesDrop.vue
+++ b/packages/web-app-files/src/views/FilesDrop.vue
@@ -47,6 +47,7 @@ import { usePublicLinkPassword, useStore } from 'web-pkg/src/composables'
 import { eventBus } from 'web-pkg/src/services/eventBus'
 import { linkRoleUploaderFolder } from 'web-client/src/helpers/share'
 import { defineComponent } from '@vue/composition-api'
+import { authService } from 'web-runtime/src/services/auth'
 
 export default defineComponent({
   components: {
@@ -151,7 +152,7 @@ export default defineComponent({
         .catch((error) => {
           // likely missing password, redirect to public link password prompt
           if (error.statusCode === 401) {
-            return this.$authService.handleAuthError(this.$router.currentRoute)
+            return authService.handleAuthError(this.$router.currentRoute)
           }
           console.error(error)
           this.errorMessage = error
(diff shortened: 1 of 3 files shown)


dschmidt (approved):
The change is okay, I'm just a bit confused, why the earlier version did not work ... :-\

kulmann (approved):

packages/web-app-files/src/views/FilesDrop.vue
```diff
@@ -151,7 +152,7 @@ export default defineComponent({
         .catch((error) => {
           // likely missing password, redirect to public link password prompt
           if (error.statusCode === 401) {
-            return this.$authService.handleAuthError(this.$router.currentRoute)
```
kulmann, line 154: Is the authService not available anymore as `this.$authService`? 🤔

JammingBen: Working with `prototype` seems to be the right way though: https://v2.vuejs.org/v2/cookbook/adding-instance-properties.html. I'll add it.

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -322,6 +322,7 @@ export const announceAuthService = ({
   router: VueRouter
 }): void => {
   authService.initialize(configurationManager, clientService, store, router)
+  vue.prototype.$authService = authService
   set(vue, '$authService', authService)
```
kulmann, line 326: I guess you can remove this line then.

JammingBen: Thought so too, but it gives me: `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'initializeContext')` 🤔  It's the same for `$permissionManager`.

JammingBen: Because of this here: https://github.com/owncloud/web/blob/fix-missing-public-password-form/packages/web-runtime/src/router/setupAuthGuard.ts#L14

dschmidt: What a mess. You're a hero!

pascalwengerter: 👏

Issue #7946, opened 2022-11-10
Recipients disappear after re-sharing to the group


ScharfViktor:
ownCloud Web UI 6.0.0-rc.6
Infinite Scale 2.0.0-rc.1+736f513cf Community -local

Steps:

- admin creates file and shares to marie
- marie accepts file and re-shares it to the users and the group 

Actual: after re-sharing to the group again, all recipients or public links disappear, but still remain after reloading the page

https://user-images.githubusercontent.com/84779829/201134209-e8df26b8-2a62-4616-a90d-ac0aba9ae550.mov

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8009.

PR #8009, opened 2022-11-22
Prevent shares from disappearing after sharing with groups

JammingBen:
## Description
The disappearing of shares after sharing with groups on the "Shared with me"-page has been fixed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7946

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-app-files/src/store/actions.ts
```diff
@@ -236,49 +236,21 @@ export default {
...
-          context.commit(
-            'CURRENT_FILE_OUTGOING_SHARES_UPSERT',
-            buildCollaboratorShare(
-              share.shareInfo,
-              context.getters.highlightedFile,
-              allowSharePermissions(context.rootGetters)
-            )
-          )
-          context.dispatch('updateCurrentFileShareTypes')
-          context.dispatch('loadIndicators', { client, currentFolder: path })
```
JammingBen, line 257: Fyi: the issue was the `storageId` here.

Since success- and error-handling after sharing with users and groups is identical, I decided to reduce code duplication (which probably "caused" the issue in the first place).

packages/web-app-files/src/store/actions.ts
```diff
@@ -236,49 +236,21 @@ export default {
...
-            {
-              title: $gettext('Error while sharing.'),
-              desc: e,
-              status: 'danger'
-            },
-            { root: true }
-          )
-        })
-      return
+    const isGroupShare = shareType === ShareTypes.group.value
```
kulmann, line 239: Could you add unit tests for this? At least checking that the correct (mocked) function of the sdk is called would be good.

Why can you omit the mutation completely?

JammingBen: Sure 👍  Which mutation do you mean?

kulmann: A few lines below there was a `CURRENT_FILE_OUTGOING_SHARES_UPSERT` mutation.

JammingBen: It's still present (a few lines below), I just removed the code duplication.

packages/web-app-files/tests/unit/store/actions.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  commit: jest.fn(),
+  dispatch: jest.fn(),
+  getters: {},
+  rootGetters: {}
+}
+
+describe('vuex store actions', () => {
+  beforeEach(() => {
+    // because of the $gettext warnings
+    jest.spyOn(console, 'warn').mockImplementation(() => undefined)
```
dschmidt, line 18: What gettext warnings? Can we fix them instead?

dschmidt (approved):

packages/web-app-files/tests/unit/store/actions.spec.ts
JammingBen: `No translations found for en`.

The actions use `$getttext` from here: https://github.com/owncloud/web/blob/master/packages/web-app-files/src/gettext.js. Honestly, I'm not sure if this is the right way to go in general. If I remember correctly, we want to get rid of this file anyway in the future. Hence I figured to not spend more time finding another solution 😅

dschmidt: Yeah, I'm just not a great fan of hiding console output ... I've spent way too much time in my life looking for missing output, when I wanted to debug something and probably even worse I'm afraid we might miss something more interesting. :-/

I've approved the PR, so if you decide it's not worth it, then just merge it

JammingBen: I changed the way how it is mocked to `jest.mock('../../../src/gettext')`. This way it only affects `$gettext` and not all console warnings :)

packages/web-app-files/src/store/actions.ts
kulmann: Whoops. Wanted to say "didn't see it in the diff", but it actually is the last visible line 😆

kulmann (approved):

Issue #8011, opened 2022-11-22
Assigning Roles stopped working


micbar:
### Steps to reproduce
1. Set up oCIS with LDAP à la https://ocis.ocis-ldap.latest.owncloud.works
2. Go to user management app
3. Change Role of user einstein from user to space-admin

### Expected behaviour
Role assignment should work and be confirmed

### Actual behaviour

The role assignment creates an error

![1669122294539](https://user-images.githubusercontent.com/4378513/203321200-5e42a85f-f1de-4b09-b9f8-f8bd5c31427e.png)

#### Api Request that fails

```sh
curl 'https://ocis.ocis-ldap.latest.owncloud.works/graph/v1.0/users/4c510ada-c86b-4815-8820-42cdf82c3d51' \
  -X 'PATCH' \
  --data-raw '{"displayName":"Albert Einstein","id":"4c510ada-c86b-4815-8820-42cdf82c3d51","mail":"einstein@example.org","onPremisesSamAccountName":"einstein","memberOf":[],"passwordProfile":{"password":""}}' \
  --compressed
```

#### Response

Status: `500`

```json
{"error":{"code":"notAllowed","innererror":{"date":"2022-11-22T13:05:08Z","request-id":"6ab12c33-2630-49ce-97dd-a2d04adf7326"},"message":"server is configured read-only"}}

```

This seems wrong, the correct API call should be

```sh
curl 'https://ocis.ocis-ldap.latest.owncloud.works//api/v0/settings/assignments-add' \
  -H 'Accept: application/json, text/plain, */*' \
  --data-raw '{"account_uuid":"f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c","role_id":"d7beeea8-8ff4-406b-8fb6-ab2dd81e6b11"}' \
  --compressed \
  --insecure
```

### Environment general
**Operating system**:

**Backend (ownCloud Core or Infinite Scale)**:

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

### Environment ownCloud Infinite Scale
**Version:**

**Updated from an older Infinite Scale or fresh install:**

**Where did you install Infinite Scale from:**

### Environment ownCloud Core
**Web server:**

**Database:**

**PHP version:**

**Version:** (see ownCloud admin page)

**Updated from an older ownCloud or fresh install:**

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
[...]

PR #8013, opened 2022-11-22
Prevent unnecessary request when saving a user

JammingBen:
## Description
We've fixed a bug where changing the role of a user without changing any other data would cause an unnecessary request. Also, error handling for saving the user role and all other data has been separated. That means, in case one save request fails, the other one still proceeds.

Note that the error messages should be more clear, but we don't add new translations on the `stable-6.0` branch.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8011

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-app-user-management/src/views/Users.vue
```diff
@@ -415,42 +415,54 @@ export default defineComponent({
...
-          if (editUser.id === this.currentUser.uuid) {
-            // Load current user quota
-            this.UPDATE_SPACE_FIELD({
-              id: editUser.drive.id,
-              field: 'spaceQuota',
-              value: updateDriveResponse.data.quota
-            })
-          }
+      if (!isEqual(actualUser.drive, editUser.drive)) {
+        const updateDriveResponse = await this.graphClient.drives.updateDrive(
```
kulmann, line 449: Could you wrap this in a try catch block as well?

kulmann (approved):

kulmann (approved):

Issue #7962, opened 2022-11-14
Overwriting file in not possible on web if there is enough quota


ScharfViktor:
ownCloud Web UI 6.0.0-rc.6
Infinite Scale 2.0.0-rc.1+736f513cf Community -local

Steps see this api test: 
- https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiSpaces/quota.feature#L56-L60
- create project space with quota = 10 `curl -k -X POST 'https://localhost:9200/graph/v1.0/drives' -u admin:admin -d '{"name":"marketing", "quota": {"total": 10}}' -vk`
- upload file 9 Bytes (if you created file- you need reload page #7959 )
- change this file on your PC to 5 bytes and upload it again

Actual: web checking
<img width="318" alt="Screenshot 2022-11-14 at 13 21 58" src="https://user-images.githubusercontent.com/84779829/201662516-ecfe5734-a780-403b-904b-6b4e7d9752ff.png">
but it is possible. 

Expect: The web check appears after clicking the "Keep Both" button. If we choose "merging" - it should be possible to overwrite

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8015.

PR #8015, opened 2022-11-23
Fix quota check when replacing files

JammingBen:
## Description
An issue with the quota check when replacing an existing file while having enough quota has been fixed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/7962

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/helpers/resource/actions/upload.ts
+++ b/packages/web-app-files/src/helpers/resource/actions/upload.ts
@@ -210,6 +210,13 @@ export class ResourcesUpload extends ConflictDialog {
     const uploadSizeSpaceMapping = uppyResources.reduce((acc, uppyResource) => {
       let targetUploadSpace
 
+      const existingFile = this.currentFiles.find(
+        (c) => !uppyResource.meta.relativeFolder && c.name === uppyResource.name
+      )
+      if (existingFile?.size >= uppyResource.data.size) {
+        return acc
+      }
+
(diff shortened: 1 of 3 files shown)


kulmann (approved):

packages/web-app-files/src/helpers/resource/actions/upload.ts
```diff
@@ -210,6 +210,13 @@ export class ResourcesUpload extends ConflictDialog {
     const uploadSizeSpaceMapping = uppyResources.reduce((acc, uppyResource) => {
       let targetUploadSpace
 
+      const existingFile = this.currentFiles.find(
+        (c) => !uppyResource.meta.relativeFolder && c.name === uppyResource.name
+      )
+      if (existingFile?.size >= uppyResource.data.size) {
+        return acc
+      }
```
dschmidt, line 218: Hmm - wouldn't it be more accurate to substract the difference from acc in this case? 

Can't we do something like this?
```javascript
if(existingFile?.size) {
  return acc - existingFile.size + uppyResource.data.size
}
```

dschmidt (approved):

packages/web-app-files/src/helpers/resource/actions/upload.ts
JammingBen: Ohh yes, makes much more sense actually! Thx

PR #8024, opened 2022-11-25
Add new translations after merging stable-6.0

JammingBen:
Adds new translations after https://github.com/owncloud/web/pull/8023 has been merged to master.

### DIFF
--- a/packages/web-app-text-editor/src/App.vue
+++ b/packages/web-app-text-editor/src/App.vue
@@ -152,18 +152,14 @@ export default defineComponent({
             )
             if (space) {
               errorPopup(
-                // FIXME: translation
-                // $gettextInterpolate(
-                //   'There is not enough quota on "%{spaceName}" to save this file',
-                //   { spaceName: space.name }
-                // )
-                $gettext('Error when contacting the server')
+                $gettextInterpolate(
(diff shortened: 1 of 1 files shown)


packages/web-app-text-editor/src/App.vue
```diff
@@ -152,18 +152,14 @@ export default defineComponent({
...
             if (space) {
               errorPopup(
-                // FIXME: translation
-                // $gettextInterpolate(
-                //   'There is not enough quota on "%{spaceName}" to save this file',
-                //   { spaceName: space.name }
-                // )
-                $gettext('Error when contacting the server')
+                $gettextInterpolate(
+                  'There is not enough quota on "%{spaceName}" to save this file',
```
kulmann, line 156:
```suggestion
                  $gettext('There is not enough quota on "%{spaceName}" to save this file'),
```

`$gettextInterpolate` doesn't do any translations. It only inlines variables. You still need to wrap the lang string in a `$gettext` call.

kulmann (changes requested):

packages/web-app-text-editor/src/App.vue
JammingBen: Oh boy, time for the weekend to come 😬

kulmann (approved):

PR #8026, opened 2022-11-25
Fix text editor appearance

JammingBen:
## Description
The text editor now looks normal again after its appearance had changed due to the inclusion of the ODS in the web repository.

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-text-editor-appearance
+++ b/changelog/unreleased/bugfix-text-editor-appearance
@@ -1,0 +1,5 @@
+Bugfix: Text editor appearance
+
+The text editor now looks normal again after its appearance had changed due to the inclusion of the ODS in the web repository.
+
+https://github.com/owncloud/web/pull/8026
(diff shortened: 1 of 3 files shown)


kulmann (changes requested):
Vertical sizing seems to be broken, see screenshot (that's the lower left corner).

Screenshot 2022-11-25 at 16 24 13

JammingBen:
@kulmann Do you mean there should be a margin-bottom because of the rounded corners? The inner container doesn't have a padding, so the textarea takes its full height - that was the same before.

kulmann:
> @kulmann Do you mean there should be a margin-bottom because of the rounded corners? The inner container doesn't have a padding, so the textarea takes its full height - that was the same before.

The crisp edge of the text editor should be centered both horizontally and vertically in the rounded corner of the app content container.

JammingBen:
Should be fine now:
image

kulmann (approved):

PR #8039, opened 2022-11-29
[full-ci] chore: prepare release v6.0.0 final

kulmann:
Prepare `all the things` ™️  for `v6.0.0`


JammingBen (approved):

mmattel:
I just have seen that there were about 20 translations to German missing - fixed just now.
If you want to have that in you need to pull translations...

kulmann:
> I just have seen that there were about 20 translations to German missing - fixed just now.
> If you want to have that in you need to pull translations from TX...

Translation changes can't be pulled into the stable branch. The current state of the stable branch has all the german translations, because we made sure that no additional strings were added to the stable branch. Transifex only reflects the current web master.

PR #8041, opened 2022-11-29
[full-ci] Dedupe CSS

dschmidt:
## Description
This PR aims to remove the duplicated CSS caused by imports in the design-system.

~~It's not perfect yet, breadcrumbs are broken and input fields (like the search input) look wrong, otherwise it seems to work fairly ok.~~

~~I would be glad if someone with more knowledge of the styles in this project to pick this up and fix the remaining issues.~~

Thanks to @JammingBen for stepping up and taking care of the issues I had left open!🏅 

I have applied the same approach in my vite PR, there everything is even a little more sophisticated ~~(and has the same issues)~~ works the same way, so I'd say finishing this, would be a good preparation for solving this in Vite as well.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Since we imported the ODS into this repo, we have a lot of duplicated styles through imports in the ODS components. This slows down the dev tools a lot and (if not stripped in production builds) bloats the built assets.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/styles/styles.scss
+++ b/packages/design-system/src/styles/styles.scss
@@ -20,3 +20,5 @@
 @import "theme/oc-visibility";
 @import "theme/oc-width";
 @import "theme/tooltip";
+
+/* SHITTY MARKER */
--- a/packages/design-system/src/styles/styles.scss
+++ b/packages/design-system/src/styles/styles.scss
@@ -21,4 +21,4 @@
 @import "theme/oc-width";
 @import "theme/tooltip";
 
-/* SHITTY MARKER */
+/* STYLES STRIP IMPORTS MARKER */
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -44,6 +44,18 @@ const __filename = fileURLToPath(import.meta.url)
 const projectRootDir = dirname(__filename)
 
 const plugins = [
+  {
+    transform(src, id) {
+      if (id.includes('rollup-plugin-vue=styles')) {
+        const split = src.split('/* SHITTY MARKER */')
+        const newSrc = split[split.length-1]
+        return {
+          code: newSrc,
+          map: null
+        }
+      }
+    }
+  },
   postcss({
     extract: path.join('css', 'web.css'),
     minimize: production,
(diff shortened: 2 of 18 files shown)


ownclouders:
Results for <strong>oC10iPhone2</strong> https://drone.owncloud.com/owncloud/web/30212/50/1
:boom: The oC10iPhone2 tests pipeline failed. The build has been cancelled.

JammingBen:
I replaced the `@extend` usages in our components with actual CSS and/or mixins. Most of the extends were unnecessary anyway IMO.

The only real exception is the extension of the `oc-input` class which was kinda nice. We could do that via mixin in the future, but that requires quite some refactoring. For now, I manually added the `oc-input` class when `oc-text-input` is being used. It's wrapped in our `OcTextInput` component anyway, so ideally you use this one.

Let's see what pipeline says.

packages/design-system/src/styles/styles.scss
```diff
@@ -20,3 +20,5 @@
 @import "theme/oc-visibility";
 @import "theme/oc-width";
 @import "theme/tooltip";
+
```
kulmann, line 23: Comment about where the marker is used would be helpful (reference to rollup.config.mjs).

rollup.config.mjs
```diff
@@ -44,6 +44,18 @@ const __filename = fileURLToPath(import.meta.url)
 const projectRootDir = dirname(__filename)
 
 const plugins = [
+  {
+    transform(src, id) {
+      if (id.includes('rollup-plugin-vue=styles')) {
+        const split = src.split('/* STYLES STRIP IMPORTS MARKER */')
```
kulmann, line 50: Comment about the location in the code (full path to the `styles.scss` file) and the reason why we're doing this would be helpful.

kulmann (changes requested):

packages/design-system/src/styles/styles.scss
dschmidt: Added reference

kulmann:
The only failing tests are unrelated to this PR, see https://github.com/owncloud/web/issues/8042
I'll force-merge now.

dschmidt:
Sweeeeet!

Issue #8034, opened 2022-11-28
Empty sidebar when opening a space from search results


ScharfViktor:
ownCloud Web UI 6.0.0-rc.8
Infinite Scale 2.0.0-rc.2 Community 

Step: 
- admin creates project space "new"
- admin goes to "personal" and tries to find project space by name "new"
- admin opens "new" space via search result
- admin tries to open `details` or `members`:

Expected: sidebar is not empty
Actual: message: sidebar is empty and message: "Select a file or folder to view details" 

https://user-images.githubusercontent.com/84779829/204304347-9a18c472-d962-45a0-9ddf-7271c20f204d.mov

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8045.

PR #8045, opened 2022-11-30
Fix resolving drives in search results

JammingBen:
## Description
We've fixed a bug where folder listing was not reloaded when being in a space/share root and navigating into another space/share root via search.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8034

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/views/spaces/GenericSpace.vue
+++ b/packages/web-app-files/src/views/spaces/GenericSpace.vue
@@ -318,6 +318,11 @@ export default defineComponent({
       handler: function () {
         this.performLoaderTask(true)
       }
+    },
+    space: {
+      handler: function () {
+        this.performLoaderTask(true)
+      }
     }
   },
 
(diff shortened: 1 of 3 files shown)


kulmann (changes requested):
Can you think of a regression-test in jest? We could write one for the `GenericSpace` view that checks the call to the loader task whenever the `item` or `space` changes.

kulmann (approved):
💯

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -348,6 +353,10 @@ export default defineComponent({
     ]),
 
     async performLoaderTask(sameRoute: boolean, path?: string, fileId?: string | number) {
+      if (this.loadResourcesTask.isRunning) {
```
kulmann, line 356: For the record: If `space` and `item` (= relative item path in that space) **both** changed, the loader task is called twice. This is skipping one of the task executions.

JanAckermann (approved):

PR #8053, opened 2022-12-01
[full-ci] ci: only deploy from merge commits to master

kulmann:
In https://github.com/owncloud/web/pull/8048 I accidentally also added the `stable-*` trigger to the continuous deployment. That should only happen from master. Omitted the tags as well, because we don't want to deploy tags on the stable branch either.


micbar (approved):
LGTM

JammingBen (approved):

PR #7952, opened 2022-11-13
[full-ci] Vite

dschmidt:
## Description
This makes oC Web build with [Vite](https://vitejs.dev/).

To test this you need to do `docker-compose up -d`.
OAuth2 Application in oC 10 and OIDC Client in oCIS are setup automatically now.

- `pnpm vite` or `pnpm vite:ocis` -> https://host.docker.internal:9201/
   -> currently uses a shitty workaround because of https://github.com/owncloud/ocis/issues/5108
  -> possibly needs more dirs added to `server.proxy` in `vite.config.ts`
- `pnpm vite:oc10` -> http://host.docker.internal:8081/

## Motivation and Context
rollup builds are rather slow and we spend a lot of time waiting for builds during development.
vite supports Hot Module Reloading which makes changes instant. If there are errors they are shown in the browser and we don't need to switch to the terminal to notice them.
Moreover it's the standard tool in the Vue world by now.
It uses rollup as a bundling backend in production mode and has compatible plugin api, so we need to worry less about standard rollup plugins (because they are abstracted away by vite) and can still use more exotic rollup plugins.

## Open tasks:

- [x] Fix asset loading
- [x] Use a shared index.html for development and production: https://github.com/vbenjs/vite-plugin-html
- [x] Implement `oidc-callback.html`
- [x] Fix `process.env.PACKAGE_VERSION` replacement
- [x] Figure out how to replace define/requirejs for external apps
     - [x] during development, kinda works
       * [x] fix external app loading
       * [x] make sure development and production apps can be loaded
     - [x] in production
     - try to load `.js` and `.cjs` files as commonjs modules, use dynamic import only for `.mjs`. Possibly worth accepting TypeScript variants of those extensions to ease development of external apps with vite
 - [x] wire up with docker setup
 - [x] Check which rollup plugins are still needed and remove others
 - [x] Port `scripts` in `package.json`
 - [x] Import ODS?
 - [x] Apply owncloud-sdk code mod? -> https://www.npmjs.com/package/vite-plugin-replace
 - [x] Remove explicit style imports in design-system: https://vitejs.dev/config/shared-options.html#css-preprocessoroptions
 - [x] Fix `TypeError: r.pbkdf2Sync is not a function` on file download (in oCIS?) -> https://github.com/sodatea/vite-plugin-node-stdlib-browser ?
 - [x] Fix preview app
 - [x] Log warning when `@extend` is being used in a component
[...]

PR #8099, opened 2022-12-09
Fix Vite regressions

dschmidt:
## Description
`config.json` should not be copied to `dist/` in production builds.

I reenabled `rollup-plugin-gzip` after taking a quick look at `vite-plugin-compress` and `vite-plugin-compression`: one supports only brotli (and no standard gzip) and both don't seem to be as well maintained as `rollup-plugin-gzip`  and I could not see any benefit over our former solution, so I decided to go back to what we used anyway.

## Related Issue
- Fixes regressions introduced in #7952 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   ],
   "dependencies": {
     "keycode": "^2.2.1",
+    "rollup-plugin-gzip": "^3.1.0",
     "vite-plugin-static-copy": "^0.12.0",
     "vite-plugin-treat-umd-as-commonjs": "0.1.3",
     "vue": "2.7.14"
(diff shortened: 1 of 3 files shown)


kulmann (approved):

dschmidt:
Apparently not all html assets are gzipped, blocker or should we look at that in a follow up?

kulmann:
> Apparently not all html assets are gzipped, blocker or should we look at that in a follow up? 

I did the "answering by merging". 😁 seriously, not shipping the config.json is by far more important than two html files not being compressed. Fine as a followup. 🙂

PR #7976, opened 2022-11-15
Add viewMode toggle for condensed resource table

pascalwengerter:
## Description
Adds route-query based toggle between "normal" and "condensed" resource table

## Related Issue
- Fixes part of #6380

## Screenshots (if appropriate):
Screenshot 2022-11-15 at 22 34 49

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Open tasks:
- [ ] Review/feedback/approval by @tbsbdr / @kulmann / @elizavetaRa

PR #8105, opened 2022-12-12
Make viewoption modes configurable via appBar || Extend OcResourceIcon for spaces

pascalwengerter:
## Description
Follow-up to https://github.com/owncloud/web/pull/7976

## Types of changes

@kulmann @tbsbdr do you want the condensed option on the favorites view also?


packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -28,7 +28,7 @@
         </oc-breadcrumb>
         <shares-navigation v-if="hasSharesNavigation" />
         <div v-if="hasViewOptions || hasSidebarToggle" class="oc-flex">
-          <view-options v-if="hasViewOptions" />
+          <view-options v-if="hasViewOptions" :additional-view-modes="viewModeOptions" />
```
kulmann, line 31: I'd prefer `view-modes` as prop name and list all available view modes with it. Otherwise the next dev/reader needs to understand first what `additional` means.

kulmann (changes requested):

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
```diff
@@ -41,30 +44,43 @@ export default {
...
-      validator: (value) => {
-        return value.match(/(xsmall|small|medium|large|xlarge|xxlarge|xxxlarge)/)
+      validator: (value: string): boolean => {
+        return AVAILABLE_SIZES.some((e) => e === value)
       }
     }
   },
   computed: {
-    iconName() {
+    iconName(): string {
```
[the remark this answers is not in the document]
pascalwengerter: Was playing around because some of the types were not super intuitive, but happy to remove them also ;)

packages/web-app-files/tests/unit/components/AppBar/ViewOptions.spec.js
```diff
@@ -98,22 +98,41 @@ describe('ViewOptions', () => {
...
       localVue,
       stubs: stubs,
       directives: { OcTooltip }
     })
-    const viewModeSwitchButtons = wrapper.find('[data-testid="viewmode-switch-buttons"]')
 
+    expect(wrapper.find('[data-testid="viewmode-switch-buttons"]').exists()).toBeFalsy()
+  })
+
+  it('shows a viewmode switcher if more than one viewModes are passes', () => {
```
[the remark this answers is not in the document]
pascalwengerter: Good eye, thx!

pascalwengerter:
@kulmann tests are green, ready for re-review :)

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
```diff
@@ -41,30 +44,43 @@ export default {
...
-      validator: (value) => {
-        return value.match(/(xsmall|small|medium|large|xlarge|xxlarge|xxxlarge)/)
+      validator: (value: string): boolean => {
+        return AVAILABLE_SIZES.some((e) => e === value)
       }
     }
   },
   computed: {
     iconName() {
+      if (this.isSpace) return defaultSpaceIcon
```
JammingBen, line 54: I know it was like this already, but we decided to not use one-line-if-blocks in the project because they are somewhat ugly to read.

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
```diff
@@ -41,30 +44,43 @@ export default {
...
+      if (this.isSpace) return 'oc-resource-icon-space'
+      if (this.isFolder) return 'oc-resource-icon-folder'
+      return 'oc-resource-icon-file'
+    },
     isFolder() {
-      return this.resource.isFolder
+      // fallback is necessary since
+      // sometimes resources without a type
+      // but with `isFolder` are being passed
+      return this.resource.type === 'folder' || this.resource.isFolder
```
JammingBen, line 74: The class checks above only check for `if (this.isFolder) {...}`. Can we omit the `type` check here as well and always use the "fallback"?

JammingBen (commented):
I really like that we have a space resource icon now! Should we maybe use the same color as in the sidebar and the projects view? A brighter blue is being used there.

image

dschmidt (changes requested):
Hrm now the `additional-view-modes` moved one level up, I'm not sure that any better than before.
The issue is the same: one needs to look up what it gets added to and one cannot remove the defaults, maybe we should just push the whole list from wherever AppBar is used?

Also I find it weird to pass an arbitrary array of strings into the component where just a few hardcoded ones are handled. If we stick to the hardcoded list of options, I would argue we should instead just use boolean props to enable/disable them.

If we want to pass in a list to make it more flexible, we should pass in an array of mode switch objects, which provide their "query name"(aria), labels, icons,  and which we can loop over.

edit:
Currently you're using this list just to check whether it's longer than 1 element or not... that could be a single boolean ;-)

Depending on how flexible you want to make it (and how much effort you want to put in), I'd prefer the pass an array of objects approach, but a single bool param to enable/disable the switcher or individual bool props to disable the currently hardcoded mode buttons would be fine for me as well.

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
dschmidt: .... and because they make debugging more cumbersome and make diffs harder to read when changed later ;-)

kulmann:
For the future use case of having view modes registered through the extension system it would indeed be beneficial to define a `ViewMode` interface and pass a list of view mode objects (with icon, label, some id/name prop) through and don't have any hardcoded values in the `ViewOptions.vue` component.

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
pascalwengerter: `if (this.isFolder) return 'oc-resource-icon-folder'` => checks the computed property `isFolder`
while
`return this.resource.type === 'folder' || this.resource.isFolder` checks the resource passed as prop for its `isFolder` attribute ;)

Ideally, any resource passed to the OcResourceIcon has a `type`, but in some cases it only has a `isFolder` (and I'm not too sure how official from an api-standpoint the `isFolder` resource attribute is anyhow)...just checked and there's some usage of `isFolder` across the conflict&sharing helpers, actions mixins and upload components so most likely out of scope for this PR

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
pascalwengerter: So just plain old

```js
if (true) {
  return "something"
}
```

then, just for clarification?

And what about `return "something" ? condition : "fallback"`, are those types of statements allowed (since I see them used in the same component also)?

pascalwengerter: > I really like that we have a space resource icon now! Should we maybe use the same color as in the sidebar and the projects view? A brighter blue is being used there.
> 
> image

Great eye! I've updated it and would need from @tbsbdr if a ResourceIcon of type "space" should have a more distinct color down the road - for now it'll be the default "passive swatch" color, as it already is in other places right now

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { shallowMount } from '@vue/test-utils'
```
dschmidt, line 1: Please import from web-test-helpers

packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -132,6 +137,12 @@ export default defineComponent({
         this.selectedFiles.length
       )
       return this.$gettextInterpolate(translated, { amount: this.selectedFiles.length })
+    },
+    viewModes() {
+      if (!this.displayViewModeSwitch) {
+        return []
+      }
+      return [ViewModeConstants.condensedTable, ViewModeConstants.default]
     }
```
dschmidt, line 146: Me gusta

packages/web-app-files/tests/unit/components/AppBar/ViewOptions.spec.ts
```diff
@@ -33,9 +34,28 @@ describe('ViewOptions component', () => {
...
+    )
+    const viewModeSwitchButtons = wrapper.find('[data-testid="viewmode-switch-buttons"]')
+    expect(viewModeSwitchButtons).toMatchSnapshot()
+  })
 })
 
-function getWrapper({ perPage = '100' } = {}) {
+function getWrapper(
+  { perPage = '100' } = {},
+  props: { [key: string]: any } = {
```
dschmidt, line 55: No need for this type

packages/web-app-files/src/composables/viewMode/useViewMode.ts
```diff
@@ -9,7 +9,7 @@ export function useViewMode<T>(options: ComputedRef<string>): ComputedRef<string
 
   const viewModeQuery = useRouteQueryPersisted({
     name: ViewModeConstants.queryName,
-    defaultValue: ViewModeConstants.default
+    defaultValue: ViewModeConstants.default.name
```
dschmidt, line 12: I've seen this code a bunch of times, maybe worth putting into its own composable

packages/web-app-files/src/components/AppBar/ViewOptions.vue
```diff
@@ -1,17 +1,19 @@
...
+      class="oc-button-group oc-visible@s oc-mr-s"
+    >
       <oc-button
-        :appearance="viewModeCurrent === ViewModeConstants.default ? 'filled' : 'outline'"
-        @click="setViewMode(ViewModeConstants.default)"
+        v-for="viewMode in viewModes"
+        :key="viewMode.name"
+        v-oc-tooltip="viewMode.label"
+        :appearance="viewModeCurrent === viewMode.name ? 'filled' : 'outline'"
+        :label="viewMode.label"
```
dschmidt, line 13: I think you need to pass this to $gettext again to make translations actually work

dschmidt (commented):
Love where this going 🤩

packages/web-app-files/tests/unit/components/AppBar/ViewOptions.spec.ts
pascalwengerter: VSCode wanted one pretty badly, but I'll try without...

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
JammingBen: Ahh sorry, didn't watch closely enough. Yeah, it would be great to streamline that, but definitely out of scope here :D

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
JammingBen: `return "something" ? condition : "fallback"` or `const someVariable = "something" ? condition : "fallback"` is fine. We just don't want to have the "one-line-return/continue".

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
pascalwengerter: I get it though, I had fun in this component. Things don't get better as a space also has a `isFolder === true` while a resource-folder apparently can potentially have either `isFolder` or `type` of undefined  🥴 riddle me this

packages/web-app-files/tests/unit/components/AppBar/ViewOptions.spec.ts
```diff
@@ -52,8 +53,8 @@ describe('ViewOptions component', () => {
 
 function getWrapper(
   { perPage = '100' } = {},
-  props: { [key: string]: any } = {
-    viewModes: []
+  props?: {
+    viewModes: ViewMode[]
```
dschmidt, line 57: Unused

pascalwengerter: Not sure how to resolve this - my IDE complains in line 46 of the same file if I remove the type annotations. If you don't mind and there's no other issues let's get this merged and let me continue, not worth another 45min of acceptance test pipelines

dschmidt: oh misread the diff, nevermind.

dschmidt (approved):

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.spec.ts
kulmann: Still valid / unresolved.

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+import { AVAILABLE_SIZES } from '../../helpers/constants'
+import { OcResourceIcon } from '..'
+
+const resources = ['file', 'folder', 'space']
+
+function getWrapper(props) {
```
kulmann, line 7: Please move the `getWrapper` function below the tests (for consistency with other tests in `web`).

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  return shallowMount(OcResourceIcon, {
```
kulmann, line 8: Please return an object with key `wrapper` and your `shallowMount` call as value, i.e.
```js
return {
  wrapper: shallowMount(...
```
for consistency with other tests in `web`.

packages/web-app-files/src/views/Favorites.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="oc-flex">
     <files-view-wrapper>
-      <app-bar :side-bar-open="sideBarOpen" />
+      <app-bar :display-view-mode-switch="true" :side-bar-open="sideBarOpen" />
```
kulmann, line 4: Side note: This already is an example where we don't want the `condensed` view mode (or at least it doesn't make too much sense, since it is not different from the default list view mode), but will want to have the `tiles` view mode available in the future. :-/ Let's keep it like you have it here for the time being.

packages/web-app-files/src/components/FilesList/ResourceTable.vue
```diff
@@ -326,7 +326,7 @@ export default defineComponent({
     },
     viewMode: {
       type: String,
-      default: ViewModeConstants.default
+      default: ViewModeConstants.default.name
```
kulmann, line 329: You have already introduced the `defaultModeName` const for this

packages/web-app-files/tests/unit/components/AppBar/ViewOptions.spec.ts
```diff
@@ -33,9 +35,28 @@ describe('ViewOptions component', () => {
...
+  })
+  it('shows a viewmode switcher if more than one viewModes are passed', () => {
+    const { wrapper } = getWrapper(
+      {},
+      {
+        viewModes: [ViewModeConstants.condensedTable, ViewModeConstants.default]
+      }
+    )
+    const viewModeSwitchButtons = wrapper.find('[data-testid="viewmode-switch-buttons"]')
+    expect(viewModeSwitchButtons).toMatchSnapshot()
```
kulmann, line 50: Would like to avoid a snapshot test for this. Should already be sufficient to check the presence of the element.
If you ever find a situation where a snapshot test is useful, please do so on the `.html()` result of the element. Will be required shortly.

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
```diff
@@ -4,24 +4,27 @@
...
-      { 'oc-resource-icon-file': !isFolder, 'oc-resource-icon-folder': isFolder }
-    ]"
+    :class="['oc-resource-icon', iconTypeClass]"
   />
 </template>
 
-<script>
+<script lang="ts">
+import { defineComponent, PropType } from 'vue'
+import { Resource } from 'web-client'
```
kulmann, line 13: Really love that you're using the `Resource` interface here. Finally that's possible! But please add `"web-client": "npm:@ownclouders/web-client"` to the list of `peerDependencies` of the design system (package.json).

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
```diff
@@ -4,24 +4,27 @@
...
+
 import OcIcon from '../OcIcon/OcIcon.vue'
-import iconNameMap from '../../helpers/resourceIconExtensionMapping'
-import iconColorMap from '../../helpers/resourceIconColorExtensionMapping'
+import { AVAILABLE_SIZES } from '../../helpers/constants'
```
kulmann, line 16:
```suggestion
import { AVAILABLE_SIZES } from '../../helpers'
```

(at least that's what my IDE is complaining about ;-) )

kulmann (changes requested):
Found some more nitpick-material, sorry 😅

packages/design-system/src/components/OcResourceIcon/OcResourceIcon.vue
pascalwengerter: It's possible but can eventually lead to circular import (stumbled upon this last week), will fix for now but keep in mind pls :)

packages/web-app-files/src/views/Favorites.vue
pascalwengerter: _soon_

packages/web-app-files/src/components/FilesList/ResourceTable.vue
pascalwengerter: Apparently slipped through, sry about that

packages/web-app-files/tests/unit/components/AppBar/ViewOptions.spec.ts
pascalwengerter: I'm somewhat torn about this - I know the oC web team in general is not a huge fan of snapshot tests, however in this case it's not only asserting that there is an element with a `viewmode-switch-buttons` data-testid, but also that there's two buttons, each with an icon inside (and checks for the *momentarily* correct CSS classes etc)

kulmann: Also fine to keep it... I'd go the way of asserting that there are `x` buttons and ignoring the icon assertion for example.

But: please change line 50 to `expect(viewModeSwitchButtons.html()).toMatchSnapshot()`

kulmann (approved):
🤘

PR #8127, opened 2022-12-16
Use wrapper.html() for generating snapshots

dschmidt:
Vue Test Utils 2 does not support expect(wrapper).toMatchSnapshot() but requires us to call expect(wrapper.html().toMatchSnapshot()

## Motivation and Context
Vue Test Utils 2 do not support `expect(wrapper).toMatchSnapshot()` but require us to call `expect(wrapper.html().toMatchSnapshot()`. VTU1 support it as well, so let's do it in stable to keep diffs small.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-external/tests/unit/app.spec.ts
+++ b/packages/web-app-external/tests/unit/app.spec.ts
@@ -53,7 +53,7 @@ describe('The app provider extension', () => {
   it('should show a loading spinner while loading', async () => {
     const { wrapper } = createShallowMountWrapper()
     await wrapper.vm.$nextTick()
-    expect(wrapper).toMatchSnapshot()
+    expect(wrapper.html()).toMatchSnapshot()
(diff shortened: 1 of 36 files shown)


kulmann (approved):

JammingBen:
I assume it needs to be done for `wrapper.find(...)` as well? Because there is still a usage in https://github.com/owncloud/web/blob/wrapper-html/packages/web-runtime/tests/unit/pages/resolvePublicLink.spec.ts#L58.

JammingBen (approved):

PR #8135, opened 2022-12-19
feat: support browsers that are 1 year old

kulmann:
## Description
This PR tackles two things:
- we want to support more than the `last 2 versions`, because `last 2 versions` frequently let's e.g. firefox extended support releases (esr) run into the unsupported browser situation. Chrome is probably similar. We expect users to be on a recent-is browser, but `last 2 versions` is too harsh.
- our vite config was actually ignoring the `browserslist` config, as it only uses esbuild targets. Which is a different thing. Added a plugin to transform `browserslist` into `esbuild targets`. If we don't want to do that we need to move away from `browserslist`.

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

cc @mmattel as this is relevant for the docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "vite:oc10": "OCIS=false vite"
   },
   "browserslist": [
-    "last 2 versions",
+    "last 1 year",
     "> .2%",
     "not dead",
     "not Explorer > 0",
@@ -69,6 +69,7 @@
     "autoprefixer": "10.4.13",
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "29.3.1",
+    "browserslist-to-esbuild": "^1.2.0",
     "browserslist-useragent-regexp": "^3.0.2",
     "commander": "8.3.0",
     "core-js": "3.26.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -40,6 +40,7 @@ importers:
       autoprefixer: 10.4.13
       babel-core: 7.0.0-bridge.0
       babel-jest: 29.3.1
+      browserslist-to-esbuild: ^1.2.0
       browserslist-useragent-regexp: ^3.0.2
       caf: '*'
       commander: 8.3.0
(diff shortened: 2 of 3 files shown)


mmattel:
> relevant for the docs

Can you tell how docs can support?

kulmann:
> > relevant for the docs
> 
> Can you tell how docs can support?

If I remember correctly you have a list of supported browsers somewhere in the docs. This PR changes supported versions (but only for web master, not for the stable GA stream).

mmattel:
For the reference, this is located at: https://doc.owncloud.com/webui/next/owncloud_web/requirements.html#desktop-requirements

dschmidt (approved):
I thought I had already approved this

kulmann:
> For the reference, this is located at: https://doc.owncloud.com/webui/next/owncloud_web/requirements.html#desktop-requirements

Thanks. I'll hit the merge button now, since the PR is approved. Now it's not clear to me what should happen to the docs page. It shows `Latest 2 versions` for Chrome, Firefox and Edge. But with this PR the web master supports `all versions from 1 year ago until now` for those browsers. For the stable product we stay at `Latest 2 versions`.

mmattel:
Q: I just don't know it, does web show somewhere its version on the screen?

kulmann:
> Q: I just don't know it, does web show somewhere its version on the screen?

Yes, in the browser console. But unfortunately at the moment both web `stable-6.0` and `master` show `6.0.0` (because that was the latest release). `master` will switch to `7.0.0` whenever we do the next official release from `master`.

mmattel:
ok, there is atm no "user-friedly" version accessible.  I have to think about how to solve this.

Note that it shoud be considered making this info available to ordinary users - exact for such cases where we have to reference in the documentation to stuff that has changed and is mentionable. To be considered that we bundle with ocis and have an app for oc10...

PR #8136, opened 2022-12-19
Feature: #8134 - copy to clipboard

### LINKED ISSUE #8134: [FR] Use Clipboard API instead of Document.execCommand for Clipboard Actions
Currently the JavaScript Code uses the follow module: https://www.npmjs.com/package/copy-to-clipboard. This module uses https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand# under the hood which is not working properly on e.g. Safari as the user is just presented a popup to copy the link himself.

Meanwhile there is a Clipboard API which is also supported on Safari, see comparison in https://www.w3docs.com/snippets/javascript/how-to-copy-the-text-to-the-clipboard-with-javascript.html.

This feature request is to replace the module and use one with uses the Clipboard API under the hood to support all browser properly with working Clipboard interactions.

pascalwengerter:
## Description
Figured I'd treat myself to this quickwin before going to bed :D Caniuse approves: https://caniuse.com/?search=navigator.clipboard.writeText

## Related Issue
- Fixes #8134 

## Motivation and Context
I couldn't test the hypothesis of #8134 locally since Safari complained about my dev setup (Chrome/Firefox work fine), please review carefully.

## How Has This Been Tested?
- Locally

## Types of changes

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -95,6 +95,7 @@
                 v-text="file.path"
               />
               <oc-button
+                v-if="isSupported"
                 v-oc-tooltip="copyEosPathLabel"
                 :aria-label="copyEosPathLabel"
                 appearance="raw"
@@ -123,6 +124,7 @@
                 v-text="directLink"
               />
               <oc-button
+                v-if="isSupported"
                 v-oc-tooltip="copyDirectLinkLabel"
                 :aria-label="copyDirectLinkLabel"
                 appearance="raw"
@@ -160,7 +162,7 @@
   </div>
 </template>
 <script lang="ts">
-import { ComputedRef, defineComponent, inject } from 'vue'
+import { ComputedRef, defineComponent, inject, ref } from 'vue'
 import { mapActions, mapGetters } from 'vuex'
 import { ImageDimension } from '../../../constants'
 import { loadPreview } from 'web-pkg/src/helpers/preview'
@@ -173,10 +175,11 @@ import {
   useCapabilityFilesTags,
   usePublicLinkContext,
   useStore,
+  useTranslations,
   useUserContext
 } from 'web-pkg/src/composables'
 import { getIndicators } from '../../../helpers/statusIndicators'
-import copyToClipboard from 'copy-to-clipboard'
+import { useClipboard } from '@vueuse/core'
 import { encodePath } from 'web-pkg/src/utils'
 import { formatDateFromHTTP, formatFileSize } from 'web-pkg/src/helpers'
 import { eventBus } from 'web-pkg/src/services/eventBus'
@@ -190,8 +193,40 @@ export default defineComponent({
   name: 'FileDetails',
   setup() {
     const store = useStore()
+    const { $gettext } = useTranslations()
+
+    const copiedDirect = ref(false)
+    const copiedEos = ref(false)
+    const { copy, copied, isSupported } = useClipboard({ legacy: true, copiedDuring: 550 })
+
+    const copyEosPathToClipboard = () => {
+      copy(inject<ComputedRef<Resource>>('displayedItem').value.path)
+      copiedEos.value = copied.value
+      store.dispatch('showMessage', {
+        title: $gettext('EOS path copied'),
+        desc: $gettext('The EOS path has been copied to your clipboard.')
+      })
+    }
+
+    const copyDirectLinkToClipboard = () => {
+      copy(
+        `${store.getters.configuration.server}files/spaces/personal/home${
+          inject<ComputedRef<Resource>>('displayedItem').value.path
+        }`
+      )
+      copiedDirect.value = copied.value
+      store.dispatch('showMessage', {
+        title: $gettext('Direct link copied'),
+        desc: $gettext('The direct link has been copied to your clipboard.')
+      })
+    }
 
     return {
+      copiedEos,
+      copyEosPathToClipboard,
+      copiedDirect,
+      copyDirectLinkToClipboard,
+      isSupported,
       isUserContext: useUserContext({ store }),
       isPublicLinkContext: usePublicLinkContext({ store }),
       accessToken: useAccessToken({ store }),
@@ -200,12 +235,8 @@ export default defineComponent({
       hasTags: useCapabilityFilesTags()
     }
   },
-
   data: () => ({
-    loading: false,
-    copiedDirect: false,
-    copiedEos: false,
-    timeout: null
+    loading: false
   }),
   computed: {
     ...mapGetters('runtime/spaces', ['spaces']),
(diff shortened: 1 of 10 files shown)


ownclouders:
Results for <strong>acceptance oC10</strong> https://drone.owncloud.com/owncloud/web/30836/37/1

<p>


### webUISharingPublicManagement-shareByPublicLink_feature-L71.png
  


</p></details>
:boom: The oC10SharingPublic2 tests pipeline failed. The build has been cancelled.

packages/web-app-files/tests/unit/components/SideBar/PrivateLinkItem.spec.ts
```diff
@@ -11,13 +11,21 @@ describe('PrivateLinkItem', () => {
...
+
     const { wrapper } = getWrapper()
     const spyShowMessage = jest.spyOn(wrapper.vm, 'showMessage')
     expect(spyShowMessage).not.toHaveBeenCalled()
 
     await wrapper.trigger('click')
     expect(wrapper.html()).toMatchSnapshot()
+    expect(window.navigator.clipboard.writeText).toHaveBeenCalledWith(
+      wrapper.vm._provided.displayedItem.privateLink
+    )
```
dschmidt, line 28: Please no private api. Can't you just hardcode the value?

dschmidt (changes requested):

packages/web-app-files/tests/unit/components/SideBar/PrivateLinkItem.spec.ts
pascalwengerter: Yeah I was more concerned with whether I could rather than whether I should :D fixed it by extracting the mocked folder, thx

pascalwengerter:
I tried running the acceptance tests according to the guide in the docs, but it wouldn't work on my M1 Mac (failed with "outdated browser")

Failing in https://drone.owncloud.com/owncloud/web/30836/36/19 in CI

packages/web-app-files/tests/unit/components/SideBar/PrivateLinkItem.spec.ts
```diff
@@ -5,6 +5,16 @@ import { createStore, defaultPlugins, mount, defaultStoreMockOptions } from 'web
 
 jest.useFakeTimers()
 
+const folder = mockDeep<Resource>({
```
dschmidt, line 8: We are getting rid of mockDeep where not strictly neccessary, it has weird side effects and can trigger stack overflows with vue 3

dschmidt (changes requested):

packages/web-app-files/tests/unit/components/SideBar/PrivateLinkItem.spec.ts
dschmidt: Otherwise good if you fix the tests 💫

pascalwengerter: We'll have to pass some sort of resource to mount the component - I don't think this test in particular will be especially picky about how the resource gets built (I've just extracted what was being used on the `getWrapper` function further down in the file 🤓 ), so should be fine when the refactoring time comes ;)

dschmidt: Just use mock instead of mockDeep

dschmidt (approved):

kulmann (approved):
Nice! 👍 Can you look into the reason why acceptance tests are failing?

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/30887/53/1
:boom: The oCISWebdavLockProtection tests pipeline failed. The build has been cancelled.

packages/web-app-files/tests/unit/components/SideBar/PrivateLinkItem.spec.ts
pascalwengerter: Done

pascalwengerter:
> Nice! 👍 Can you look into the reason why acceptance tests are failing?

I will, as pointed out above I couldn't get them to run locally as it worked in the past? I'll re-check one of these days, maybe I messed up some docker thing with the M1 chip

pascalwengerter:
Needed to dig a bit deeper here - turns out using the native browser clipboard is only available when using localhost or a secure connection (and hence fails with a self-signed cert/running acceptance tests over HTTP). I searched around and used the workaround mentioned [here](https://github.com/amatsuda/heavens_door/issues/15#issuecomment-565882201) - using the native API if available, and falling back to the 3rd party implementation that was already used before if it's not available

dschmidt:
If the lib doesn't abstract both approaches away for us, is it worth pulling it in at all? It's just a single function invocation anyway, maybe just inline it?

pascalwengerter:
> If the lib doesn't abstract both approaches away for us, is it worth pulling it in at all? It's just a single function invocation anyway, maybe just inline it?

What do you think of https://vueuse.org/core/useClipboard/ with the `legacy` flag? :)

dschmidt:
Perfect, obviously 💫

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -190,8 +193,40 @@ export default defineComponent({
...
+    const { copy, copied, isSupported } = useClipboard({ legacy: true, copiedDuring: 550 })
+
+    const copyEosPathToClipboard = () => {
+      copy(inject<ComputedRef<Resource>>('displayedItem').value.path)
+      copiedEos.value = copied.value
+      store.dispatch('showMessage', {
+        title: $gettext('EOS path copied'),
+        desc: $gettext('The EOS path has been copied to your clipboard.')
+      })
+    }
```
dschmidt, line 209: I know you just copied existing code, but do we really want to expose the `EOS` string/name to a user? 

Should we revisit this, while you're at it?

pascalwengerter: > Perfect, obviously 💫

I'm not quite sure if I get the `isSupported` right though - reading https://github.com/vueuse/vueuse/blob/70f55d4e68fd8f100fe844dd883c89b6d6594d03/packages/core/useClipboard/index.ts#L67 it seems like I could omit it since it is basically supported through the legacy option anyways?

Now it's "only" fixing unit tests, no opinion about the EOS translation thought...it's a CERN-internal thing anyways and practically no-one will be able to "casually spin up Reva/oCIS/oc-web on an EOS and then be confused about those strings" ;P

pascalwengerter: And from what I remember they literally send each other EOS paths when referencing resources

pascalwengerter:
Closed as superseded by #8173, thanks @JammingBen for taking over!

PR #8175, opened 2023-01-03
Rename 'user management' app to 'admin settings'

JammingBen:
## Description
The "user management" app has been renamed to "admin settings" because the app will contain a variety of general settings in the future.

## Related Issue

- https://jira.owncloud.com/browse/OCIS-3190

## Follow-up
- [ ] Update the web config in oCIS after a new Web RC has been created

## Types of changes

- [ ] Technical debt
- [ ] Tests


changelog/unreleased/enhancement-rename-user-management
```diff
@@ -0,0 +1,5 @@
+Enhancement: Rename "user management" app
+
+The "user management" app has been renamed to "admin settings" because the app will contain a variety of general settings in the future.
+
+https://github.com/owncloud/web/pull/8175
```
dschmidt, line 5: It's a breaking change, as it potentially needs a config adjustment by the user. 


We could also do a graceful deprecation and keep the old name working in config but pring a warning when it's used... 

Then it should also be added to the deprecations ticket

JammingBen: Hmm I'm not sure actually, according to the Jira ticket, it's not a breaking change due to https://github.com/owncloud/web/pull/8175/files#diff-9521d586d9cc2dae8e0510f697d29d0fc9b164419ebc855a5bca4e216414e96e. That also prints a warning

dschmidt: Oh you're right, apparently scrolled over that! Happy with that, nvm!

dschmidt: Can you still add it to the deprecations list?

JammingBen: Done so, thx for the hint!

fschade (approved):
LGTM

PR #8187, opened 2023-01-04
Feature/8166 - Show current upload speed and upload details in tooltip

### LINKED ISSUE #8166: Show file upload speed
Add the ability to display the upload speed in the files extension. Perhaps as a separate parameter in the settings. I want to switch from nextcloud to ocis. It is important for me to see in the web interface the speed with which I upload files to my cloud. How can I see her?

pascalwengerter:
## Description
Quick draft for #8166, happy to have someone else pick up the ball here :)

## Related Issue
- Fixes #8166 but needs further work (tests, translations, changelog item, ...)

## Motivation and Context
Community request

## Types of changes

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -295,6 +309,14 @@ export default defineComponent({
       this.filesInEstimation[file.meta.uploadId] = progress.bytesUploaded
 
       const timeElapsed = +new Date() - this.timeStarted
+
+      // "bytesTotal" is a wrong type documentation,
+      // fix in progress via https://github.com/transloadit/uppy/pull/4263
+      this.uploadSpeed = getSpeed({
```
pascalwengerter, line 315: I had the content of `getSpeed` copied to a local method while figuring out why it wouldn't do what I expected it to, using it directly still returns `NaN` and breaks the tooltip so further investigation needed/potentially a type error somewhere

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -167,13 +170,24 @@ export default defineComponent({
...
   }),
   computed: {
     ...mapGetters(['configuration']),
 
+    uploadDetails() {
+      if (this.uploadSpeed === 0) {
+        return ''
+      }
+      const todo = filesize(this.bytesUploaded)
+      const done = filesize(this.bytesTotal)
```
dschmidt, line 186: Naming of these two variables seems off

packages/web-runtime/src/components/UploadInfo.vue
micbar: upstream fix has been released.

pascalwengerter: Ah nice, thanks for the reminder! I'll see if I can get this PR done this week then :)

pascalwengerter:
@JammingBen @kulmann could someone give this a look and potentially test it on oC10 also? Don't think I'll find more time to invest into this, but so far it was looking okay

Screenshot 2023-02-06 at 21 42 08
Screenshot 2023-02-06 at 21 42 20

packages/web-runtime/src/components/UploadInfo.vue
pascalwengerter: Updated, thx! (no clue what I was thinking 😶‍🌫️ )

changelog/unreleased/enhancement-show-upload-speed
```diff
@@ -0,0 +1,8 @@
+# Show upload speed
```
kulmann, line 1:
```suggestion
Enhancement: Show upload speed
```

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -167,13 +170,24 @@ export default defineComponent({
...
 
+    uploadDetails() {
+      if (this.uploadSpeed === 0) {
+        return ''
+      }
+      const uploadedBytes = filesize(this.bytesUploaded)
+      const totalBytes = filesize(this.bytesTotal)
+      const currentUploadSpeed = filesize(this.uploadSpeed)
+      // needs $gettext interpolation
+      return `${uploadedBytes} of ${totalBytes} (${currentUploadSpeed}/s)`
```
kulmann, line 189: that's a `TODO` that should be solved before merging. :-) fyi, in the vue3 version of gettext the interpolation happens directly in `$gettext` (grep for examples in the code)

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -167,13 +170,24 @@ export default defineComponent({
...
     remainingTime: undefined
   }),
   computed: {
     ...mapGetters(['configuration']),
 
+    uploadDetails() {
+      if (this.uploadSpeed === 0) {
+        return ''
+      }
+      const uploadedBytes = filesize(this.bytesUploaded)
```
kulmann, line 185: Please use our `formatFileSize` helper from `web-pkg`. For the use case of showing upload speed I'd say you should introduce an optional third argument to that helper, `precision?: number`, with which you should override the internally calculated precision to `2`.
Reason: a) unified file size formatting across ownCloud Web and b) it forces you to pass the users locale into the formatting.

kulmann (changes requested):
The tooltip feels **very** jumpy because of the rapid text changes and a non-monospaced font. I don't think there is much we can do about it... one thing you could try is using only 1 decimal instead of 2. And the other option would be debouncing the text changes by like 200-500ms, so that the changes are frequent but not as frequent as they are now. I would definitely like to keep the upload speed in the tooltip. Maybe we should indicate somehow that a tooltip is available in this case...

kulmann:
Forgot to say, oc10 works fine as well 👍

packages/web-runtime/src/components/UploadInfo.vue
pascalwengerter: Do you mean override the precision to `1` (which "limiting the decimal to 1 instead of 2" from your comment above suggests) since from how I understand https://github.com/avoidwork/filesize.js/blob/d41818f78b0e2023eec98e5f9964aaeefaf91e8a/src/filesize.js#L178 the default of the wrapped lib is already 2? :)

kulmann: our `formatFileSize` helper has a default of `1` for values smaller than 1 mb, and a default of `0` for values greater than 1 mb. That's somehow our default when showing file sizes (was decided by PM 1 or 2 years ago). I meant overriding **that** default, not the one of the filesize lib ;-)

pascalwengerter: Got it, thx for the swift reply!

packages/web-runtime/src/components/UploadInfo.vue
pascalwengerter: Updated, thx

changelog/unreleased/enhancement-show-upload-speed
pascalwengerter: Updated

pascalwengerter:
I've updated this PR to use `$gettext` and `formatFileSize`, feels better now if you ask me. I've also added a catch to prevent the tooltip from being displayed after the download finishes (or has been cancelled).
Regarding debouncing - I took a look at https://vueuse.org/shared/watchDebounced/, but that would mean we'd have to introduce `@vueuse/core` also (could also be done in a subsequent PR)
Regarding discoverability - not sure what possibilities we'd have, given that the tooltip is also not visible on mobile (and not sure how to present it to screenreaders in a way that doesn't "spam" them with information 😅 )

kulmann (approved):
nice improvement! thank you! 💪

kulmann:
> I've updated this PR to use `$gettext` and `formatFileSize`, feels better now if you ask me. I've also added a catch to prevent the tooltip from being displayed after the download finishes (or has been cancelled).
> Regarding debouncing - I took a look at https://vueuse.org/shared/watchDebounced/, but that would mean we'd have to introduce `@vueuse/core` also (could also be done in a subsequent PR)

The "level of jumpiness" is much better now with less decimal places. IMO sufficient without the debounce.


> Regarding discoverability - not sure what possibilities we'd have, given that the tooltip is also not visible on mobile (and not sure how to present it to screenreaders in a way that doesn't "spam" them with information 😅 )

It's good that the tooltip is not accessible for screen readers anyway at the moment. Don't see that this is valuable for screen reader in the current state. A hidden announcer could be an option, but the information is so "nice to have" that I wouldn't try that at the moment.

PR #8192, opened 2023-01-09
[full-ci] Add admin space sidebar details

lookacat:
## Description
See https://github.com/owncloud/web/issues/8219

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8219

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -155,6 +158,41 @@ export default defineComponent({
       toggleSelectSpace,
       unselectAllSpaces
     }
+  },
+  computed: {
```
JammingBen, line 162: Please use the Composition API. For computed props, you can use `computed(() => { ... })` (like done above) and methods can be simply written as arrow-functions.

In general, we want to use the Composition API in the future because it's the recommended way in Vue 3.

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -155,6 +158,41 @@ export default defineComponent({
...
+        {
+          app: 'SpaceDetails',
+          icon: 'layout-grid',
+          title: this.$gettext('Space details'),
+          component: this.sidebarSpacePanelComponent,
+          default: true,
+          enabled: true,
+          componentAttrs: {
+            spaceResource: this.selectedSpaces[0],
+            spaceCount: this.selectedSpaces.length
```
JammingBen, line 180:
```suggestion
```

JammingBen (changes requested):
I find the panels a bit confusing at the moment. We have `SpaceDetails` if one space is selected and we have `DetailsPanel` if no or >1 spaces are selected. I would go for one of the following:

* Use one panel (`SpaceDetails`) to handle all cases. Problem with this approach is that the `SpaceDetails` panel is used in the files-app as well, where we don't want this. Hence I'd rather do it like it's done in the files-app already:
* Have 3 separate panels for all 3 cases: `NoSelection` (0 spaces), `SpaceDetails` (1 space), `SpaceDetailsMultiple` (2 or more spaces). Those components can live in `web-pkg` because we might re-use them in the files app as well. The check which panels needs to be displayed would then be done in `packages/web-app-admin-settings/src/views/Spaces.vue` via the `enabled` prop of each panel.

And we also need to make sure to not display the space image in `SpaceDetails` in the admin-settings. You could just introduce another property `showImage` which defaults to `true`.

packages/web-app-admin-settings/src/views/Spaces.vue
lookacat: sí 👍🏼

changelog/unreleased/enhancement-add-admin-space-details-sidebar
```diff
@@ -0,0 +1,6 @@
+Enhancement: Add space details for admins
```
JammingBen, line 1: You can just add your PR link to the existing changelog item for the spaces management. It makes more sense from a user perspective since we're releasing it together as one feature.

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -126,6 +130,44 @@ export default defineComponent({
...
+          default: false,
+          enabled: unref(selectedSpaces).length > 1
+        }
+      ]
+    })
+
+    const sideBarActivePanel = computed(() => {
+      const t = unref(sideBarAvailablePanels).find((e) => e.enabled).app
+      console.log(t)
+      return t
```
JammingBen, line 168:
```suggestion
      return unref(sideBarAvailablePanels).find((e) => e.enabled).app
```

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -145,16 +187,27 @@ export default defineComponent({
...
+  methods: {
+    selectPanel(panel) {
+      this.sideBarActivePanel = panel || 'DetailsPanel'
+    },
+    closeSideBar() {
+      this.sideBarOpen = false
+    },
+    toggleSideBar() {
+      this.sideBarOpen = !this.sideBarOpen
+    }
```
JammingBen, line 210: Please move those to `setup()` as well.

JammingBen (changes requested):
Nice 👍  Just a few minor things I've noticed.

pnpm-lock.yaml
```diff
@@ -506,7 +506,7 @@ importers:
       web-runtime: link:../web-runtime
     devDependencies:
       '@jest/globals': 29.3.1
-      '@vueuse/core': 9.8.2_vue@3.2.45
+      '@vueuse/core': 9.8.2
```
kulmann, line 509: Do you know where this change came from? 🤔 cc @dschmidt

packages/web-pkg/src/constants.ts
```diff
@@ -0,0 +1,15 @@
+export const EVENT_TROW_MOUNTED = 'rowMounted'
+export const EVENT_FILE_DROPPED = 'fileDropped'
+export const EVENT_TROW_CONTEXTMENU = 'contextmenuClicked'
+
+export abstract class ImageDimension {
```
kulmann, line 5: You only copied this file, right? There is the same file in `web-app-files/src/constants.ts`. Please make sure that this is not duplicated. I.e. please delete the one in `web-app-files` and change all the imports of the file to the `web-pkg` version.

pnpm-lock.yaml
lookacat: probably because of my rebase, vscode had a bad day yesterday and did weird things, totally not my fault :trollface:

packages/web-pkg/src/components/sideBar/Details/SpaceNoSelection.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
kulmann, line 1: I think the `Details` folder should have another parent folder named `Spaces`. And then please move `SpaceNoSelection.vue` one level up. Result should be this layout:

```
- Spaces
  - Details
    - SpaceDetails.vue
    - SpaceDetailsMultiple.vue
  - SpaceNoSelection.vue
```

In a followup PR you could make use of the `SpaceNoSelection` component in the `Projects.vue` view of the files app.

kulmann (changes requested):

packages/web-pkg/src/components/sideBar/Details/SpaceDetails.vue
```diff
@@ -77,11 +77,16 @@ import SpaceQuota from '../../SpaceQuota.vue'
...
 import { configurationManager } from 'web-pkg/src/configuration'
 import { eventBus } from 'web-pkg/src/services/eventBus'
-import { SideBarEventTopics } from '../../../composables/sideBar'
 
 export default defineComponent({
   name: 'SpaceDetails',
   components: { SpaceQuota },
+  props: {
+    spaceResource: {
+      type: Object,
```
kulmann, line 86: Please use the correct type: 
```js
type: Object as PropType<SpaceResource>
```
Also: any reason why you didn't just name the variable `space`?

lookacat: because space is already used ^^

kulmann: yeah ok, sorry 😆 then keep the name like it is. But please specify the type.

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -126,6 +129,49 @@ export default defineComponent({
...
+          component: SpaceNoSelection,
+          default: true,
+          enabled: unref(selectedSpaces).length === 0
+        },
+        {
+          app: 'SpaceDetails',
+          icon: 'layout-grid',
+          title: $gettext('Space details'),
+          component: SpaceDetails,
+          default: false,
```
kulmann, line 147:
```suggestion
          default: true,
```

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -126,6 +129,49 @@ export default defineComponent({
...
+          componentAttrs: {
+            spaceResource: unref(selectedSpaces)[0]
+          }
+        },
+        {
+          app: 'SpaceDetailsMultiple',
+          icon: 'layout-grid',
+          title: $gettext('Space details'),
+          component: SpaceDetailsMultiple,
+          default: false,
```
kulmann, line 158:
```suggestion
          default: true,
```

packages/web-app-admin-settings/src/views/Spaces.vue
kulmann: because it's a root level panel. that's what the `default` means.

packages/web-app-admin-settings/src/views/Spaces.vue
kulmann: because it's a root level panel. that's what the `default` means.

packages/web-pkg/src/components/sideBar/Details/SpaceDetails.vue
```diff
@@ -183,6 +188,16 @@ export default defineComponent({
       return formatDateFromISO(this.space.mdate, this.$language.current)
     },
     ownerUsernames() {
+      // TODO: Find a better solution for reactiveness
```
kulmann, line 191: Please add a comment line why the solution is not good in your opinion.

packages/web-pkg/src/components/sideBar/Details/SpaceDetailsMultiple.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <div class="oc-mt-xl">
+    <div class="oc-flex space-info">
+      <oc-icon name="layout-grid" size="xxlarge" />
+      <p v-translate>Multiple spaces selected</p>
```
kulmann, line 5: Please open a followup issue, the multi selected space panel could show some more info. E.g. how many spaces are selected and what their combined quota values are (total, remaining, used).

kulmann (changes requested):

packages/web-pkg/src/components/sideBar/Details/SpaceDetails.vue
lookacat: yes

lookacat:
Followup: https://github.com/owncloud/web/issues/8222

packages/web-pkg/tests/unit/components/sidebar/Details/SpaceDetails.spec.ts
```diff
@@ -1,4 +1,4 @@
-import SpaceDetails from '../../../../../src/components/SideBar/Details/SpaceDetails.vue'
+import SpaceDetails from 'web-pkg/src/components/SideBar/Details/SpaceDetails.vue'
```
JammingBen, line 1: Should fix the pipeline error for now, I think. We should rename `sideBar` to `SideBar`, but then we need to adjust a lot of imports... So maybe better to do this as a follow-up.

```suggestion
import SpaceDetails from 'web-pkg/src/components/sideBar/Details/SpaceDetails.vue'
```

lookacat: aaaah damn thx, thought i lost my mind

kulmann (approved):
https://github.com/owncloud/web/pull/8192/files#r1067973359 is still unresolved. Other than that good now. Please also add tests in a followup.

JammingBen (approved):

PR #8201, opened 2023-01-10
Vue 3: Remove `vue-async-computed` extension

JammingBen:
## Description
We've removed the `vue-async-computed` extension because it was broken in the right sidebar and we want to use vue-concurrency for such things anyway.

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/package.json
+++ b/packages/web-app-files/package.json
@@ -24,7 +24,6 @@
     "sanitize-html": "^2.7.0",
     "semver": "^7.3.8",
     "uuid": "^9.0.0",
-    "vue-async-computed": "^3.9.0",
     "vue-concurrency": "4.0.0",
     "vue-gettext": "2.1.12",
     "vue-resize": "^1.0.1",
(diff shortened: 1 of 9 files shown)


dschmidt (approved):
Hooray 🥳

kulmann (changes requested):
Doesn't work to full extent: when I click multiple images in a row without waiting for the previews to load, I sometimes get the preview of a previously selected file.

JammingBen:
@kulmann Thx for the hint! Could you please try again?

dschmidt:
@kulmann when you re-test, can you check whether you think `yield new Promise((resolve) => setTimeout(resolve, 500))` is still neccessary? I don't see any flickering

kulmann:
> @kulmann when you re-test, can you check whether you think `yield new Promise((resolve) => setTimeout(resolve, 500))` is still neccessary? I don't see any flickering

I agree, doesn't seem to be needed anymore. And the assignment to `preview.value` could be inlined instead of assigning `yield loadPreview(...)` to a variable and then assigning that to `preview.value`.

Issue #8275, opened 2023-01-19
Loading additional user data does not work when opening sidebar via select-toggle


JammingBen:
### Steps to reproduce
1. In the users table, select your own user via checkbox
2. Toggle the sidebar via the sidebar toggle on the top right side
3. Open the edit panel -> You'll see nothing regarding quota
4. Now click on the pencil-quick action for your user -> sidebar panel reloads -> quota can be edited

This happens because the `loadAdditionalUserDataTask` does not get called when working via the select-toggle.

PR #8276, opened 2023-01-19
Fix sidebar data loading for users

JammingBen:
## Description
This PR improves the loading of additional user data which is needed in the sidebar.

Before, this was handled by a custom `showPanel` event, that first loaded the additional data, then opened the sidebar. Problems here:

1. This event was only fired on action click, not on select (see https://github.com/owncloud/web/issues/8275).
2. We might have deeper component structures in the future. When some deep component (or a composable) wants to emit this event, we'd need to propagate the event all the way to the parent view component.
3. There is no loading-indicator for the user. This might be no issue here, but in general this is bad.

The new approach uses the `eventBus` which is coupled with the `useSidebar` composable. Opening/closing the sidebar and setting the active panels is all being handled by the composable. Furthermore, the additional data is being loaded when a user gets selected. This way, it works on select-toggle as well.

Also, the sidebar now uses the loading state properly to show a loading spinner during data loading.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8275

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-loading-additional-user-data
+++ b/changelog/unreleased/bugfix-loading-additional-user-data
@@ -3,3 +3,4 @@ Bugfix: Loading additional user data
 Loading additional user data in the admin settings when opening the sidebar via select-toggle has been fixed.
 
 https://github.com/owncloud/web/pull/8276
+https://github.com/owncloud/web/issues/8275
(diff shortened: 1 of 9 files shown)


dschmidt (approved):
Nice 👍🏻

PR #8282, opened 2023-01-19
Vue 3: Remove `$on` listeners on instances

JammingBen:
See https://v3-migration.vuejs.org/breaking-changes/events-api.html.

I moved quite some parts to `setup()` in the course of this. Also, the `OcAccordion` component was removed because we don't use it and it's quite painful to make it work without the listener.

## Types of changes

- [x] Technical debt
- [ ] Tests


packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -202,21 +211,23 @@ export default defineComponent({
     const instance = getCurrentInstance().proxy as any
     const uppyService = useService<UppyService>('$uppyService')
     const store = useStore()
+    const filesSelectedSub = ref()
+    const uploadCompletedSub = ref()
```
dschmidt, line 215: Doesn't look like we need this as a Ref?

dschmidt (approved):

Issue #8307, opened 2023-01-24
Sidebar is broken for resources with public links with `pnpm vite`


JammingBen:
### Steps to reproduce
1. Open the sharing panel for a resource
2. Create a public link (not quick link) -> sidebar is broken

Issue #8283, opened 2023-01-19
Indirect links can't be expanded


JammingBen:
### Steps to reproduce
1. Open the sidebar for a resource which has an indirect link
2. Click on "Show" at the bottom -> nothing happens

![image](https://user-images.githubusercontent.com/50302941/213537428-dc76e51a-4795-4634-a609-2c302c18cf6c.png)

PR #8308, opened 2023-01-24
Fix indirect link toggling

JammingBen:
## Description
Toggling indirect links in the sidebar (as well as the broken sidebar state with public links with `pnpm vite`) has been fixed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8283
- Fixes https://github.com/owncloud/web/issues/8307

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
@@ -155,9 +155,7 @@ export default defineComponent({
     const store = useStore()
 
     const linkListCollapsed = !store.getters.configuration.options.sidebar.shares.showAllOnLoad
-    const indirectLinkListCollapsed = ref(
-      !store.getters.configuration.options.sidebar.shares.showAllOnLoad
-    )
+    const indirectLinkListCollapsed = ref(linkListCollapsed)
 
     return {
       ...useGraphClient(),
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -155,8 +155,9 @@ export default defineComponent({
     const store = useStore()
 
     const linkListCollapsed = !store.getters.configuration.options.sidebar.shares.showAllOnLoad
-    const indirectLinkListCollapsed =
+    const indirectLinkListCollapsed = ref(
       !store.getters.configuration.options.sidebar.shares.showAllOnLoad
+    )
```
dschmidt, line 160: Does this really work? Just wrapping a value in ref() shouldn't make it reactive I thought, doesnt this need to be wrapped in computed()?

JammingBen: This works, yes. My idea was to make it mutable, because it needs to be written. `showAllOnLoad ` is just the initial value.

kulmann: Sounds legit @JammingBen we indeed don't want to mutate the config option in the store. using it as initial value and not persist the change in memory is a good solution. well spotted though @dschmidt I also had the same raised eyebrow feeling.

kulmann: Ah, one idea for improvement though:
```suggestion
const indirectLinkListCollapsed = ref(linkListeCollapsed)
```
Otherwise you have the same store reference twice in consecutive lines.

kulmann (approved):
👍

Issue #8284, opened 2023-01-20
A file (already shared with a user as a viewer) when opened through a public link with `internal` role is shown as a folder.


roshanlc:
### Description of issue:

When a file (already shared with a user as a viewer) is opened through a public link with `internal` role by that user, it is shown as a folder.

![image](https://user-images.githubusercontent.com/6222781/213674807-8daf8a1e-a9b9-40a7-8868-f782d99e32b1.png)

### Steps to reproduce
1. Create a folder (say `folderToShare`) and a file (say `fileToShare`) at the root of personal space.
2. Then share both of them with user (say `einstein`) as `Viewer`
3. User `einstein` has accepted both the shares.
 
![image](https://user-images.githubusercontent.com/6222781/213675048-4b037dd4-93a1-4f22-a80f-8b37bb280352.png)

4. Create public links for both the objects ( `folderToShare` and `fileToShare` )  and change the role of both the public links to `internal`
5. Now, try opening the both the links as user `einstein`

### Actual behaviour
For folder, it is shown properly.

![image](https://user-images.githubusercontent.com/6222781/213675268-25c0d1b5-21ed-4e68-a51f-116764b6fe7d.png)

For file, it is shown in a similar way as that of a folder.

![image](https://user-images.githubusercontent.com/6222781/213675355-86cbe3fd-e795-4cfa-8f79-15f02d529366.png)

### Environment general

**Backend (ownCloud Core or Infinite Scale)**: Infinite Scale

### Environment ownCloud Infinite Scale
**Version:**
Tried with two versions:
- Infinite Scale 2.0.0+a4a6c4bf7
- Infinite Scale 2.0.0+a0fee6e04

### Environment ownCloud WebUI
**Version:**
Tried with two versions:
- ownCloud Web UI 6.0.0
- ownCloud Web UI 7.0.0-rc.6

**Operating system:**
```bash
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
```

### Client configuration
**Browser:** 
- Firefox Version 109.0 (64-bit)
- Chromium Version 109.0.5414.74 (Official Build) snap (64-bit)

tbsbdr:
tahnks @roshanlc for bringing up this issue!
@kulmann @JanAckermann reproduced on wopi latest - the internal link does not work for files at all. can you confirm?

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8310.

PR #8310, opened 2023-01-24
Fix resolving internal file links

JammingBen:
## Description
Resolving internal file links when the original file is a shared resource has been fixed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8284

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/pages/resolvePrivateLink.vue
+++ b/packages/web-runtime/src/pages/resolvePrivateLink.vue
@@ -94,7 +94,7 @@ export default defineComponent({
     const resolvePrivateLinkTask = useTask(function* (signal, id) {
       let path
       let matchingSpace = getMatchingSpace(id)
-      let resourceIsInsideShare = false
+      let resourceIsNestedInShare = false
       if (matchingSpace) {
         path = yield clientService.owncloudSdk.files.getPathForFileId(id)
         resource.value = yield clientService.webdav.getFileInfo(matchingSpace, { path })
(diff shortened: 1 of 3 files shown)


JammingBen:
@kulmann I didn't find a convenient way to check if a resource is inside a received share, so I check if we run into our hacky `while`-loop. Maybe you've got a better idea?

packages/web-runtime/src/pages/resolvePrivateLink.vue
```diff
@@ -105,6 +106,7 @@ export default defineComponent({
         const sharePathSegments = mountPoint ? [] : [unref(resource).name]
         let tmpResource = unref(resource)
         while (!mountPoint) {
+          resourceIsInsideShare = true
```
kulmann, line 109: Would be sufficient to set `resourceIsInsideShare = !mountPoint` once after line 104 instead of setting it to `true` in each iteration.

kulmann: While you're at it, what do you think about `resourceIsShareRoot` or `resourceIsNestedInShare` as var name?

kulmann:
> @kulmann I didn't find a convenient way to check if a resource is inside a received share, so I check if we run into our hacky `while`-loop. Maybe you've got a better idea?

No better idea at the moment, no. The hacky while loop is fine for now.

kulmann (approved):
👍

JanAckermann (approved):
LGTM

PR #8313, opened 2023-01-25
[full-ci] Refactor sidebar resource usage in the files-app

JammingBen:
## Description

* `Sidebar.vue` now provides `resource` and `space`:
  * `resource` is the fully loaded resource which should be used by all panels and underlying components.
  * `space` is the space we're currently in.
* I reworked the resource loading a bit as well. It still feels complicated, but I don't know how to simplify it more.
  * Each panel that needs `resource` is now disabled until the `resource` is fully loaded, so we guarantee that prop is always available when needed.
  * The sidebar is in a loading state as long as `currentFolder` is not set (except when being in the root of a space)

We could move all this parts to `setup()`, but first I want to know if this approach makes sense. 

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/SideBar/TagsPanel.vue
+++ b/packages/web-app-files/src/components/SideBar/TagsPanel.vue
@@ -73,8 +73,8 @@ export default defineComponent({
     const store = useStore()
     const { graphClient } = useGraphClient()
 
-    const displayedItem = inject<Resource>('displayedItem')
-    const resource = computed(() => unref(displayedItem))
+    const injectedResource = inject<Resource>('resource')
+    const resource = computed(() => unref(injectedResource))
     const selectedTags = ref([])
     const availableTags = ref([])
     const tagSelect = ref(null)
@@ -116,7 +116,6 @@ export default defineComponent({
           value: [...unref(selectedTags)]
         })
 
-        unref(resource).tags = [...unref(selectedTags)]
         eventBus.publish('sidebar.entity.saved')
       } catch (e) {
         console.error(e)
@@ -124,7 +123,7 @@ export default defineComponent({
     }
 
     watch(resource, () => {
-      if (unref(resource).canEditTags()) {
+      if (unref(resource)?.canEditTags()) {
         revertChanges()
       }
     })
--- a/packages/web-app-admin-settings/src/components/Spaces/SideBar/ActionsPanel.vue
+++ b/packages/web-app-admin-settings/src/components/Spaces/SideBar/ActionsPanel.vue
@@ -3,7 +3,7 @@
     <quota-modal
       v-if="quotaModalIsOpen"
       :cancel="closeQuotaModal"
-      :space="selectedSpaces[0]"
+      :space="resources[0]"
       @space-quota-updated="spaceQuotaUpdated"
     />
     <oc-list id="oc-spaces-actions-sidebar" class-name="oc-mt-s">
@@ -11,7 +11,7 @@
         v-for="(action, index) in actions"
         :key="`action-${index}`"
         :action="action"
-        :items="selectedSpaces"
+        :items="resources"
         class="oc-py-xs"
       />
     </oc-list>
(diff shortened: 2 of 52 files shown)


kulmann (commented):
Yes, this makes total sense, thank you for cleaning up this mess. My reasoning:
- `provide/inject` makes sense for `space` and `resource` because this is (nearly) always the context of the right sidebar and potentially needed in every nesting level of the sub-components.
- `provide/inject` is possible, because we know that we only ever have zero or one right sidebar at any time (`provice/inject` doesn't work from the view down to e.g. the rows in the ResourceTable, because you would inject stuff that's not needed most of the time.
- having the `space` object available is often beneficial, because it's the/a file context root. `space` is a good/consistent name for that injected var.
- for `resource` I'd prefer the more generic var name `item`, because I think that `resource` doesn't fit well vor e.g. users or groups in the admin settings app.

And a little bit of context: the fact that we had var names like `displayedItem` before was due to a reactivity fuckup/hack, that is not needed anymore with the composition api syntax of provide/inject. So yeah, let's have consistent, short var names for the injected context.

JammingBen:
Thx for the feedback 👍 

> for resource I'd prefer the more generic var name item, because I think that resource doesn't fit well vor e.g. users or groups in the admin settings app.

Agreed in general, but we're only in the files-app context here. So `resource` is more fitting, don't you think?

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/32172/72/1
:boom: The oCISTrashbinJourney tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/32172/69/1

ownclouders:
Results for <strong>acceptance oC10</strong> https://drone.owncloud.com/owncloud/web/32172/73/1
:boom: The oC10IntegrationNotifications tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/32172/67/1
:boom: The oCISSharingPerm2 tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oC10</strong> https://drone.owncloud.com/owncloud/web/32172/74/1
:boom: The oC10IntegrationApp1 tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/32172/70/1
:boom: The oCISSharingPublic3 tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/32172/68/1
:boom: The oCISSharingPublic1 tests pipeline failed. The build has been cancelled.

JammingBen:
@kulmann Also, how do you feel about the `highlightedFile` store-getter? We could remove it now, as it's only used by `Sidebar.vue` (and `Projects.vue`, but that can be overcome easily).

kulmann:
> @kulmann Also, how do you feel about the `highlightedFile` store-getter? We could remove it now, as it's only used by `Sidebar.vue` (and `Projects.vue`, but that can be overcome easily).

Kill it with fire 🔥

JammingBen:
> > @kulmann Also, how do you feel about the `highlightedFile` store-getter? We could remove it now, as it's only used by `Sidebar.vue` (and `Projects.vue`, but that can be overcome easily).
> 
> Kill it with fire 🔥

Ahh damn, I was wrong. It's also used when adding/editing shares and, way more painful, when loading share indicators and shares tree 😢 I guess it has to wait then until we refactored those things (I remember that this has to be done as well, right?).

JammingBen:
Alright, this is ready 4 review from my side. Would be nice to have some besides mit smoke-testing this, with a focus on reactivity and the "Shares"-page.

packages/web-app-files/src/components/SideBar/TagsPanel.vue
```diff
@@ -73,8 +73,8 @@ export default defineComponent({
     const store = useStore()
     const { graphClient } = useGraphClient()
 
-    const displayedItem = inject<Resource>('displayedItem')
-    const resource = computed(() => unref(displayedItem))
+    const injectedResource = inject<Resource>('resource')
```
kulmann, line 76: Not needed. You can reduce this to one line:
```js
const resource = inject<Resource>('resource');
```

JammingBen: Yeah I thought so too, but then the unit tests fail horribly with `Cannot read properties of undefined (reading 'isRunning')`. I don't know why this as any effect on the task, but it has something to do with the watcher we have on `resource`.

kulmann (approved):
Nice cleanup 🧹

PR #8320, opened 2023-01-26
Add expiration for space members and apply new permission structure

JammingBen:
## Description
We've added expiration for space members and made Web compatible with the new permission structure that was introduced by https://github.com/owncloud/ocis/pull/5453.

## Types of changes

- [x] Technical debt
- [ ] Tests

To-Dos:

- [x] e2e tests

### DIFF
--- a/packages/web-client/src/helpers/share/space.ts
+++ b/packages/web-client/src/helpers/share/space.ts
@@ -31,6 +31,6 @@ export function buildSpaceShare(s, storageId): Share {
     permissions,
     role,
     expiration: s.expirationDate,
-    expires: new Date(s.expirationDate)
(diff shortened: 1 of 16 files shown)


packages/web-client/src/helpers/share/space.ts
```diff
@@ -29,6 +29,8 @@ export function buildSpaceShare(s, storageId): Share {
       additionalInfo: null
     },
     permissions,
-    role
+    role,
+    expiration: s.expirationDate,
+    expires: s.expirationDate
```
C0rby, line 34: I had another look and the backend is returning the date in RFC3339 format.
If we convert it to a js Date here then `DateTime.fromJSDate(expires)` should work too.
```suggestion
    expires: new Date(s.expirationDate)
```

Issue #8347, opened 2023-02-01
refactor e2e test so that the quick link copied is accessed from clipboard 


SwikritiT:
We have a test step
```feature
And "Brian" copies quick link of the resource "shared_folder" from the context menu
```

Buth this step only clicks the quick link creation button and doesn't get the copied value from clipboard and store it in the `link store` 

Refactor this function so that the URL is stored as well.

- [x] PR to stable (https://github.com/owncloud/web/pull/8499)
- [x] Port to master (https://github.com/owncloud/web/pull/8530)

SwikritiT:
we can get value from the clipboard  through the clipboard api provided by `playwright`

SwikritiT:
Roshan will make a PR for this

SwikritiT:
What if we need to access the link? I would expect that after copying the link we should be able to access the link as well. In the current scenario, we cannot access the link copied from the quick link if we use it in conjunction with this step
```
 When "Anonymous" opens the public link "QuickLink"
```

I would expect we should be able to use them together or is there another way we can access the quick link?

That's my opinion.
 cc @SagarGi @amrita-shrestha @saw-jan 

Others can also add theirs

saw-jan:
As of now, we don't have tests related to quick-link so maybe implement it along with the test as well.

PR #8352, opened 2023-02-02
[tests-only][full-ci] Add a e2e test for quick link

roshanlc:
## Description
With this PR
-  when a quicklink is created then the corresponding url value is added into links environment with name `Quicklink` by extracting it from the clipboard

## Related Issue
- https://github.com/owncloud/web/issues/8347

## Motivation and Context
We have a test step
```feature
And "Brian" copies quick link of the resource "shared_folder" from the context menu
```
In the current scenario, we cannot access the link copied from the quick link if we use it together with this step
```feature
 When "Anonymous" opens the public link "QuickLink"
``` 
and in my expectation, we should be able to do so. As we have the same button to create+copy the quick link, the first time the button is clicked the link gets created and copied, and from the second time onwards the link is copied. Thus if the key `Quicklink` already exists while creating the link we do not add it to the link store otherwise the link is stored in the link store also.

## How Has This Been Tested?
- Locally
- CI

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/e2e/cucumber/features/smoke/link.feature
+++ b/tests/e2e/cucumber/features/smoke/link.feature
@@ -53,3 +53,24 @@ Feature: link
     And "Anonymous" should not be able to open the old link "myPublicLink"
     And "Anonymous" logs out
     And "Alice" logs out
+
+
+  Scenario: Quick link
+    Given "Admin" creates following users
+      | id    |
+      | Alice |
+    When "Alice" logs in
+    And "Alice" opens the "files" app
+    And "Alice" creates the following resources
+      | resource     | type   |
+      | folderPublic | folder |
+    And "Alice" uploads the following resources
+      | resource  | to           |
+      | lorem.txt | folderPublic |
+    And "Alice" copies quick link of the resource "folderPublic" from the context menu
+    When "Anonymous" opens the public link "Quicklink"
+    And "Anonymous" downloads the following public link resources using the sidebar panel
+      | resource     | type |
+      | lorem.txt    | file |
+    And "Anonymous" logs out
+    And "Alice" logs out
--- a/tests/e2e/support/objects/app-files/share/index.ts
+++ b/tests/e2e/support/objects/app-files/share/index.ts
@@ -16,6 +16,7 @@ import {
 import { resourceIsNotOpenable, isAcceptedSharePresent } from './utils'
 import { copyLinkArgs } from '../link/actions'
 import { LinksEnvironment } from '../../../environment'
+import { linkStore } from '../../../store'
 
 export class Share {
   #page: Page
@@ -23,7 +24,6 @@ export class Share {
 
   constructor({ page }: { page: Page }) {
     this.#page = page
-    this.#linksEnvironment = new LinksEnvironment()
   }
 
   async create(args: Omit<createShareArgs, 'page'>): Promise<void> {
@@ -70,12 +70,15 @@ export class Share {
   }
 
   async copyQuickLink(args: Omit<copyLinkArgs, 'page'>): Promise<void> {
-    await copyQuickLink({ ...args, page: this.#page })
-    const quickLink = await this.#page.evaluate(() => navigator.clipboard.readText())
-    this.#linksEnvironment.createLink({
-      key: 'Quicklink',
-      link: { name: 'Quicklink', url: quickLink }
-    })
+    this.#linksEnvironment = new LinksEnvironment()
+    const url = await copyQuickLink({ ...args, page: this.#page })
+    const key = 'Quicklink'
+    if (!linkStore.has(key)) {
+      this.#linksEnvironment.createLink({
+        key: key,
+        link: { name: key, url: url }
+      })
+    }
   }
 
   async resourceIsNotOpenable(resource): Promise<boolean> {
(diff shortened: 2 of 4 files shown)


tests/e2e/support/objects/app-files/share/index.ts
```diff
@@ -15,12 +15,15 @@ import {
...
 import { copyLinkArgs } from '../link/actions'
+import { LinksEnvironment } from '../../../environment'
 
 export class Share {
   #page: Page
+  #linksEnvironment: LinksEnvironment
 
   constructor({ page }: { page: Page }) {
     this.#page = page
+    this.#linksEnvironment = new LinksEnvironment()
```
grgprarup, line 26:
```suggestion
```
We don't need to initialize the links environment for every `Share object`.

tests/e2e/support/objects/app-files/share/index.ts
```diff
@@ -68,6 +71,11 @@ export class Share {
 
   async copyQuickLink(args: Omit<copyLinkArgs, 'page'>): Promise<void> {
     await copyQuickLink({ ...args, page: this.#page })
+    const quickLink = await this.#page.evaluate(() => navigator.clipboard.readText())
+    this.#linksEnvironment.createLink({
+      key: 'Quicklink',
+      link: { name: 'Quicklink', url: quickLink }
+    })
```
amrita-shrestha, line 78: i think we should createLink from action.ts rather than index.ts

amrita-shrestha, line 78: i think this action is actually createAndCopyQuickLink

grgprarup: The `copyQuickLink` method is used to copy links through UI, which won't create links every time. If there is no quick link created then it will create a new link and copy the link otherwise it will just copy the link. This might fail when the same step is used more than 1 time.

SwikritiT: @grgprarup  Isn't `copyQuickLink` just used to copy the quick links? In that case, it should work because you cannot create two quick links at once. Whenever you click the quick link button the quick link either already created or newly created is copied. I don't understand what you're trying to say 😅

grgprarup:
```
this.#linksEnvironment.createLink({
      key: 'Quicklink',
      link: { name: 'Quicklink', url: quickLink }
    })
```

This will try to create a link every time and step `And "Brian" copies quick link of the resource "shared_folder" from the context menu` might fail if a step is used more than 1 time in a single scenario.

SwikritiT: yes true

SwikritiT: Before adding the link to the link store we might want to check if the link store has the link with the key `quick-link` and if the key exists don't add it to the link store.

SwikritiT: we should also clear the link store after the execution of the scenario

ownclouders:
Results for <strong>e2e-tests oC10</strong> https://drone.owncloud.com/owncloud/web/32420/11/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/32420/tracing/file-alice-2023-2-6-08-44-42.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/32420/tracing/file-brian-2023-2-6-08-44-54.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/32420/tracing/quick-link-alice-2023-2-6-08-42-18.zip 


</p></details>

amrita-shrestha (commented):
lgtm

tests/e2e/cucumber/features/smoke/link.feature
```diff
@@ -53,3 +53,24 @@ Feature: link
...
+      | id    |
+      | Alice |
+    When "Alice" logs in
+    And "Alice" opens the "files" app
+    And "Alice" creates the following resources
+      | resource     | type   |
+      | folderPublic | folder |
+    And "Alice" uploads the following resources
+      | resource  | to           |
+      | lorem.txt | folderPublic |
```
saw-jan, line 69: Can be done via API steps

SwikritiT: yes. I didn't realize it. I'll refactor them

tests/e2e/support/objects/app-files/share/index.ts
```diff
@@ -67,7 +70,15 @@ export class Share {
   }
 
   async copyQuickLink(args: Omit<copyLinkArgs, 'page'>): Promise<void> {
-    await copyQuickLink({ ...args, page: this.#page })
+    this.#linksEnvironment = new LinksEnvironment()
+    const url = await copyQuickLink({ ...args, page: this.#page })
```
SagarGi, line 74:
```suggestion
    const url = await copyAndGetQuickLinkUrl({ ...args, page: this.#page })
```

can we name this function like this if possible? both have same function name i guess if i am not wrong.

SwikritiT:
I've created a separate PR here https://github.com/owncloud/web/pull/8499 with the same changes.

PR #8356, opened 2023-02-02
perf: slightly faster key up/down

kulmann:
## Description
Keyboard interaction with file listings needs a major rework (should be a composable instead of a component + is not compatible with the tiles view), so I didn't dig deeper. This PR is just a slight improvement of the render performance of the current implementation of up/down key uses, since it was a low hanging fruit.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/tests/unit/components/FilesList/KeyboardActions.spec.ts
+++ b/packages/web-app-files/tests/unit/components/FilesList/KeyboardActions.spec.ts
@@ -70,14 +70,14 @@ describe('KeyboardActions', () => {
     })
   })
   describe('global shortcuts', () => {
-    beforeEach(() => {
-      document.body.innerHTML =
-        '<table>' +
-        '<tr class="oc-tbody-tr" data-item-id="0"><td></td></tr>' +
-        '<tr class="oc-tbody-tr" data-item-id="1"><td></td></tr>' +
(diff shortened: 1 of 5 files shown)


packages/web-app-files/tests/unit/components/FilesList/KeyboardActions.spec.ts
```diff
@@ -70,14 +70,14 @@ describe('KeyboardActions', () => {
...
-        '</table>'
-    })
+    // beforeEach(() => {
+    //   document.body.innerHTML =
+    //     '<table>' +
+    //     '<tr class="oc-tbody-tr" data-item-id="0"><td></td></tr>' +
+    //     '<tr class="oc-tbody-tr" data-item-id="1"><td></td></tr>' +
+    //     '<tr class="oc-tbody-tr" data-item-id="2"><td></td></tr>' +
+    //     '</table>'
+    // })
```
JammingBen, line 80:
```suggestion
```

JammingBen (changes requested):
I found an edge case:

1. Select files via "shift" + "arrow down" until you reach the end of the files list.
2. Now try to navigate up again without shift -> will work one time, then it's broken.

JammingBen (approved):

Issue #8368, opened 2023-02-04
[Tiles view] Resize tiles


pascalwengerter:
### Expected behaviour
Tiles can be resized using a slider component (HTML `range` input, potentially to be added to the ODS). 
Details for tile sizing have been sketched out in https://github.com/owncloud/web/issues/6126 and https://github.com/owncloud/web/pull/7991 . Independently of the current tile size on bigger viewports, https://github.com/owncloud/web/issues/8367 should be respected on mobile from my understanding

Open tasks:

- [x] Implement slider
- [x] Resize the default file icon with the tile size
- [x] Fix view when using bigger tile-sizes and having a small screen and/or opening the sidebar. Content may float out of the right screen currently.
- [x] Only show tile size slide when tile size is supported

**Bonus:** Use e.g. https://vueuse.org/core/useresizeobserver/#useresizeobserver to load & update currently viewed tiles with bigger thumbnails upon resizing on certain thresholds (only when enlarging tiles, keeping the bigger size of already loaded tiles when decreasing the size should be fine).

ChrisEdS:
Great stuff! I know this is work in progress but, for the record, I saw the following on main branch:

1. Set the tiles view slider to the second position
2. Open a document in OnlyOffice
3. Switch back to files view inside the new tab

### Expected
The tiles view still has the size I set before

### Current
The slider has still the second position but the tiles view is set back to the not enlarged state

<img width="775" alt="image" src="https://user-images.githubusercontent.com/33026403/218250945-066c051b-74d3-4424-ba1a-1a63a4e5ff89.png">

ChrisEdS:
I don't know if it's in scope but as a power user I would like to be able to select images (tick the boxes) with the keyboard.

kulmann:
> I don't know if it's in scope but as a power user I would like to be able to select images (tick the boxes) with the keyboard.

Not really in scope at the moment, but wanted behaviour. I created an issue for it https://github.com/owncloud/web/issues/8429

PR #5052, opened 2021-04-29
Make html container resize-able

pascalwengerter:
## Description
See changelog item

PR #8372, opened 2023-02-05
Fix ResourceTiles on mobile and add basic resize slider for tiles

### LINKED ISSUE #8367: [Tiles view] Fix mobile display of tiles grid
### Expected behaviour
Tiles should fill 100% of the display width on mobile display

### Actual behaviour
One tile gets displayed, but doesn't fill the full width

pascalwengerter:
## Description
Didn't invest the time to create a `OcSlider` component but kept it inline, also doesn't fetch bigger tiles thumbnails since the current resolution seems quite reasonable even on bigger tiles.

## Related Issue
- Fixes #8367
- Addresses parts of #8368

### DIFF
--- a/packages/design-system/src/tokens/ods/size.yaml
+++ b/packages/design-system/src/tokens/ods/size.yaml
@@ -22,3 +22,8 @@ size:
     value: 22px
   form-check-default:
     value: 14px
+  tiles:
+    resizeable:
+      value: 12rem
+    default:
+      value: 14rem
--- a/packages/design-system/src/tokens/ods/size.yaml
+++ b/packages/design-system/src/tokens/ods/size.yaml
@@ -23,7 +23,7 @@ size:
   form-check-default:
     value: 14px
   tiles:
-    resizeable:
+    resize-step:
       value: 12rem
     default:
       value: 14rem
--- a/changelog/unreleased/enhancement-resources-tiles-view
+++ b/changelog/unreleased/enhancement-resources-tiles-view
@@ -2,6 +2,7 @@ Enhancement: Add tiles view for resource display
 
 We've added a switch to change from the known resource table to a tiles view.
 The change gets saved to the url and persisted across resource navigation.
+Tiles can be dynamically resized on screens bigger than mobile, using the "display customization options" dropdown.
 
 https://github.com/owncloud/web/pull/7991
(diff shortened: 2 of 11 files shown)


packages/design-system/src/tokens/ods/size.yaml
```diff
@@ -22,3 +22,8 @@ size:
     value: 22px
   form-check-default:
     value: 14px
+  tiles:
+    resizeable:
```
kulmann, line 26:
```suggestion
    resize-step:
```
or something similar? resizable (typo btw) sounds like a boolean.

kulmann (changes requested):

packages/design-system/src/tokens/ods/size.yaml
pascalwengerter: Great idea, and thanks for the educate - turns out past me already made this mistake :D https://github.com/owncloud/web/pull/5052

packages/design-system/src/tokens/ods/size.yaml
```diff
@@ -22,3 +22,8 @@ size:
     value: 22px
   form-check-default:
     value: 14px
+  tiles:
+    resize-step:
+      value: 12rem
+    default:
```
kulmann, line 28: The default is not in use. Delete it?

packages/design-system/src/tokens/ods/size.yaml
```diff
@@ -22,3 +22,8 @@ size:
     value: 22px
   form-check-default:
     value: 14px
+  tiles:
+    resize-step:
```
kulmann, line 26: If `resize-step` is the only size value for tiles, we could omit the nesting an just have something like
```yml
  tiles-step:
    value: 12rem
```

packages/design-system/src/tokens/ods/size.yaml
pascalwengerter: Good point, thx. It's meant to be used for non-resizable tiles (like the ones on the spaces-projects view), I added the token without actually using it in the component 🙈 see latest commit in this PR

kulmann: aaaah, now it makes sense 😅

kulmann (approved):

PR #8378, opened 2023-02-06
Add groups filter to the users list

JammingBen:
## Description
Add the possibility to filter users by their group assignments.

Needs https://github.com/owncloud/ocis/pull/5533 for the server.

## Screenshot

image

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8377

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-admin-settings/src/components/Users/UsersList.vue
+++ b/packages/web-app-admin-settings/src/components/Users/UsersList.vue
@@ -8,12 +8,7 @@
       autocomplete="off"
     />
     <slot name="filter" />
-    <no-content-message
-      v-if="!users.length"
-      id="admin-settings-users-empty"
-      class="files-empty"
-      icon="user"
-    >
+    <no-content-message v-if="!users.length" icon="user">
       <template #message>
         <span v-text="$gettext('No users in here')" />
       </template>
--- a/packages/web-app-admin-settings/src/components/Users/UsersList.vue
+++ b/packages/web-app-admin-settings/src/components/Users/UsersList.vue
@@ -4,9 +4,10 @@
       id="users-filter"
       v-model="filterTerm"
       class="oc-ml-m oc-my-s"
-      :label="$gettext('Filter users')"
+      :label="$gettext('Search')"
       autocomplete="off"
     />
+    <slot name="filter" />
     <oc-table
       ref="tableRef"
       class="users-table"
--- a/packages/web-app-admin-settings/src/components/Users/UsersList.vue
+++ b/packages/web-app-admin-settings/src/components/Users/UsersList.vue
@@ -8,7 +8,18 @@
       autocomplete="off"
     />
     <slot name="filter" />
+    <no-content-message
+      v-if="!users.length"
+      id="admin-settings-users-empty"
+      class="files-empty"
+      icon="user"
+    >
+      <template #message>
+        <span v-text="$gettext('No users in here')" />
+      </template>
+    </no-content-message>
     <oc-table
+      v-else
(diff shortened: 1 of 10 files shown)


JammingBen:
Blocked until the server side implementation is finished: https://github.com/owncloud/ocis/issues/5487.

packages/web-app-admin-settings/src/components/Users/UsersList.vue
```diff
@@ -4,10 +4,17 @@
       id="users-filter"
       v-model="filterTerm"
       class="oc-ml-m oc-my-s"
-      :label="$gettext('Filter users')"
+      :label="$gettext('Search')"
```
JanAckermann, line 7: Could we make it consistent, for groups and spaces as well?

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -2,7 +2,6 @@
   <div>
     <app-template
       ref="template"
-      :loading="loadResourcesTask.isRunning || !loadResourcesTask.last"
```
JanAckermann, line 5: why was this removed?

JanAckermann (changes requested):

packages/web-app-admin-settings/src/components/Users/UsersList.vue
JammingBen: Good point!

packages/web-app-admin-settings/src/views/Users.vue
JammingBen: It just moved from there to the current view, because the filter triggers the user loading. It should only re-load the users-list, not the entire page.

JanAckermann: can we do the same with the other views and purge it from the app template?

JanAckermann (changes requested):

packages/web-app-admin-settings/src/views/Users.vue
JammingBen: Yes we should do that, but I'd rather do it in a separate PR to keep this PR here about the filter only.

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -156,6 +191,11 @@ export default defineComponent({
       return data
     })
 
+    const filterGroups = (groups) => {
+      const groupIds = groups.map((g) => g.id)
+      loadResourcesTask.perform(false, groupIds)
```
JanAckermann, line 196: loadUserTask instead pls

JanAckermann (changes requested):

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -138,11 +163,21 @@ export default defineComponent({
...
 
-    const loadResourcesTask = useTask(function* (signal) {
-      const usersResponse = yield unref(graphClient).users.listUsers('displayName')
+    const loadResourcesTask = useTask(function* (signal, loadGroups = true, groupIds = null) {
+      const groupFilter = groupIds
+        ?.reduce((acc, id) => {
+          acc += `memberOf/any(m:m/id eq '${id}') and `
+          return acc
+        }, '')
+        .slice(0, -5)
```
JanAckermann, line 172: Would love to see join here

JanAckermann (changes requested):

JanAckermann (approved):
LGTM :1st_place_medal:

PR #8383, opened 2023-02-06
Introduce ancestor meta data to store

JammingBen:
## Description
This PR introduces the `ancestorMetaData` state in the store. It holds information about all ancestor folders: `fileId`, `sharesTypes`, `space`, `parentFolderId`.

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/services/folder/loaderSpace.ts
+++ b/packages/web-app-files/src/services/folder/loaderSpace.ts
@@ -89,7 +89,7 @@ export class FolderLoaderSpace implements FolderLoader {
         }
 
         yield store.dispatch('Files/loadAncestorMetaData', {
-          path: currentFolder.path,
+          folder: currentFolder,
           space,
           client: webdav
         })
--- a/packages/web-app-files/src/services/folder/loaderSpace.ts
+++ b/packages/web-app-files/src/services/folder/loaderSpace.ts
@@ -88,6 +88,12 @@ export class FolderLoaderSpace implements FolderLoader {
           }
         }
 
+        yield store.dispatch('Files/loadAncestorMetaData', {
+          path: currentFolder.path,
+          space,
+          client: webdav
+        })
+
         store.commit('Files/LOAD_FILES', {
           currentFolder,
           files: resources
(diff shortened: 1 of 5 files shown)


packages/web-app-files/src/store/actions.ts
```diff
@@ -522,6 +523,40 @@ export default {
     if (preview) {
       commit('UPDATE_RESOURCE_FIELD', { id: resource.id, field: type, value: preview })
     }
+  },
+
+  loadAncestorMetaData({ commit, state }, { path, space, client }) {
+    const ancestorMetaData = {}
+    const promises = []
+    const parentPaths = getParentPaths(path, true)
```
kulmann, line 531:
```suggestion
    const parentPaths = getParentPaths(path, false)
```

... since we only want to propfind parents, right?

packages/web-app-files/src/store/actions.ts
```diff
@@ -522,6 +523,40 @@ export default {
...
+        ancestorMetaData[parentPath] = cachedData
+        continue
+      }
+
+      promises.push(
+        client
+          .listFiles(
+            space,
+            { path: parentPath },
+            { depth: 0, davProperties: [DavProperty.FileId, DavProperty.ShareTypes] }
```
kulmann, line 545:
```suggestion
            { depth: 0, davProperties: [DavProperty.FileId, DavProperty.FileParent, DavProperty.ShareTypes] }
```

would make sense to grab the parent file id as well - or in a subsequent PR that adds the parent file id to the breadcrumbs.

packages/web-app-files/src/store/actions.ts
JammingBen: As discussed: Let's keep the current folder because we need it anyways.

packages/web-app-files/src/services/folder/loaderSpace.ts
```diff
@@ -88,6 +88,12 @@ export class FolderLoaderSpace implements FolderLoader {
           }
         }
 
+        yield store.dispatch('Files/loadAncestorMetaData', {
```
kulmann, line 91: I'm curious if we can omit the `yield` here. Please try with and without in your share indicators refactoring PR. In theory Vue3 should be much more powerful in updating the file list with the share indicators after initial rendering than Vue2 was.

kulmann (approved):
💪

Issue #8409, opened 2023-02-10
Sorting does not work when `sortBy` and `sortDir` both change


JammingBen:
### Steps to reproduce
1. Go to your personal space and make sure, you default sorting: `sortBy` = `name` and `sortDir` = `asc`
2. Click on the `size` table header to sort by size -> it won't sort by size, but by `name` `desc`

This happens because only one query param is being updated at the time.

dschmidt:
Probably a regression introduced by the vue-router 4 update, which made all navigations async.

PR #8414, opened 2023-02-10
Fix updating route params async

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8409

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-pkg/src/composables/router/useRouteQuery.ts
+++ b/packages/web-pkg/src/composables/router/useRouteQuery.ts
@@ -1,6 +1,8 @@
-import { computed, Ref, unref } from 'vue'
+import { computed, nextTick, Ref, unref } from 'vue'
 import { useRouter } from './useRouter'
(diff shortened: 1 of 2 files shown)


JammingBen:
Current status:

The implementation works, but it reveals another problem: Some query params don't get cleaned up after route change. E.g. when navigating from "Shares" to "Spaces", params like `declined-sort-by` are still present afterwards.

This happens because of the watcher in `useRouteQueryPersisted`. The watcher detects that the query changes on route change (from something to `undefined`), and will then set the default value. Basically, we need a way to tell the watcher to drop certain query params.

The issue exists without this PR as well, but it's not noticable because the last query update overwrites all previous query params right now.

JammingBen:
Closing for now, as we fixed https://github.com/owncloud/web/issues/8409 via https://github.com/owncloud/web/pull/8675.

PR #8431, opened 2023-02-14
[full-ci] Implement permission concept for handling user abilities

JammingBen:
## Description
Until now, Web handles permission checks only based on role names. This PR adds support for permissions that come from the server. The lib `CASL` has been introduced for defining and enforcing those permissions.

The permission concept is a bit complicated because it has 2 "layers": Permissions that are only related to the user and permissions that additionally relate to a resource (like share permissions, space roles). We only care about the former here! That distinction is very important to fully understand how the permissions are structured.

They can be found in [packages/web-runtime/src/services/auth/abilities.ts](https://github.com/owncloud/web/pull/8431/files#diff-3303332eb1a585a4784118a717240526526d4a03dfdc70ab38f009e1ae1db49e) and consist of an `action` and a `subject`. Actions have one of two suffixes: `-all` (system-wide permission) or `-own` (restricted to the current user). E.g.:

* `{ action: 'create-all', subject: 'Account' }` defines the permission to read all accounts of this system
* `{ action: 'read-all', subject: 'Space' }` defines the permission to read all spaces on this system
* `{ action: 'create-own', subject: 'Space' }` defines the permission to create a personal space on this system

All available actions and subjects are defined in [packages/web-pkg/src/utils/types.ts](https://github.com/owncloud/web/pull/8431/files#diff-b1a2d1e3a03629b941fa2093b97ee12c2b812df0bc9c8a3c87d53ae92db86e28).

The `PermissionManager` has been removed as we don't need it anymore.

**Note:** Requires the latest server docker image!

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8428

## Types of changes

- [x] Technical debt
- [ ] Tests


ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/32810/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/32810/tracing/multiple-spaces-can-be-managed-at-once-in-the-admin-settings-via-the-batch-actions-alice-2023-2-15-12-48-10.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/32810/tracing/spaces-can-be-managed-in-the-admin-settings-alice-2023-2-15-12-47-02.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/32810/tracing/unstructured-collection-of-testable-space-interactions-alice-2023-2-15-12-54-25.zip 


</p></details>

packages/web-app-admin-settings/tests/unit/views/Spaces.spec.ts
```diff
@@ -12,14 +12,6 @@ import {
...
 import Spaces from '../../../src/views/Spaces.vue'
 
-window.ResizeObserver =
-  window.ResizeObserver ||
-  jest.fn().mockImplementation(() => ({
-    disconnect: jest.fn(),
-    observe: jest.fn(),
-    unobserve: jest.fn()
-  }))
-
```
JammingBen, line 22: Fyi this has been moved to `jest.init.ts` so we don't have to add this for a dozen of tests.

packages/web-client/src/helpers/space/functions.ts
```diff
@@ -169,20 +169,20 @@ export function buildSpace(data): SpaceResource {
     canDownload: function () {
       return true
     },
-    canBeDeleted: function ({ user }: { user?: User } = {}) {
-      return this.disabled && (user?.role?.name === 'spaceadmin' || this.isManager(user))
+    canBeDeleted: function ({ user, ability }: { user?: User; ability?: any } = {}) {
+      return this.disabled && (ability?.can('delete', 'Space') || this.isManager(user))
```
JammingBen, line 173: I don't really like that we mix the user- and the resource-permission-layers here. IMO this place should be resource-related only.

But I don't know an elegant way to separate this without refactoring more code.

kulmann: If we don't use the permissions here I only see the option to stick to the space role in the resource (i.e. space admin can't do anything if he/she is only viewer in the space) and then let the space admin use the `admin-settings` app for administrative tasks. But then again, I don't know if it would feel inconsistent to be able to do stuff in the `admin-settings` app but not in the files app.

packages/web-client/src/helpers/space/functions.ts
```diff
@@ -169,20 +169,20 @@ export function buildSpace(data): SpaceResource {
...
     },
     canRename: function ({ user }: { user?: User } = {}) {
-      return user?.role?.name === 'spaceadmin' || this.isManager(user)
+      return this.isManager(user)
```
kulmann, line 176: Why did you decide to only remove the role/permission check from some of the functions?

packages/web-app-admin-settings/tests/unit/index.spec.ts
```diff
@@ -50,23 +50,43 @@ describe('admin settings index', () => {
...
+        ability.can.mockReturnValueOnce(false)
+        ability.can.mockReturnValueOnce(true)
+        expect(
+          index
+            .routes({ $ability: ability })
             .find((n) => n.path === '/')
             .redirect().name
         ).toEqual('admin-settings-spaces')
       })
+      it('should throw an error if permission are insufficient', () => {
```
kulmann, line 85:
```suggestion
      it('should throw an error if permissions are insufficient', () => {
```

packages/web-app-admin-settings/src/mixins/general/resetLogo.ts
```diff
@@ -11,6 +11,9 @@ export default {
           label: () => {
             return this.$gettext('Reset logo')
           },
+          isEnabled: () => {
+            return this.$can('update', 'Logo')
```
kulmann, line 15: Shouldn't this be `delete`? 🤔

kulmann (changes requested):
Please add a changelog item. The removal of the permission manager is a breaking change for extension developers.

packages/web-client/src/helpers/space/functions.ts
JammingBen: Because editing is apparently not supported for space admins (except when they are managers of the space). At least that's currently the case...

I'm still a bit confused by the permissions the backend sends, and there are also bugs. We might re-introduce that later. But I don't know how that's going so look like, so I decided to kill it for now.

packages/web-app-admin-settings/src/mixins/general/resetLogo.ts
JammingBen: I wanted to keep it close to the backend permissions (which is `change-logo` only). And technically, you update the logo by resetting it to default. But `delete` kinda applies as well... 🤷

JammingBen:
@kulmann This is in a finished state from my side (for now at least).

The permissions coming from the server still have weird names, that's going to change in the future. Also, there still seem to be bugs with the default permissions of the `space admin` role. I've tracked them here: https://github.com/owncloud/ocis/issues/5414#issuecomment-1434667541.

packages/web-app-admin-settings/src/index.ts
```diff
@@ -15,14 +17,20 @@ const appInfo = {
...
         return { name: 'admin-settings-general' }
       }
-      return { name: 'admin-settings-spaces' }
+      if ($ability.can('read-all', 'Account')) {
+        return { name: 'admin-settings-users' }
+      }
+      if ($ability.can('read-all', 'Space')) {
+        return { name: 'admin-settings-spaces' }
+      }
+      throw Error('Insufficient permissions')
```
kulmann, line 33: isn't there also a `read-all` for `Groups`? 🤔 At least there was a `group-management` permission backend side at some point... if that's the case you could add the default redirect to the group management view as well.

Edit: yes, you're already using the `read-all` for `Groups` for displaying the `Groups` nav item. So it's missing here. :-)

packages/web-client/src/helpers/space/functions.ts
kulmann: ok, thanks for the input 👍

packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -14,20 +15,23 @@ export class AuthService {
...
 
   public hasAuthErrorOccured: boolean
 
   public initialize(
     configurationManager: ConfigurationManager,
     clientService: ClientService,
     store: Store<any>,
-    router: Router
+    router: Router,
+    $ability: Ability
```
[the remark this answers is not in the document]
kulmann: I'm confused. You resolved the conversation, gave me a thumbs up, but only changed the var name in the `userManager`, but not in the `authService`. 🙈 I meant both the `userManager` and the `authService`.

kulmann (changes requested):

kulmann (approved):
🥳 🚀

packages/web-runtime/src/services/auth/authService.ts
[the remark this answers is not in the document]
JammingBen: Ahh sorry, forgot to change it there as well!

packages/web-app-admin-settings/src/mixins/general/resetLogo.ts
kulmann: thought about it again... `delete` is actually too technical. Good that you kept it at `update`. 👍

PR #8387, opened 2023-02-07
Spaces/User admin list edit quota batch actions

lookacat:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #8438, opened 2023-02-15
#8387 Spaces/User admin list edit quota batch actions - Followup

lookacat:
## Description
See https://github.com/owncloud/web/pull/8387

## Issue
Fixes https://github.com/owncloud/web/issues/8417

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-admin-settings/src/views/Spaces.vue
+++ b/packages/web-app-admin-settings/src/views/Spaces.vue
@@ -135,7 +135,7 @@ export default defineComponent({
     const { sideBarOpen, sideBarActivePanel } = useSideBar()
 
     const loadResourcesEventToken = ref(null)
-    const updateQuotaForSpaceEventToken = ref(null)
+    const updateQuotaForSpaceEventToken = null
     const template = ref(null)
     const listHeaderPosition = ref(0)
     const selectedSpaces = ref([])
--- a/packages/web-app-admin-settings/src/views/Spaces.vue
+++ b/packages/web-app-admin-settings/src/views/Spaces.vue
@@ -135,7 +135,7 @@ export default defineComponent({
     const { sideBarOpen, sideBarActivePanel } = useSideBar()
 
     const loadResourcesEventToken = ref(null)
-    const updateQuotaForSpaceEventToken = null
+    let updateQuotaForSpaceEventToken
     const template = ref(null)
     const listHeaderPosition = ref(0)
     const selectedSpaces = ref([])
@@ -265,7 +265,7 @@ export default defineComponent({
       window.addEventListener('resize', calculateListHeaderPosition)
       resizeObserver.observe(unref(appBarActionsRef))
 
-      updateQuotaForSpaceEventToken.value = eventBus.subscribe(
+      updateQuotaForSpaceEventToken = eventBus.subscribe(
         'app.admin-settings.spaces.space.quota.updated',
         ({ spaceId, quota }) => {
           const space = unref(spaces).find((s) => s.id === spaceId)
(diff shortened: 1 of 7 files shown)


packages/web-pkg/src/components/Spaces/QuotaModal.vue
```diff
@@ -69,7 +69,11 @@ export default defineComponent({
     }
   },
   mounted() {
-    this.selectedOption = this.spaces[0].spaceQuota.total || 0
+    if (!this.spaces.length) {
+      this.selectedOption = 0
+      return
+    }
+    this.selectedOption = this.spaces[0].spaceQuota?.total || 0
```
kulmann, line 76:
```suggestion
    this.selectedOption = this.spaces[0]?.spaceQuota?.total || 0
```
With this the early return above is not needed. `this.spaces[0]` is still valid when the array is empty.

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -135,7 +135,7 @@ export default defineComponent({
     const { sideBarOpen, sideBarActivePanel } = useSideBar()
 
     const loadResourcesEventToken = ref(null)
-    const updateQuotaForSpaceEventToken = ref(null)
+    const updateQuotaForSpaceEventToken = null
```
kulmann, line 138: Needs to stay `ref(null)`. Why did you change it? 🤔

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -219,11 +219,12 @@ export default defineComponent({
...
-      const requests = []
-      unref(selectedUsers).forEach((user) => {
-        requests.push(loadAdditionalUserDataTask.perform(user))
-      })
-      await Promise.all(requests)
+      const requests = unref(selectedUsers).map((user) => loadAdditionalUserDataTask.perform(user))
+      const results = await Promise.allSettled<Array<unknown>>(requests)
+      const failedRequests = results.filter((result) => result.status === 'rejected')
+      if (failedRequests.length > 0) {
+        console.error('Failed to load additional user data', failedRequests)
```
kulmann, line 226: This will only fail for users who never logged in, right? Since that's expected, shouldn't we avoid the console noise? Maybe reducing the log level to `debug` instead of `error` is an option

packages/web-app-admin-settings/src/views/Spaces.vue
JammingBen: I think that refers to a comment I made in the last PR, where I said that this does not need to be a `ref` because it's not reactive. Although you need to make it a `let` and remove `.value` (and/or `unref()`) down below.

kulmann: Ah ok. Yes, simply removing the wrapping `ref()` is not sufficient. That's why I was confused.

packages/web-app-admin-settings/src/views/Users.vue
lookacat: sounds good, just felt bad to not have any error handling

fschade (approved):
one last thing, please update the related changelog to contain this pr too.
Everything else, LGTM

kulmann (approved):
👍

kulmann:
> one last thing, please update the related changelog to contain this pr too.
> Everything else, LGTM

From what I see that was already the case :-) If I overlooked something please followup again

PR #8441, opened 2023-02-15
add custom search provider example

fschade:
The skeleton app has been part of the project for a long time, but with the conversion to vite it has since been ignored and no longer transpiled.
Due to the change, the app is now taken into account again, but must be explicitly enabled. For this please see the associated APP README.

Also new is a search example that is now included and uses GitHub to show how a custom search provider can be developed.


diocas:
hi fschade I tried your gh serach provider.. I know it's just a demo, but found some issues and decided to report it...

Basically, the API returns thousands of results, but in the dropdown it only says "show 3 results". Then, when you click it, it takes you to the results page, but it crashes and shows an empty page.

Btw, it would be great to be able to disable the default provider.. at CERN's fork we replaced it, but once we take your skeleton as an inspiration and put it as a separate app, it would be great to only show ours.

fschade:
> hi @fschade I tried your gh serach provider.. I know it's just a demo, but found some issues and decided to report it...
> 
> Basically, the API returns thousands of results, but in the dropdown it only says "show 3 results". Then, when you click it, it takes you to the results page, but it crashes and shows an empty page.
> 
> Btw, it would be great to be able to disable the default provider.. at CERN's fork we replaced it, but once we take your skeleton as an inspiration and put it as a separate app, it would be great to only show ours.

@diocas thanks for your feedback, unfortunately I got sick and won’t be able to work for the rest of this week. I fix the mentioned points next week.

fschade:
@diocas im back from my sick leave.
Thanks for the hint `show 3 results`, i removed it if the provider does not have a listSearch component.

@kulmann is it ok for you to add a option which search providers are available?
Then we can make the `files.sdk` provider default but overwritable. Thoughts?

kulmann:
> @kulmann is it ok for you to add a option which search providers are available? Then we can make the `files.sdk` provider default but overwritable. Thoughts?

Being able to customize the search providers (with `files.sdk` being the default) is a good thing, yes 👍

fschade:
@kulmann THX, ill create a follup up pr after the sprint tasks are done

packages/web-app-skeleton/src/index.ts
```diff
@@ -8,7 +10,11 @@ const appInfo = {
   extensions: []
 }
 
-const injectExtensions = async (api) => {
+const injectSearch = (): void => {
+  eventBus.publish('app.search.register.provider', new GitHUbSearch())
```
lookacat, line 14: Would prefer `GithubSearch` as name, looks smoother or at least `GitHubSearch`

packages/web-app-skeleton/src/search/github/PreviewComponent.vue
```diff
@@ -0,0 +1,34 @@
...
+    :href="repository.html_url"
+  >
+    {{ repository.name }}
+  </oc-button>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+import { Repository } from './client'
+import { PropType } from 'vue'
```
lookacat, line 17: would combine first import and this import into one

packages/web-app-skeleton/src/search/github/PreviewComponent.vue
```diff
@@ -0,0 +1,34 @@
...
+
+export default defineComponent({
+  props: {
+    searchResult: {
+      type: Object as PropType<Repository>,
+      default: function () {
+        return {}
+      }
```
lookacat, line 25: i think just `default: {}` should be fine

packages/web-app-skeleton/src/search/github/client.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface RepositoriesSearch {
+  total_count: number
+  items: Repository[]
+}
+
+export class GitHubClient {
+  private readonly httpClient: HttpClient
+
+  constructor(httpClient: HttpClient) {
```
lookacat, line 19: you can use short hand constructor in typescript like this:
`constructor(
    private readonly httpClient: HttpClient
  ){}`

lookacat: Then you don't can remove line 17

lookacat (approved):
Some small minor things, but fine to me 👍🏼

packages/web-app-skeleton/src/search/github/client.ts
fschade: not a fan of that shorthand, in my opinion wayyyy to implicit.

lookacat: fine for me too :)

packages/web-app-skeleton/src/search/github/PreviewComponent.vue
fschade: (vue/require-valid-default-prop)[https://eslint.vuejs.org/rules/require-valid-default-prop.html]

packages/web-app-skeleton/src/search/github/PreviewComponent.vue
fschade: vue/require-valid-default-prop, https://eslint.vuejs.org/rules/require-valid-default-prop.html.

Must be of type function

fschade:
@diocas, follow up issue -> https://github.com/owncloud/web/issues/8524

PR #8442, opened 2023-02-15
WIP: new file editor view

diocas:
cc @pascalwengerter

PR #8447, opened 2023-02-16
[full-ci] Refactor extension appBar design/concept

### LINKED ISSUE #8422: Universal "app context" top bar
when opening a file in a viewer or editor in ownCloud web, it should display an area in the top bar that shows the file name, a "three dot" menu offering actions of the editor, and possibly a close button that closes the editor.

The idea is to integrate that into the topmost darkblue bar of oC Web to not waste vertical space for that.

CERN has a prototype implementation that should be integrated into the product.

@tbsbdr @diocas @elizavetaRa

pascalwengerter:
## Description
Details (esp. design) tbd, just wanted to make this take into consideration the existing apps (and try to limit the codechanges needed)

## Related Issue
- Supersedes #8442
- Fixes #8422


ownclouders:
Results for <strong>e2e-tests oC10</strong> https://drone.owncloud.com/owncloud/web/33136/11/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/33136/tracing/search-in-personal-spaces-alice-2023-2-27-07-09-58.zip 

#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/33136/tracing/search-in-personal-spaces-brian-2023-2-27-07-09-49.zip 


</p></details>
:boom: The e2e-oc10 tests pipeline failed. The build has been cancelled.

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,24 +1,52 @@
...
+  <portal to="app.runtime.header.left">
+    <div class="oc-app-top-bar oc-p-xs oc-flex oc-flex-center">
+      <div class="open-file-bar oc-flex oc-mr-s">
+        <oc-resource
+          v-if="resourceReady"
+          id="open-file-bar-resource"
+          :is-thumbnail-displayed="false"
+          :resource="resource"
+        />
+        <oc-resource v-else :resource="{ name: 'Loading' }" />
```
pascalwengerter, line 11: What to do with the type error here? The `<oc-resource />` expects a full resource

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,24 +1,52 @@
...
+
+          <oc-drop
+            drop-id="oc-openfile-contextmenu"
+            toggle="#oc-openfile-contextmenu-trigger"
+            mode="click"
+            close-on-click
+            padding-size="small"
+            @click.stop.prevent
+          >
+            <slot name="contextActions" />
```
pascalwengerter, line 30: @kulmann slots or creating a list by looping over an array of a clearly defined interface/type?

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,24 +1,52 @@
...
+          </oc-drop>
+        </template>
+      </div>
+
+      <div class="oc-flex main-actions oc-my-s">
+        <slot name="mainActions" />
```
pascalwengerter, line 36: @kulmann slots or creating a list by looping over an array of a clearly defined interface/type?

packages/web-pkg/src/components/AppTopBar.vue
kulmann: We have the interfaces for actions now :-) Would prefer looping over actions instead of a slot.

packages/web-pkg/src/components/AppTopBar.vue
kulmann: Same as above, even the same type for the prop, but different prop. :-)

packages/web-pkg/src/components/AppTopBar.vue
pascalwengerter: Decided to try and go with a `<span>` and give it the `.oc-resource` class for styling/positioning purposes

packages/web-app-external/src/App.vue
```diff
@@ -64,12 +63,13 @@ export default defineComponent({
...
+    appUrl: '',
+    errorMessage: '',
+    formParameters: {},
     loading: false,
     loadingError: false,
-    errorMessage: '',
-    appUrl: '',
     method: '',
-    formParameters: {}
+    resource: <Resource>{}
```
pascalwengerter, line 72: I figured we'd need the resource for the `AppTopBar` 😇  and since we do an API call in the `created()` hook anyways we might as well fetch the whole thing

packages/web-app-external/tests/unit/__snapshots__/app.spec.ts.snap
```diff
@@ -41,6 +44,7 @@ exports[`The app provider extension should fail for unauthenticated users 1`] =
 exports[`The app provider extension should show a loading spinner while loading 1`] = `
 <main class="oc-height-1-1 oc-flex oc-flex-center oc-flex-middle">
   <h1 class="oc-invisible-sr">"exampleApp" app page</h1>
+  <app-top-bar-stub resource="undefined"></app-top-bar-stub>
```
pascalwengerter, line 47: Kinda confused because I thought I'd be able to mock `data` within the wrapper, but couldn't get it to work properly

ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/36230/58/1

<p>


### webUITextEditor-textFile_feature-L21.png
  

### webUITextEditor-textFile_feature-L29.png
  


</p></details>

packages/web-app-external/tests/unit/__snapshots__/app.spec.ts.snap
lookacat: maybe `wrapper.vm.resource = xyz ` before the snapshot works? But just a wild guess

packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -155,9 +152,18 @@ export default {
...
+  data: function () {
+    return {
+      contentOnLeftPortal: false
+    }
+  },
   computed: {
     ...mapGetters(['configuration', 'user']),
 
+    showCenterSlot() {
+      return !this.contentOnLeftPortal
```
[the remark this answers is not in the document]
lookacat: a nevermind in now get it ^^

lookacat:
Screenshot (39)
on small thing i noticed is the search icon margin to the right is off at smaller screens

packages/web-runtime/src/components/Topbar/TopBar.vue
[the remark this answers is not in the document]
pascalwengerter: Yeah it's sadly not intuitive, very open to suggestions for simplification/more explicitness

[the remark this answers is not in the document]
pascalwengerter: Yeah it's sadly not intuitive, very open to suggestions for simplification/more explicitness

pascalwengerter:
> Screenshot (39) on small thing i noticed is the search icon margin to the right is off at smaller screens

Looks more like the smaller `gap` between the other icons got lost on the way ;) I'll take a look later

JammingBen (commented):
Nice, I really like the feature! A few things I noticed though:

* Is it really intended that the search bar goes missing? I would at least expect the search icon like in the mobile view.
* Is it possible to align the tabs to the left? Having them centered feels a bit weird to me.
* The disabled-/enabled-states of the "save"-button in the text-editor hardly differ from each other, especially in light mode. That makes it hard to tell if the button is clickable or not.

packages/web-app-external/tests/unit/__snapshots__/app.spec.ts.snap
JammingBen: When `resource` is mocked via `mock` or `mockDeep` it will result in `undefined` in the snapshots unfortunately. Probably because those return proxy objects, but I can't give you the exact reasoning.

dschmidt: Interesting, maybe we can fix that in the snapshot lib

kulmann:
> Nice, I really like the feature! A few things I noticed though:
> 
> * Is it really intended that the search bar goes missing? I would at least expect the search icon like in the mobile view.
> * Is it possible to align the tabs to the left? Having them centered feels a bit weird to me.
> * The disabled-/enabled-states of the "save"-button in the text-editor hardly differ from each other, especially in light mode. That makes it hard to tell if the button is clickable or not.

Strongly agree with @JammingBen's opinion on all three points... the first and second are blockers for me on the course of getting this merged.

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,45 +1,151 @@
...
 
-<style lang="scss" scoped>
-.app-top-bar {
-  align-items: center;
-  justify-content: space-between;
+<style lang="scss">
+.oc-app-top-bar {
+  align-self: end;
+  grid-column: 1 / 4;
+  grid-row: secondRow;
```
kulmann, line 105: nice 😎

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,45 +1,151 @@
...
+  components: {
+    ContextActionMenu
+  },
   props: {
+    dropDownActions: {
+      type: Array as PropType<Action[]>,
+      default: () => []
+    },
+    mainActions: {
+      type: Array as PropType<Action[]>,
```
kulmann, line 81: should we limit this to a certain number of actions? if yes, then you could enforce that via a validator.

diocas:
Pasting my feedback from ownCloud talk:
> I'm trying it and is not working that well... i.e external apps don't show anything (not even the logo). But, since we haven't used the latest web, I might need to change settings.json...

> And did we lose the extra actions like download or add to favs? I suppose this is app dependent? it shows on the photo viewer, but not on pdf viewer, for example.

> finally, the radius seems unproportionate to the others in the UI

packages/web-pkg/src/components/AppTopBar.vue
pascalwengerter: I'm open for the suggestion, should I limit it to...one? Two?

pascalwengerter:
> Pasting my feedback from ownCloud talk:
> 
> > I'm trying it and is not working that well... i.e external apps don't show anything (not even the logo). But, since we haven't used the latest web, I might need to change settings.json...
> 
> > And did we lose the extra actions like download or add to favs? I suppose this is app dependent? it shows on the photo viewer, but not on pdf viewer, for example.
> 
> > finally, the radius seems unproportionate to the others in the UI

Thanks for porting your feedback ;)

Responses in order
- haven't tested it with external apps myself, current setup doesn't work for me (but will debug now)
- actions need to be registered per extension - media viewer and PDF viewer e.g. don't get a dedicated "download" action since they both feature a download button within the actual extension "body"
- will update the radius, good point

pascalwengerter:
> > Nice, I really like the feature! A few things I noticed though:
> > 
> > * Is it really intended that the search bar goes missing? I would at least expect the search icon like in the mobile view.
> > * Is it possible to align the tabs to the left? Having them centered feels a bit weird to me.
> > * The disabled-/enabled-states of the "save"-button in the text-editor hardly differ from each other, especially in light mode. That makes it hard to tell if the button is clickable or not.
> 
> Strongly agree with @JammingBen's opinion on all three points... the first and second are blockers for me on the course of getting this merged.

Thanks for the feedback. In order:
- _search bar missing_ - from the original designs I figured that's a tradeoff everyone is fine with (at least for now), adding this in a sane way will be a pain but here we go I guess^^
- _align tabs to the left_ - looking at https://github.com/owncloud/web/issues/8422 again I realise you're right and the apptopbar is displayed aligned to the left, however without background color (which was requested from oC) it felt a bit lost which is why I think I centered it...will re-align it to the left
- _action visibility_ - true, will take another look. Got lost in the appearance/variation/configurable background color jungle and will need to adjust `ActionMenuItem` component/type to expect and apply more configuration options

packages/web-pkg/src/components/AppTopBar.vue
kulmann: on second thought: it's the choice of the developer. If it looks awful or is unusable, they will reduce it themselves based on user feedback. So might be overengineered to limit the amount of main actions... and we don't need to document anywhere that it's limited to x main actions. :trollface: so let's keep it unlimited for now. sorry for the noise

tbsbdr:
relates to https://github.com/owncloud/web/pull/9159/commits/31e2ad66bfa301a145b4c14c0c77080c742f8618

JammingBen:
@pascalwengerter After talking with @tbsbdr : it's okay to drop the search bar, at least for now. Alignment to the left and the button color are still valid points. Also, we think it might be better to make the tab more decent looking. Goal: it should be clear that actions belong to the tab. Let's do it like that:

image

Just ping me if I can help you with some of these points as we want to get that merged soon.

> I understand, but each app has it's own place for the download button. Having a consistent place (even if duplicated) for users to search for it would be an improvement, in my view.

I agree, although I'd suggest doing that in a follow-up PR.

packages/web-app-preview/src/App.vue
```diff
@@ -190,23 +179,46 @@ export default defineComponent({
...
       })
     }
+    const isFileContentLoading = ref(true)
+
+    const triggerActiveFileDownload = () => {
+      if (isFileContentLoading.value) {
+        return
+      }
+
+      useDownloadFile().downloadFile(activeFilteredFile.value)
```
dschmidt, line 189: while this probably works here, composables should be instantiated on top level in `setup()`

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,45 +1,147 @@
...
+      default: () => []
+    },
     resource: {
       type: Object as PropType<Resource>,
-      default: null
+      default: () => {}
```
dschmidt, line 82: Why are we having a default at all? Should we just make it required?

packages/web-pkg/src/components/ContextActions/ContextActionMenu.vue
```diff
@@ -21,11 +21,11 @@
 <script lang="ts">
 import { defineComponent, PropType } from 'vue'
 import ActionMenuItem from './ActionMenuItem.vue'
-import { ActionOptions } from '../../composables/actions'
+import { Action, ActionOptions } from '../../composables/actions'
 
 type MenuSection = {
   name: string
-  items: any[]
+  items: Action[]
```
dschmidt, line 28: 🎉

dschmidt (approved):
Cannot say much about the style changes, the code looks mostly very good.

I'm just a bit unhappy with the empty default resource objects, those are not valid resources - why are we having them? Let's just make the resource mandatory everywhere?

pascalwengerter:
> Cannot say much about the style changes, the code looks mostly very good.
> 
> I'm just a bit unhappy with the empty default resource objects, those are not valid resources - why are we having them? Let's just make the resource mandatory everywhere?

I'm not happy with it either tbh, but I'm not sure how I'd handle the loading state "inside" the component while still having the resource as a required prop. Any suggestions?

packages/web-pkg/src/components/AppTopBar.vue
```diff
@@ -1,45 +1,130 @@
...
-  emits: ['close']
+  emits: ['close'],
+  computed: {
+    contextMenuLabel() {
+      return this.$gettext('Show context menu')
+    },
+    closeButtonLabel() {
+      return this.$gettext('Close')
+    }
+  }
```
dschmidt, line 75: Can we add new code in compostion api please?

packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -155,9 +152,18 @@ export default {
       appMenuItems
     }
   },
+  data: function () {
+    return {
+      contentOnLeftPortal: false
+    }
+  },
```
dschmidt, line 159: Can we add new code in composition api?

dschmidt (commented):
In general I'm happy with this, just get rid of the empty resource objects and if by any means reasonable add new code in composition api...

JammingBen (commented):
The buttons look very nice now 👍  The tabs are still aligned in the center and missing their border though (see https://github.com/owncloud/web/pull/8447#issuecomment-1604006063).

pascalwengerter:
> The buttons look very nice now 👍 The tabs are still aligned in the center and missing their border though (see [#8447 (comment)](https://github.com/owncloud/web/pull/8447#issuecomment-1604006063)).

Alignment was a major PITA and yeah, I did mess up the ODS color token/theme situation but should be fixed now ;)

packages/web-pkg/src/components/AppTopBar.vue
pascalwengerter: Done ;) thx

packages/web-runtime/src/components/Topbar/TopBar.vue
pascalwengerter: Done ;) thx

packages/web-app-pdf-viewer/src/App.vue
```diff
@@ -33,7 +34,7 @@ export default defineComponent({
     loading: true,
     loadingError: false,
     url: '',
-    resource: null
+    resource: <Resource>{}
```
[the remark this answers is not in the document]
pascalwengerter: Done ;) thx

packages/web-app-external/src/App.vue
```diff
@@ -64,12 +65,13 @@ export default defineComponent({
...
+    appUrl: '',
+    errorMessage: '',
+    formParameters: {},
     loading: false,
     loadingError: false,
-    errorMessage: '',
-    appUrl: '',
     method: '',
-    formParameters: {}
+    resource: <Resource>null
```
dschmidt, line 74: You don't need to cast null

packages/web-app-pdf-viewer/src/App.vue
```diff
@@ -33,7 +34,7 @@ export default defineComponent({
     loading: true,
     loadingError: false,
     url: '',
-    resource: null
+    resource: <Resource>null
```
dschmidt, line 37: Same

saw-jan:
The drone-ci failure is due to the new version availability of the notifications app. We can either upgrade the app with `occ upgrade` or pull from the commit that has the previous version.

JammingBen (approved):
Awesome! 🥳

PR #8460, opened 2023-02-16
Fix fileExtensionShown toggle for ResourceTiles

pascalwengerter:
## Description
Random find, worth the fix I suppose

### DIFF
--- a/packages/web-app-files/src/components/FilesList/ResourceTiles.vue
+++ b/packages/web-app-files/src/components/FilesList/ResourceTiles.vue
@@ -41,6 +41,7 @@
           :resource="resource"
           :resource-route="getRoute(resource)"
           :is-resource-selected="isResourceSelected(resource)"
+          :is-extension-displayed="areFileExtensionsShown"
           @vue:mounted="
             $emit('rowMounted', resource, tileRefs.tiles[resource.id], ImageDimension.Tile)
           "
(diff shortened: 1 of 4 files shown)


packages/web-app-files/src/components/FilesList/ResourceTiles.vue
```diff
@@ -287,6 +289,9 @@ export default defineComponent({
     return {
       ImageDimension
     }
+  },
+  computed: {
+    ...mapState('Files', ['areFileExtensionsShown'])
   }
```
JammingBen, line 295: You can do that via Composition API as well: `store.getters['Files/areFileExtensionsShown']` 🙂

JammingBen (changes requested):

packages/web-app-files/src/components/FilesList/ResourceTiles.vue
pascalwengerter: Applied, good point

JammingBen: Ahh sorry, apparently there is no `getter`, so you need to go via the `state`: `store.state.Files.areFileExtensionsShown`.

And you need to move it into a computed to be reactive then.

pascalwengerter: Now we should be good

PR #8473, opened 2023-02-17
Update CERN links

diocas:
Updated the links displayed in the sidebar for CERN's deployment.

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -87,8 +87,8 @@
             />
           </td>
         </tr>
-        <tr v-if="runningOnEos" data-testid="eosPath">
-          <th scope="col" class="oc-pr-s oc-font-semibold" v-text="$gettext('EOS Path')" />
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -207,20 +237,37 @@ export default defineComponent({
...
+      copy(getSambaPath(unref(resource).path))
+      copiedSamba.value = unref(copied)
+      store.dispatch('showMessage', {
+        title: $gettext('Windows path copied'),
+        desc: $gettext('The Windows path has been copied to your clipboard.')
       })
     }
 
+    const getSambaPath = (path) => {
+      const pathComponents = path?.split('/').filter(Boolean)
```
kulmann, line 264: Can you please define the `pathMappings` inline here instead of having them in their own file? Completely CERN specific, so a dedicated file is probably more confusing than just having it inline here.

diocas: Done, please let me know if it needs to be somewhere else. Also rebased to latest upstream.

kulmann (approved):
👍

PR #8492, opened 2023-02-23
Add role filter to users view

JammingBen:
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-admin-settings/src/views/Users.vue
+++ b/packages/web-app-admin-settings/src/views/Users.vue
@@ -62,7 +62,25 @@
                   :allow-multiple="true"
                   display-name-attribute="displayName"
                   :filterable-attributes="['displayName']"
+                  class="oc-mr-s"
                   @selection-change="filterGroups"
+                >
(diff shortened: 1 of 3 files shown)


JanAckermann (approved):

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -71,6 +72,23 @@
...
+                  :filter-label="$gettext('Roles')"
+                  :items="roles"
+                  :show-filter="true"
+                  :allow-multiple="true"
+                  display-name-attribute="displayName"
+                  :filterable-attributes="['displayName']"
+                  @selection-change="filterRoles"
+                >
+                  <template #image="{ item }">
+                    <avatar-image :width="32" :userid="item.id" :user-name="item.displayName" />
```
JanAckermann, line 86: do we need i18n here?

JammingBen: Good catch, yes we do!

PR #8499, opened 2023-02-24
[tests-only][full-ci] Add a e2e test for quick link

SwikritiT:
## Description
With this PR
-  when a quicklink is created then the corresponding url value is added into links environment with name `Quicklink` by extracting it from the clipboard

## Related Issue
- https://github.com/owncloud/web/issues/8347

## Motivation and Context
We have a test step
```feature
And "Brian" copies quick link of the resource "shared_folder" from the context menu
```
In the current scenario, we cannot access the link copied from the quick link if we use it together with this step
```feature
 When "Anonymous" opens the public link "QuickLink"
``` 
and in my expectation, we should be able to do so. As we have the same button to create+copy the quick link, the first time the button is clicked the link gets created and copied, and from the second time onwards the link is copied. Thus if the key `Quicklink` already exists while creating the link we do not add it to the link store otherwise the link is stored in the link store also.

## How Has This Been Tested?
- Locally
- CI

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/e2e/cucumber/features/smoke/link.feature
+++ b/tests/e2e/cucumber/features/smoke/link.feature
@@ -59,18 +59,18 @@ Feature: link
     Given "Admin" creates following users
       | id    |
       | Alice |
+    And "Alice" creates the following folders in personal space using API
+      | name         |
+      | folderPublic |
+    And "Alice" creates the following files into personal space using API
+      | pathToFile             | content     |
+      | folderPublic/lorem.txt | lorem ipsum |
     When "Alice" logs in
     And "Alice" opens the "files" app
-    And "Alice" creates the following resources
-      | resource     | type   |
-      | folderPublic | folder |
-    And "Alice" uploads the following resources
-      | resource  | to           |
-      | lorem.txt | folderPublic |
     And "Alice" copies quick link of the resource "folderPublic" from the context menu
     When "Anonymous" opens the public link "Quicklink"
     And "Anonymous" downloads the following public link resources using the sidebar panel
-      | resource     | type |
-      | lorem.txt    | file |
+      | resource  | type |
+      | lorem.txt | file |
     And "Anonymous" logs out
     And "Alice" logs out
--- a/tests/e2e/cucumber/features/smoke/link.feature
+++ b/tests/e2e/cucumber/features/smoke/link.feature
@@ -53,3 +53,24 @@ Feature: link
     And "Anonymous" should not be able to open the old link "myPublicLink"
     And "Anonymous" logs out
     And "Alice" logs out
+
+
+  Scenario: Quick link
+    Given "Admin" creates following users
+      | id    |
+      | Alice |
+    When "Alice" logs in
+    And "Alice" opens the "files" app
+    And "Alice" creates the following resources
+      | resource     | type   |
+      | folderPublic | folder |
+    And "Alice" uploads the following resources
+      | resource  | to           |
+      | lorem.txt | folderPublic |
+    And "Alice" copies quick link of the resource "folderPublic" from the context menu
+    When "Anonymous" opens the public link "Quicklink"
+    And "Anonymous" downloads the following public link resources using the sidebar panel
+      | resource     | type |
+      | lorem.txt    | file |
+    And "Anonymous" logs out
+    And "Alice" logs out
--- a/tests/e2e/cucumber/features/smoke/link.feature
+++ b/tests/e2e/cucumber/features/smoke/link.feature
@@ -65,7 +65,7 @@ Feature: link
     And "Alice" creates the following files into personal space using API
       | pathToFile             | content     |
       | folderPublic/lorem.txt | lorem ipsum |
-    When "Alice" logs in
+    And "Alice" logs in
     And "Alice" opens the "files" app
     And "Alice" copies quick link of the resource "folderPublic" from the context menu
     When "Anonymous" opens the public link "Quicklink"
--- a/tests/e2e/cucumber/features/smoke/link.feature
+++ b/tests/e2e/cucumber/features/smoke/link.feature
@@ -67,8 +67,8 @@ Feature: link
       | folderPublic/lorem.txt | lorem ipsum |
     And "Alice" logs in
     And "Alice" opens the "files" app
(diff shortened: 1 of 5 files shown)


tests/e2e/support/objects/app-files/share/actions.ts
```diff
@@ -187,10 +193,17 @@ export const hasPermissionToShare = async (
...
+export const copyAndGetQuickLinkUrl = async (args: copyLinkArgs): Promise<string> => {
   const { page, resource, via } = args
+  let url = null
   if (via === 'CONTEXT_MENU') {
     await clickActionInContextMenu({ page, resource }, 'create-quicklink')
+    if (config.ocis) {
+      url = await page.evaluate(() => navigator.clipboard.readText())
+    } else {
+      url = await page.locator(util.format(publicLinkInputField, 'Quicklink')).textContent()
+    }
```
SwikritiT, line 205: Separating the logic here for oc10 and ocis because according to MDN [Clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) docs
> This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility).

In case of `oc10` the values is directly read from the input field

tests/e2e/cucumber/features/smoke/link.feature
```diff
@@ -53,3 +53,24 @@ Feature: link
...
+    And "Alice" creates the following folders in personal space using API
+      | name         |
+      | folderPublic |
+    And "Alice" creates the following files into personal space using API
+      | pathToFile             | content     |
+      | folderPublic/lorem.txt | lorem ipsum |
+    When "Alice" logs in
+    And "Alice" opens the "files" app
+    And "Alice" copies quick link of the resource "folderPublic" from the context menu
+    When "Anonymous" opens the public link "Quicklink"
```
amrita-shrestha, line 71:
```suggestion
    And "Anonymous" opens the public link "Quicklink"
```

SwikritiT: This should be the when step I'll change the above one

saw-jan (commented):
.

tests/e2e/support/environment/actor/shared.ts
```diff
@@ -20,6 +20,7 @@ export interface ActorOptions extends ActorsOptions {
 export const buildBrowserContextOptions = (options: ActorOptions): BrowserContextOptions => {
   const contextOptions: BrowserContextOptions = {
     acceptDownloads: options.context.acceptDownloads,
+    permissions: ['clipboard-read', 'clipboard-write'],
```
saw-jan, line 23: Can we only open read permission as we don't need write permissions as of now.
```suggestion
    permissions: ['clipboard-read'],
```

tests/e2e/support/objects/app-files/share/index.ts
```diff
@@ -66,8 +69,16 @@ export class Share {
     return await hasPermissionToShare({ page: this.#page, resource })
   }
 
-  async copyQuickLink(args: Omit<copyLinkArgs, 'page'>): Promise<void> {
-    await copyQuickLink({ ...args, page: this.#page })
+  async copyAndGetQuickLinkUrl(args: Omit<copyLinkArgs, 'page'>): Promise<void> {
+    this.#linksEnvironment = new LinksEnvironment()
```
saw-jan, line 73: need to preserve it?
```suggestion
    const linksEnvironment = new LinksEnvironment()
```

tests/e2e/support/objects/app-files/share/index.ts
```diff
@@ -66,8 +69,16 @@ export class Share {
...
+    const url = await copyAndGetQuickLinkUrl({ ...args, page: this.#page })
+    const key = 'Quicklink'
+    if (!linkStore.has(key)) {
+      this.#linksEnvironment.createLink({
+        key: key,
```
saw-jan, line 78:
```suggestion
        key,
```

tests/e2e/support/objects/app-files/share/index.ts
```diff
@@ -66,8 +69,16 @@ export class Share {
...
+        link: { name: key, url: url }
+      })
```
saw-jan, line 80:
```suggestion
    if (!linkStore.has(key)) {
      this.#linksEnvironment.createLink({
        key,
        link: { name: key, url }
      })
```

SagarGi (commented):
LGTM :+1:

tests/e2e/support/objects/app-files/share/actions.ts
```diff
@@ -187,10 +194,25 @@ export const hasPermissionToShare = async (
...
+  let url = ''
+  const linkName = 'Quicklink'
+
   if (via === 'CONTEXT_MENU') {
     await clickActionInContextMenu({ page, resource }, 'create-quicklink')
-    await waitForPopupNotPresent(page)
   }
+
+  if (config.backendUrl.startsWith('https')) {
+    url = await page.evaluate(() => navigator.clipboard.readText())
```
ScharfViktor, line 207: Just to clarify. function `navigator.clipboard.readText()` doesn't work for http?
in oc10 where we use http quick link copies to clickboard

ScharfViktor (approved):
looks nice

tests/e2e/support/objects/app-files/share/actions.ts
saw-jan: yeah, `navigator.clipboard` is only available for secure context (i.e. `https`). Also, see [comment](https://github.com/owncloud/web/pull/8499#discussion_r1116562349), [Clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard)

amrita-shrestha (approved):
lgtm

individual-it (approved):

saw-jan:
TODO:
- [ ] forward port to master

PR #8432, opened 2023-02-14
Option to inject custom css and js files

fschade:
Add option to include own header scripts and styles.
This function is currently still experimental and there is a possibility that the api will change.

For the reasons mentioned, the functionality is not yet documented in the official documentation, but can be used as follows:

* to inject custom css add the following property to your `config.json`, `"styles": [{ "href": "css/custom.css", }]`.
* to inject custom scripts add the following property to your `config.json`, `"scripts": [{ "src": "js/custom.js", "async": true, }]`.

The feature was requested in https://github.com/owncloud/web/issues/4735 and backported from the experimental branch.

PR #8512, opened 2023-02-28
fix: move custom styles and scripts to config options

kulmann:
oCIS parses the config.json file and ignores keys that are not known in the backend struct. Hence moving styles and scripts to "options" because that already is accepted as a (potentially nested) map in the config struct.

The other option would've been to define the `scripts` and `styles` entries in the backend struct. Not much of a difference effort wise, but I don't want to impose the bad design of the config.json on the backend code. 😅 

This is a followup of https://github.com/owncloud/web/pull/8432


micbar (approved):

fschade:
Thanks for taking care 😘

PR #8518, opened 2023-03-01
[full-ci] Refactor notifications and make them work with oCIS

JammingBen:
Needs https://github.com/owncloud/ocis/pull/5699.

~Currently known issues on the backend side:~

* ~Timestamps always show the current time~
* ~`share removed` notification does not work yet (shows `removed from space` event instead)~
* ~`space deleted` notification does not work yet~
* ~Capabilities are missing~

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8519
- Fixes https://github.com/owncloud/web/issues/8520

## Screenshots
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## To-Dos:
- [x] e2e tests (see https://github.com/owncloud/web/pull/8533)


packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -1,117 +1,337 @@
...
-      this.$client.requests
-        .ocs({
+    const setAdditionalData = () => {
+      loading.value = true
+      for (const notification of unref(notifications)) {
+        notification.computedMessage = getMessage(notification)
+        notification.computedLink = getLink(notification)
+      }
+      loading.value = false
+    }
```
JammingBen, line 173: I decided to separate the additional data from the `fetchNotificationsTask` and only load it when the notifications drop menu is open. The reason behind this is because I don't now how expensive it will get with a bunch of notifications. Web polls notifications actively every 30 seconds, there is no need to load additional data every time.

I don't see any disadvantages with that approach, but I also don't now if it's necessary. Feedback welcome!

JammingBen:
Status: currently waiting for the backend to finish the rest of the implementation.

kulmann (changes requested):

kulmann (approved):

Issue #8469, opened 2023-02-17
[admin-settings] User: Filter by multiple groups, unexpected outcome


hurradieweltgehtunter:
### Steps to reproduce
1. Open admin settings, “Users” Tab
2. Click on groups filter, select multiple groups.

### Expected behaviour
My expectation was to get a result with users, who belong to one of the selected groups (OR).

### Actual behaviour
Users got filtered by groups with AND => Only users who belong to _all_ selected groups are displayed.

Not a bug, but unexpected outcome. If this is not done on purpose, it should be considered what the general expectation of this filter is.
**OR**
Add a selection “Show users belonging to ONE of the selected groups” / “Show users belonging to all selected groups”

https://ocis.ocis-wopi.latest.owncloud.works

![screenshot_000447](https://user-images.githubusercontent.com/26610733/219732420-fd1cf1b5-6004-4c24-acf8-c4d185d1303c.png)

tbsbdr:
added image to initial comment

tbsbdr:
Should be changed to "OR" concatenation (as this logic meets most usecases)

JammingBen:
An `or` filter needs to be implemented on the server first, currently it only supports `and`.

cc @micbar @rhafer

rhafer:
Ok. `or` support for groupMemberShip should be easy enough to add.

@tbsbdr Probably a bit off topic for this issue, but one question regarding the roleAssignment filtering (which will get introduced soon): Which behavior do you envision if a group and a role is picked for filtering in the admin UI. I guess that should then have `AND` behaviour, right? So it should only display users that are member of one of the selected groups AND have the selected role assigned, correct?

Also do we need to allow selecting multiple roles for the roles filter as well? It's possible, but it would need some extra implementation effort in the backend, that I'd rather defer for now.

tbsbdr:
hey @rhafer 
1) filter concatenation: correct, filters should be and-concatenated amongst each other (show people who are in group XY AND have the role guest)
2) Single select for role selection is sufficient for the first step

PR #8525, opened 2023-03-02
Change behaviour of user filters to `OR`

JammingBen:
## Description
Change user filter for groups and roles to have an `OR` behaviour. So if you select both groups `physics-lovers` and `physics-haters`, it will result in users that are in either of those groups. Combining different filters like groups and roles still concatenates via `AND`.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8469

## Types of changes

- [ ] Technical debt
- [ ] Tests


kulmann (approved):
Nice 💪

PR #8527, opened 2023-03-02
[full-ci] E2E Tests: Admin settings logo upload

lookacat:
## Description
See https://github.com/owncloud/web/issues/8514


ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/33287/69/1

JanAckermann (approved):

tests/e2e/support/objects/app-admin-settings/general/actions.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  await page.waitForSelector('text=Logo was uploaded successfully')
+}
+
+export const resetLogo = async (page): Promise<void> => {
+  await page.click('#logo-context-btn')
+
+  await page.click('.oc-general-actions-reset-logo-trigger')
+
+  await page.waitForSelector('text=Logo was reset successfully. Reloading page...')
```
kulmann, line 15: Please avoid checks on text if possible. What about waiting until the page is reloaded and then check the img src? The way it's implemented right now has the downside of not knowing if the reset really worked or it just printed the success message (as a false positive).

tests/e2e/support/objects/app-admin-settings/general/actions.ts
```diff
@@ -0,0 +1,16 @@
+export const uploadLogo = async (path, page): Promise<void> => {
+  await page.click('#logo-context-btn')
+
+  const logoInput = await page.$('#logo-upload-input')
+  await logoInput.setInputFiles(path)
+
+  await page.waitForSelector('text=Logo was uploaded successfully')
```
kulmann, line 7: Please avoid checks on text if possible. What about waiting until the page is reloaded and then check the img src? The way it's implemented right now has the downside of not knowing if the upload really worked or it just printed the success message (as a false positive).

kulmann (changes requested):
Looks good and straightforward. Not happy with the details of the two verifications in the actions though (see comments). Could you research that part please?

kulmann (commented):
Ok, so post merge review: Looks good now!
Note: for running the test in local dockerized dev env, the `:ro` flag on the `dist` volume mount needs to be removed. Otherwise the logo upload fails (which is expected on a readonly `dist` folder mount).

Issue #8585, opened 2023-03-10
User Management should not break when role is empty


micbar:
### Steps to reproduce

1. Add an education user via the provisioning API
2. Browse to the user management page
3. Try to change his role

### Expected behaviour

Empty Role setting should be visible in the UI (Proposal `-`)
The role assigment in the right sidebar should still work.

### Actual behaviour
The Table layout is broken because the Role is missing
The right sidebar role assigment options list is empty.

#### Screenshot

![Bildschirmfoto 2023-03-07 um 22 00 36](https://user-images.githubusercontent.com/4378513/224248133-f3faa085-bf00-44d5-a5b0-4113dd6782e7.png)

### Environment general
**Operating system**:

**Backend (ownCloud Core or Infinite Scale)**:

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

### Environment ownCloud Infinite Scale
**Version:**

**Updated from an older Infinite Scale or fresh install:**

**Where did you install Infinite Scale from:**

### Environment ownCloud Core
**Web server:**

**Database:**

**PHP version:**

**Version:** (see ownCloud admin page)

**Updated from an older ownCloud or fresh install:**

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results into https://gist.github.com/ and put the link here.
```

**The content of config/config.php:**

```
Log in to the web-UI with an administrator account and click on
'admin' -> 'Generate Config Report' -> 'Download ownCloud config report'
This report includes the config.php settings, the list of activated apps
and other details in a well-sanitized form.

or 

If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

*ATTENTION:* Do not post your config.php file in public as is. Please use one of the above
methods whenever possible. Both, the generated reports from the web-ui and from occ config:list
consistently remove sensitive data. You still may want to review the report before sending.
If done manually then it is critical for your own privacy to diligently
remove *all* hostnames, passwords, usernames, salts and other credentials before posting.
You should assume that attackers find such information and will use them against your systems.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder.
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

### Client configuration
[...]

PR #8590, opened 2023-03-10
Fix loading/editing users that do not have a role assignment

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8585

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-admin-settings/src/components/Users/UsersList.vue
+++ b/packages/web-app-admin-settings/src/components/Users/UsersList.vue
@@ -347,7 +347,7 @@ export default defineComponent({
       const assignedRole = user.appRoleAssignments[0]
 
       return this.$gettext(
-        this.roles.find((role) => role.id === assignedRole.appRoleId)?.displayName || ''
+        this.roles.find((role) => role.id === assignedRole?.appRoleId)?.displayName || ''
       )
     }
   }
--- a/packages/web-app-admin-settings/src/components/Users/UsersList.vue
+++ b/packages/web-app-admin-settings/src/components/Users/UsersList.vue
@@ -346,8 +346,10 @@ export default defineComponent({
     getRoleDisplayNameByUser(user) {
       const assignedRole = user.appRoleAssignments[0]
 
-      return this.$gettext(
-        this.roles.find((role) => role.id === assignedRole?.appRoleId)?.displayName || ''
+      return (
+        this.$gettext(
+          this.roles.find((role) => role.id === assignedRole?.appRoleId)?.displayName || ''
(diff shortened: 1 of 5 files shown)


JanAckermann:
How does this happen, in which case a user doesn't have a role? This is quite new to me.

JammingBen:
You can test this by running oCIS locally and clearing the `metadata` folder on your storage. I can't give you a real-world example though, maybe @micbar knows more about that.

But as the field can be empty and the API allows that, I guess we shouldn't assume that it's always set.

JanAckermann:
image

image

We might want to show a '-' in the details and also in the table

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -693,7 +696,7 @@ export default defineComponent({
     updateUserAppRoleAssignments(user, editUser) {
       return this.graphClient.users.createUserAppRoleAssignment(user.id, {
         appRoleId: editUser.appRoleAssignments[0].appRoleId,
-        resourceId: editUser.appRoleAssignments[0].resourceId,
+        resourceId: editUser.appRoleAssignments[0].resourceId || this.applicationId,
```
JanAckermann, line 699: could this always be this.applicationId to avoid confusion ?

JanAckermann (changes requested):

JammingBen:
> We might want to show a '-' in the details and also in the table

Agreed, also keeps it compatible with the Quota field 👍

packages/web-app-admin-settings/src/views/Users.vue
JammingBen: Should work that way, yes!

packages/web-app-admin-settings/src/components/Users/UsersList.vue
```diff
@@ -346,8 +346,10 @@ export default defineComponent({
     getRoleDisplayNameByUser(user) {
       const assignedRole = user.appRoleAssignments[0]
 
-      return this.$gettext(
-        this.roles.find((role) => role.id === assignedRole?.appRoleId)?.displayName || ''
+      return (
+        this.$gettext(
+          this.roles.find((role) => role.id === assignedRole?.appRoleId)?.displayName || ''
+        ) || '-'
```
JanAckermann: is the '' )  || '-' not irrelevant?

JanAckermann (changes requested):

JanAckermann (approved):

Issue #6183, opened 2021-12-21
[Web] Add Loading Indicator


tbsbdr:
# Description
Provide immediate feedback for ui-interactions and system status

## User Stories

* > As a user I want to see that the system is working on my command so that I'm reassured that my order is being worked on.

## Value

1. They reassure the user that the system is working and reduce the user’s uncertainty. 
2. Empower the user to do the right actions (if the user knows that the system is copying files, he/she acts different than if she/he thinks that the system is doing nothing)
3. They give the user something to look at while waiting, which makes the waiting period easier to tolerate. 
4. They offer a reason to wait for the system to finish— simply because something appears to be happening.
5. They can reduce users’ perception of time: Users devote some cognitive resources to the feedback itself, and they pay less attention to the wait itself.
6. Enables us to let the user continue their work while a long running process runs in the background

( source: https://www.nngroup.com/articles/progress-indicators/ )
## Acceptance Criteria
- [ ] This progress indicator should be used for any action (with a debounce, see below). At least the potentially long running operations copy, move, batch delete, archive download. 🫣
- [ ] show top loading bar if a user-triggered action takes longer than 500 ms (debounce threshold needs some fine tuning when it can be tried out)
- [ ]  Style: use primary color; check YouTube example video attached.
- [ ]  Indeterminate State: Looped indicator if percent-done is not possible
- [ ]  Multiple long running actions can happen at the same time. ;) Keep loading indicator visible until all of them are finished.

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] codre review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

## Research / examples
Red loading bar on top:
![2021-12-21_11-16-20 (1)](https://user-images.githubusercontent.com/26610733/147040935-fd2ffa06-c7eb-4954-ae9d-ed52ee598a38.gif)

Blue loading bar on top:
![2022-01-03_17-46-08 (1)](https://user-images.githubusercontent.com/26610733/147956913-3a13d62f-c38b-4204-90ae-34bb0edc4d8f.gif)

tbsbdr:
Added YouTube example

tbsbdr:
Added box com example

lookacat:
[...]

Issue #2134, opened 2019-10-02
[QA] Show progress bar when delete or rename takes longer


jnweiger:
Steps:

* Click on the "Rename" or "Delete" button of any folder for which the user is permitted
* Type another name and click on OK

ACTUAL: After clicking on "OK" the pop-up will be closed. Ther is no loading bar displayed. After a few seconds the entered name is displayed correctly.

EXPECTED: After the pop-up is closed, a loading bar should be displayed besides the navigation bar while updating the folder name.

jnweiger:
Same when adding a favorite:

Steps:

*  Click on the Favorite icon on any folder/file

ACTUAL: See attached video - After clicking on favorite icon, for a short time nothing happens. There is no loading bar displayed. After a few seconds the folder/file is flagged as favorite correctly.

EXPECTED: After clicking on favorite icon, a loading bar should be displayed besides the navigation bar while updating the content area.

PVince81:
favorite already reported here: #1903

PVince81:
I see showing visual feedback as an important feature to make the UI look polished and avoid people wondering whether something is broken.

dj4oC:
I think my issue is related to this one, but I'm not sure.
When I delete a file or a folder, the DELETE gets disabled after clicking and user needs to wait until action is finished.
but: there is no progress or "please wait" or something else shown to the user and the CANCEL Button is still enabled.
So in my opinion, there should be a busy-spinner and the CANCEL-Button needs to be disables

PR #8611, opened 2023-03-15
[full-ci] Add global loading indicator for long running tasks

JammingBen:
## Description
Adds a global loading indicator for long running actions at the top of the page.

## How to use it (for devs)

Simply wrap your Promise in the `addTask` method provided by the `loadingService`:

```
loadingService.addTask(() => {
  return new Promise((resolve, reject) => {
    resolve(true)
  })
})
```

The method also provides a callback `setProgress` which can be used to update the progress of a task.

```
loadingService.addTask((setProgress) => {
  const promises = []
  const actions = [1, 2, 3, 4, 5]

  for (const [i, action] of actions.entries()) {
    promises.push(
      new Promise((resolve, reject) => {
        setProgress({ total: actions.length, current: i + 1 })
        resolve(true)
      })
    )
  }

  return Promise.all(promises)
}, { indeterminate: false })
```

The following options can be provided to `addTask`:

* `debounceTime` - Time until the loading state for a task is set active. Defaults to `200` ms
* `indeterminate` - Determines if the task is in an indeterminate state. Defaults to `true`.

Note that the loading bar won't show a progress when at least one task is in an indeterminate state.

Thx to @dschmidt for helping me out with the task adding & handling!

## Video

https://user-images.githubusercontent.com/50302941/225225397-f99075c4-859b-49da-8b13-66d4fc422188.mov

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6183
- Fixes https://github.com/owncloud/web/issues/2134

## Types of changes

- [ ] Technical debt
- [ ] Tests


ownclouders:
Results for <strong>acceptance oC10</strong> https://drone.owncloud.com/owncloud/web/33656/32/1
- webUITrashbinDelete/trashbinDelete.feature:36

<p>


### webUITrashbinDelete-trashbinDelete_feature-L110.png
  

### webUITrashbinDelete-trashbinDelete_feature-L123.png
  

### webUITrashbinDelete-trashbinDelete_feature-L85.png
  


</p></details>
:boom: The oC10Trashbin tests pipeline failed. The build has been cancelled.

packages/web-pkg/src/services/loadingService.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (!task.state) {
+      task.state = { total: 0, current: 0 }
+    }
+    task.state.total = total
+    task.state.current = current
+    eventBus.publish(LoadingEventTopics.setProgress)
+  }
+}
+
+export const loadingService = new LoadingService()
```
dschmidt, line 100: I would prefer not to export a "default instance", but to make it available via `app.config.globalProperties.$loadingService` and to access that through a `useLoadingService` composable

Yes, I know we create these default instances for all or at least most of the other services, but I still think it's pretty ugly and we should stop 😅

packages/web-pkg/tests/unit/services/loadingService.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    expect(service.isLoading).toBeFalsy()
+  })
+  it('adds a task and sets it active after the debounce', () => {
+    const service = new LoadingService()
+    const action = new Promise((resolve) => {
+      resolve(true)
+    })
+
+    service.addTask(() => action)
+    jest.runAllTimers()
```
dschmidt, line 23: Maybe add another check *before* the timers are run to check `isLoading` is falsy?

packages/web-pkg/src/services/loadingService.ts
```diff
@@ -0,0 +1,100 @@
...
+    return Math.round((progress / tasks.length) * 100)
+  }
+
+  public addTask(
+    callback: ({ setProgress }: { setProgress: (args: LoadingTaskState) => void }) => Promise<any>,
+    {
+      debounceTime = DEFAULT_DEBOUNCE_TIME,
+      indeterminate = true
+    }: { debounceTime?: number; indeterminate?: boolean } = {}
+  ): Promise<any> {
```
dschmidt, line 60: I think, we can improve the types a little:

```typescript
// not a great name, better ideas appreciated
// maybe `LoadingTaskProgress` and then a `.set` method?
export interface LoadingTaskCallbackArguments {
  setProgress: (args: LoadingTaskState) => void
}
// …
  public addTask<T>(
    callback: ({ setProgress }: LoadingTaskCallbackArguments) => Promise<T>,
    {
      debounceTime = DEFAULT_DEBOUNCE_TIME,
      indeterminate = true
    }: { debounceTime?: number; indeterminate?: boolean } = {}
  ): Promise<T> {
  // …
  }
```

That way 
- we can more easily pass on the `LoadingTaskCallbackArguments` to helper classes/functions, which I prefer over just passing around the `setProgress` function, as it makes it easier to add more features later (`.advanceOne()`, `.advance(2)` or something like that e.g.  :)) 
- we (more importantly) don't lose the return type of the callback function

dschmidt (commented):
I really like where this is going - thanks for taking the time to try the "`addTask` approach" I suggested.

packages/web-pkg/tests/unit/services/loadingService.spec.ts
JammingBen: I already have a separate test that covers exactly this scenario. I like to have that more explicitly :)

tbsbdr:
Visual improvement proposal
- @JammingBen et al: movement catches our attention by itself; thus I'd propose to make the bar more decent; do you think this would be an improvement?


https://user-images.githubusercontent.com/26610733/225310499-fc3242e1-ecf8-4f74-8c87-cc5f8291786d.mp4

dschmidt (approved):

packages/web-app-files/src/store/actions.ts
```diff
@@ -205,6 +204,9 @@ export default {
             { root: true }
           )
         })
+        .finally(() => {
+          setProgress({ total: files.length, current: removedFiles.length })
```
dschmidt, line 211: Hm - looking at this again: a file is pushed only to the removedFiles after success, right? So just moving this to finally should have no effect I guess.

JammingBen: Or simply do it like in the other places via `for (const [i, file] of files.entries())` and use `i`?

packages/web-app-files/src/composables/actions/files/useFileActionsRestore.ts
```diff
@@ -256,7 +262,12 @@ export const useFileActionsRestore = ({ store }: { store?: Store<any> } = {}) =>
       resource.path = urlJoin(parentPath, resolvedName)
       resolvedResources.push(resource)
     }
-    return restoreResources(space, resolvedResources, missingFolderPaths)
+    return loadingService.addTask(
+      ({ setProgress }) => {
+        return restoreResources(space, resolvedResources, missingFolderPaths, { setProgress })
+      },
```
dschmidt, line 268: No need to destructure and build a new object, but not so important

dschmidt (approved):

PR #8554, opened 2023-03-07
Update form input css

hurradieweltgehtunter:
## Description
Updates layout for form inputs, labels & textareas

## Motivation and Context
Text inputs looked condensed and not unified styled.

## How Has This Been Tested?
- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] IE

### Form layout
Before: 
grafik

After:
grafik

### Input focus highlight
The highlight border uses the gradient of web primary button gradient background.
Before:
grafik

After:
grafik

### Comboboxes/Select uses same layout like input
Before:
grafik

After:
grafik

### Disabled comboboxes look more disabled & cursor is disabled
Before:
grafik

After:
grafik

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

Issue #7330, opened 2022-07-26
Notifications block right sidebar


tbsbdr:
### Steps to reproduce
1. login to https://ocis.ocis-web.latest.owncloud.works
2. create > 5 links for a file
3. open sidebar for that file
4. ex. batch-delete 100 files so that you get some notifications
5. right sidebar elements are blocked by notifications

![Screenshot 000062@2x](https://user-images.githubusercontent.com/26610733/180934412-8ef4e6dc-a2fe-483a-8354-67b8e23b4641.png)

![Screenshot 000068@2x](https://user-images.githubusercontent.com/26610733/180941592-05862a9d-9c80-4803-83cf-973aa3fcf9ab.png)

### Expected behaviour
- UI Elements should not be blocked / blocking
- Possible solution:
  - make notifications dismissible
  - move them to the bottom left
  - aggregate multiple notifications into a single overlay so that they don't take up to much space. on click shows all notifications.

### Actual behaviour
- UI Elements are blocked by notifications, esp. when there are many notifications.

---------------------------
---------------------------
---------------------------

## Note on our upload widget
UI is also blocked by the upload widget, but as you can collapse it and as information there are more continuously relevant, its not as disturbing as the notifications. btw: onedrive did a quite well job on the integration of the upload widget: collapsed version is integrated (not overlayed) into the ui; expanded version is implemented as a modal.

### Collapsed and Integrated:
![Screenshot 000063@2x](https://user-images.githubusercontent.com/26610733/180937716-460a15e9-2700-40ad-88f4-99ad3d337baf.png)

### Expanded and modal:
<img width="1500" alt="Screenshot 000064@2x" src="https://user-images.githubusercontent.com/26610733/180937747-a5f32e7e-d5c4-4be0-98ec-3259ca2ab2a7.png">

hurradieweltgehtunter:
https://github.com/owncloud/web/pull/8613 fixes this a bit by making notifications dismissable

JanAckermann:
Notifications are now dismissible. Closing.

PR #8613, opened 2023-03-15
add OcAlert component

hurradieweltgehtunter:
Continuing this topic from https://github.com/owncloud/web/pull/8554

I was missing an alert component, which nearly every css framework offers. So I built it.

Edit: I found some changelogs, where an ocAlert component got removed. What's up with that?

## Description
grafik

grafik

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/design-system/changelog/unreleased/enhancement-add-OcAlert
+++ b/packages/design-system/changelog/unreleased/enhancement-add-OcAlert
@@ -3,4 +3,4 @@ Enhancement: Add `OcAlert` component
 The component can to be used to display messages as information, success, warning or danger.
 It can also be equipped with an icon, a custom icon and a close button.
 
-https://github.com/owncloud/web/pull/8554/
+https://github.com/owncloud/web/pull/8613
--- a/packages/design-system/changelog/unreleased/enhancement-add-OcAlert
+++ b/packages/design-system/changelog/unreleased/enhancement-add-OcAlert
@@ -1,0 +1,6 @@
+Enhancement: Add `OcAlert` component
+
+The component can to be used to display messages as information, success, warning or danger.
+It can also be equipped with an icon, a custom icon and a close button.
+
+https://github.com/owncloud/web/pull/8554/
--- a/packages/design-system/src/components/OcAlert/OcAlert.spec.ts
+++ b/packages/design-system/src/components/OcAlert/OcAlert.spec.ts
@@ -1,0 +1,198 @@
+import { mount } from 'web-test-helpers'
+import OcAlert from './OcAlert.vue'
+
+describe('OcAlert', () => {
+  const selectors = {
+    closeButton: '.oc-alert-close-button',
+    icon: '.oc-alert-icon'
+  }
+  describe('prop value', () => {
+    describe('when alert is closeable', () => {
+      it('should render close button and remove alert when close button is clicked', async () => {
+        const wrapper = getWrapper({
+          closeable: true
+        })
+
+        const alertCloseButton = wrapper.find(selectors.closeButton)
+        expect(alertCloseButton.exists()).toBeTruthy()
+        await alertCloseButton.trigger('click')
+
+        expect(wrapper.find('.oc-alert').exists()).toBe(false)
+      })
+    })
+
+    describe('when alert is not closeable', () => {
+      it('should not render close button', () => {
+        const wrapper = getWrapper({
+          closeable: false
+        })
+
+        const alertCloseButton = wrapper.find(selectors.closeButton)
+        expect(alertCloseButton.exists()).toBeFalsy()
+      })
+    })
+
+    describe('when closeable is not set', () => {
+      it('should not render close button', () => {
+        const wrapper = getWrapper({
+          closeable: undefined
+        })
+
+        const alertCloseButton = wrapper.find(selectors.closeButton)
+        expect(alertCloseButton.exists()).toBeFalsy()
+      })
+    })
+
+    // Test icons
+    describe('custom icon is set', () => {
+      it('should render the given icon', () => {
+        const wrapper = getWrapper({
+          customIcon: 'test-tube'
+        })
+
+        const icon = wrapper.find(selectors.icon)
+        expect(icon.exists()).toBeTruthy()
+      })
+    })
+
+    describe('hasIcon is set', () => {
+      it('should render the icon', () => {
+        const wrapper = getWrapper({
+          hasIcon: true
+        })
+
+        const icon = wrapper.find(selectors.icon)
+        expect(icon.exists()).toBeTruthy()
+      })
+    })
+
+    describe('hasIcon is false', () => {
+      it('should not render the icon', () => {
+        const wrapper = getWrapper({
+          hasIcon: false
+        })
+
+        const icon = wrapper.find(selectors.icon)
+        expect(icon.exists()).toBe(false)
+      })
+    })
+
+    describe('hasIcon not set', () => {
+      it('should not render the icon', () => {
+        const wrapper = getWrapper({
+          hasIcon: undefined
+        })
+
+        const icon = wrapper.find(selectors.icon)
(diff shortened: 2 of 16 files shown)


kulmann:
> Edit: I found some changelogs, where an ocAlert component got removed. What's up with that?

Answer to that is simple: it was unused. Since our primary goal is to develop ownCloud Web, not a design system, we're only implementing and maintaining what we need.

So my main question is: What's the use case for the alert component? Happy to re-introduce an alert component if we have good reasons for it. Otherwise I want to keep the design system as slim as possible.

Another question: what's the use case of closing an alert? From my experience the fine line between an alert and a notification/toast is, that you can close a notification and it looks more standalone, while you can't close an alert and it is inlined into some content.

hurradieweltgehtunter:
> > Edit: I found some changelogs, where an ocAlert component got removed. What's up with that?
> 
> Answer to that is simple: it was unused. Since our primary goal is to develop ownCloud Web, not a design system, we're only implementing and maintaining what we need.
> 
> So my main question is: What's the use case for the alert component? Happy to re-introduce an alert component if we have good reasons for it. Otherwise I want to keep the design system as slim as possible.
> 
> Another question: what's the use case of closing an alert? From my experience the fine line between an alert and a notification/toast is, that you can close a notification and it looks more standalone, while you can't close an alert and it is inlined into some content.

Ok, understandable.
Usecase was in the form-update PR but as you know we decided to go without using the alert. I don't have a specific usecase. General idea was to include it because it is a standard element in css frameworks and IMO a nice possibility to display feedback to the user. But yeah, we build a product, not a framework.

I didn't really think about the closeable alerts, just copied the mechanism from other examples.

diocas:
You are developing a product, but wouldn't it make sense to have all of these options for extensions to use? (We have already one that would benefit from an alert, closable toast notification)
Sorry for commenting, but when I saw the mockups I immediately wanted to use them! (I guess this is for the extensions epic, but having a way to generate notifications from extensions would be great)

hurradieweltgehtunter:
> You are developing a product, but wouldn't it make sense to have all of these options for extensions to use? (We have already one that would benefit from an alert, closable toast notification) Sorry for commenting, but when I saw the mockups I immediately wanted to use them! (I guess this is for the extensions epic, but having a way to generate notifications from extensions would be great)

IMO that's a valid point

diocas:
> The notifications are auto-closing after a while instead...

If this is hard to do, I understand, but I actually see a point in having an `x`. Sometimes, when I'm doing things in the sidebar, I find myself wanting one... Invariably, the notification is on top of some button I want to press and the only thing I can do is wait for it to disappear. It's a bit frustrating, but nothing major :)

Thanks for the tip!

kulmann:
> > The notifications are auto-closing after a while instead...
> 
> If this is hard to do, I understand, but I actually see a point in having an `x`. Sometimes, when I'm doing things in the sidebar, I find myself wanting one... Invariably, the notification is on top of some button I want to press and the only thing I can do is wait for it to disappear. It's a bit frustrating, but nothing major :)
> 
> Thanks for the tip!

Yep, annoying as hell. Tracked here: https://github.com/owncloud/web/issues/7330 Please add a comment that it's annoying for you and your users :-)

hurradieweltgehtunter:
> The one thing that I don't understand actually is the close button on the OcAlert component.

It's optional and it comes in hand for the notification messages for example.
For Inline usage it won't makes sense, you are right.

tbsbdr:
Please note: we moved the notifications from the `top middle` to the `bottom righ`t because  we received feedback that `top middle` (in front of search bar) was blocking users from accessing the search bar. now we realized that buttom right is blocking users with normal screensizes like 1280 x 800 as well =D

we could also try to place it bottom left to get it out of the way.
proper solution is imo ex how onedrive solved it (https://github.com/owncloud/web/issues/7330)

- close-x is generally an improvement 
- proposal: place it bottom left to have it non-blocking - what do you think @kulmann ?

hurradieweltgehtunter:
IMO moving them to left is not solving the root issue. Also, my guess would be that no one expects a notification on the left bottom corner.
I like the onedrive proposal, but this relates to different things. We do already show only a condensed view for uploading many files.
Our problem here IMO is to have different notifications, which take a lot of space and are (were) not dismissable.

kulmann:
Ah, so that basically means that you could remove the OcNotificationMessage component entirely. Or in other words, you didn't introduce a new component but restyled, enhanced functionality and of and renamed OcNotificationMessage? Wouldn't it make sense then to do that in the existing component?

Introducing a `close` button introduces an accessibility violation. I can accept that for a very short while, because the usability is in fact much improved by that (meaning: doesn't need to be solved in the same PR). But we will soon need to fix the focus management requirements on the notification then. More precisely: when a notification message appears it needs to receive the browser focus and when it gets closed the focus needs to be reverted to the element that was focussed before the notification caught the focus.

Also, what's important and seems to be missing now is the auto-close of the notification messages. Most of the time you don't want to close the notification toasts manually. You want them to just disappear. Closing them manually comes in handy when the notification blocks an underlying UI element that you want to use.

kulmann:
> IMO moving them to left is not solving the root issue. Also, my guess would be that no one expects a notification on the left bottom corner.
> I like the onedrive proposal, but this relates to different things. We do already show only a condensed view for uploading many files.
> Our problem here IMO is to have different notifications, which take a lot of space and are (were) not dismissable.

I agree, moving them to the left wouldn't help. Allowing the user to close the notification does help and seems to be the right way (but as stated above, introduces other issues).

JanAckermann:
Sorry, I need to intervene here. 

I don't think the OCAlert component as it right now looks well adjusted to the overall web appearance. 
They look somehow pretty old school and don't fit too well, moreover, they look a little distractive. 

I think we need to find a better way to make those happen.

hurradieweltgehtunter:
> Wouldn't it make sense then to do that in the existing component?

Yes, this could be true and it came to my attention, too, while hacking around. But since the `OcAlert` component can also be used as standalone component (for example as from feedback) it might be better to have it as an own component and include it in `OcNotificationMessage`. I'm not sure, though, maybe we can combine everything in one.


> Also, what's important and seems to be missing now is the auto-close of the notification messages.

No, it's not (at least not on purpose). The auto-close mechanism is defined in `OcNotificationMessage` which I didn't touch.

> Introducing a close button introduces an accessibility violation.

That's a valid point. How was the focus management handled so far?

> moreover, they look a little distractive.

I think that's the purpose of an alert/notification component :) 

> I also don't get it when we want to use OcAlert

From tailwind CSS:
```
The alert component can be used to provide information to your users such as success or error messages, but also highlighted information complementing the normal flow of paragraphs and headers on a page.
```
From my understanding, this represents the one and only component for dynamic user feedback.

To summarize, we have the following questions, sorted by priority:

1. Do we need an OcAlert component? Which use case does it solve?
Yes. It acts as the default component to display feedback to the user. Also, it provides a useful and defacto standard (compare: css frameworks) element for future OC app developers (see @diocas comment)

2. How does it behave?
It comes with different variants and features to be adaptable to many scenarios. 
- It must fulfil all accessibility requirements. 
- It must be optionally closeable 
- It must contain an optional icon
- It must contain an optional body for longer texts
- It might have to contain CTAs (TBD)

3. How does it look?
The design right now is based on how input fields look (rounded corners, border, ...). There are many proposals, from more discreet (https://flowbite.com/docs/components/alerts/) to more prominent (https://daisyui.com/components/alert/) designs. More: https://dribbble.com/search/alerts

IMO the design should be based on other, comparable elements. Comparable in this case means to me: Elements the user interacts with (buttons, inputs, ...)

I think we first should focus on question 1 and 2. If we all agree on that, focus on question 3.

LukasHirt:
We're not going to move forward with this for now due to resources - closing. If we need alert component, we can revive this.

PR #8620, opened 2023-03-15
Use new Drive.ReadWriteQuota.Project Permission

fschade:
## Description
use Drive.ReadWriteQuota.Project from https://github.com/owncloud/ocis/pull/5662/commits/8e4bf92a94b89d4a89c553432ffc230593e91adb

## Related Issue

- Needed by https://github.com/owncloud/ocis/pull/5662

### DIFF
--- a/packages/web-runtime/src/services/auth/abilities.ts
+++ b/packages/web-runtime/src/services/auth/abilities.ts
@@ -39,7 +39,7 @@ export const getAbilities = (
       { action: 'update-all', subject: 'Space' }
     ],
     'list-all-spaces.all': [{ action: 'read-all', subject: 'Space' }],
-    'set-space-quota.all': [{ action: 'set-quota-all', subject: 'Space' }]
+    'Drive.ReadWriteQuota.Project.all': [{ action: 'set-quota-all', subject: 'Space' }]
   }
 
   return Object.keys(abilities).reduce((acc, permission) => {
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/services/auth/abilities.ts
```diff
@@ -39,7 +39,7 @@ export const getAbilities = (
       { action: 'update-all', subject: 'Space' }
     ],
     'list-all-spaces.all': [{ action: 'read-all', subject: 'Space' }],
-    'set-space-quota.all': [{ action: 'set-quota-all', subject: 'Space' }]
+    'Drive.ReadWriteQuota.Project.all': [{ action: 'set-quota-project', subject: 'Space' }]
```
JammingBen, line 42: It got me thinking: shouldn't the action be the same like before: `set-quota-all`? Because `project` is already covered by the subject `Space` here. And `-all` is coming from the server, hence we should add it as well.

fschade: hmmm.... for now i'm fine using `set-quota-all`... but if we get a second type then we need to change it.
The backend decided to have it explicit `Drive.ReadWriteQuota.Project.all`, than we should too.

Fine using `set-quota-project-all`, what do you think?

JammingBen (approved):
👍

PR #8633, opened 2023-03-16
[test-only] fix removing user from group test

ScharfViktor:
fix origin fixFlakyRemovingUsersFromGroup


butonic (approved):
🥇

JammingBen:
@ScharfViktor Should we do similar when adding users to groups as well? 🤔

ScharfViktor:
> @ScharfViktor Should we do similar when adding users to groups? 🤔

Yes, to make more confidence. thank you)

JammingBen (approved):

PR #8658, opened 2023-03-20
[test-only] change notification test after fixing #5865

ScharfViktor:
related [#5865](https://github.com/owncloud/ocis/issues/5865)

- user should not get notification about own action

@kobergj I ran a notification test agains your branch `kobergj:NoNotificationForOwnActions `and found that the user does not receive notifications about disabling and deleting space. Could you please check this


ScharfViktor:
@kobergj I changed test

saw-jan (commented):
LGTM

PR #8660, opened 2023-03-21
Language header follow up

JammingBen:
## Description
Follow-up of https://github.com/owncloud/web/pull/8621

* Remove superfluous `useGraphClient` composable
* Bump `owncloud-sdk` to `3.1.0-alpha.4`


packages/web-app-admin-settings/src/components/Users/CreateUserModal.vue
```diff
@@ -134,7 +135,8 @@ export default defineComponent({
       }
 
       try {
-        await this.graphClient.users.getUser(this.user.onPremisesSamAccountName)
+        const client = this.clientService.graphAuthenticated
+        await client.users.getUser(this.user.onPremisesSamAccountName)
```
kulmann, line 139: is the response irrelevant here? I know, was like that before, but still wondering now...

packages/web-app-admin-settings/src/components/Users/SideBar/EditPanel.vue
```diff
@@ -271,7 +272,8 @@ export default defineComponent({
 
       if (this.user.onPremisesSamAccountName !== this.editUser.onPremisesSamAccountName) {
         try {
-          await this.graphClient.users.getUser(this.editUser.onPremisesSamAccountName)
+          const client = this.clientService.graphAuthenticated
+          await client.users.getUser(this.editUser.onPremisesSamAccountName)
```
kulmann, line 276: Same here, response irrelevant? Why do we need the call then?

packages/web-app-admin-settings/src/components/Users/CreateUserModal.vue
JammingBen: Got me wondering as well, but now I remember: We request the user here to validate that the username does not yet exist. If the request succeeds, we throw a validation error. So the response itself is irrelevant.

kulmann: Could you add a comment in the code explaining this?

kulmann (approved):
💪

PR #8694, opened 2023-03-23
Bugfix: Broken re-login after logout fails

fschade:
## Description
After a user logged out, it was no longer possible to re-login without reloading the ocis root domain,
this has now been fixed and only whitelisted query-params are taken into account in the logout url-replace.

## How Has This Been Tested?
- unit test
- manual testing

## Types of changes

- [ ] ~~New feature (non-breaking change which adds functionality)~~
- [ ] ~~Breaking change (fix or feature that would cause existing functionality to change)~~
- [ ] ~~Technical debt~~
- [X] Tests

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] ~~Acceptance tests added~~
- [ ] ~~Documentation ticket raised~~

### DIFF
--- a/changelog/unreleased/bugfix-oidc-logout-query-param-replication
+++ b/changelog/unreleased/bugfix-oidc-logout-query-param-replication
@@ -1,0 +1,6 @@
+Bugfix: Broken re-login after logout
+
+After a user logged out, it was no longer possible to login without reloading the ocis root domain,
+this has now been fixed and only allowed query-params are taken into account.
+
+https://github.com/owncloud/web/pull/8694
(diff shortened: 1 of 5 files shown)


fschade:
i found another case where it happens, need to fix that too.

JammingBen (approved):
LGTM and works just fine. Thanks for fixing! 🤗

kulmann:
Out of curiosity, @fschade was it one specific query param that was breaking the login? Because apparently the `details` query param doesn't do any harm.

fschade:
> Out of curiosity, @fschade was it one specific query param that was breaking the login? Because apparently the `details` query param doesn't do any harm.

yes, the `state (oidc)`, we had this multiple times in the query-param ...

PR #8661, opened 2023-03-21
[tests-only][full-ci]Add test to check the list of members in a space in the admin settings via sidebar panel

SwikritiT:
## Description
This PR does the following
- Adds a scenario to check the list of members in the space
- Adds API step to add member in a space
- Changes the vue component to add some `data-testid` needed for tests

## Related Issue

- Part of: https://github.com/owncloud/web/issues/8601

## How Has This Been Tested?

- Local
- CI

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #8696, opened 2023-03-24
[tests-only]Add assertion in a e2e tests

SwikritiT:
PR https://github.com/owncloud/web/pull/8661 didn't add assertions in then step, maybe I messed up while rebasing and removed the assertion so this PR fixes that.


saw-jan (approved):
LGTM

Issue #8710, opened 2023-03-27
Customize input fields on FilterItem


hurradieweltgehtunter:
### Steps to reproduce
1. Got to admin settings -> Users
2. Click on filter button "groups" or "roles"

### Expected behaviour
Input field shows customized placeholder, like "Filter groups..."

<img width="331" alt="grafik" src="https://user-images.githubusercontent.com/16665512/227935376-80b437ba-69cf-41e1-96c2-868f4bdfec35.png">

### Actual behaviour
Placeholder reads generic "Filter"

<img width="323" alt="grafik" src="https://user-images.githubusercontent.com/16665512/227935259-0aa9ede5-1b1a-44dc-a9a8-fe3695b5ea59.png">

PR #8711, opened 2023-03-27
FilterItem filter: Add label and make label and placeholder customizable

hurradieweltgehtunter:
## Description
Make the placeholder on text input field for the ItemFilter component editable via prop

## Related Issue
- Fixes https://github.com/owncloud/web/issues/8710

## Motivation and Context
Adds possibility to customized the filter input

### before
grafik

#### with custom placeholder and label
grafik

####
default
grafik

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


JammingBen (changes requested):
> How do you normally handle code coverage fails?

SonarCloud can be quite buggy when determining the coverage on new code. You tested the 2 new properties perfectly fine.

packages/web-pkg/src/components/ItemFilter.vue
```diff
@@ -72,6 +73,16 @@ export default defineComponent({
...
+    placeholder: {
+      type: String,
+      required: false,
+      default: ''
+    },
+    label: {
+      type: String,
+      required: false,
+      default: ''
+    },
```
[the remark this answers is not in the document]
hurradieweltgehtunter: I see your point and I think it's valid from a "inside the component"-perspective. On the other hand, when using the component ("outside the component perspective"), there is only one label and one placeholder. My thinking was, why make it unnecessary complex then?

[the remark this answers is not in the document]
JammingBen: > On the other hand, when using the component ("outside the component perspective"), there is only one label and one placeholder.

Well, not necessarily, because that's strictly bound to the `showFilter` property. Without explicitly setting that to `true`, the new props won't show at all. So one might wonder why these props are not being displayed despite setting them. By grouping them via some name prefix like `optionFilter`, this makes it more clear.

packages/web-pkg/src/components/ItemFilter.vue
```diff
@@ -204,18 +215,26 @@ export default defineComponent({
...
+      showDrop,
+      toggleItemSelection
+    }
+  },
+  computed: {
+    getPlaceholder() {
+      return this.placeholder === '' ? this.$gettext('Enter term') : this.placeholder
+    },
+    getLabel() {
+      return this.label === '' ? this.$gettext('Filter list') : this.label
```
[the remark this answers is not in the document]
JammingBen: > Declaring computed inside the setup() method threw an error with the $gettext method (not defined).

You need to make use of the Gettext-composable first within `setup`: `const { $gettext } = useGettext()`.

> Also, I always try to separate layout definitions (inside template tags) and functional definitions (inside script tag).

While I agree for most situations, such simple and label-related checks are fine in the template IMO. Plus you save yourself from finding a name for the computed props, because the `get...` prefix like you used before is usually reserved for methods.

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -62,14 +62,16 @@
...
                 </div>
                 <item-filter
-                  filter-name="groups"
+                  :allow-multiple="true"
                   :filter-label="$gettext('Groups')"
+                  :filterable-attributes="['displayName']"
                   :items="groups"
+                  :optionFilterPlaceholder="$gettext('Enter group name')"
+                  :optionFilterLabel="$gettext('Filter groups')"
                   :show-filter="true"
```
JammingBen, line 71:
```suggestion
                  :show-option-filter="true"
```

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -80,13 +82,15 @@
...
                 </item-filter>
                 <item-filter
-                  filter-name="roles"
+                  :allow-multiple="true"
                   :filter-label="$gettext('Roles')"
+                  :filterable-attributes="['displayName']"
                   :items="roles"
+                  :optionFilterPlaceholder="$gettext('Enter role name')"
+                  :optionFilterLabel="$gettext('Filter roles')"
                   :show-filter="true"
```
JammingBen, line 91:
```suggestion
                  :show-option-filter="true"
```

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -62,14 +62,16 @@
                   <span v-text="$gettext('Filter:')" />
                 </div>
                 <item-filter
-                  filter-name="groups"
+                  :allow-multiple="true"
                   :filter-label="$gettext('Groups')"
+                  :filterable-attributes="['displayName']"
                   :items="groups"
+                  :optionFilterPlaceholder="$gettext('Enter group name')"
+                  :optionFilterLabel="$gettext('Filter groups')"
```
JammingBen, line 70:
```suggestion
                  :option-filter-placeholder="$gettext('Enter group name')"
                  :option-filter-label="$gettext('Filter groups')"
```

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -80,13 +82,15 @@
                   </template>
                 </item-filter>
                 <item-filter
-                  filter-name="roles"
+                  :allow-multiple="true"
                   :filter-label="$gettext('Roles')"
+                  :filterable-attributes="['displayName']"
                   :items="roles"
+                  :optionFilterPlaceholder="$gettext('Enter role name')"
+                  :optionFilterLabel="$gettext('Filter roles')"
```
JammingBen, line 90:
```suggestion
                  :option-filter-placeholder="$gettext('Enter role name')"
                  :option-filter-label="$gettext('Filter roles')"
```

JammingBen (changes requested):

packages/web-app-admin-settings/src/views/Users.vue
hurradieweltgehtunter: 🤦

JammingBen (approved):
Nice 👍  Another linter-run is required though.

hurradieweltgehtunter:
> Nice 👍 Another linter-run is required though.

you were too fast :D

done

hurradieweltgehtunter:
hu, SonarCloud gug? buggy code is not mine 🤷 Ignore?

PR #8727, opened 2023-03-28
[full-ci] Introduce read-only options to OcSelect

JammingBen:
## Description
This is required for the user-group-assignment select because a user can have read-only groups that must not be removed.

## Related Issue

- Needed for https://github.com/owncloud/web/issues/8729

image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/design-system/src/components/OcSelect/OcSelect.vue
+++ b/packages/design-system/src/components/OcSelect/OcSelect.vue
@@ -8,6 +8,7 @@
       :loading="loading"
       :searchable="searchable"
       :clearable="clearable"
+      :multiple="multiple"
       class="oc-select"
       style="background: transparent"
       v-bind="additionalAttributes"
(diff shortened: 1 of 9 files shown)


packages/web-app-admin-settings/src/components/Users/SideBar/EditPanel.vue
JammingBen, line 1: Changes in this file are just for testing purposes.

packages/design-system/src/components/OcSelect/OcSelect.vue
```diff
@@ -19,10 +20,30 @@
       <template v-for="(index, name) in $slots" #[name]="data">
         <slot v-if="name !== 'search'" :name="name" v-bind="data" />
       </template>
-      <template #no-options><div v-translate>No options available.</div></template>
+      <template #no-options><div :v-text="$gettext('No options available.')" /></template>
```
lookacat, line 23: I think you don't need `:` before `v-text`

JammingBen: True, thx! :)

dschmidt (approved):
Love it! 🚀 ❤️

PR #8741, opened 2023-03-29
Implement GDPR export for users

JammingBen:
## Description
Adds a GDPR export on the account page of each user. Note that this is only supported when running oCIS as backend.

The post-processing state can be tested when starting the server with:

```
POSTPROCESSING_DELAY: "30s"
STORAGE_USERS_OCIS_ASYNC_UPLOADS: "true"
```

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8738

image

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-client/src/graph.ts
```diff
@@ -133,17 +138,16 @@ export const graph = (baseURI: string, axiosClient: AxiosInstance): Graph => {
       deleteUser: (userId: string) => userApiFactory.deleteUser(userId),
       listUsers: (orderBy?: any, filter?: string) =>
         usersApiFactory.listUsers(
-          0,
-          0,
```
JammingBen, line 137: Unrelated, but some parameters seems to have gone after updating the graph client.

JammingBen:
This is ready for review, although the server does not ship the capability yet. So for testing, you need to set it to `true` manually: https://github.com/owncloud/web/pull/8741/files#diff-f2ac05790ff955307ad01dea9de39462b2f5f017959394934eaad70208f0b1a7R143

dschmidt (approved):
works like a charm and the code LGTM. :shipit:

Issue #8420, opened 2023-02-13
OIDC token refresh failing


wkloucek:
### Steps to reproduce
1. Log in to https://ocis.ocis-web.daily.owncloud.works/
2. Wait some minutes
3. See screen that you have been logged out. If you click on "login again" you will be immediately logged in again 

If you watch the browser console you will see this:

![image](https://user-images.githubusercontent.com/34452982/218449543-b595501f-b9b1-4d99-9f0f-ec405573ec62.png)

### Expected behaviour

Token refresh works and indeed it does on https://ocis.ocis-traefik.daily.owncloud.works

### Actual behaviour
The "logged out" screen appears

### Environment general
Latest oCIS master with latest web

tbsbdr:
Console Screenshot in higher resolution:
![Clipboard - February 13, 2023 12_36 PM](https://user-images.githubusercontent.com/26610733/218460330-22358c9c-8232-4543-a759-f4cafeb89179.png)

```
A » [IFrameWindow] navigate: no state found in response url
• • (oidcClient] readSigninResponseState: Error: No state in response at Pfe. readSigninResponseState (vendor-aceee24.mjs:2:96800)
at Pfe.processSigninResponse (vendor-aceee24.mis:2:97145)
at fo. signinEnd (vendor-ac6eee24.mjs:2:117839)
at fo. signin (vendor-ac6eee24.mjs:2:117349)
at async fo.signinSilent (VM242 vendor-aceee24.mjs:2:114818)
at async Ufe. tokenExpiring (VM242 vendor-aceee24.mjs:2:111289)
© > (SilentRenewService] tokenExpiring: Error from signinSilent: Error: No state in response
at Pfe. readSigninResponseState (vendor-aceee24.mis:2:96800)
at Pfe.processSigninResponse (vendor-aceee24.mis:2:97145)
at fo. signinEnd (vendor-ac6eee24.mjs:2:117839)
at fo. signin (vendor-ac6eee24.mjs:2:117349)
at async fo.signinSilent (VM242 vendor-ac6eee24.mjs:2:114818)
at async Ufe. tokenExpiring (VM242 vendor-aceee24.mjs:2:111289)
© Silent Renew Error: Error: No state in response
at Pfe. readSigninResponseState (vendor -aceee24.mjs:2:96800) at pfe.processSigninResponse (vendor-ac6eee24.mis:2:97145)
at fo. signinEnd (vendor-aceee24.mis:2:117839)
[...]

PR #8762, opened 2023-03-31
fix: oidc token renewal

kulmann:
## Description
this PR fixes multiple oidc related things:

- Our dockerized dev environment was missing the redirect uris for silent token renewal `/oidc-silent-redirect.html`. The silent renewal landed in a (silent :trollface: ) access_denied.
- silent token renewal had a url format that the oidc-client-lib could not parse when in vue router hash mode
- we now stop the application bootstrap early in the silent redirect callback page. This mitigates an ocis bug in the authentication middleware. Will post an ocis issue as well later on. 

As a result you should not get redirected to the access denied page anymore. That was happening on token renewals before. 

## Issues
Fixes https://github.com/owncloud/web/issues/8420

### DIFF
--- a/packages/web-runtime/src/services/auth/authService.ts
+++ b/packages/web-runtime/src/services/auth/authService.ts
@@ -4,7 +4,12 @@ import { Store } from 'vuex'
 import { ClientService } from 'web-pkg/src/services'
 import { ConfigurationManager } from 'web-pkg/src/configuration'
 import { RouteLocation, Router } from 'vue-router'
-import { extractPublicLinkToken, isPublicLinkContext, isUserContext } from '../../router'
+import {
+  extractPublicLinkToken,
+  isAnonymousContext,
+  isPublicLinkContext,
+  isUserContext
+} from '../../router'
 import { unref } from 'vue'
 import { Ability } from 'web-pkg/src/utils'
 import { Language } from 'vue3-gettext'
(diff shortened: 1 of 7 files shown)


kulmann:
Converted back to draft. There are other issues... redirect to logged out page still happening. I see 401 errors from requests in the token renewal iframe in the network tab. Investigating.

packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -74,7 +79,9 @@ export class AuthService {
         ability: this.ability,
         language: this.language
       })
+    }
 
+    if (!this.userManager.areEventHandlersRegistered && !isAnonymousContext(this.router, to)) {
```
dschmidt, line 84: You could make it
```
if (!isAnonymousContext(this.router, to)) {
   if (!this.userManager.areEventHandlersRegistered) {
   …
   }
  const accessToken = await this.userManager.getAccessToken()
  …
}
```
and save the duplicated `!isAnonymousContext(this.router, to)` check

dschmidt (approved):

packages/web-runtime/src/services/auth/authService.ts
kulmann: nice 👍

kulmann:
@dschmidt I re-introduced the redirect uris `https://host.docker.internal:9200/` (and for the other ports as well) because those are needed for redirect after logout. At least that's what we've hardcoded in the userManager config. ;-)

kulmann:
While debugging this @dschmidt and I also discovered an ocis bug: https://github.com/owncloud/ocis/issues/5986

Issue #6731, opened 2022-04-04
Long breadcrumb: truncate and add drop menu


tbsbdr:
### Steps to reproduce
1. login on web.owncloud.com, switch to new design
2. create many folders and open them until the breadcrumb line-breaks

### Expected behaviour
Breadcrumb should not line-break

### Actual behaviour
Breadcrumb Line-breaks:
<img width="1200" alt="image" src="https://user-images.githubusercontent.com/26610733/161613051-3b772cf9-1741-4509-b99b-4feea303cbe0.png">

### Proposal
- truncate the breadcrumb with `...` on the left, so that it does not line-break
- click on `...` opens a drop menu with only the **remaining** i.e. truncated (hidden) folders.
- click on item navigates to the respective folder
- root elements are not truncated, i.e. `Personal`, `Spaces > Mars` stays visible, truncation starts after that

Scribble:
<img width="1095" alt="image" src="https://user-images.githubusercontent.com/26610733/161613567-054c72bc-d07f-4b3d-a703-6ecac8e70606.png">

fschade:
maybe the easiest solution

```html
<div id="parent">
  <ul id="child" class="breadcrumb">...</ul>
</div>
```

#parent -> overflow-x: hidden
#child -> whitespace: nowrap

window.onResize: #child.outerWitdh >= #parent.outerWitdh => child.li -> pop

lookacat:
@fschade i would also propose that solution, i can't think of any other solution that wouldn't be hacky. You would probably need to calculate each items width before rendering and then sort out which elements are displayed and which are in the three dots menu. If anyone has Deep CSS knowledge and knows a solution for it please let me know :D

lookacat:
Would propose doing it like flo said, I tried to find another solution but I'm not sure if the three dots menu is possible that easy :/

kulmann:
> Would propose doing it like flo said, I tried to find another solution but I'm not sure if the three dots menu is possible that easy :/

You could have a look at what the classic ui in oc10 does. That has the truncation like we want to have it (not the three dots menu though, but that could be solved in a separate PR).

hurradieweltgehtunter:
depending issue: creating a new folder in a deep nested folder with line-breaking breadcrumb results in this:

<img width="862" alt="grafik" src="https://user-images.githubusercontent.com/16665512/188114954-9af7b917-55cf-4004-bbd6-277ac3a8bc8c.png">

Gets resolved on changeing the windows width or reloading the page.

[...]

PR #8769, opened 2023-04-03
Add option to set max length of a breadcrumb item

elizavetaRa:
## Description
Adds config option `breadcrumbItemMaxLength` for max length of a breadcrumb item (desktop version)

- With the option set to > 0 the item text gets cut to max length + "..."
- Shorten items get a tooltip with the full item text

## Related Issue
https://github.com/owncloud/web/issues/6731

## Motivation and Context
UI improvement

Long items without pr
before

Long items with pr
after

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] Should the last item always be fully visible?
- [ ] Should max length be also used for mobile version (drop)? But it doesn't have the same problem as a multiline list
- [ ] Now added only to breadcrumb in `AppBar.vue `. Where else it makes sense to add?

### DIFF
--- a/changelog/unreleased/enhancement-max-length-breadcrumb-item
+++ b/changelog/unreleased/enhancement-max-length-breadcrumb-item
@@ -1,0 +1,5 @@
+Enhancement: Option for the max length of a breadcrumb item
+
+We've added a configuration option `options.breadcrumbItemMaxLength` to set the max length of a breadcrumb items with exeption of the last item.
+
+https://github.com/owncloud/web/pull/8769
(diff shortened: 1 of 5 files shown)


elizavetaRa:
@tbsbdr Could you give your opinion on the following:
- Should the last item always be fully visible? Currently, it can also be shorten with this option.
- Should max length also be used for mobile version (drop)? But it doesn't have the same problem as a multiline list, so, currently, not.
- Now added only to the web-app-files breadcrumb in `AppBar.vue`. Where else it makes sense to add?

tbsbdr:
> * Should the last item always be fully visible? Currently, it can also be shorten with this option.

I'd say: last item is always **not** truncated (but tbh I have no super strong opinion about it)

> * Should max length also be used for mobile version (drop)? But it doesn't have the same problem as a multiline list, so, currently, not.

jep👍 I think the same

> * Now added only to the web-app-files breadcrumb in `AppBar.vue`. Where else it makes sense to add?

Administration Settings: I think its not necessary to truncate all breadcrumbs in `Administration settings`, having the truncation in the files app is only is perfectly sufficient

kulmann:
Superseded by https://github.com/owncloud/web/pull/8984

PR #8790, opened 2023-04-11
Init custom translations

JanAckermann:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/index.ts
+++ b/packages/web-runtime/src/index.ts
@@ -4,6 +4,8 @@ import { configurationManager } from 'web-pkg/src/configuration'
 import { createHead } from '@vueuse/head'
 import { abilitiesPlugin } from '@casl/vue'
 import { createMongoAbility } from '@casl/ability'
+import merge from 'lodash-es/merge'
+import { v4 as uuidV4 } from 'uuid'
 
 import {
   announceConfiguration,
@@ -39,6 +41,7 @@ import PortalVue, { createWormhole } from 'portal-vue'
 
 import Avatar from './components/Avatar.vue'
 import focusMixin from './mixins/focusMixin'
+import fs from 'fs'
 
 export const bootstrapApp = async (configurationPath: string): Promise<void> => {
   const app = createApp(pages.success)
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/index.ts
```diff
@@ -60,8 +62,31 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> =>
   })
   const themePromise = announceTheme({ store, app, designSystem, runtimeConfiguration })
   await Promise.all([applicationsPromise, themePromise])
+  const customTranslations = {}
+
+  for (const customTranslation of configurationManager.customTranslations) {
```
kulmann, line 67: Could you please move fetching and collecting/merging the custom translations into its own method? Something like `fetchCustomTranslations` which then returns the `customTranslations` object. Otherwise this is an abstraction level mismatch. All the other code in this `bootstrapApp` method is on a higher abstraction level.

kulmann:
I already went ahead and added `customTranslations` to the web config struct in ocis, see https://github.com/owncloud/ocis/pull/6032

I needed the `scripts` and `styles` keys, so adding `customTranslations` in the same PR was a low hanging fruit... hope that's fine for you.

packages/web-runtime/src/index.ts
```diff
@@ -60,8 +63,13 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> =>
   })
   const themePromise = announceTheme({ store, app, designSystem, runtimeConfiguration })
   await Promise.all([applicationsPromise, themePromise])
+  const customTranslations = await loadCustomTranslations({ configurationManager })
```
kulmann, line 66: Last nitpick: tiny performance improvement possible by splitting this into promise creation and awaiting the resolved promise together with other promises. This is a good thing because the promise execution starts immediately with creation, so this enables parallel processing instead of awaiting promises sequentially.

In other words:
Please split this line into:
```js
const customTranslationsPromise = loadCustomTranslations({ configurationManager })
```
for promise creation and add that to the `Promise.all`. You can access the resolved custom translations like this:

```js
const [customTranslations] = await Promise.all([customTranslationsPromise, applicationsPromise, themePromise])
```

kulmann (approved):

kulmann:
Linter complains :-)

kulmann:
Screenshot 2023-04-12 at 13 31 27


Great value for any oCIS admin out there 🚀

Issue #8815, opened 2023-04-14
User quota edit is not reactive in right sidebar


JammingBen:
### Steps to reproduce
1. Go to the users list
2. Open the right sidebar for a user
3. Edit the user's quota via the context action (not in the sidebar) -> quota in sidebar remains unchanged

Issue #8690, opened 2023-03-23
Some batch actions try/catch prohibits logic from occurring in case of one item failing


lookacat:
### Problem description
`useUserActionsDelete.ts`
`useGroupActionsDelete.ts`
(there may be some more files where the same logic is applied)

The problem here is that if one item in promise.all failes the catch block is called. This causes needed code to not to have been called. (showing the succeeded items, reloading the list)

### Expected behaviour
If i batch edit 4 items and 2 fail, i would expect two messages, one for the succeeded and one for the failed.
Also we need to make sure that no code in the try block is skipped because of one failing item.

### Actual behaviour
If i batch edit 4 items and 2 fail, I only get an error message.

PR #8816, opened 2023-04-14
Fix user quota reactivity

JammingBen:
## Description
Fixes user quota reactivity in right sidebar and implements proper messaging on user- and group-batch-delete.

I had to refactor `useSpaceActionsEditQuota.ts` a bit and removed `selectedSpace` and the method which updates it. This was built to deal with one given space, but now that the quota can be edited via batch it doesn't make sense anymore.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8815
- Fixes https://github.com/owncloud/web/issues/8690

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-admin-settings/src/composables/actions/groups/useGroupActionsDelete.ts
+++ b/packages/web-app-admin-settings/src/composables/actions/groups/useGroupActionsDelete.ts
@@ -14,33 +14,42 @@ export const useGroupActionsDelete = ({ store }: { store?: Store<any> }) => {
     const graphClient = clientService.graphAuthenticated
     const promises = groups.map((group) => graphClient.groups.deleteGroup(group.id))
 
-    try {
-      await Promise.all(promises)
-      store.dispatch('hideModal')
-      store.dispatch('showMessage', {
-        title: $ngettext(
-          'Group "%{group}" was deleted successfully',
-          '%{groupCount} groups were deleted successfully',
-          groups.length,
-          { groupCount: groups.length, group: groups[0].displayName },
-          true
-        )
-      })
-      eventBus.publish('app.admin-settings.list.load')
-    } catch (error) {
-      console.error(error)
-      store.dispatch('hideModal')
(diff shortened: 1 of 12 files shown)


packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -448,6 +460,10 @@ export default defineComponent({
           editLoginModalIsOpen.value = true
         }
       )
+      editLoginActionEventToken = eventBus.subscribe(
+        'app.admin-settings.users.user.quota.updated',
+        updateSpaceQuota
+      )
```
JanAckermann, line 466: incorrect var name?

JanAckermann (changes requested):

packages/web-app-admin-settings/src/views/Users.vue
JammingBen: Arghs yes 😅

packages/web-app-admin-settings/src/composables/actions/groups/useGroupActionsDelete.ts
```diff
@@ -14,33 +14,42 @@ export const useGroupActionsDelete = ({ store }: { store?: Store<any> }) => {
...
+      const title =
+        succeeded.length === 1 && groups.length === 1
+          ? $gettext('Group "%{group}" was deleted successfully', { group: groups[0].displayName })
+          : $ngettext(
+              '%{groupCount} group was deleted successfully',
+              '%{groupCount} groups were deleted successfully',
+              succeeded.length,
+              { groupCount: succeeded.length.toString() },
+              true
+            )
```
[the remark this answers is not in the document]
JammingBen: Because this is bound to the condition `failed.length === 1 && groups.length === 1`. I don't think we can solve that via `ngettext` only.

JanAckermann (approved):

Issue #8277, opened 2023-01-19
Prevent drag and drop on RO public links


diocas:
We had a support ticket from a very confused user that was dragging and dropping a file in a public link (mistakenly set to RO) and the only error he saw was "1 item failed".

Since we disable the upload button, the drag and drop area should also be disabled to prevent these situations.

PR #8825, opened 2023-04-17
Remove drop target in read-only folders

JammingBen:
## Description
Removes the drop target in read-only folders

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8277

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-remove-drop-target-in-read-only-folders
+++ b/changelog/unreleased/bugfix-remove-drop-target-in-read-only-folders
@@ -1,0 +1,6 @@
+Bugfix: Remove drop target in read-only folders
+
+The drop target in read-only folders has been removed.
+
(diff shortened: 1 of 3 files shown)


kulmann (approved):
I see another chance for improvement, but that can/should happen in a followup:
Both the read only public link view and the read only share view suggest to the user that they can drag & drop files for upload when navigating into an empty folder (see screenshots). That subtitle of the `No resources found` message could be removed on read only folders. But again, also fine as a followup.

This PR itself LGTM 💪 

Screenshot 2023-04-17 at 15 14 01
Screenshot 2023-04-17 at 15 15 29

JammingBen:
Ohhh yeah, good catch! I've created https://github.com/owncloud/web/issues/8834 for this.

PR #8835, opened 2023-04-17
[test-only] flaky e2e while uploading

ScharfViktor:
https://drone.owncloud.com/owncloud/ocis/21380/68/10
https://drone.owncloud.com/owncloud/ocis/21392/88/9

and more ....

Screenshot 2023-04-17 at 16 03 55

click by 'body' https://github.com/owncloud/web/blob/master/tests/e2e/support/objects/app-files/resource/index.ts#L73-L74 hit right on the button `decline`

cc @kobergj 

second commit - to accelerate the test

Screenshot 2023-04-17 at 17 11 48

### DIFF
--- a/tests/e2e/support/objects/app-files/resource/index.ts
+++ b/tests/e2e/support/objects/app-files/resource/index.ts
@@ -70,6 +70,9 @@ export class Resource {
     const startUrl = this.#page.url()
     await uploadResource({ ...args, page: this.#page })
     await this.#page.goto(startUrl)
+    if (!config.ocis) {
+      await this.#page.locator('body').click()
+    }
   }
 
   async download(args: Omit<downloadResourcesArgs, 'page'>): Promise<Download[]> {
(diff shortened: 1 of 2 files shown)


kulmann (approved):

tests/e2e/support/objects/app-files/resource/index.ts
```diff
@@ -70,8 +70,9 @@ export class Resource {
     const startUrl = this.#page.url()
     await uploadResource({ ...args, page: this.#page })
     await this.#page.goto(startUrl)
-    // why? o_O
-    await this.#page.locator('body').click()
+    if (!config.ocis) {
+      await this.#page.locator('body').click()
```
kulmann, line 74: did you find out why the `body` click is needed in oc10 environment?

saw-jan: IMO, it can be removed from here. But I have no idea why that was added before

ScharfViktor: Yes. After changing the link role from uploder to editor  -> on the UI the role does not change (I couldn't reproduce it localy). I tried to set `page.reload()` but sometimes it fails. needs to be investigated and fixed in another PR
changing role - looks okay
Screenshot 2023-04-18 at 08 28 28

checking role - role does not change
Screenshot 2023-04-18 at 08 29 11

PR #8839, opened 2023-04-17
Enable readOnly prop for OcSelect and OcTextInput

JanAckermann:
## Related Issue

- Needed for https://github.com/owncloud/web/issues/8840

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

image
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcModal/__snapshots__/OcModal.spec.ts.snap
+++ b/packages/design-system/src/components/OcModal/__snapshots__/OcModal.spec.ts.snap
@@ -10,7 +10,7 @@ exports[`OcModal displays input 1`] = `
       </div>
       <div class="oc-modal-body">
         <!--v-if-->
-        <oc-text-input-stub class="oc-modal-body-input" clearbuttonaccessiblelabel="" clearbuttonenabled="false" disabled="false" fixmessageline="true" id="oc-textinput-1" label="Folder name" modelvalue="New folder" type="text"></oc-text-input-stub>
+        <oc-text-input-stub class="oc-modal-body-input" clearbuttonaccessiblelabel="" clearbuttonenabled="false" disabled="false" fixmessageline="true" id="oc-textinput-1" label="Folder name" modelvalue="New folder" readonly="false" type="text"></oc-text-input-stub>
         <!--v-if-->
         <!--v-if-->
       </div>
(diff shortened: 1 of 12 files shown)


JanAckermann:
🤯

JanAckermann:
snapshots look weird, should they have icon inside?

JammingBen (commented):
I'd suggest putting the icon in the `no-options` and in the `selected-option-container` slot. Then you don't have to fiddle around with absolute positioning.

It also breaks currently if you use the `header` slot:

image

JanAckermann:
Not so happy with that as if the drop down opens the icon won't be shown. This would also mean a different type of implementation than the oc-text-input one

JammingBen:
> Not so happy with that as if the drop down opens the icon won't be shown. 

Hmm then add it to the `option` container as well? 😄 Unfortunately, `vue-select` does not provide a global slot for such use cases.

> This would also mean a different type of implementation than the oc-text-input one

Different component = different solution, no? I mean, we have to find a solution. Or we need to document that the `header` slot can't be used together with an icon, which I really don't like.

JanAckermann:
I mean I can check if a header is provided and adjust the margins. 
I don't think it's quite elegant to repetitive add icons in n slots

dschmidt:
What is the semantic difference between ` readonly`  and ` disabled`?

JanAckermann:
For example:
The login(allowed) select might be disabled because you are not allowed to change this for your own account. 
It is readOnly if the server doesn't allow to change this option for any user

dschmidt:
and maybe you can extend this table: https://github.com/owncloud/web/issues/8857

So we know what property name we used here and that we might want to add it to additional components as well.

dschmidt:
> For example: The login(allowed) select might be disabled because you are not allowed to change this for your own account. It is readOnly if the server doesn't allow to change this option for any user

Might be worth documenting/giving this example in the prop documentation

JanAckermann:
I don't know this would be very case specific, I think we shouldn't go too much in detail here, for me its enough if we say sth like: "same as disabled but adds a lock item"

dschmidt:
> I don't know this would be very case specific, I think we shouldn't go too much in detail here, for me its enough if we say sth like: "same as disabled but adds a lock icon"

I would argue: it's a design system, it should guide its users on how to use it. One can decide to not follow the usual rules, but for that one needs to know the rules.

JanAckermann:
I'll add the example, can't block that any longer because its needed for a sprint task

dschmidt (approved):
Haven't tested it, but I'm happy with the explanation now ;-)
Thanks!

JammingBen (approved):

PR #8799, opened 2023-04-12
Implement batch action for editing the login state of users

JammingBen:
## Description
Adds a batch action to the users list in the admin settings which allows the admin to batch-edit the login state of users.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8798

## Screenshots
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

Issue #8579, opened 2023-03-09
[Admin Settings] Unify Dialogs


tbsbdr:
Dialogs are currently presented in different schemas and wordings.
This is a proposal to unify the dialogs.
Generally the dialogs should be as precise and as short as possible.

- Show "Affected number" if applicable in
   - Title
   - Description (optional)
   - ~~Button~~

- Show "Affected number" only, if at least 2 items are selected
  - Delete Space ✅
  - Delete 2 Spaces ✅
  - ~~Delete 1 Space~~

![image](https://user-images.githubusercontent.com/26610733/224056926-b0465467-cb2f-48f3-afad-aa7f8e5c46b3.png)

hurradieweltgehtunter:
Same goes for CTA Buttons in dialogs. I came across different primary CTAs:
- disable space -> CTA reads “Disable” (this refers to the context of my desired action)
- rename space -> CTA reads “Rename” (this refers to the context of my desired action)
- many other actions -> CTA reads “Confirm” (this refers to the dialog itself and not the action, which is the correct context IMO)

IMO primary CTA's should be aligned based on the context of the dialog:
- Is the dialog something I need to confirm? for example:  Delete a space -> CTA should read: Confirm
- Is the dialog something I need to enter a something in an input/textfield? for example, change a space name -> CTA should read: Save
- CTA's should never contain the action itself, for example disable a space -> “Confirm” instead of “Disable” (current)

hurradieweltgehtunter:
Another finding: Remove a member from a space.
1. Open space
2. remove member

Modal asks: "Are you sure you want to remove this share?"
CTA reads -> "Remove" instead of "Confirm"

JanAckermann:
@hurradieweltgehtunter good finding @tbsbdr can you make a decision on that?

hurradieweltgehtunter:
> Another finding: Remove a member from a space.
> 
>     1. Open space
> 
>     2. remove member
> 
> 
> Modal asks: "Are you sure you want to remove this share?" CTA reads -> "Remove" instead of "Confirm"

Apparently the second part of my comment got lost. I added sth like:
 
Action in sidebar should read: "Remove member | { user | group }" instead of "Remove share"
Following modal should read: 
Title: "Remove member | { user | group }"
Body: "Are you sure you want to remove this member | { user | group }?"
CTA: "Yes" (Apparently "Confirm" is not to be used according to our design guidelines on owncloud.design)
Cancel CTA: "No"

[...]

PR #8853, opened 2023-04-18
Enhance bulk edit login

hurradieweltgehtunter:
Follow-up for https://github.com/owncloud/web/pull/8799

## Description
Re-use the select box label from the right sidebar sidebar ("Login") instead of introducing a new one ("Login status") which saves an additional translation and stays consistent with editing a single user in the right sidebar.

Also, add an informational text line in the modal to explain what happens here.

### Before
grafik

### After
grafik

### DIFF
--- a/packages/web-app-admin-settings/src/components/Users/LoginModal.vue
+++ b/packages/web-app-admin-settings/src/components/Users/LoginModal.vue
@@ -10,7 +10,7 @@
     <template #content>
       <oc-select
         :model-value="selectedOption"
-        :label="$gettext('Login status')"
+        :label="$gettext('Login')"
         :options="options"
         :placeholder="$gettext('Select...')"
         :warning-message="
--- a/packages/web-app-admin-settings/src/components/Users/LoginModal.vue
+++ b/packages/web-app-admin-settings/src/components/Users/LoginModal.vue
@@ -8,6 +8,7 @@
     @confirm="$emit('confirm', { users, value: selectedOption.value })"
   >
     <template #content>
+      <p v-text="$gettext('Specify for the selected users whether they can log in.')" />
       <oc-select
         :model-value="selectedOption"
         :label="$gettext('Login')"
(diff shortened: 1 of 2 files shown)


JammingBen:
We recently tried to unify modals and removed texts above input fields/moved such texts to the modal title (see https://github.com/owncloud/web/issues/8579). IMO the text here doesn't justify to break with this approach.

"Login status" -> "Login" I agree, then we have it consistent with the right sidebar 👍

hurradieweltgehtunter:
> We recently tried to unify modals and removed texts above input fields/moved such texts to the modal title (see #8579). IMO the text here doesn't justify to break with this approach.
> 
> "Login status" -> "Login" I agree, then we have it consistent with the right sidebar 👍

Alright, didn't knew that. Removed it.

JammingBen (approved):

hurradieweltgehtunter:
hm build fails because of documentation `Unable to locate config file`. What to do here?

packages/web-app-admin-settings/src/components/Users/LoginModal.vue
kulmann, line 1: Don't know if this is a rebase-gone-wrong, but I don't see a code change for the new message in the modal. Only the changed label string.

Judging by the screenshot, please use `$ngettext` and provide different strings for a single selected user or multiple selected users.

kulmann (changes requested):

packages/web-app-admin-settings/src/components/Users/LoginModal.vue
hurradieweltgehtunter: Message got removed based on a comment by @JammingBen https://github.com/owncloud/web/pull/8853#issuecomment-1513150489

kulmann: ah ok, thanks. makes sense!

kulmann (approved):

Issue #6768, opened 2022-04-11
Pagination works incorrect after changing the number of elements on page 2


ScharfViktor:
branch v5.4.0-rc.4 
web agains ocis docker v 1.19.0

Precondition:
- default pagination. Items per page: 100

Steps:

- Administrator uploads a folder with small files (over 200)
- Go to page "2"
- Change the number of items on the page to 500 or 1000

Expected result:
- Administrator can see all 200+ files on the page

Actual result:
- The page is empty. But If you set up pagination on the first page - works correct

https://user-images.githubusercontent.com/84779829/162714482-e1daa57d-5c8f-4367-be64-784c3c3dd3e6.mov

bishwasojha:
__still getting this issue__

- *probable scenario:* For 1000 images , If `Items per page` is 100, there are 10 pages and,  if `Items per page` is 500 there are 2 pages. Now, if we alter the value of `Items per page` from 100 to 500 on page 3 (or any page other than page 1, 2) we see the error message `No resource found.`. Page 3 won't contain an image because page 1 and 2 can accommodate 1000 images.

[Screencast from 16-11-22 02:33:06 अपराह्न +0545.webm](https://user-images.githubusercontent.com/33860671/202133253-d79147da-6c00-436e-9969-0a7ce3f7b499.webm)

PR #8854, opened 2023-04-18
Fix pagination after increasing items per page

JammingBen:
## Description
Fixes an issue where the file list incorrectly showed no items after paginating and increasing the amount of items per page. This was achieved by resetting the `page` query back to `1` if the user is on a page > 1.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6768

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-pagination-after-increasing-items-per-page
+++ b/changelog/unreleased/bugfix-pagination-after-increasing-items-per-page
@@ -1,0 +1,6 @@
+Bugfix: Pagination after increasing items per page
+
+An issue where the file list incorrectly showed no items after paginating and increasing the amount of items per page has been fixed.
+
+https://github.com/owncloud/web/issues/6768
+https://github.com/owncloud/web/pull/8854
--- a/packages/web-app-files/src/components/AppBar/ViewOptions.vue
+++ b/packages/web-app-files/src/components/AppBar/ViewOptions.vue
@@ -149,12 +149,11 @@ export default defineComponent({
     }
 
     const setItemsPerPage = (itemsPerPage: string) => {
-      const resetPagination = itemsPerPage > unref(itemsPerPageQuery) && unref(currentPage) > 1
       return router.replace({
         query: {
           ...unref(currentRoute).query,
           'items-per-page': itemsPerPage,
-          ...(resetPagination && { page: '1' })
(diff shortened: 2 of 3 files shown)


packages/web-app-files/src/components/AppBar/ViewOptions.vue
```diff
@@ -130,6 +148,21 @@ export default defineComponent({
       }
     }
 
+    const setItemsPerPage = (itemsPerPage: string) => {
+      const resetPagination = itemsPerPage > unref(itemsPerPageQuery) && unref(currentPage) > 1
```
kulmann, line 152: IMO we should go back to page 1 whenever the items per page change. So not only if the items per page increase, but instead on any change

kulmann (changes requested):

packages/web-app-files/src/components/AppBar/ViewOptions.vue
kulmann: I thought we already had this kind of reset in the past. Don't know if it got lost when we moved the pagination params from the store into the composable+query. 🤔 do you know or did you find any orphaned code fragments regarding the page reset?

kulmann: can't find anything either

changelog/unreleased/bugfix-pagination-after-increasing-items-per-page
```diff
@@ -0,0 +1,6 @@
+Bugfix: Pagination after increasing items per page
+
+An issue where the file list incorrectly showed no items after paginating and increasing the amount of items per page has been fixed.
```
kulmann, line 3:
```suggestion
An issue where the file list incorrectly showed no items after paginating and changing the amount of items per page has been fixed.
```

kulmann (approved):

changelog/unreleased/bugfix-pagination-after-increasing-items-per-page
JammingBen: Was about to do the same, but the bug report is still right though, no? The issue only occurred when increasing the amount of items.

kulmann: yes :D fine by me either way

Issue #8860, opened 2023-04-19
Fuse filter does not show results in certain scenarios


JammingBen:
### Steps to reproduce
1. Make sure your ownCloud has our demo users
2. Go to the users list
3. Search for `marie` -> one result -> good
4. Search for `curie` -> no result -> bad
5. Search for `phillips` -> one result -> good
6. Search for `ips` -> no result -> bad

It not only affects this filter but basically all search inputs where we use Fuse.

PR #8863, opened 2023-04-19
Fix fuzzy fuse filter results

JammingBen:
## Description
We want the Fuse `threshold` to be `0` to disable any fuzziness. Additionally, `ignoreLocation` has to be set to `true`. Otherwise the locations of the search terms would play a role as well (which we don't want). See https://fusejs.io/concepts/scoring-theory.html#scoring-theory for more details.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8860

## Types of changes

- [ ] Technical debt
- [ ] Tests


kulmann (approved):
👍

PR #8867, opened 2023-04-19
Fix client service requests with new access token

JammingBen:
## Description
We have to be careful not to destruct the `useClientService` composable in `setup`. The reactivity needed for getting the access token will get lost by doing this.

This is a regression from https://github.com/owncloud/web/pull/8621.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8861

## Types of changes

- [ ] Technical debt
- [ ] Tests


kulmann (approved):
oof size is large 😨

PR #8884, opened 2023-04-20
[full-ci] Webfinger redirect app

kulmann:
## Description
This PR introduces a new app which is responsible for performing an authenticated webfinger discovery and either redirecting the user to their ownCloud instance or showing a comprehensible error message.

In its current state the app is meant for being run in a standalone ownCloud Web deployment - the full fledged ocis instance is the redirect target after the webfinger discovery. ;-) I still decided to put the app into our mono-repo because in the future this will transition into a multi-tenancy instance selection and that makes sense to be run alongside our other apps then.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8883
- Blocked by https://github.com/owncloud/ocis/issues/6102

## How Has This Been Tested?
Not at all, yet. Need to write unit tests...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:
- [x] Nicer state messages ( discovery / error / redirect)
- [x] Change web-runtime to allow a defaultExtension that doesn't have a nav item 😠 

## Followup tasks:
- [ ] Documentation how to set it up
- [ ] Unit tests


packages/web-app-webfinger/src/discovery/webfingerDiscovery.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor(serverUrl: string, clientService: ClientService) {
+    this.serverUrl = serverUrl
+    this.clientService = clientService
+  }
+
+  public async discoverOwnCloudInstances(): Promise<OwnCloudInstance[]> {
+    const client = this.clientService.httpAuthenticated
+    const url =
+      urlJoin(this.serverUrl, '.well-known', 'webfinger') + `?resource=${encodeURI(this.serverUrl)}`
+    const response: OwnCloudInstancesResponse = (await client.get(url)).data
```
kulmann, line 25: fyi the request currently fails in a setup with self signed certificate, see https://github.com/owncloud/ocis/issues/6102

packages/web-app-webfinger/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const routes = () => [
+  {
+    name: 'root',
+    path: '/',
+    redirect: () => {
+      return { name: 'webfinger-resolve' }
+    },
+    meta: {
```
kulmann, line 23: probably too early here 😁 I'm planning to allow defining a root route of an app, so that we don't need to rely on apps having a navItem (which is used right now to identify the first route of an app when setting the app as `options.defaultExtension`).

packages/web-app-webfinger/src/discovery/types.ts
dschmidt, line 1: I think an Instance suffix is a bit weird for a type, maybe Installation or address or something like that would be less confusing as it does not overload oop terms

dschmidt (commented):
So far so good!

packages/web-app-webfinger/src/discovery/types.ts
kulmann: oh true, didn't think of that 👀

ownclouders:
Results for <strong>e2e-tests oCIS</strong> https://drone.owncloud.com/owncloud/web/35103/12/1

<p>


#### npx playwright show-trace https://cache.owncloud.com/public/owncloud/web/35103/tracing/space-members-can-be-invited-with-an-expiration-date-alice-2023-4-25-09-35-02.zip 


</p></details>

packages/web-app-webfinger/src/discovery/webfingerDiscovery.ts
kulmann: doesn't fail anymore 👍

packages/web-app-webfinger/src/discovery/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface OwnCloudInstance {
+  rel: string
+  href: string
+  titles?: any
```
JammingBen, line 4: Can we specify the type a bit more, or can it literally be anything?

packages/web-app-webfinger/src/views/Resolve.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <div
+    class="webfinger-resolve oc-height-viewport oc-flex oc-flex-column oc-flex-center oc-flex-middle"
+  >
+    <div class="oc-card oc-card-body oc-text-center oc-width-large">
+      <template v-if="hasError">
+        <h2 key="webfinger-resolve-error">
```
JammingBen, line 7: Don't we need a `h1` to be semantically correct?

packages/web-app-webfinger/src/views/Resolve.vue
```diff
@@ -0,0 +1,86 @@
...
+
+export default defineComponent({
+  name: 'WebfingerResolve',
+  setup() {
+    const configurationManager = useConfigurationManager()
+    const clientService = useClientService()
+    const loadingService = useLoadingService()
+
+    const ownCloudInstances = ref<OwnCloudInstance[]>([])
+    const isLoading = computed(() => loadingService.isLoading)
```
JanAckermann, line 36: what is this for?

packages/web-app-webfinger/src/views/Resolve.vue
```diff
@@ -0,0 +1,86 @@
...
+      if (instances.length === 0) {
+        return
+      }
+      // we can't deal with multi-instance results. just pick the first one for now.
+      window.location.href = ownCloudInstances.value[0].href
+    })
+
+    return {
+      ownCloudInstances,
+      isLoading,
```
JanAckermann, line 62: same

JanAckermann (changes requested):

packages/web-app-webfinger/src/views/Resolve.vue
```diff
@@ -0,0 +1,86 @@
...
+  <div
+    class="webfinger-resolve oc-height-viewport oc-flex oc-flex-column oc-flex-center oc-flex-middle"
+  >
+    <div class="oc-card oc-card-body oc-text-center oc-width-large">
+      <template v-if="hasError">
+        <h2 key="webfinger-resolve-error">
+          <span v-text="$gettext('Sorry!')" />
+        </h2>
+        <p v-text="$gettext('Something went wrong.')" />
+        <p v-text="$gettext('(We could not resolve the destination)')" />
```
JanAckermann, line 11:
```suggestion
        <p v-text="$gettext('(We could not resolve the destination.)')" />
```

JanAckermann (changes requested):

packages/web-app-webfinger/src/discovery/types.ts
kulmann: was just too lazy during development 😁  changed to `Record<string, string>`

packages/web-app-webfinger/src/views/Resolve.vue
kulmann: added the `h1` as screenreader-only page title, like we do in basically all other views.

packages/web-app-webfinger/src/views/Resolve.vue
kulmann: dev leftover :-) removed it

packages/web-app-webfinger/src/discovery/types.ts
kulmann: changed to `OwnCloudServer`

JanAckermann (approved):
😍

JammingBen (approved):

Issue #8881, opened 2023-04-20
[space trashbin] opening the sidebar of disabled space results in failing request


hurradieweltgehtunter:
### Steps to reproduce
1. Create space
2. Disable space
3. Go to "Deleted files" -> select space trash bin
4. Click on sidebar toggle icon

### Expected behaviour
Tell us what should happen

### Actual behaviour
Loading spinner keeps spinning. Request fails:
<img width="978" alt="grafik" src="https://user-images.githubusercontent.com/16665512/233439413-76c61fa4-bae7-4c50-b0c2-f502585a0e51.png">

Opening sidebar of enabled space shows:
<img width="425" alt="grafik" src="https://user-images.githubusercontent.com/16665512/233440164-59fdd2ce-d824-44a4-bfbe-07f792ec63c9.png">

JanAckermann:
Huh? Shoudn't be possible at all do navigate into the trashbin of a disabled space, let me check =)

hurradieweltgehtunter:
interestingly, entering a trash bin of a disabled space won't show any deleted files.
But I can see, you are already on it :)

PR #8885, opened 2023-04-21
Don't show disabled spaces in the trash overview

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8881

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/views/trash/Overview.vue
+++ b/packages/web-app-files/src/views/trash/Overview.vue
@@ -187,7 +187,6 @@ export default defineComponent({
     const getSpaceAttributes = (space: SpaceResource) => {
       if (isProjectSpaceResource(space) && space.disabled === true) {
         return {
-          type: 'span',
           disabled: true
         }
       }
--- a/packages/web-app-files/src/views/trash/Overview.vue
+++ b/packages/web-app-files/src/views/trash/Overview.vue
@@ -177,7 +177,7 @@ export default defineComponent({
       if (isPersonalSpaceResource(space)) {
         return $gettext('Personal')
       }
-      if (space.disabled !== false) {
+      if (space.disabled === true) {
         return `${space.name} (${$gettext('disabled')})`
       }
 
--- a/packages/web-app-files/src/views/trash/Overview.vue
+++ b/packages/web-app-files/src/views/trash/Overview.vue
@@ -185,7 +185,7 @@ export default defineComponent({
     }
 
     const getSpaceAttributes = (space: SpaceResource) => {
-      if (isProjectSpaceResource(space) && space.disabled !== false) {
+      if (isProjectSpaceResource(space) && space.disabled === true) {
         return {
           disabled: true
         }
--- a/packages/web-app-files/src/views/trash/Overview.vue
+++ b/packages/web-app-files/src/views/trash/Overview.vue
@@ -178,7 +178,7 @@ export default defineComponent({
         return $gettext('Personal')
       }
       if (space.disabled === true) {
-        return `${space.name} (${$gettext('disabled')})`
+        return $gettext('%{spaceName} (disabled)', { spaceName: space.name })
       }
 
       return space.name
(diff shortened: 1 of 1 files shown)


hurradieweltgehtunter:
Is hiding really the best solution? For spaces we also do not hide disabled ones. Maybe better grey out and add `(disabled)` to the name?

JanAckermann:
@hurradieweltgehtunter hm depends, if you have n disabled spaces your trashbin list gets longer and longer and you can't really do anything in this view with the space.

JanAckermann:
@kulmann any opinion?

kulmann:
> @kulmann any opinion?

I would do the same as in the spaces overview. That would mean: show it and disable it. At some point in time it will probably make sense that we get a `hide disabled spaces` option, which should then behave consistently in the spaces overview and the trashbin. But that's an issue for future-Jan.

JanAckermann:
FFS rip future-jan

JanAckermann:
Adjusted 

image

packages/web-app-files/src/views/trash/Overview.vue
```diff
@@ -175,7 +174,27 @@ export default defineComponent({
     ])
 
     const getSpaceName = (space: SpaceResource) => {
-      return isPersonalSpaceResource(space) ? $gettext('Personal') : space.name
+      if (isPersonalSpaceResource(space)) {
+        return $gettext('Personal')
+      }
+      if (space.disabled !== false) {
```
kulmann, line 180:
```suggestion
      if (space.disabled === true) {
```

easier to read :-) and this way the absence of the `disabled` prop in the space object would result in the space being shown as enabled. Which sounds correct to me.

packages/web-app-files/src/views/trash/Overview.vue
```diff
@@ -175,7 +174,27 @@ export default defineComponent({
...
+        return `${space.name} (${$gettext('disabled')})`
```
kulmann, line 181:
```suggestion
        return $gettext('%{spaceName} (disabled)', { spaceName: space.name })
```

Please put as much as possible into transifex, always. 🙈  There are rare cases where concatenation of translated strings is not avoidable. But most of the time it is possible and should be the preferred solution. Translations are a complicated topic, this makes it a little bit easier.

kulmann (changes requested):

packages/web-app-files/src/views/trash/Overview.vue
```diff
@@ -175,7 +174,27 @@ export default defineComponent({
...
+      }
+      if (space.disabled === true) {
+        return $gettext('%{spaceName} (disabled)', { spaceName: space.name })
+      }
+
+      return space.name
+    }
+
+    const getSpaceAttributes = (space: SpaceResource) => {
+      if (isProjectSpaceResource(space) && space.disabled !== false) {
```
kulmann, line 188:
```suggestion
      if (isProjectSpaceResource(space) && space.disabled === true) {
```
oh sorry, here as well

kulmann (approved):
thank you 👏

packages/web-app-files/src/views/trash/Overview.vue
```diff
@@ -175,7 +174,27 @@ export default defineComponent({
...
+        return $gettext('%{spaceName} (disabled)', { spaceName: space.name })
+      }
+
+      return space.name
+    }
+
+    const getSpaceAttributes = (space: SpaceResource) => {
+      if (isProjectSpaceResource(space) && space.disabled === true) {
+        return {
+          disabled: true
```
kulmann, line 190: Is `type: 'span'` needed?

```suggestion
          type: 'span',
          disabled: true
```

JanAckermann: Could you check that in the ui, I don't see any reason why it should be

JanAckermann: Hehe, committed w/o checking and nope the ui representation is quite wrong with a span, because it doesn't represent the disabled state, reverting

kulmann: I WAS TESTING YOU :trollface:

JanAckermann: TRY HARD 🔥🔥🔥

Issue #8902, opened 2023-04-25
[QA][OCIS]e2e test failing in nightly 


SwikritiT:
The following test is failing on CI
```feature
Scenario: space members can be invited with an expiration date (attempt 1, retried) # tests/e2e/cucumber/features/smoke/spaces/memberExpiry.ocis.feature:4
   ✔ Before # tests/e2e/cucumber/environment/index.ts:36
   ✔ Given "Admin" creates following users using API # tests/e2e/cucumber/steps/api.ts:10
       | id    |
       | Alice |
       | Brian |
   ✔ And "Admin" assigns following roles to the users using API # tests/e2e/cucumber/steps/api.ts:24
       | id    | role        |
       | Alice | Space Admin |
   ✔ And "Alice" creates the following project space using API # tests/e2e/cucumber/steps/api.ts:125
       | name | id     |
       | team | team.1 |
   ✔ When "Alice" logs in # tests/e2e/cucumber/steps/ui/session.ts:24
   ✔ And "Alice" opens the "files" app # tests/e2e/cucumber/steps/ui/application.ts:10
   ✔ And "Alice" navigates to the projects space page # tests/e2e/cucumber/steps/ui/spaces.ts:11
   ✔ And "Alice" navigates to the project space "team.1" # tests/e2e/cucumber/steps/ui/spaces.ts:23
   ✔ And "Alice" adds following users to the project space # tests/e2e/cucumber/steps/ui/spaces.ts:54
       | user  | role   | kind |
       | Brian | editor | user |
   ✖ And "Alice" sets the expiration date of the member "Brian" of the project space to "+5 days" # tests/e2e/cucumber/steps/ui/spaces.ts:132
       locator.click: Timeout 30000ms exceeded.
       =========================== logs ===========================
       waiting for locator('.vc-day.id-2023-04-30').first()
         locator resolved to visible <div class="vc-day id-2023-04-30 day-30 day-from-end-…>…</div>
       attempting click action
         waiting for element to be visible, enabled and stable
           element is not stable - waiting...
         element is visible, enabled and stable
         scrolling into view if needed
         done scrolling
         performing click action
         <span tabindex="-1" role="button" aria-disabled="true"…>23</span> from <div class="vc-day id-2023-04-23 day-23 day-from-end-…>…</div> subtree intercepts pointer events
       retrying click action, attempt #1
         waiting for element to be visible, enabled and stable
         element is visible, enabled and stable
         scrolling into view if needed
         done scrolling
[...]

Issue #7513, opened 2022-08-24
Scroller scrolls up automatically while trying to select an expiry date for public link


sushmita56:
### Steps to reproduce
1. First of all reduce the screen resolution 
2. Log in to the WebUI and create a resource
3. Then, create a public link for the resource
4. And click on the public link edit button and then click on the `Add expiration date` option
5. Leave year and month as it is and now try to set the expiration date to the very last date of the month like 28/29/30/31

To select the last date of the month you'll have to scroll down. Whenever you scroll down and are about to click on the date, `say 30`, the scroller scrolls up automatically. 

Discovered such behavior while working on the issue: https://github.com/owncloud/web/issues/7320

### Actual behaviour
The scroller moves up instantly as we hover over the date.

### Expected behaviour
Unless we select the date the scroller was not assumed to shift upward.

### Screenshots

https://user-images.githubusercontent.com/66173400/186376131-e49668a2-09f8-4c03-b1f5-614f37a7a03f.mp4

### Environment general
**Operating system**: Ubuntu

**Backend (ownCloud Core or Infinite Scale)**: Core

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

JammingBen:
Phew, at a first glance, the `v-calendar` lib which is used here does not provide any options to prevent this from happening. One easy solution would be to place the calendar always at the top. What do you think @tbsbdr ?

![image](https://user-images.githubusercontent.com/50302941/186412135-abf4595c-228b-44dc-a7b3-74be00595808.png)

SwikritiT:
unskip this test in oc10 after this issue is fixed https://github.com/owncloud/web/pull/7645/files#diff-8f1afe20f5e88baf2ebfbef0ad7d5870137b6eebb5ea62edc0fa541a077152e0

grgprarup:
Tests failed on nightly https://drone.owncloud.com/owncloud/web/32014/36/18, https://drone.owncloud.com/owncloud/web/32014/67/13, and PRs due to this issue.
Unskip the test https://github.com/owncloud/web/pull/8301/files#:~:text=in%20the%20server-,%40skip,-Write after this issue is fixed.

SwikritiT:
This test failure is also because of this issue https://github.com/owncloud/web/issues/8902

SwikritiT:
Unskip the test skipped in https://github.com/owncloud/web/pull/8905 after this bug is fixed

PR #8905, opened 2023-04-25
[tests-only][full-ci]Skip the expiry date related e2e test

SwikritiT:
The skipped test started to fail because we have a bug in the calendar that the scroller will automatically scroll up while trying to select the end-of-the-month dates https://github.com/owncloud/web/issues/7513 .
This PR skips the test so that it won't obstruct the CI

Also, read the comment below https://github.com/owncloud/web/pull/8905#issuecomment-1521214490

Fixes: https://github.com/owncloud/web/issues/8902

### DIFF
--- a/tests/e2e/cucumber/features/smoke/spaces/memberExpiry.ocis.feature
+++ b/tests/e2e/cucumber/features/smoke/spaces/memberExpiry.ocis.feature
@@ -1,7 +1,7 @@
 Feature: spaces member expiry
 
   # skipped because scroller scrolls up automatically while
-  # tring to select the end of the month dates
+  # trying to select the end of the month dates
   # this test will work at the beginning/mid of a month
   @skip @issue-7513
   Scenario: space members can be invited with an expiration date
--- a/tests/e2e/cucumber/features/smoke/spaces/memberExpiry.ocis.feature
+++ b/tests/e2e/cucumber/features/smoke/spaces/memberExpiry.ocis.feature
@@ -1,6 +1,9 @@
 Feature: spaces member expiry
 
-
+  # skipped because scroller scrolls up automatically while
+  # tring to select the end of the month dates
+  # this test will work at the beginning/mid of a month
+  @skip @issue-7513
   Scenario: space members can be invited with an expiration date
     Given "Admin" creates following users using API
       | id    |
--- a/.drone.star
+++ b/.drone.star
@@ -1191,7 +1191,7 @@ def e2eTests(ctx):
                      "image": OC_CI_NODEJS,
                      "environment": environment,
                      "commands": [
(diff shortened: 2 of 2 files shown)


SwikritiT:
I've skipped the test in this PR but I think we made a decision to not skip the tests in e2e (don't remember the exact reason). Or maybe it was a decision about not adding `@skipOnOcis`, `@skipOnOC10` tags, either way, my memory is cloudy.
So I want to make sure that what's done in this PR is okay. If it's not, other paths we can follow is
1. Comment out the test (which is equivalent to skipping)
2. hard code the date to some distant future that'll be at the begging of the month. (this is like dodging the bug but the bug is known and the test will also run throughout the month)

@individual-it @ScharfViktor @saw-jan

tests/e2e/cucumber/features/smoke/spaces/memberExpiry.ocis.feature
```diff
@@ -1,6 +1,9 @@
 Feature: spaces member expiry
 
-
+  # skipped because scroller scrolls up automatically while
+  # tring to select the end of the month dates
```
phil-davis, line 4:
```suggestion
  # trying to select the end of the month dates
```

ScharfViktor:
> I've skipped the test in this PR but I think we made a decision to not skip the tests in e2e (don't remember the exact reason). Or maybe it was a decision about not adding `@skipOnOcis`, `@skipOnOC10` tags, either way, my memory is cloudy. So I want to make sure that what's done in this PR is okay. If it's not, other paths we can follow is
> 
> 1. Comment out the test (which is equivalent to skipping)
> 2. hard code the date to some distant future that'll be at the beginning of the month. (this is like dodging the bug but the bug is known and the test will also run throughout the month)
> 
> @individual-it @ScharfViktor @saw-jan

honestly, I wouldn't skip test

Maybe increase page size?  using `await page.setViewportSize`

https://user-images.githubusercontent.com/84779829/234213459-ac747c83-887e-4d73-b7b0-91f5f00586bf.mov

SwikritiT:
> Maybe increase page size? using await page.setViewportSize

yes, that would probably work but I want to make sure that we are well aware that we are just tiptoeing over the bug.
So I should increase the size? @ScharfViktor

SwikritiT:
By setting the viewport to 
```
await page.setViewportSize({
    width: 1280,
    height: 800
  })
```

I can make the test pass locally

kulmann:
As soon as there are some shares / members in the list above the increased viewport height doesn't help anymore again... kind of prefer to skip it to unblock PRs against master.

ScharfViktor (approved):
we need add `--tags ~@skip` in ocis too

Issue #8917, opened 2023-04-26
Some items in files context menu cannot be reached


individual-it:
### Steps to reproduce
1. create a space (I could not reproduce the issue with the personal space)
2. upload multiple files to the space
3. check the context menu of every item

### Expected behaviour
all items of the context menu of every item should be reachable

### Actual behaviour
There is one particular file, where the bottom of the context menu is not reachable
![image](https://user-images.githubusercontent.com/2425577/234547844-26dbd2cd-efdb-4199-bd9c-775e97d4f376.png)

### Environment general
**Operating system**:
tried on Ubuntu

**Backend (ownCloud Core or Infinite Scale)**:
infinite scale

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
no

### Environment ownCloud Infinite Scale
**Version:**
3.0.0rc1

**Updated from an older Infinite Scale or fresh install:**
fresh

**Where did you install Infinite Scale from:**
https://download.owncloud.com

### Client configuration
**Browser:**
FF 112
 
**Operating system:**
tried on Ubuntu & Win 10

kulmann:
Does the context menu not become scrollable?

individual-it:
no, when I scroll, the whole file-list scrolls out of view, but the context menu stays where it is
![image](https://user-images.githubusercontent.com/2425577/234550911-0a97bff6-82d6-450a-bc3d-f5899ee1a087.png)

PR #8919, opened 2023-04-26
Fix context menu placement issue caused by offset

JammingBen:
## Description
Fixes another edge case with the context menu position. I removed the default offset modifier as it seems to interfere with the mechanism to find out if the menu should be placed top or bottom. A custom offset can still be defined by passing `popperOptions` to `OcDrop`.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8917

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -171,7 +171,7 @@ export default defineComponent({
       plugins: [hideOnEsc],
       theme: 'none',
       maxWidth: 400,
-      offset: this.offset,
+      offset: this.offset ? this.offset : 0,
```
kulmann, line 174:
```suggestion
      offset: this.offset ?? 0,
```

PR #8921, opened 2023-04-26
[full-ci] Enhancement: Add whitespace context-menu

lookacat:
## Description
We've added a generic context-menu for right clicking on whitespace.
See https://github.com/owncloud/web/issues/5861

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5861

Screenshot (32)

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] https://github.com/owncloud/web/pull/8938

### DIFF
--- a/packages/web-app-files/src/views/spaces/GenericSpace.vue
+++ b/packages/web-app-files/src/views/spaces/GenericSpace.vue
@@ -400,7 +400,7 @@ export default defineComponent({
       const filesViewWrapper = document.getElementsByClassName('files-view-wrapper')[0]
       filesViewWrapper?.addEventListener('contextmenu', (event) => {
         const { target } = event
-        if ((target as HTMLElement).closest('.no-whitespace')) {
+        if ((target as HTMLElement).closest('.has-item-context-menu')) {
           return
         }
         event.preventDefault()
--- a/packages/design-system/src/components/OcTable/OcTable.vue
+++ b/packages/design-system/src/components/OcTable/OcTable.vue
@@ -35,7 +35,7 @@
         </oc-th>
       </oc-tr>
     </oc-thead>
-    <oc-tbody class="no-whitespace">
+    <oc-tbody class="has-item-context-menu">
       <oc-tr
         v-for="(item, trIndex) in data"
         :key="`oc-tbody-tr-${itemDomSelector(item) || trIndex}`"
--- a/packages/web-app-admin-settings/tests/unit/components/Spaces/__snapshots__/SpacesList.spec.ts.snap
+++ b/packages/web-app-admin-settings/tests/unit/components/Spaces/__snapshots__/SpacesList.spec.ts.snap
@@ -97,7 +97,7 @@ exports[`SpacesList should render all spaces in a table 1`] = `
         </th>
       </tr>
     </thead>
-    <tbody class="no-whitespace">
+    <tbody class="has-item-context-menu">
       <tr class="oc-tbody-tr oc-tbody-tr-1" data-item-id="1" draggable="false" tabindex="-1">
         <td class="oc-table-cell oc-table-cell-align-left oc-table-cell-align-middle oc-table-cell-width-shrink oc-td oc-table-data-cell oc-table-data-cell-select oc-pl-s">
           <oc-checkbox-stub class="oc-ml-s" disabled="false" hidelabel="true" id="oc-checkbox-2" label="Select 1 Some space" modelvalue="false" option="[object Object]" outline="false" size="large"></oc-checkbox-stub>
--- a/packages/web-app-files/src/components/FilesList/ResourceTiles.vue
+++ b/packages/web-app-files/src/components/FilesList/ResourceTiles.vue
@@ -35,7 +35,7 @@
       </oc-drop>
     </div>
     <oc-list class="oc-tiles oc-flex" :class="resizable ? 'resizableTiles' : ''">
-      <li v-for="resource in data" :key="resource.id" class="oc-tiles-item no-whitespace">
+      <li v-for="resource in data" :key="resource.id" class="oc-tiles-item has-item-context-menu">
         <oc-tile
           :ref="(el) => (tileRefs.tiles[resource.id] = el)"
           :resource="resource"
--- a/packages/web-app-files/tests/unit/components/FilesList/__snapshots__/ResourceTiles.spec.ts.snap
+++ b/packages/web-app-files/tests/unit/components/FilesList/__snapshots__/ResourceTiles.spec.ts.snap
@@ -13,10 +13,10 @@ exports[`ResourceTiles component renders an array of spaces correctly 1`] = `
 <div>
   <!--v-if-->
   <oc-list class="oc-tiles oc-flex">
-    <li class="oc-tiles-item no-whitespace">
+    <li class="oc-tiles-item has-item-context-menu">
       <oc-tile draggable="false" is-resource-selected="false" resource="[object Object]" resource-icon-size="xlarge" resource-route="[object Object]"></oc-tile>
     </li>
-    <li class="oc-tiles-item no-whitespace">
+    <li class="oc-tiles-item has-item-context-menu">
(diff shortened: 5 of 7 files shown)


lookacat:
Refactoring of the CreateAndUpload component into composables needed before this PR
https://github.com/owncloud/web/pull/8938

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -393,13 +397,36 @@ export default defineComponent({
...
+        })
     })
 
     onBeforeUnmount(() => {
       visibilityObserver.disconnect()
       eventBus.unsubscribe('app.files.list.load', loadResourcesEventToken)
     })
 
+    const genericContextMenu = ref(null)
+    const showContextMenu = (event) => {
```
lookacat, line 422: someone mentioned it would be better to do all this logic in one component which i am not a fan of since this logic toggles the whole component

tbsbdr:
cool one 👍
A remark on Icon+wording: I'd suggest du directly show the folder icon (instead of the "+") and just call it "New Folder" (`create new` is semantically redundant)

image

kulmann (changes requested):
Didn't look into the code, yet, but tried it out. Really awesome stuff! Two things that I'd like to have different:
- click on whitespace should clear the selection
- click on "details" in the whitespace context menu doesn't show the details of the current folder

Could you work on that as well?

lookacat:
@kulmann to the first point yes possible, the second point: Did you try it in the space root? in subfolders it should work, my assumption was there is no detail on the e.g. personal space root? Or did i miss something?

lookacat:
@kulmann alright shouldl be doable, the first point was right, i just fixed it ^^

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -393,13 +397,36 @@ export default defineComponent({
           performLoaderTask(true, path, fileId)
         }
       )
+      const filesViewWraapper = document.getElementsByClassName('files-view-wrapper')[0]
```
JammingBen, line 400:
```suggestion
      const filesViewWrapper = document.getElementsByClassName('files-view-wrapper')[0]
```

packages/web-app-files/src/components/Spaces/SpaceGenericContextMenu.vue
kulmann, line 1: Why did you decide to not use the `ContextActionMenu.vue` component? 🤔

packages/web-app-files/src/views/spaces/GenericSpace.vue
kulmann, line 400:
```suggestion
      const filesViewWrapper = document.getElementsByClassName('files-view-wrapper')[0]
```

packages/web-app-files/src/components/Spaces/SpaceGenericContextMenu.vue
kulmann, line 1: Could you please use a component name which makes clear that it's hooked to whitespace? Don't have a good name in mind, yet though...

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -393,13 +397,36 @@ export default defineComponent({
...
+        showContextMenu(newEvent)
+      })
     })
 
     onBeforeUnmount(() => {
       visibilityObserver.disconnect()
       eventBus.unsubscribe('app.files.list.load', loadResourcesEventToken)
     })
 
+    const genericContextMenu = ref(null)
```
kulmann, line 420: `genericContextMenu` is not a good name IMO. `whitespaceContextMenu` maybe?

packages/web-app-files/src/components/Spaces/SpaceGenericContextMenu.vue
lookacat: but will try to use it, shouldn't be much difference between those two besides the sections

lookacat: arg, my assumption was wrong, can't get the ContextActionMenu to work at this point because _tippy is always undefined, would rename it to WhitespaceContextMenu for now :/

lookacat: don't see how to quickly fix that, also there are some differences between those two component

packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
```diff
@@ -115,12 +115,19 @@ export const useFileActionsCreateNewFolder = ({
...
         },
-        isEnabled: ({ resources }) => {
+        isEnabled: ({}) => {
+          if (
+            isProjectSpaceResource(space) &&
+            !space.isEditor(store.getters.user) &&
+            !space.isManager(store.getters.user)
+          ) {
+            return false
+          }
```
JammingBen, line 130: This works for spaces, however, not for shares. You can simply use `space.canUpload()` I believe to cover both cases.

packages/web-app-files/src/views/spaces/GenericSpace.vue
JammingBen, line 400: There is still a typo 😄

packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
lookacat: oh thanks

packages/web-app-files/src/views/spaces/GenericSpace.vue
lookacat: oops

JammingBen (approved):

packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
```diff
@@ -115,13 +115,13 @@ export const useFileActionsCreateNewFolder = ({
...
+        icon: 'folder',
         handler,
         label: () => {
-          return $gettext('Create new Folder')
+          return $gettext('New Folder')
         },
-        isEnabled: ({ resources }) => {
-          return true
+        isEnabled: () => {
+          return unref(currentFolder)?.canUpload({ user: store.getters.user })
```
kulmann, line 124: this should use `canCreate` instead of  `canUpload`. Weirdly both `canUpload` and  `canCreate` are implemented in the same way (they check if the dav permission for folder creation is present), so you could argument that it doesn't make a difference in functionality... but semantically the creation check is the right one for folder creation. 

Side note: For the other composable (creating files) I prefer the `canUpload` check (you already have it) because if one can upload a file that's the same as creating a file. So that one should stay like it is. 👍

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -393,13 +397,36 @@ export default defineComponent({
           performLoaderTask(true, path, fileId)
         }
       )
+      const filesViewWrapper = document.getElementsByClassName('files-view-wrapper')[0]
+      filesViewWrapper?.addEventListener('contextmenu', (event) => {
+        const { target } = event
+        if (
+          (target as HTMLElement).closest('tbody') ||
```
kulmann, line 404: I would prefer if you would a) define a new dummy class, b) apply that to both the tbody and all the tile list items and c) use that new dummy class as selector here. That way you don't get a dependency on specific file view modes. If we ever implement a third file view mode it would be sufficient to give the correct item your dummy class to make the whitespace context menu work.

kulmann (changes requested):
Found two more things. Also, what about unit tests, at least for opening the whitespace context menu (not for the actions inside it)?

lookacat:
@kulmann done

kulmann (approved):
Awesome stuff! Love the UX of this 😍

PR #8950, opened 2023-04-28
feat: idp auth context

kulmann:
## Description

Added an `idp` authContext type and used it for the webfinger resolve page. Idea is that `idp` requires a valid login in contrast to `user` which also requires an ownCloud user. The webfinger app doesn't need an ownCloud backend, only the webfinger service.

As a consequence the `idp` context being ready doesn't lead to loading user, capabilities, spaces, etc.

This again increases the urgency of documenting at least parts of the extension system. The authentication requirements in route definitions are thought through to a good extend, this needs to be written down and explained. Out of scope for this PR though...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-runtime/src/index.ts
+++ b/packages/web-runtime/src/index.ts
@@ -118,19 +118,6 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> => 
 
   store.watch(
     (state, getters) => {
-      return getters['runtime/auth/isIdpContextReady']
-    },
-    async (idpContextReady) => {
-      if (!idpContextReady) {
-        return
-      }
-      // fake spaces being loaded so that the ApplicationLayout doesn't stay in loading state
-      store.commit('runtime/spaces/SET_SPACES_INITIALIZED', true)
-    }
-  )
-
-  store.watch(
-    (state, getters) => {
       return getters['runtime/auth/isUserContextReady']
     },
     async (userContextReady) => {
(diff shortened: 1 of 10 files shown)


packages/web-runtime/src/index.ts
```diff
@@ -115,6 +116,19 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> =>
...
+  store.watch(
+    (state, getters) => {
+      return getters['runtime/auth/isIdpContextReady']
+    },
+    async (idpContextReady) => {
+      if (!idpContextReady) {
+        return
+      }
+      // fake spaces being loaded so that the ApplicationLayout doesn't stay in loading state
+      store.commit('runtime/spaces/SET_SPACES_INITIALIZED', true)
```
JanAckermann, line 128: ughhhhhhhh, maybe we add something else like idpContext and check on that in ApplicationLayout
like that runtime/auth/isIdpContextReady

JanAckermann (changes requested):

packages/web-runtime/src/index.ts
kulmann: uh, you're right, thanks! In the `Application`-layout I can just add `idp` to the list of authContext types that ignore if spaces are loaded. 🙈  Will do!

kulmann: done

JanAckermann (approved):
Codewise LGTM

Issue #8895, opened 2023-04-24
[account settings] wrong language selected


hurradieweltgehtunter:
Make sure to have a non-english browser language set, in my example german

### Steps to reproduce
1. Login with new user (first login)
2. Go to account settings

### Expected behaviour
Language selection box should match the language in use

### Actual behaviour
UI is in german but language selector shows "english"

![grafik](https://user-images.githubusercontent.com/16665512/233955541-c4a44b09-4afb-4d6f-99ab-77ada2d97063.png)

kulmann:
> As far as I know this happens because English is being set as default language in oCIS: https://github.com/owncloud/ocis/blob/master/services/settings/pkg/service/v0/settings.go#L161.
> 
> @kulmann What do you think, maybe we just ignore that setting in Web and display the current language if no language is set for the user?

The code you referenced defines basically the "recommended default option", but oCIS doesn't apply the recommended default. IMO fine to ignore the default. I see two options:
1. select the browser language if it is present in the list. con: what do you select if the browser language is not present in the list?
2. add a new option `browser language` which is selected if the backend doesn't have a value for the setting (= the default). con: users will expect that they can switch back to the undefined value / `browser language` option and to my knowledge that's not possible.

Pick any of the two as a short term solution. ;-)

My preferred option is 3: convert the single choice setting to a string setting in the settings service code base and just show the languages in a dropdown in web that we already have defined in the web runtime. That way we wouldn't need to define in the ocis code base which languages are supported by web (which is quite controversial at the moment) and could easily fall back to an empty string for using the browser language.

tbsbdr:
Adding to @kulmann `s option 1)
> what do you select if the browser language is not present in the list?

How about showing then showing a "Select ..." as placeholder?
![screenshot_000846](https://user-images.githubusercontent.com/26610733/235076381-67b724ba-c370-41f3-bf34-29a2dbc10551.png)

JanAckermann:
We could hack a little more, this applies if no lang is set in accounts settings service (bakend):

1. extract locale by navigator.language 
2. lookup if the language is existing in the lang settings (backend?
3. If yes, pre-select in select box, if now -> please choose

JammingBen:
> extract locale by navigator.language

This is actually handled already when the translations are being loaded. Can't remember exactly how it's called, `currentLanguage` or so. But that needs to be the fallback IMO! So your 1. and 2. need to be switched, otherwise the browser language overwrites the setting, no?

[...]

PR #8958, opened 2023-04-28
Set select value to browser lang if not set in accounts service

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8895

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/pages/account.vue
+++ b/packages/web-runtime/src/pages/account.vue
@@ -77,6 +77,7 @@
         <dd data-testid="language">
           <oc-select
             v-if="languageOptions"
+            :placeholder="$gettext('Please choose...')"
             :model-value="selectedLanguageValue"
             :clearable="false"
             :options="languageOptions"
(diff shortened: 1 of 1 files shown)


packages/web-runtime/src/pages/account.vue
```diff
@@ -335,11 +336,16 @@ export default defineComponent({
...
-            )
-          : unref(languageOptions)?.find((o) => o.default)
+
+        if (languageConfiguration) {
+          selectedLanguageValue.value = unref(languageOptions)?.find(
+            (lO) => lO.value === languageConfiguration.value?.listValue?.values?.[0]?.stringValue
+          )
+        } else {
+          selectedLanguageValue.value = unref(languageOptions).find(
+            (lO) => lO.value === navigator.language
```
JammingBen, line 346: This is what I meant. Please no `navigator.language` - that's already handled when the translations are being initialized.

```suggestion
            (lO) => lO.value === langauge.defaultLanguage
```

JanAckermann: yes, saw that, but I even dont know if we take this solution at all, this ha also some downsides

JanAckermann: not language.defaultLanguage but language.current

JanAckermann: I don't think thats the way to go, I guess current will fallback to en if its a language we don't support, so we will have set English there istead of Please choose, idk

JanAckermann: ok checked with zh, current is fine

Issue #8962, opened 2023-05-02
Current language is not highlighted in select field


JammingBen:
"English" should be highlighted:

<img width="831" alt="image" src="https://user-images.githubusercontent.com/50302941/235611620-512b0b34-01c9-4ee1-a8da-4dea2c6a6a50.png">

JanAckermann:
not again :_;

JammingBen:
Yeah :( I think our recent refactoring caused this.

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8972.

PR #8911, opened 2023-04-25
Disable notification emails 

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8904

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #8972, opened 2023-05-03
Fix highlighting of the currently selected language

JammingBen:
## Description
Fixes a regression due to the refactoring in https://github.com/owncloud/web/pull/8911. Also adds proper types for the settings stuff.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8962

## Screenshots:
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/helpers/settings.ts
+++ b/packages/web-runtime/src/helpers/settings.ts
@@ -1,0 +1,53 @@
+export interface SettingValue {
+  identifier: {
+    bundle: string
+    extension: string
+    setting: string
+  }
+  value: {
+    accountUuid: string
+    bundleId: string
+    id: string
+    resource: {
+      type: string
+    }
+    settingId: string
+    boolValue?: boolean
+    listValue?: {
+      values: {
+        stringValue: string
+      }[]
+    }
+  }
+}
+
+export interface AccountBundleSetting {
+  description: string
+  displayName: string
+  id: string
+  name: string
+  resource: {
+    type: string
+  }
+  singleChoiceValue?: {
+    options: Record<string, any>[]
+  }
+  boolValue?: Record<string, any>
+}
+
+export interface AccountBundle {
+  displayName: string
+  extension: string
+  id: string
+  name: string
+  resource: {
+    type: string
+  }
+  settings: AccountBundleSetting[]
+  type: string
+}
+
+export interface LanguageOption {
+  label: string
+  value: string
+}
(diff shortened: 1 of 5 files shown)


JanAckermann (approved):
:star_struck:

dschmidt (approved):

packages/web-runtime/src/helpers/settings.ts
```diff
@@ -0,0 +1,53 @@
...
+  resource: {
+    type: string
+  }
+  singleChoiceValue?: {
+    options: Record<string, any>[]
+  }
+  boolValue?: Record<string, any>
+}
+
+export interface AccountBundle {
```
kulmann, line 38:
```suggestion
export interface SettingsBundle {
```

packages/web-runtime/src/helpers/settings.ts
```diff
@@ -0,0 +1,53 @@
...
+    boolValue?: boolean
+    listValue?: {
+      values: {
+        stringValue: string
+      }[]
+    }
+  }
+}
+
+export interface AccountBundleSetting {
```
kulmann, line 24:
```suggestion
export interface Setting {
```

packages/web-runtime/src/helpers/settings.ts
```diff
@@ -0,0 +1,53 @@
+export interface SettingValue {
```
kulmann, line 1:
```suggestion
export interface SettingsValue {
```

packages/web-runtime/src/pages/account.vue
```diff
@@ -152,10 +153,10 @@ export default defineComponent({
     const { $gettext } = language
     const clientService = useClientService()
     const configurationManager = useConfigurationManager()
-    const valuesList = ref()
-    const bundlesList = ref()
-    const selectedLanguageValue = ref()
-    const disableEmailNotificationsValue = ref()
+    const valuesList = ref<SettingValue[]>()
+    const bundlesList = ref<AccountBundle>()
```
kulmann, line 157: Why is the bundles list not of type `ref<AccountBundle[]>`? 🤔

kulmann (commented):
Not too happy with the settings related interface names. Why `Account`? 🤔

packages/web-runtime/src/pages/account.vue
JammingBen: Because `bundlesList` is actually not a list of bundles but the account bundle. Let my fix that name in a follow-up, just like the new interface names :)

kulmann: thank you 💪

Issue #8932, opened 2023-04-27
[Preview app] Shows images that are not in the current file list


JammingBen:
### Steps to reproduce
1. Upload 3 images to your personal space
2. Create a public link for one of them
3. Go to the `Shared via link` page and open the image with the preview app

### Expected behaviour
There should only be this one image.

### Actual behaviour
The app lets you switch between all 3 images form your personal space. The same goes for the `Shared with others` and the `Search` view.

jacob-nv:
@JammingBen Unable to replicate

dschmidt:
I'm rather sure it should be still an issue.

It was one of the reasons the concept of folder loaders were introduced so the context for images could always be retrieved in especially the files app (and then we never got round to push it forward). 

You basically need to find a place were an image is shown not from it's original location but from a "virtual" place like a share, search or maybe public link. Then reload the page so the context needs to be loaded and that should trigger the faulty behavior.

JammingBen:
Yes, this is still an issue:

https://github.com/owncloud/web/assets/50302941/818c0c23-7dc0-48a8-a6da-fe4900e0e49f

jacob-nv:
Ah, sorry, that's misunderstanding on my side.

JammingBen:
Solution for this is being described here: https://github.com/owncloud/web/pull/11893#discussion_r1839834883

PR #8975, opened 2023-05-04
Fix preview app file loading

JammingBen:
## Description
Fixes a bug where the preview app would load the files of an image's source location instead of the files in the actual file list.

This happened because the preview app called `loadFolderForFileContext` which loads all files of the current file's source location. This is not needed IMO (at least currently) because you can only open the preview app from a files list where files are already loaded.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8932

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/bugfix-preview-file-loading
+++ b/changelog/unreleased/bugfix-preview-file-loading
@@ -1,0 +1,6 @@
+Bugfix: Preview app file loading
+
+A bug where the preview app would load the files of an image's source location instead of the files in the actual file list has been fixed.
+
+https://github.com/owncloud/web/issues/8932
+https://github.com/owncloud/web/pull/8975
--- a/packages/web-app-preview/src/App.vue
+++ b/packages/web-app-preview/src/App.vue
@@ -266,11 +266,10 @@ export default defineComponent({
     }
   },
 
-  async mounted() {
+  mounted() {
     // keep a local history for this component
     window.addEventListener('popstate', this.handleLocalHistoryEvent)
     document.addEventListener('fullscreenchange', this.handleFullScreenChangeEvent)
-    await this.loadFolderForFileContext(this.currentFileContext)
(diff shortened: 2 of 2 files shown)


JanAckermann (approved):
💪

kulmann (changes requested):
Unfortunately not that simple, sorry. Works when you come from the files app (from whatever view you had), but is entirely broken in case you do a page reload / load from a bookmark.

JammingBen:
> Unfortunately not that simple, sorry. Works when you come from the files app (from whatever view you had), but is entirely broken in case you do a page reload / load from a bookmark.

Arghs, I feared some scenario like that would exist 😄

JammingBen:
Unless somebody has an easy workaround, I'd like to postpone that fix because it gets quite complicated.

The folder loader which is used to load the files does this based on the given space. While this is fine for generic space views, it is not for views like `Shared via link`. As a result, we need to load the files individually based on the current view, ideally by re-using (parts of) the corresponding folder loader - which of course lives in the files app.

kulmann:
> Unless somebody has an easy workaround, I'd like to postpone that fix because it gets quite complicated.
> 
> The folder loader which is used to load the files does this based on the given space. While this is fine for generic space views, it is not for views like `Shared via link`. As a result, we need to load the files individually based on the current view, ideally by re-using (parts of) the corresponding folder loader - which of course lives in the files app.

I suppose that @dschmidt has an opinion on this. ;-) Loading the files for the correct context (including the loader tasks) should be part of the `useAppDefaults` composable. Or in other words: extension developers need this kind of functionality. Quite some refactoring involved. I don't see an easy workaround.

dschmidt:
I'm happy if F5 works 😬

JammingBen:
Closing here for now as the solution will probably look totally different from this PR.

Issue #8599, opened 2023-03-11
Hard to see file names in file listing in webui


gennaios:
## Is your feature request related to a problem? Please describe.

As is, from the number of columns displayed, there is only so much space for the file name such that maybe around 20 characters are shown. I have numerous files where name could be 50–100+ characters.

## Describe the solution you'd like

Perhaps something closer to how file listings are done in a more typical file manager, with some columns optional, or using some other way such as icons grouped together under some metadata column, or more button, or something. I think file name perhaps could take up 50% to 2/3’s or so of the panel width, with then optional columns date modified, date created, and other. Maybe current icons towards the right are better placed under the more button/menu.

## Additional context

![B8BAB4F9-3F21-44A5-91AF-615CCD9B8311](https://user-images.githubusercontent.com/12172/224464382-e14cc8c8-1c46-49e5-ae18-a6999efe32fc.jpeg)

tbsbdr:
just for comparison if we'd want to stick with the auto-size approach: google drive solved it quite well (i like it):

## Auto-columns in google drive
Name-Column always has precedence:

https://user-images.githubusercontent.com/26610733/224774780-84d70136-d6af-4754-871b-5aad43471eb2.mp4



## Web UI currently

https://user-images.githubusercontent.com/26610733/224774347-89f87182-b2e8-443d-b03b-e84797fe076c.mp4

gennaios:
Having configurable columns is an option but I'm not sure if that is best.

Perhaps such is a good time to rethink what one should be able to do from the web. Comparing various UIs such as Dropbox, or other table interfaces such as web mail:

Some prioritize name/title and hardly anything else. Dropbox on phones adapts as such where only appears a checkbox, file name, and an action button.

![image](https://user-images.githubusercontent.com/12172/224871613-fedaf39e-304b-4c4b-a6f2-0335449e5b97.jpeg)

Other UIs show a date column or sometimes each table row representing one file is two rows with file name and then date modified or other below. I think Dropbox mobile app is as such.

Maybe checkbox, file name, date modified, and one action button is the minimum to show (on larger screens), with date modified not shown on smaller screens such as mobile phones? Other UI elements currently shown in the file list such as rename either already are in or could be moved to the action menu. I'm just concerned with showing what is more essential and having such adapt to various screens.

On multiple select, Dropbox also offers move and share. That too seems useful.

tbsbdr:
cc @lookacat

PR #8978, opened 2023-05-04
[full-ci] feat: shrink most table columns to their minimum

kulmann:
## Description
- shrinks table columns that took a weirdly high amount of horizontal space
- adds a prop validator for allowed view modes in the resource table
- invisible sort buttons in the table header no don't take up horizontal space anymore (otherwise alignment looks weird for shrinked columns)
- gave more horizontal padding to the table cells so that the overall look and feel of the table is not crimped

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8599

### now:
Screenshot 2023-05-04 at 12 09 54

### before:
Screenshot 2023-05-04 at 12 10 55

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/enhancement-shrink-table-columns
+++ b/changelog/unreleased/enhancement-shrink-table-columns
@@ -1,0 +1,6 @@
+Enhancement: Shrink table columns
+
+We've decided to shrink most table columns in the file list to their absolute minimum (Shares, Size, Modified/Shared/Delete date, Actions) so that the file name can take up more horizontal space.
+
+https://github.com/owncloud/web/issues/8599
+https://github.com/owncloud/web/pull/8978
--- a/packages/design-system/src/components/OcAvatars/OcAvatars.vue
+++ b/packages/design-system/src/components/OcAvatars/OcAvatars.vue
@@ -164,7 +164,7 @@ export default defineComponent({
 .oc-avatars {
   display: inline-flex;
   box-sizing: border-box;
-  flex-flow: row wrap;
+  flex-flow: row nowrap;
   gap: var(--oc-space-xsmall);
   width: fit-content;
 
--- a/packages/design-system/src/components/OcTable/OcTable.vue
+++ b/packages/design-system/src/components/OcTable/OcTable.vue
@@ -19,7 +19,7 @@
           <oc-button
             v-if="field.sortable"
             :aria-label="getSortLabel(field.name)"
-            :class="{ 'oc-invisible': sortBy !== field.name }"
+            :class="{ 'oc-invisible-sr': sortBy !== field.name }"
             class="oc-button-sort"
             variation="passive"
             appearance="raw"
--- a/packages/design-system/src/components/_OcTableCell/_OcTableCell.vue
+++ b/packages/design-system/src/components/_OcTableCell/_OcTableCell.vue
@@ -58,7 +58,7 @@ export default defineComponent({
 <style lang="scss">
 .oc-table-cell {
   /* padding is not configurable until we need it */
-  padding: 0 var(--oc-space-xsmall);
+  padding: 0 var(--oc-space-small);
   position: relative;
 
   &-align {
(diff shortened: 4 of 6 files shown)


kulmann:
Targeting this enhancement to `master` as it's clearly not a bugfix... ;-)

Also, I know that @lookacat you discussed this yesterday with @tbsbdr - heard about it too late and already had most of the changes ready... sorry that I took this issue away from you. 🤗 

@tbsbdr Please try out this PR. Especially if the click on sortable headers feels jumpy for you (table headers that are aligned to the right get pushed/pulled horizontally a little bit when the sort icon appears/disappears).

tbsbdr (changes requested):
woha, nice!

1. click on sortable header: yes, feels a little bit jumpy; If you figure out a solution on how to avoid the jumpynes within reasonable efforts, I would be happy to avoid jumpynes. If a fix is too expensive, I would trade jumpynes for better readibility of the filename.
2. spacing between checkbox and resource icon is increased; can't judge anymore if this is an improvement as I'm too much used to the ui. Only thing that would speak against increasing it is so save space. up to you, this is just my 2 cents. 

screenshot_000885

tbsbdr:
especially makes a difference on small screens ❤️ best to test if one decreases resoultion to somewhat 1200 px width

kulmann:
@tbsbdr regarding jumpiness and paddings:

### jumpiness

We can avoid the jumpiness by keeping the width of the sorting icons even in the invisible state. The effect is that column headers that are aligned to the right have an offset from the right cell border (width of an oc-icon basically). Header text and body cell texts don't appear aligned anymore then. See screenshot (e.g. the `Modified` column). I would prefer to use the correct right alignment and accept the jumpiness.

Screenshot 2023-05-04 at 14 49 19

### paddings

With smaller horizontal paddings I had the feeling that all the shrinked columns feel very crowded. See screenshot, especially the table headers. But I'm fine with both the smaller (`xsmall`) and bigger (`small`) horizontal padding.

Screenshot 2023-05-04 at 14 48 47

tbsbdr (approved):

tbsbdr:
ah okay, thanks for the explanation. then all fine from my side 👍

JammingBen:
Hmm, maybe it's because I need to get used to it, but I'm not the biggest fan, especially on bigger screens. The right side looks totally crowded. The benefits are mainly for smaller screens, maybe we could limit it to those? @tbsbdr 

image

tbsbdr:
> especially on bigger screens.

saw that as well and thought its mainly my viewing habit + my guess is, that most desktop users got =< 1920 screenwidth (ex https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide )

> The benefits are mainly for smaller screens, maybe we could limit it to those?

I'm not against improving if effort is reasonable - just a remark that we might move into the 20% zone of pareto, but ... 🤠

JammingBen (approved):
Fair enough. It would probably need quite some work and would make things overall more complicated :)

Pipeline stills fails though, not sure if related 🤔

kulmann:
> Fair enough. It would probably need quite some work and would make things overall more complicated :)
> 
> Pipeline stills fails though, not sure if related 🤔

The quick actions caused line wraps before on small resolutions and now don't because I "fixed" that (it was looking ugly). Now the three dots button scrolls out of the window horizontally on the XGA resolution. Will need to adjust how columns get hidden on smaller resolutions to fix this issue and resolve CI pipelines.

gennaios:
Nice. Thanks. I'm the filer of the issue of hard to see file names. I'll test this out sometime.

Thinking maybe in the future there could be more options or refinement. Some of the buttons like share or link, perhaps some other popular cloud services have made different decisions such as hiding them under the more menu, or presenting some different set of options? Overall my impression is that the UI could be cleaner and some of the options might necessarily be used all the time, e.g. create link, add share, etc. and we might not need to see them all the time in file lists..

tbsbdr:
@gennaios you can test it on:

- https://ocis.ocis-web.latest.owncloud.works/
- user: einstein
- password: relativity

gennaios:
@tbsbdr looks good. I think add people/shareand copy link icons could be removed? I think one isn't going to use it that often to need the icon on every row. It's not like one is going to use it all the time I think to warrant a separate button. As they're both in the context menu along with other options such as rename, add tags, etc., I think such is good enough as one isn't going to use them as often as browsing, searching, etc. Same as with a local desktop file manager, some options are not always used and like a file manager, I prefer a clean look.

PR #8984, opened 2023-05-04
[full-ci] Implement long breadcrumb strategy

lookacat:
## Description
Breadcrumb should not line-break. This is a new strategy to deal with long breadcrumbs and long folder names.
See https://github.com/owncloud/web/issues/6731

Maybe this helps
breadcrumb drawio

## Screenshots
Before (old):
old

After (new):
after

## Related Issue

- Fixes https://github.com/owncloud/web/issues/6731

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] Fix `...` button
- [x] vertical alignment of breadcrumb items is off. The last breadcrumb item has a lower baseline as all the other elements. More precisely, the breadcrumb items (except the last one) have slightly moved up. You can see that when you compare to the old state and watch the distance between text baseline and the chevron right icon.
- [x] The breadcrumbs are truncated far too early. See video. There is a lot of space left but the breadcrumbs already get shrinked down to a minimum. That feels wrong.

### DIFF
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -11,14 +11,15 @@
           'oc-flex-middle',
           {
             hide:
-              hiddenItems.indexOf(item) !== -1 || (item.text === '...' && hiddenItems.length === 0)
+              hiddenItems.indexOf(item) !== -1 ||
+              (item.isPreviousHiddenFolder && hiddenItems.length === 0)
           }
         ]"
       >
         <router-link
           v-if="item.to"
           :aria-current="getAriaCurrent(index)"
-          :to="item.text === '...' ? lastHiddenItem.to : item.to"
+          :to="item.isPreviousHiddenFolder ? lastHiddenItem.to : item.to"
         >
           <span class="oc-breadcrumb-item-text">{{ item.text }}</span>
         </router-link>
@@ -240,7 +241,8 @@ export default defineComponent({
         allItemsIncludingThreeDots.value.splice(1, 0, {
           text: '...',
           allowContextActions: false,
-          to: {}
+          to: {},
+          isPreviousHiddenFolder: true
         })
       }
       console.log(allItemsIncludingThreeDots.value)
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -208,8 +208,7 @@ export default defineComponent({
         return
       }
       // Remove from the left side
-      let removed
-      removed = visibleItems.value.splice(offsetIndex, 1)
+      const removed = visibleItems.value.splice(offsetIndex, 1)
 
       hiddenItems.value.push(removed[0])
       reduceBreadcrumb(offsetIndex)
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -276,6 +276,10 @@ export default defineComponent({
     align-items: baseline;
     flex-wrap: nowrap;
 
+    span {
+      white-space: nowrap;
+    }
+
     #oc-breadcrumb-contextmenu-trigger > span {
       vertical-align: middle;
       border: 3px solid transparent;
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -203,11 +203,7 @@ export default defineComponent({
       const breadcrumbMaxWidth = props.calculateBreadcrumbMaxWidth()
       const totalBreadcrumbWidth = calculateTotalBreadcrumbWidth()
 
-      console.log('breadcrumbMaxWidth', breadcrumbMaxWidth)
-      console.log('totalBreadcrumbWidth', totalBreadcrumbWidth)
-
       const isOverflowing = breadcrumbMaxWidth < totalBreadcrumbWidth
-      console.log(totalBreadcrumbWidth)
       if (!isOverflowing) {
         return
       }
@@ -245,7 +241,6 @@ export default defineComponent({
           isPreviousHiddenFolder: true
         })
       }
-      console.log(allItemsIncludingThreeDots.value)
       visibleItems.value = allItemsIncludingThreeDots.value.slice()
       hiddenItems.value = []
 
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -2,7 +2,7 @@
   <nav :id="id" :class="`oc-breadcrumb oc-breadcrumb-${variation}`">
     <ol class="oc-breadcrumb-list oc-flex oc-m-rm oc-p-rm">
       <li
-        v-for="(item, index) in items"
+        v-for="(item, index) in itemsWithDropdown"
         :key="index"
         :data-key="index"
         :class="{
@@ -10,7 +10,7 @@
           'oc-flex': true,
           'oc-flex-middle': true,
           hide:
-            hiddenItems.indexOf(item) !== -1 || (item.type === 'hidden' && hiddenItems.length === 0)
+            hiddenItems.indexOf(item) !== -1 || (item.text === '...' && hiddenItems.length === 0)
         }"
       >
         <router-link v-if="item.to" :aria-current="getAriaCurrent(index)" :to="item.to">
@@ -32,7 +32,7 @@
           <span>{{ item.text }}</span>
         </oc-button>
         <span v-else :aria-current="getAriaCurrent(index)" tabindex="-1" v-text="item.text" />
-        <template v-if="showContextActions && index === items.length - 1">
+        <template v-if="showContextActions && index === itemsWithDropdown.length - 1">
           <oc-button
             id="oc-breadcrumb-contextmenu-trigger"
             v-oc-tooltip="contextMenuLabel"
@@ -55,7 +55,7 @@
       </li>
     </ol>
     <oc-button
-      v-if="items.length > 1"
+      v-if="itemsWithDropdown.length > 1"
       appearance="raw"
       type="router-link"
       :aria-label="$gettext('Navigate one level up')"
@@ -65,7 +65,7 @@
       <oc-icon name="arrow-left-s" fill-type="line" size="large" class="oc-mr-m" />
     </oc-button>
   </nav>
-  <div v-if="items.length > 1" class="oc-breadcrumb-mobile-current">
+  <div v-if="itemsWithDropdown.length > 1" class="oc-breadcrumb-mobile-current">
     <span class="oc-text-truncate" aria-current="page" v-text="currentFolder.text" />
   </div>
 </template>
@@ -160,6 +160,7 @@ export default defineComponent({
     const { $gettext } = useGettext()
     const visibleItems = ref([])
     const hiddenItems = ref([])
+    const itemsWithDropdown = ref(props.items.slice())
 
     const getBreadcrumbElement = (key): HTMLElement => {
       return document.querySelector(`[data-key="${key}"]`)
@@ -187,13 +188,23 @@ export default defineComponent({
         return
       }
       // Remove from the left side
-      const removed = visibleItems.value.splice(offsetIndex, 1)
+      let removed
+      if (visibleItems.value[offsetIndex].text === '...') {
+        removed = visibleItems.value.splice(offsetIndex + 1, 1)
+      } else {
+        removed = visibleItems.value.splice(offsetIndex, 1)
+      }
       hiddenItems.value.push(removed[0])
+      console.log('hidden', hiddenItems.value)
       reduceBreadcrumb(offsetIndex)
     }
 
     const renderBreadcrumb = () => {
-      visibleItems.value = props.items.slice()
+      itemsWithDropdown.value = props.items.slice()
+      if (itemsWithDropdown.value.length > 1) {
+        itemsWithDropdown.value.splice(1, 0, { text: '...', allowContextActions: false, to: {} })
+      }
+      visibleItems.value = itemsWithDropdown.value.slice()
       hiddenItems.value = []
 
       nextTick(() => {
@@ -247,7 +258,8 @@ export default defineComponent({
       getAriaCurrent,
       visibleItems,
       hiddenItems,
-      renderBreadcrumb
+      renderBreadcrumb,
+      itemsWithDropdown
     }
   }
 })
@@ -261,7 +273,6 @@ export default defineComponent({
   overflow: hidden;
   //width: 50vw;
   //width: 70%;
-  background-color: green;
 
   &-mobile-current,
   &-mobile-navigation {
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -163,6 +163,10 @@ export default defineComponent({
       }
     },
 
+    /**
+     * Function to calculate the max width of the breadcrumb.
+     * Defaults to 500px.
+     */
     calculateBreadcrumbMaxWidth: {
       type: Function,
       required: false,
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -2,7 +2,7 @@
   <nav :id="id" :class="`oc-breadcrumb oc-breadcrumb-${variation}`">
     <ol class="oc-breadcrumb-list oc-flex oc-m-rm oc-p-rm">
       <li
-        v-for="(item, index) in itemsWithDropdown"
+        v-for="(item, index) in itemsWithThreeDots"
         :key="index"
         :data-key="index"
         :class="{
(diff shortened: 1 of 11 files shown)


ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/35559/50/1

<p>


### webUIPreview-mediaPreview_feature-L143.png
  


</p></details>
:boom: The oCISBasic tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oC10</strong> https://drone.owncloud.com/owncloud/web/35580/20/1

<p>


### webUIPreview-mediaPreview_feature-L143.png
  


</p></details>

JammingBen:
Haven't looked into the code yet, but what I've noticed testing this:

* Breadcrumb items get "cut" when exceeding a specific length. See for example "a folder with a very long name" -> "a folder with a":
image

* Clicking on the `...` doesn't render a drop menu with the omitted items, but just goes one level up. Is that intentional?
* The mobile navigation looks broken:
image

lookacat:
> Clicking on the ... doesn't render a drop menu with the omitted items, but just goes one level up. Is that intentional?

yes, it should navigate to the previous hidden folder, so when you e.g. only have one folder in the breadcrumb you can navigate one up.

Will look into mobile

lookacat:
will replace the ids with refs as far as possible ^^

JammingBen (commented):
You are obviously much deeper into the topic, but one thing got me wondering: Do you really need to pass `calculateBreadcrumbMaxWidth` as a method to `OcBreadcrumb`? Couldn't you just pass a number property (e.g. `availableWidth`) to the component? You'd simply need to watch this value and re-evaluate the breadcrumb on change. Also, it would make the resize observer obsolete (we already have one in `Appbar.vue`).

kulmann (changes requested):
Didn't look into the code, yet. Just tried it out. Two things that caught my eye:

1. vertical alignment of breadcrumb items is off. The last breadcrumb item has a lower baseline as all the other elements. More precisely, the breadcrumb items (except the last one) have slightly moved up. You can see that when you compare to the old state and watch the distance between text baseline and the chevron right icon. See two screenshots:

### before
Screenshot 2023-05-17 at 14 54 48


### after
Screenshot 2023-05-17 at 14 54 54


2. The breadcrumbs are truncated far too early. See video. There is a lot of space left but the breadcrumbs already get shrinked down to a minimum. That feels wrong.


https://github.com/owncloud/web/assets/3532843/ca12b981-6779-46f1-a0be-22fe9cfd60a0

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +186,82 @@ export default defineComponent({
...
+      () => props.maxWidth,
+      () => renderBreadcrumb()
+    )
+    watch(
+      () => props.items,
+      () => renderBreadcrumb()
+    )
+    onMounted(() => {
+      renderBreadcrumb()
+      window.addEventListener('resize', renderBreadcrumb)
```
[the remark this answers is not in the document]
lookacat: nope, if you change route props.items changes and we need to update the breadcrumb (e.g. displayItems)

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +186,82 @@ export default defineComponent({
   },
   setup(props) {
     const { $gettext } = useGettext()
+    const visibleItems = ref([])
+    const hiddenItems = ref([])
+    const displayItems = ref(props.items.slice())
```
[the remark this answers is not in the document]
lookacat: yes

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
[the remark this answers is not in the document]
JammingBen: You already cover that with the `watch` - I was talking about the resize event listener on `window`. But I see you already removed that :)

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -20,12 +35,27 @@
...
         >
-          <span>{{ item.text }}</span>
+          <span
+            :class="[
+              'oc-breadcrumb-item-text',
+              {
+                'oc-breadcrumb-item-text-last': index === displayItems.length - 1
+              }
+            ]"
+            >{{ item.text }}</span
```
JanAckermann, line 48: I like v-text a little more

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -132,6 +161,12 @@ export default defineComponent({
         return [...AVAILABLE_SIZES, 'remove'].some((e) => e === value)
       }
     },
+
+    maxWidth: {
+      type: Number,
+      required: false,
+      default: 99999
```
JanAckermann, line 168: Default looks a little off, can't we change that too 100 or if it needs to be a magic number -1 + add a comment ?

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +177,78 @@ export default defineComponent({
...
+        return
+      }
+      // Remove from the left side
+      const removed = visibleItems.value.splice(offsetIndex, 1)
+
+      hiddenItems.value.push(removed[0])
+      reduceBreadcrumb(offsetIndex)
+    }
+
+    const lastHiddenItem = computed(() => {
```
JanAckermann, line 214: Can be optimised to one liner

JanAckermann (changes requested):

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +177,78 @@ export default defineComponent({
...
+
+      nextTick(() => {
+        nextTick(() => {
+          reduceBreadcrumb(2)
+        })
+      })
+    }
+
+    watch(
+      () => props.maxWidth,
```
JanAckermann, line 242: In vue3 watcher can watch multiple props, if in composition api, if this is not possible you can add a computed and watch this

packages/design-system/src/components/OcBreadcrumb/types.ts
```diff
@@ -1,8 +1,10 @@
 import { RouteLocationRaw } from 'vue-router'
 
 export interface BreadcrumbItem {
+  id?: string
   text: string
   to?: RouteLocationRaw
   allowContextActions?: boolean
   onClick?: () => void
+  isPreviousHiddenFolder?: boolean
```
JanAckermann, line 9: IsPrevioudFolderHidden ?

JanAckermann (changes requested):

packages/design-system/src/components/OcBreadcrumb/types.ts
lookacat: @JanAckermann its the name for the "...", don't know a better name

JanAckermann: isPreviousHiddenFolder -> IsPrevioudFolderHidden i tought

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
lookacat: ah ok now i know

JanAckermann (changes requested):

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +182,71 @@ export default defineComponent({
   },
   setup(props) {
     const { $gettext } = useGettext()
+    const visibleItems = ref<BreadcrumbItem[]>([])
+    const hiddenItems = ref<BreadcrumbItem[]>([])
+    const displayItems = ref<BreadcrumbItem[]>(props.items.slice())
+
+    const getBreadcrumbElement = (id): HTMLElement => {
+      return document.querySelector(`[data-item-id="${id}"]`)
```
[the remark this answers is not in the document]
lookacat: nope as stated earlier you would have to add and remove dynamic refs each time the list of item changes e.g. navigating in and out of a folder

JanAckermann (approved):

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+    }
+
+    const calculateTotalBreadcrumbWidth = () => {
+      let totalBreadcrumbWidth = 100 // 100px margin to the right to avoid breadcrumb from getting too close to the controls
+      visibleItems.value.forEach((item, index) => {
```
kulmann, line 194: `index` is unused, please remove it

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
   },
   setup(props) {
     const { $gettext } = useGettext()
+    const visibleItems = ref<BreadcrumbItem[]>([])
+    const hiddenItems = ref<BreadcrumbItem[]>([])
+    const displayItems = ref<BreadcrumbItem[]>(props.items.slice())
+
+    const getBreadcrumbElement = (id): HTMLElement => {
+      return document.querySelector(`[data-item-id="${id}"]`)
```
kulmann, line 189: Changing the key from `data-id` to `data-item-id` was not what I meant... or at least that's not more specific than before. What I had in mind was that you make the document query more specific to the breadcrumbs, e.g. by adding the `oc-breadcrumb-list` class to the selector. That should be `.oc-breadcrumb-list [data-id="${id}"]`. That way the document query only checks for elements inside the breadcrumbs. If another part of the document also has a `data-id` attribute in the future (for different purposes) it will not cause conflicts for your breadcrumb item document query.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+      })
+      return totalBreadcrumbWidth
+    }
+
+    const reduceBreadcrumb = (offsetIndex) => {
+      const breadcrumbMaxWidth = props.maxWidth
+      if (!breadcrumbMaxWidth) {
+        return
+      }
+      document.getElementById(props.id)?.style.setProperty('--max-width', `${breadcrumbMaxWidth}px`)
```
kulmann, line 207: What is this property needed for? Can't find it being used anywhere in your code.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+      }
+      visibleItems.value = displayItems.value.slice()
+      hiddenItems.value = []
+
+      nextTick(() => {
+        reduceBreadcrumb(2)
+      })
+    }
+
+    watch([() => props.maxWidth, () => props.items], renderBreadcrumb)
```
kulmann, line 243:
```suggestion
    watch([() => props.maxWidth, () => props.items], renderBreadcrumb, { immediate: true })
```

kulmann: Using `immediate: true` you can remove the `onMounted` hook (the three lines below) again. Same effect, less code.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+      hiddenItems.value.push(removed[0])
+      reduceBreadcrumb(offsetIndex)
+    }
+
+    const lastHiddenItem = computed(() =>
+      hiddenItems.value.length >= 1 ? unref(hiddenItems)[unref(hiddenItems).length - 1] : { to: {} }
+    )
+
+    const renderBreadcrumb = () => {
+      displayItems.value = props.items.slice()
```
kulmann, line 226:
```suggestion
      displayItems.value = [...props.items]
```

If I don't miss anything both create a shallow copy of `props.items`. I find `[...props.items]` more intuitive.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+      if (displayItems.value.length > 1) {
+        displayItems.value.splice(1, 0, {
+          text: '...',
+          allowContextActions: false,
+          to: {},
+          isTruncated: true
+        })
+      }
+      visibleItems.value = displayItems.value.slice()
```
kulmann, line 235:
```suggestion
      visibleItems.value = [...displayItems.value]
```
Same here as a few lines above for the IMO easier readability of making a shallow copy.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+    )
+
+    const renderBreadcrumb = () => {
+      displayItems.value = props.items.slice()
+      if (displayItems.value.length > 1) {
+        displayItems.value.splice(1, 0, {
+          text: '...',
+          allowContextActions: false,
+          to: {},
+          isTruncated: true
```
kulmann, line 232: Could you please add `isAriaHidden: true` to this element and then use that to make the whole `<li>` tag `:aria-hidden="true"` in the template? This is important so that screen reader users don't get confused by the `...` list item. Their usage of the breadcrumbs is kind of "superior" to the seeing eye, because they can always read the full list of breadcrumb items (see comment about `oc-invisible-sr` css class above).

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -142,6 +181,69 @@ export default defineComponent({
...
+        })
+      }
+      visibleItems.value = displayItems.value.slice()
+      hiddenItems.value = []
+
+      nextTick(() => {
+        reduceBreadcrumb(2)
```
kulmann, line 239: Please introduce a param for the offset. Don't forget to use it when inserting the `...` element to the correct position. Use `2` as default for the offset param. But then set `3` for breadcrumbs in project spaces. That way you get `Spaces > SpaceName > ... > trölölö` for project spaces. Otherwise the space name will always be cut off.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
kulmann: And another one: Did you plan to use `isTruncated` in a different way? Because I only ever see it set and used for the `...` breadcrumb item. What I read from your discussion with @JanAckermann was that it was renamed to this because it was supposed to represent any of the truncated items. So `isAriaHidden: true` from my previous comment is probably not needed, you can instead rename `isTruncated` again to `isTruncationPlaceholder` or something like that. You can then use it for both checking its visibility and for setting it to `aria-hidden="true"`.

kulmann (changes requested):
Probably looks like a wall of text (sorry for that) but all of these comments are small stuff and should be easy to fix... Great work, love it! UX feels really good.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
JanAckermann: Smart 🤓

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
JanAckermann: Agree, slice seems to minimal
faster but we usually use the spread operator for copys

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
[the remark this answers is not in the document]
JammingBen: Refs can actually be defined dynamically on your elements via `:ref="(el) => (yourRefs[index] = el)"`. But IMO it's fine to work with vanilla JS in this case as you don't really profit from the Vue API here.

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
lookacat: yes right didn't see it somehow

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
lookacat: oh its a dev leftover, doesn't make any sense anymore

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
lookacat: yeah i implemented it BUT, as soon as i navigate into a space im no longer in a project space but in a generic space, the same as personal

kulmann: But you do have a `space` object via prop in the generic space view. And on that `space` object you can use the type guard `isProjectSpaceResource(props.space)` to identify if it's a project space or not. I think you also need the same with `isShareSpaceResource(props.space)` because a share breadcrumb also has an overview item and a share root item in the breadcrumbs (`Shared with me > some share name > ... > subfolder`).

kulmann:
Screenshot 2023-05-24 at 16 01 08

kulmann (approved):
🥳

Issue #8968, opened 2023-05-02
Breadcrumb root item does not update after language change


JammingBen:
### Steps to reproduce
1. Go to your account page and change the current language
2. Navigate to your personal space
3. Check the root breadcrumb item -> it still has the language you had selected before. After a page reload, it shows the correct language.

Edit: The same goes for the page title.

JammingBen:
Fine by me!

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/8992.

PR #8992, opened 2023-05-08
Fix personal space name not updating after language change

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/8968

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/pages/account.vue
+++ b/packages/web-runtime/src/pages/account.vue
@@ -158,9 +158,11 @@ export default defineComponent({
     const selectedLanguageValue = ref<LanguageOption>()
     const disableEmailNotificationsValue = ref<boolean>()
 
-    // FIXME: Use graph capability when we have it
+    // FIXME: Use settings service capability when we have it
     const isSettingsServiceSupported = useCapabilitySpacesEnabled()
+    // FIXME: Use change password capability when we have it
     const isChangePasswordEnabled = useCapabilitySpacesEnabled()
+    const spacesEnabled = useCapabilitySpacesEnabled()
     const isPersonalDataExportEnabled = useCapabilityGraphPersonalDataExport()
     const user = computed(() => {
       return store.getters.user
@@ -230,7 +232,7 @@ export default defineComponent({
     })
 
     const groupNames = computed(() => {
-      if (unref(useCapabilitySpacesEnabled())) {
+      if (unref(spacesEnabled)) {
         return unref(user)
           .groups.map((group) => group.displayName)
           .join(', ')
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/pages/account.vue
```diff
@@ -297,6 +298,14 @@ export default defineComponent({
             value
           }
         })
+        if (unref(personalSpace)) {
+          // update personal space name with new translation
+          store.commit('runtime/spaces/UPDATE_SPACE_FIELD', {
+            id: unref(personalSpace).id,
+            field: 'name',
+            value: $gettext('Personal')
```
kulmann, line 306: For backends without spaces we don't use `Personal` but `All files`.

kulmann (changes requested):

packages/web-runtime/tests/unit/pages/account.spec.ts
```diff
@@ -142,17 +142,6 @@ describe('account page', () => {
...
-  describe('computed method "isChangePasswordEnabled"', () => {
-    it('should be true if capability is enabled', () => {
-      const { wrapper } = getWrapper({ capabilities: { spaces: { enabled: true } } })
-      expect(wrapper.vm.isChangePasswordEnabled).toBeTruthy()
-    })
-    it('should be false if capability is not enabled', () => {
-      const { wrapper } = getWrapper()
-      expect(wrapper.vm.isChangePasswordEnabled).toBeFalsy()
-    })
-  })
```
JammingBen, line 155: This was just testing the result of the `useCapabilitySpacesEnabled` composable, no need to test that here.

packages/web-runtime/src/pages/account.vue
```diff
@@ -159,18 +156,18 @@ export default defineComponent({
     const disableEmailNotificationsValue = ref<boolean>()
 
     // FIXME: Use graph capability when we have it
-    const isSettingsServiceSupported = useCapabilitySpacesEnabled()
-    const isChangePasswordEnabled = useCapabilitySpacesEnabled()
+    const spacesEnabled = useCapabilitySpacesEnabled()
```
kulmann, line 159: I liked the aliases of the spaces enabled capability. Reason: It's bad that we use spaces capability for the respective checks. But the aliases limited the defect to a scope where it's easy to understand with the comment above. Using the spaces capability in the template makes it harder to understand that that's not desired...

JammingBen: Oh well, okay 😅  Then I'll add it a third time.

kulmann (approved):
👍

PR #9009, opened 2023-05-10
Fix public link preview

kulmann:
## Description
Right sidebar / single file resource details file preview was not working at all in public links.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Actions/FileActions.vue
+++ b/packages/web-app-files/src/components/SideBar/Actions/FileActions.vue
@@ -13,7 +13,7 @@
 <script lang="ts">
 import ActionMenuItem from 'web-pkg/src/components/ContextActions/ActionMenuItem.vue'
 import { useFileActions } from '../../../composables/actions/files/useFileActions'
-import { computed, defineComponent, inject, unref } from 'vue'
+import { computed, defineComponent, inject, Ref, unref } from 'vue'
 import { Resource, SpaceResource } from 'web-client'
 
 export default defineComponent({
@@ -23,7 +23,7 @@ export default defineComponent({
   },
   setup() {
     const resource = inject<Resource>('resource')
-    const space = inject<SpaceResource>('space')
+    const space = inject<Ref<SpaceResource>>('space')
     const resources = computed(() => {
       return [unref(resource)]
     })
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -2,15 +2,15 @@
   <div id="oc-file-details-sidebar">
     <div v-if="hasContent">
       <div
-        v-if="resource.thumbnail"
+        v-if="isPreviewLoading || preview"
         key="file-thumbnail"
         :style="{
-          'background-image': loadPreviewTask.isRunning ? 'none' : `url(${preview})`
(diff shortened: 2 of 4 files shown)


packages/web-app-files/src/components/SideBar/Actions/FileActions.vue
```diff
@@ -23,7 +23,7 @@ export default defineComponent({
   },
   setup() {
     const resource = inject<Resource>('resource')
-    const space = inject<SpaceResource>('space')
+    const space = inject<Ref<SpaceResource>>('space')
```
kulmann, line 26: here the injected space was correctly used (with `unref`) but the type was wrong.

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -295,13 +295,25 @@ export default defineComponent({
...
+    const isFolder = computed(() => {
+      return unref(resource).isFolder
+    })
     const loadPreviewTask = useTask(function* (signal, resource) {
+      if (unref(isFolder)) {
+        return
+      }
       preview.value = yield previewService.loadPreview({
-        space,
+        space: unref(space),
```
kulmann, line 306: This line is the actual fix in this PR. The `space` is provided by the sidebar as computed, but we were not accessing the value but the ref itself.

packages/web-app-files/src/components/SideBar/Actions/FileActions.vue
JammingBen: Could you fix the type for `resource` one line above as well while you're at it (same in `FileDetails.vue`)? 🙂

JammingBen (approved):

packages/web-app-files/src/components/SideBar/Actions/FileActions.vue
kulmann: Funny enough, the `resource` is provided as `Resource`, not as `Ref`. I don't know why we do that... but at least the type here is correct.

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
JammingBen: Oh boy, you're right, see https://github.com/owncloud/web/blob/admin-settings-pagination/packages/web-app-files/src/components/SideBar/SideBar.vue#L296

We should align that some time in the future...

packages/web-app-files/src/components/SideBar/Actions/FileActions.vue
JammingBen: Oh boy, you're right, see https://github.com/owncloud/web/blob/admin-settings-pagination/packages/web-app-files/src/components/SideBar/SideBar.vue#L296

We should align that some time in the future...

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -295,13 +295,25 @@ export default defineComponent({
...
         resource,
         dimensions: ImageDimension.Preview
       })
     }).restartable()
+    const isPreviewLoading = computed(() => {
+      if (unref(isFolder)) {
```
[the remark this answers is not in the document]
kulmann: noo I like the easier readability this way

PR #9010, opened 2023-05-10
e2e test. check deleted link

ScharfViktor:
- added step for checking deleted link

related https://github.com/owncloud/web/issues/9004

### DIFF
--- a/tests/e2e/support/objects/app-files/link/index.ts
+++ b/tests/e2e/support/objects/app-files/link/index.ts
@@ -84,9 +84,6 @@ export class Link {
   async delete(args: Omit<deleteLinkArgs, 'page'>): Promise<void> {
     const startUrl = this.#page.url()
     await deleteLink({ ...args, page: this.#page })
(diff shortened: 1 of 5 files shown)


tests/e2e/support/objects/app-files/link/index.ts
```diff
@@ -84,9 +84,6 @@ export class Link {
   async delete(args: Omit<deleteLinkArgs, 'page'>): Promise<void> {
     const startUrl = this.#page.url()
     await deleteLink({ ...args, page: this.#page })
-    this.#linksEnvironment.deleteLink({
-      key: args.name
-    })
```
SwikritiT, line 89: We still need to delete it from the internal store, don't we?

ScharfViktor: no, not nessasary. after test we have step `createdLinkStore.clear()`

SwikritiT (approved):

Issue #9014, opened 2023-05-11
[OCIS]Personal space page doesn't reflect the resources properly after batch deleting all the resources from second page


SwikritiT:
The personal space page doesn't reflect the resources properly after batch deleting the resources from the second page. Only after navigating to another page and coming back will result into the display of remaining folders on the personal page.

### Steps to reproduce
1. Create about 110 folders in personal space
```bash
#!/usr/bin/env bash
for (( c=1; c<=110; c++ ))
do 
    folderName="test$c"
    echo ${folderName}
    curl -k -X MKCOL -u admin:admin https://host.docker.internal:9200/remote.php/dav/files/admin/${folderName}
done
```
2. Go to UI, and make sure that the item per page is set to 100
3. Navigate to page two
4. Batch delete all the folders in page two

### Expected behavior
`10` folders should be deleted and the page should fallback to first page with `100` folders

### Actual behavior
The UI shows that there's no resource.  Reloading the same page won't reflect any changes. Only after navigating to another page and coming back will show the remaining `100` folders.

### Environment general
**Operating system**: Ubuntu 22.04.2 LTS

**Backend (ownCloud Core or Infinite Scale)**: ocis

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

### Environment ownCloud Infinite Scale
**Version:**  3.0.0-rc.3+1f8164c52
**Web client version :** 7.0.0-rc.36

Ocis started with https://github.com/owncloud/web/blob/master/docker-compose.yml config

## Video

https://github.com/owncloud/web/assets/41103328/fc3d5c28-d314-44f0-87c2-bfbd47a49305

JammingBen:
Fixed via https://github.com/owncloud/web/pull/9017.

PR #9017, opened 2023-05-11
Fix empty file list after deleting resources

JammingBen:
## Description
Fixes an issue where the file list would be empty after deleting all resources on the last page. The file list now resets the pagination in such scenario.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9014

## Types of changes

- [ ] Technical debt
- [ ] Tests


JammingBen:
Do you agree with resetting the pagination when deleting resources and the pagination is > 1? It would be nicer to do this only in case all resources on the current page have been removed, but that involves a bit more logic.

JanAckermann (approved):

lookacat (approved):
LGTM 👍🏼

kulmann (approved):

PR #9044, opened 2023-05-15
[full-ci] Implement tag filter on search result page

JammingBen:
## Description
Adds a tag filter to the search result page which can be used to filter the displayed search result by tags.

This was a bit more complicated than I anticipated, mainly because of 2 issues I encountered:

* The file list component's purpose was to simply display a search result, meaning the component could not trigger a search. Triggering a search was done by the parent component in the search app instead. With the filter now living inside the files component, I decided to changed this. The component in the files app now emits a `search` event to the parent component which eventually handles the search. This way, all the filter logic stays decoupled in the files list component.
* The filter chip's visual state can be changed by either interacting with the filter directly or by loading the page initially with given filter query params. On the search result page however, there is a third way to filter resources: by clicking a tag in the files list. The filter chip won't notice this change unless you tell the component to do so - which is why I exposed the method `setSelectedItemsBasedOnQuery` to its parent.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9054

## Screenshot
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -110,6 +110,7 @@ import SideBar from '../../components/SideBar/SideBar.vue'
 import { buildShareSpaceResource, SpaceResource } from 'web-client/src/helpers'
 import {
   queryItemAsString,
+  useCapabilityFilesTags,
   useClientService,
   useFileListHeaderPosition,
   useRoute,
@@ -164,6 +165,7 @@ export default defineComponent({
     const route = useRoute()
     const { y: fileListHeaderY } = useFileListHeaderPosition()
     const clientService = useClientService()
+    const hasTags = useCapabilityFilesTags()
 
     const searchTermQuery = useRouteQuery('term')
     const searchTerm = computed(() => {
@@ -216,7 +218,9 @@ export default defineComponent({
     }
 
     onMounted(async () => {
-      await loadAvailableTagsTask.perform()
+      if (unref(hasTags)) {
+        await loadAvailableTagsTask.perform()
+      }
       emit('search', buildSearchTerm())
     })
 
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -2,7 +2,11 @@
   <div class="files-search-result oc-flex">
     <files-view-wrapper>
       <app-bar :has-bulk-actions="false" :side-bar-open="sideBarOpen" />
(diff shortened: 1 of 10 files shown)


ownclouders:
Results for <strong>acceptance oCIS</strong> https://drone.owncloud.com/owncloud/web/35716/52/1

<p>


### webUISharingAcceptShares-acceptShares_feature-L174.png
  

### webUISharingAcceptShares-acceptShares_feature-L223.png
  


</p></details>
:boom: The oCISSharingBasic tests pipeline failed. The build has been cancelled.

ownclouders:
Results for <strong>acceptance oC10</strong> https://drone.owncloud.com/owncloud/web/35719/27/1
:boom: The oC10SharingIntGroupsToRoot tests pipeline failed. The build has been cancelled.

packages/web-app-files/src/components/Search/List.vue
```diff
@@ -2,6 +2,28 @@
...
+        <item-filter
+          v-if="availableTags.length"
+          ref="tagFilter"
+          :allow-multiple="true"
+          :filter-label="$gettext('Tags')"
+          :filterable-attributes="['label']"
+          :items="availableTags"
+          :option-filter-label="$gettext('Filter tags')"
+          :show-option-filter="true"
+          class="files-search-filter-tags oc-mr-s"
```
lookacat, line 15: would prefer a data-testid over class since it seems its only used as a test selector

packages/web-app-files/src/components/Search/List.vue
```diff
@@ -224,3 +326,8 @@ export default defineComponent({
   }
 })
 </script>
+<style lang="scss">
+.files-search-result .files-app-bar-actions {
+  display: none !important;
```
lookacat, line 331: you can just give the app-bar an `oc-hidden` class if I'm not wrong

JammingBen: I only want to hide the actions, not the whole `app-bar` component.

packages/web-app-files/src/components/Search/List.vue
JammingBen: Alright, as discussed in chat: Let's keep the class as we might want to style it later and there is no real need to handle this via `testid` because it doesn't have an id.

packages/web-app-files/src/components/Search/List.vue
lookacat: oh i see

JanAckermann (approved):

PR #9077, opened 2023-05-22
Add search page breadcrumbs and advanced search icon

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/9072

## Screenshots:
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -240,9 +240,9 @@ export default defineComponent({
     const breadcrumbs = computed(() => {
       return [
         {
-          text: !unref(searchTerm)
-            ? $gettext('Search')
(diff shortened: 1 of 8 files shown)


packages/web-app-files/src/components/Search/List.vue
```diff
@@ -235,6 +237,16 @@ export default defineComponent({
       return term
     }
 
+    const breadcrumbs = computed(() => {
+      return [
+        {
+          text: !unref(searchTerm)
```
kulmann, line 243: Why the `!`? 😅  You could easily omit it and flip the order of the two options... which would make it easier to read because there wouldn't be a negation.

kulmann (approved):
LGTM 👍

packages/web-app-files/src/components/Search/List.vue
JammingBen: Yep you're right 😄

Issue #9030, opened 2023-05-12
Share indicators missing after cut&paste (page reload needed)


kulmann:
With a `MOVE` request (in the ui: cut & paste) oCIS also moves the attached shares. ownCloud Web doesn't reflect that in the share indicators after successful `MOVE`.

### Steps to reproduce
1. Create a folder "test" and a folder "another test" (I did in a project space, but should be the same in a personal space)
2. Create a public link and a share on folder "test"
3. Select context action "Cut", navigate into "another test" and select "Paste"

### Expected behaviour
After successful `MOVE` the file list should reflect the share state in the share indicators because `MOVE` also moves the shares.

### Actual behaviour
You need a page reload to see the share indicators. The right sidebar sharing panel correctly shows the shares.

### Screenshots
<img width="1827" alt="Screenshot 2023-05-12 at 13 09 39" src="https://github.com/owncloud/web/assets/3532843/b396a5d5-7715-4abe-ba77-cfc6f5403aac">
<img width="1827" alt="Screenshot 2023-05-12 at 13 09 56" src="https://github.com/owncloud/web/assets/3532843/3b7d2c2e-77ac-4dd5-8949-3c4579475d62">

JammingBen:
Fixed via https://github.com/owncloud/web/pull/9110.

PR #9110, opened 2023-05-25
fix: share indicator loading after pasting resources

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/9030

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/store/actions.ts
+++ b/packages/web-app-files/src/store/actions.ts
@@ -125,7 +125,9 @@ export default {
         }
 
         return Promise.all(loadingResources).then(() => {
+          const currentFolder = context.getters.currentFolder
           context.commit('UPSERT_RESOURCES', fetchedResources)
+          context.commit('LOAD_INDICATORS', currentFolder.path)
         })
       },
       { debounceTime: 0 }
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/store/actions.ts
```diff
@@ -125,7 +125,9 @@ export default {
         }
 
         return Promise.all(loadingResources).then(() => {
+          const currentFolder = context.getters.currentFolder
           context.commit('UPSERT_RESOURCES', fetchedResources)
+          context.commit('LOAD_INDICATORS', currentFolder.path)
```
kulmann, line 130: is it possible that the `currentFolder` var can be nullish?

JammingBen: I thought about it, but I don't see a scenario where we don't have a folder to paste into. Do you?

kulmann: True. And now that I looked at it in detail - the currentFolder is already used at the top of the function. So yeah, fine like that 😅

kulmann (approved):

JanAckermann (approved):

PR #9113, opened 2023-05-25
Some a11y fixes

kulmann:
Some accessibility fixes
- `h1` tag in `GenericSpace` views was not filled
- fix `h`-tag cascading / order
- moved `div` meta element which is used for the left sidebar nav active item highlighting out of the `ul` (because `div`s  are not allowed as children of `ul`)

### DIFF
--- a/packages/web-app-files/src/components/AppBar/AppBar.vue
+++ b/packages/web-app-files/src/components/AppBar/AppBar.vue
@@ -87,9 +87,10 @@ import {
   useFileActionsMove,
   useFileActionsRestore
 } from 'web-app-files/src/composables/actions'
-import { useStore } from 'web-pkg/src'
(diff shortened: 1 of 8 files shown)


packages/web-app-files/src/components/AppBar/AppBar.vue
```diff
@@ -182,10 +184,19 @@ export default defineComponent({
       return props.breadcrumbs.length <= 1
     })
 
+    const routeMetaTitle = useRouteMeta('title')
+    const pageTitle = computed(() => {
+      if (unref(routeMetaTitle)) {
+        return $gettext(unref(routeMetaTitle))
+      }
+      return props.space.name
```
JammingBen, line 192: `space` could be `null` theoretically, it would probably make sense to catch that, what do you think?

JammingBen (approved):

PR #9145, opened 2023-06-01
Feature/7600 - Scroll to newly created folder

### LINKED ISSUE #7600: No scrolling to just created folder
### Steps to reproduce
1. Create a new folder  in the repository with many folders and name it, so that it's name is outside of the view

### Expected behaviour
The view should scroll to the created folder

### Actual behaviour
- There is no scrolling
- Users get confused if the folder was actually created (from user feedback)
- Users need to scroll manually and search for the folder

### LINKED ISSUE #7601: ScrollToResource function not working as expected
### Steps to reproduce
1a. By selected resource scroll up outside of the view with this resource and press key "down"
1b. By selected resource scroll down outside of the view with this resource and press key "up"

### Expected behaviour
The called `scrollToResource` function shows the resource in the middle of the view

### Actual behaviour
By 1a the resource is shown in the last row of the visible list. By 1b nothing happens

pascalwengerter:
## Description
Upstreams a CERN commit by @elizavetaRa  which originates from before some major refactorings. It also reduces the `scrollToResource` arguments from a full resource where only the `id` would be used to a string/number.

## Related Issue
- Fixes #7600 
- Fixes #7601

### DIFF
--- a/packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
+++ b/packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
@@ -1,6 +1,6 @@
 import { Resource, SpaceResource } from 'web-client/src/helpers'
 import { Store } from 'vuex'
-import { computed, unref } from 'vue'
+import { computed, nextTick, unref } from 'vue'
 import { useClientService, useRouter, useStore } from 'web-pkg/src/composables'
 import { FileAction } from 'web-pkg/src/composables/actions'
 import { useGettext } from 'vue3-gettext'
@@ -9,6 +9,7 @@ import { join } from 'path'
 import { WebDAV } from 'web-client/src/webdav'
 import { isLocationSpacesActive } from 'web-app-files/src/router'
 import { getIndicators } from 'web-app-files/src/helpers/statusIndicators'
+import { useScrollTo } from '../../scrollTo/useScrollTo'
 
 export const useFileActionsCreateNewFolder = ({
   store,
@@ -17,6 +18,7 @@ export const useFileActionsCreateNewFolder = ({
   store = store || useStore()
   const router = useRouter()
   const { $gettext } = useGettext()
+  const { scrollToResource } = useScrollTo()
 
   const clientService = useClientService()
   const currentFolder = computed((): Resource => store.getters['Files/currentFolder'])
@@ -72,6 +74,9 @@ export const useFileActionsCreateNewFolder = ({
       store.dispatch('showMessage', {
         title: $gettext('"%{folderName}" was created successfully', { folderName })
       })
+
+      await nextTick()
+      scrollToResource(resource.id, { forceScroll: true })
     } catch (error) {
       console.error(error)
       store.dispatch('showErrorMessage', {
--- a/packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
+++ b/packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
@@ -1,11 +1,12 @@
-import { computed, onBeforeUnmount, onMounted, ref, unref } from 'vue'
+import { Ref, computed, nextTick, onBeforeUnmount, onMounted, ref, unref } from 'vue'
 import { useStore } from 'web-pkg/src/composables'
 import { useScrollTo } from 'web-app-files/src/composables/scrollTo'
 import { Key, ModifierKey } from 'web-pkg/src/composables/keyboardActions'
 import { eventBus } from 'web-pkg'
 import { useGettext } from 'vue3-gettext'
+import { Resource } from 'web-client/src'
 
-export const useKeyboardActionsSearchTable = (keyActions, paginatedResources) => {
+export const useKeyboardActionsSearchTable = (keyActions, paginatedResources: Ref<Resource[]>) => {
   const store = useStore()
   const { scrollToResource } = useScrollTo()
   const selectionCursor = ref(0)
@@ -54,9 +55,11 @@ export const useKeyboardActionsSearchTable = (keyActions, paginatedResources) =>
       return
     }
     resetSelectionCursor()
-    await store.dispatch('Files/resetFileSelection')
+    store.dispatch('Files/resetFileSelection')
+    await nextTick()
     store.commit('Files/ADD_FILE_SELECTION', { id: nextId })
-    scrollToResource({ id: nextId } as any)
+    await nextTick()
+    scrollToResource(nextId)
   }
 
   const getNextResourceId = (previous = false) => {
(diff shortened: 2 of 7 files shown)


packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
```diff
@@ -66,12 +68,14 @@ export const useFileActionsCreateNewFolder = ({
         resource.indicators = getIndicators({ resource, ancestorMetaData: unref(ancestorMetaData) })
       }
 
-      store.commit('Files/UPSERT_RESOURCE', resource)
+      await store.commit('Files/UPSERT_RESOURCE', resource)
```
pascalwengerter, line 71: My IDE suggests there's no difference when using `await` here, however the scrolling only happens if you put it there 🤷🏽

packages/web-app-files/src/composables/scrollTo/useScrollTo.ts
```diff
@@ -21,30 +21,16 @@ export const useScrollTo = (): ScrollToResult => {
...
-    const topbarElement = document.getElementsByClassName('files-topbar')[0] as HTMLElement
-    // topbar height + th height + height of one row = offset needed when scrolling top
-    const topOffset = topbarElement.offsetHeight + resourceElement.offsetHeight * 2
-
-    // top reached
-    if (resourceElement.getBoundingClientRect().top < topOffset) {
-      const fileListWrapperElement = document.getElementsByClassName('files-view-wrapper')[0]
-      fileListWrapperElement.scrollBy(0, -resourceElement.offsetHeight)
-    }
+    resourceElement.scrollIntoView({ behavior: 'smooth', block: 'center' })
```
pascalwengerter, line 33: Not sure why all the other code was necessary tbh - scrolling to a resource and centering it on screen seems pretty straightforward, and if we need a "scroll to top" we should implement it explicitly?

kulmann: doesn't work with active pagination though. if you're on page 3 and a just created folder is sorted into page 17 you would need to navigate to that page before trying to scroll...

packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
kulmann: I don't get that either... `commit` is a synchronous call / doesn't return a promise, so `await` makes no sense at all...

packages/web-app-files/src/composables/scrollTo/useScrollTo.ts
pascalwengerter: Ah damn, I knew there was something I was missing (because otherwise that issue would've been resolved long ago already I suppose)

kulmann: fine to split it into two issues btw. for this PR: assume that using `scrollToResource` is sufficient and declare it not checking pagination as bug. Then solve using pagination in `scrollToResource` as a separate PR.

elizavetaRa:
Adding some opinion: scrolling to resource would also make sense when uploading a resource and even when creating new file that opens in new tab

packages/web-app-files/src/composables/scrollTo/useScrollTo.ts
JammingBen: > Not sure why all the other code was necessary tbh - scrolling to a resource and centering it on screen seems pretty straightforward, and if we need a "scroll to top" we should implement it explicitly?

It was there for the keyboard navigation. Now when navigating via keyboard, it always immediately scrolls to the next resource. Before, it only scrolled when the resource was outside of the viewport. Both works, but I actually like the latter because it does not feel so "floaty".

packages/web-app-files/src/composables/scrollTo/useScrollTo.ts
```diff
@@ -21,29 +21,25 @@ export const useScrollTo = (): ScrollToResult => {
     return queryItemAsString(unref(detailsQuery))
   })
 
-  const scrollToResource = (resource) => {
+  const scrollToResource = (resourceId: Resource['id'], forceScroll = false) => {
```
dschmidt, line 24: Not a fan of bool params tbh, can we make it an options obj? 

```suggestion
  const scrollToResource = (resourceId: Resource['id'], { forceScroll = false}) => {
```

packages/web-app-files/src/composables/actions/files/useFileActionsCreateNewFolder.ts
dschmidt: https://stackoverflow.com/a/55263084

What you actually want is probably `await nextTick()`

pascalwengerter: Done, thx for the hint!

packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
```diff
@@ -1,11 +1,12 @@
...
+import { Ref, computed, nextTick, onBeforeUnmount, onMounted, ref, unref } from 'vue'
 import { useStore } from 'web-pkg/src/composables'
 import { useScrollTo } from 'web-app-files/src/composables/scrollTo'
 import { Key, ModifierKey } from 'web-pkg/src/composables/keyboardActions'
 import { eventBus } from 'web-pkg'
 import { useGettext } from 'vue3-gettext'
+import { Resource } from 'web-client/src'
 
-export const useKeyboardActionsSearchTable = (keyActions, paginatedResources) => {
+export const useKeyboardActionsSearchTable = (keyActions, paginatedResources: Ref<Resource[]>) => {
```
pascalwengerter, line 9: @jacob-nv hope that little update is fine with you? :)

packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
```diff
@@ -54,9 +55,11 @@ export const useKeyboardActionsSearchTable = (keyActions, paginatedResources) =>
       return
     }
     resetSelectionCursor()
-    await store.dispatch('Files/resetFileSelection')
+    store.dispatch('Files/resetFileSelection')
+    await nextTick()
     store.commit('Files/ADD_FILE_SELECTION', { id: nextId })
-    scrollToResource({ id: nextId } as any)
+    await nextTick()
```
pascalwengerter, line 61: Not sure if we should await the nextTick within `scrollToResource` instead of spamming it outside though

dschmidt: What do you mean?

packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
jacob-nv: Completely fine, this composable is being split up in new updates coming

pascalwengerter:
Sonarcloud failure is unrelated to my changes? 🤔

elizavetaRa:
@pascalwengerter @dschmidt it doesn't scroll when new folder is on another page

JammingBen:
> @pascalwengerter @dschmidt it doesn't scroll when new folder is on another page

I've created https://github.com/owncloud/web/issues/9494 to keep track of this. It's a bit more complex to achieve that.

PR #9150, opened 2023-06-05
Cloud import (first iteration: OneDrive)

JammingBen:
## Description
Implements an action to import files from other clouds. The first iteration supports `OneDrive`. Needs https://github.com/owncloud/ocis/pull/6453 for the config part.

You need to run the `ocis` `traefik` and `companion` containers. Companion runs at `https://host.docker.internal:9200/companion` which needs to be defined as companion URL in oCIS (or hard-coded in the `ConfigurationManager` for now).

Please contact me for secret keys if you want to test it yourself. Those need to be added as env variables of the `companion` container:

```
COMPANION_ONEDRIVE_KEY: "..."
COMPANION_ONEDRIVE_SECRET: "..."
```

## Screenshots

image

image

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9151

## Types of changes

- [ ] Technical debt
- [ ] Tests

## To-Do

- [x] Basic file import
- [x] Translations
- [x] Remove "Confirm" button from modal
- [x] Config (?)
- [x] Fix unit tests
- [x] Disable on public pages

## Follow-up

- [ ] Folder upload (including hierarchy)
- [ ] Conflict handling
- [ ] Remove upload helpers and use the prepare-plugin instead
- [ ] Docs


packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -275,6 +276,14 @@ export default defineComponent({
       required: false,
       default: false
     },
+    /**
+     * Asserts whether the modal should render a confirm button
+     */
+    withoutButtonConfirm: {
```
lookacat, line 282: needed because we have no confirmation button in the import modal

packages/web-app-files/src/uppyUploadPlugin.ts
JammingBen, line 1: FYI: This plugin is going to handle all Uppy file uploads in the future, therefore it's called `HandleUpload`. We will tackle this in a follow-up PR. For now, the plugin only prepares the files for the cloud import though.

packages/web-app-files/src/uppyUploadPlugin.ts
```diff
@@ -0,0 +1,136 @@
...
+      ({ id }) => id === this.currentFolder.storageId
+    )
+  }
+
+  /**
+   * Get the relative path of the file when the file was inside a directory on the client computer.
+   * @param file
+   */
+  getRelativeFilePath = (file): string | undefined => {
+    const relativePath = file.webkitRelativePath || (file as any).relativePath
```
dschmidt, line 55: types could be better ;-)

packages/web-runtime/src/services/uppyService.ts
```diff
@@ -34,11 +34,30 @@ export class UppyService {
 
   constructor() {
     this.uppy = new Uppy({
-      autoProceed: true
+      autoProceed: true,
+      locale: {
+        strings: {
+          // for some reason this string is required and missing in uppy
+          addedNumFiles: 'Added %{numFiles} file(s)'
```
dschmidt, line 41: Don't we need to pass this through `$gettext`?

dschmidt (approved):

docker-compose.yml
```diff
@@ -164,5 +167,60 @@ services:
...
+      - "9200:443"
+    volumes:
+      - "./data_traefik/letsencrypt:/letsencrypt"
+      - "./data_traefik/configs:/configs"
+      - "/var/run/docker.sock:/var/run/docker.sock:ro"
+
 volumes:
   uploads:
+  uppy_companion_datadir:
+  traefik_letsencrypt:
```
dschmidt, line 226: Hmm  - with this vite and ocis/traefik don't share the same certs anymore, right?

packages/web-app-files/src/uppyUploadPlugin.ts
JammingBen: Totally agree! I'll tackle this in the follow up PR because it refactors the plugin by quite a lot.

packages/web-runtime/src/services/uppyService.ts
JammingBen: We don't actually use the string (which makes the error without even more confusing), but you're right, it would be correct to translate it anyways.

JammingBen:
> It would be great to have documentation on how to add secrets to docker-compose

I added a brief comment on how to do it in the PR description 👍  For the full docs I'd rather wait until we have the extension running to put it there.

PR #9180, opened 2023-06-07
Privacy statement in account menu (authenticated)

lookacat:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


JanAckermann (changes requested):

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -70,17 +71,21 @@
           </div>
         </li>
       </oc-list>
+      <div v-if="imprintUrl && privacyUrl" class="imprint-footer oc-py-s oc-mt-m oc-text-center">
```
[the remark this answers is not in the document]
lookacat: no i don't think we have an default, correct me if i am wrong @kulmann @tbsbdr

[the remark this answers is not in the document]
JanAckermann: well most of the time its self hosted, so it makes sense to have no default ^^

kulmann, line 74:
```suggestion
      <div v-if="imprintUrl || privacyUrl" class="imprint-footer oc-py-s oc-mt-m oc-text-center">
```

... and a `v-if` with the respective url on the two oc-buttons please. If any of the links exists we want to show it. If someone for whatever reason only has an imprint, but not a privacy url, we still want to show the imprint link.

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -70,17 +71,21 @@
...
+        <a v-translate :href="imprintUrl" target="_blank">Imprint</a> -
```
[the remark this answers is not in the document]
kulmann: Please don't use `v-translate` anymore, but instead `v-text="$gettext('Imprint')"`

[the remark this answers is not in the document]
kulmann: and yes, `oc-button` with href please 😍

JanAckermann, line 75: Please use interpunct (·) as a separator, like in the mockup -> because of beauty. I also highly recommend wrapping that into a span

JanAckermann (changes requested):

packages/web-runtime/src/components/Topbar/UserMenu.vue
lookacat: ah okay makes sense, will build it so only one can be shown

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -258,4 +283,15 @@ export default defineComponent({
     }
   }
 }
+
+.imprint-footer {
+  background-color: var(--oc-color-background-hover);
+  margin-left: -8px;
```
kulmann, line 289: please use the respective spacing css variable. otherwise the margin-correction breaks if someone changes their spacing css values.

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -258,4 +283,15 @@ export default defineComponent({
...
+  width: calc(100% + 16px);
```
kulmann, line 290: same here for the spacing css variable

kulmann (changes requested):
tiny nitpick (see comments). Visual result is perfect 👍

packages/web-runtime/src/components/Topbar/UserMenu.vue
lookacat: yes you are right, didn't know you could calc with vars

JanAckermann (approved):

PR #9079, opened 2023-05-22
[full-ci] Improve app loading

dschmidt:
## Description

BASED ON #9075 

Externalize web-pkg and web-client 
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #9212, opened 2023-06-13
Backport app dependency injections from #9079

dschmidt:
## Description
We don't want to backport the whole PR #9079 from master because it's too invasive, but backporting the injections gives us some level of compatibility already, which is good enough for now

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-app-loading
+++ b/changelog/unreleased/bugfix-app-loading
@@ -1,0 +1,5 @@
+Bugfix: Experimental app loading
+
+We've made a change to make our helper package "web-pkg" available in our (still experimental) extension system.
+
+https://github.com/owncloud/web/pull/9212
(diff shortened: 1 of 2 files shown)


dschmidt:
Hrm - I'm not sure about the changelog. I would rather keep it as an unofficial "happens to work"-hack. 
We have no documentation about what is supported, there is no information on the compatibility of the published web-pkg etc etc

I'm happy it fixes the richdocuments app for now, but that's all I want to guarantee and rather not mention it otherwise

kulmann:
> Hrm - I'm not sure about the changelog. I would rather keep it as an unofficial "happens to work"-hack. We have no documentation about what is supported, there is no information on the compatibility of the published web-pkg etc etc
> 
> I'm happy it fixes the richdocuments app for now, but that's all I want to guarantee and rather not mention it otherwise

I need a changelog item for making a release. ;-) only a matter of wording... I added one which doesn't sound like officially supported extension development.

lookacat (approved):

Issue #9247, opened 2023-06-15
"#" sign in directory name causes error by uploading files inside it


elizavetaRa:
### Steps to reproduce
1. Create a directory and name it "test with # folder"
2. Open directory
3. Upload a file

### Expected behaviour
File gets uploaded

### Actual behaviour
Request call POST
`https://ocis.ocis-traefik.latest.owncloud.works/remote.php/dav/spaces/0ef049e8-a2a9-4d3d-9aec-3ad647a0e1e9$1cd70506-913a-4f89-bf4c-923e354ce128/test with `:  starting with "#" url gets cut.

At CERNBox request call PUT `https://cernbox.cern.ch/remote.php/dav/files/ragozina/eos/user/r/ragozina/test with` returns 204 and creates a file "test with" in parent directory.

Possibly the same bug for some other signs?

dschmidt:
Can you "copy as cURL" the failing request from the developer tools?

elizavetaRa:
```
  curl 'https://ocis.ocis-traefik.latest.owncloud.works/remote.php/dav/spaces/0ef049e8-a2a9-4d3d-9aec-3ad647a0e1e9$1cd70506-913a-4f89-bf4c-923e354ce128/test%20with%20' \
  -H 'authority: ocis.ocis-traefik.latest.owncloud.works' \
  -H 'accept: */*' \
  -H 'accept-language: en' \
[...]

PR #9282, opened 2023-06-22
fix: cloud import into folder with special characters

JammingBen:
## Description
Fixes the cloud import when importing files into a folder that has special characters in its name.

Follow-up for https://github.com/owncloud/web/pull/9150.

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-cloud-import
+++ b/changelog/unreleased/enhancement-cloud-import
@@ -4,3 +4,5 @@ An action to import files from other external cloud providers has been added. On
 
 https://github.com/owncloud/web/issues/9151
 https://github.com/owncloud/web/pull/9150
+https://github.com/owncloud/web/pull/9282
+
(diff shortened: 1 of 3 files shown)


dschmidt (commented):
maybe encode it just once here:
https://github.com/owncloud/web/blob/445375b3e488d15788755d36d190b03b3106eccd/packages/web-app-files/src/HandleUpload.ts#LL145C54-L145C63
?

dschmidt:
Does this fix https://github.com/owncloud/web/issues/9247?

JammingBen:
> maybe encode it just once here: https://github.com/owncloud/web/blob/445375b3e488d15788755d36d190b03b3106eccd/packages/web-app-files/src/HandleUpload.ts#LL145C54-L145C63 ?

That's the directory when uploading files that are located inside a folder (= when uploading a folder). The current folder path is `trimmedUploadPath`.

> Does this fix #9247?

Unfortunately not. It won't create any files with cut-off names but it still fails with status code `500`. Might be a server issue, not sure yet though.

dschmidt (approved):

PR #9343, opened 2023-07-04
Display/hide columns for “shared with me” view on small screens

koebel:
No changes in code, just moving the change log file into the correct folder. Sorry for the mistake.


JammingBen (commented):
Merging this will still keep the existing changelog item in the `design-system/changelog` folder, so there would be duplicates.

I recommend that you hard reset this branch to master via `git reset --hard origin/master` and then simply move the existing changelog item from `design-system/changelog` to `changelog`.

JammingBen:
@koebel The pipeline has been fixed, you just need to rebase your PR(s) with current master.

changelog/unreleased/bugfix-switch-columns-shared-with-me-view
```diff
@@ -1,4 +1,4 @@
-Bugfix: switch columns displayed on small screens in "Shared with me" view
+Bugfix: switch columns displayed on small screens in "Shared with me" view
```
dschmidt, line 1: Can you remove the trailing whitespace again? ;-)

dschmidt: If you just want to re-trigger ci, you can simply run `git commit --amend --no-edit` to change the last commit hash and force push that.

koebel: thanks for the advice, I indeed did it as a work around because I wasn't aware of the --no-edit command

Issue #9224, opened 2023-06-14
Show separate tags if comma was entered


tbsbdr:
### Steps to reproduce
1. enter "hello, world" as new tag
2. (formfield shows "hello, world" as single tag)
3. press "Save"
4. filelist shows "hello, world" as single tag ❌
5. refresh
6. fielist shows "hello" and "world" as separate tags ✅

( **Step 4: filelist shows "hello, world" as single tag ❌** )
![image](https://github.com/owncloud/web/assets/26610733/38751b75-b9f8-41fc-809b-55dce54cbab9)

### Expected behaviour

immediately reflect the user, that 2 tags will be created:
![image](https://github.com/owncloud/web/assets/26610733/b8baf5c3-6aab-4c87-980f-0ffdcc039b97)
![image](https://github.com/owncloud/web/assets/26610733/83ece209-8f56-4547-8cda-480734e5a801)

-  if a comma gets entered, show separate tags
- show tag 🏷️ icon instead of circle in form field and autosuggest

### Actual behaviour
- separate tags are only shown upon refresh

superseeds https://github.com/owncloud/web/issues/9223

tbsbdr:
@lookacat 👆for you?

JammingBen:
Fixed via https://github.com/owncloud/web/pull/9348.

PR #9348, opened 2023-07-05
9224 comma separation

grimmoc:
## Description
Tags containing commas are now split up into separate tags on client side to match server side processing behaviour.

## Related Issue

- implements https://github.com/owncloud/web/issues/9224

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/SideBar/TagsPanel.vue
+++ b/packages/web-app-files/src/components/SideBar/TagsPanel.vue
@@ -11,7 +11,7 @@
         :options="availableTags"
         taggable
         push-tags
-        :select-on-key-codes="[188, 13]"
+        :select-on-key-codes="[keycode('enter'), keycode(',')]"
         :label="$gettext('Add or edit tags')"
         :create-option="createOption"
         :selectable="isOptionSelectable"
@@ -64,6 +64,7 @@ import { useClientService, useStore } from 'web-pkg/src/composables'
 import { Resource } from 'web-client'
 import diff from 'lodash-es/difference'
 import { useGettext } from 'vue3-gettext'
+import keycode from "keycode";
 
 const tagsMaxCount = 100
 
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/SideBar/TagsPanel.vue
```diff
@@ -181,7 +182,7 @@ export default defineComponent({
       createOption,
       isOptionSelectable,
       showSelectNewLabel,
-      save
+      save,
```
kulmann, line 185: Please always run `pnpm lint --fix` on save in your IDE or before you commit. ;-) This comma fails in the linting step.

packages/web-app-files/src/components/SideBar/TagsPanel.vue
```diff
@@ -11,6 +11,7 @@
         :options="availableTags"
         taggable
         push-tags
+        :select-on-key-codes="[188, 13]"
```
kulmann, line 14: No magic numbers please :-)

kulmann: We already have the dependency `keycode` (see root level package.json). You can use that to create an array with human readable keys in the setup function, return that from the setup function, and use it here in the template. (e.g. `keycode('enter')` returns `13`, you can just throw that into an array.

packages/web-app-files/src/components/SideBar/TagsPanel.vue
```diff
@@ -74,6 +76,7 @@ type TagOption = {
 
 export default defineComponent({
   name: 'TagsPanel',
+  methods: { keycode },
```
kulmann, line 79: Please don't introduce new options api code. You can return the keycode function in the setup function.

grimmoc: hmm, I should better not rely on method importing using WebStorm or at least I should double check before committing :/

JammingBen (approved):
👍

Issue #9278, opened 2023-06-20
[CERN] Buckets for Web Merge


tbsbdr:
# Core

- [x] [De-duplicate event handling to prevent errors on Draw-io](https://github.com/owncloud/web/pull/9159/commits/0a0cd9940a682d2c0c91744ea99e45e533e55f9d)  fixed via https://github.com/owncloud/web/pull/8576
- [x]  [Fix 'Show more'-action in shares panel](https://github.com/owncloud/web/pull/9159/commits/98e47e5bed3e1593000de6826240eab8e3cb61cc) fixed upstream https://github.com/owncloud/web/pull/8482
- [x]  [Fix multiline breadcrumb](https://github.com/owncloud/web/pull/9159/commits/7abee7a63662fc22b74cd12f1606f44408e35f0b) fixed via https://github.com/owncloud/web/pull/8984 https://github.com/owncloud/web/pull/8765
- [x]  [Single file share view](https://github.com/owncloud/web/pull/9159/commits/44f5d2ca0ad63e854bbb80bf9670e4d1cd765afe) fixed via https://github.com/owncloud/web/pull/8446
- [x]  [Option for max length of breadcrumb item](https://github.com/owncloud/web/pull/9159/commits/537e2327cbacb86c2ec5bde91fc139101ed7ae38) superseeded by https://github.com/owncloud/web/pull/8984 https://github.com/owncloud/web/pull/8765
- [x]  [Advanced/collapsible table](https://github.com/owncloud/web/pull/9159/commits/7dc1c55d1d0f794fddade0f649a2526b7064074c) (until oc provides a better concept via filters) in progress: https://github.com/owncloud/web/pull/8443
- [x]  [More file extensions for text-editor](https://github.com/owncloud/web/pull/9159/commits/323030a42bc6184850b8f20f3106e827b6443e63) fixed via https://github.com/owncloud/web/pull/7967
- [x] [Favorites view update on remove from favorites](https://github.com/owncloud/web/pull/9159/commits/a2dcfceeeba1fcb95ccd56ede3058240263e7b64) fixed upstream https://github.com/owncloud/web/pull/9237
- [x]  [Improvements to User search autocomplete listing look and feel](https://github.com/owncloud/web/pull/9159/commits/d04a7ae28a27a0dc5d5e5dc17aa362aeacf14694) fixed upstream https://github.com/owncloud/web/pull/9273
[...]

PR #9369, opened 2023-07-08
Limit relative quota in user menu to two decimals

### LINKED ISSUE #9259: Wrong display of Quota percentage
relates to: https://github.com/owncloud/web/pull/9159/commits/f972d1dfcb7e18737033c091bee3816045f6b4a5
```
Fix quota percentage
Should be fixed in the backend

```

### Steps to reproduce
1. (tbd)
2.
3.

### Expected behaviour
screenshot_001238

### Actual behaviour
screenshot_001236

pascalwengerter:
## Related Issue
- Fixes #9259, works towards #9278

### DIFF
--- a/packages/web-runtime/src/components/Topbar/UserMenu.vue
+++ b/packages/web-runtime/src/components/Topbar/UserMenu.vue
@@ -210,7 +210,7 @@ export default defineComponent({
     },
     quotaUsagePercent() {
       return this.useLegacyQuota
-        ? this.quota.relative
+        ? parseFloat(this.quota.relative.toFixed(2))
         : parseFloat(((this.quota.used / this.quota.total) * 100).toFixed(2))
     },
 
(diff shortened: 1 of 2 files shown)


dschmidt (approved):

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -210,7 +210,7 @@ export default defineComponent({
     },
     quotaUsagePercent() {
       return this.useLegacyQuota
-        ? this.quota.relative
+        ? parseFloat(this.quota.relative.toFixed(2))
```
kulmann, line 213: Only parse this.quota.relative to float and then to a toFixed(2) on the result.

pascalwengerter: I did that initially but then `quotaUsagePercent` is `string | number` and it gets messy further down where we're calculating how much free quota is left to change the color of the OcProgress component. 
The alternative was `parseFloat(parseFloat(this.quota.relative).toFixed(2))` but that also didn't really feel right...

pascalwengerter: We'll go ahead as-is and can adapt (= add another `parseFloat`) if EOS/Reva requires it

Issue #9361, opened 2023-07-06
Hide pause-, resume- and retry-functionality for cloud imports


JammingBen:
When importing files via the new cloud import feature, the pause-, resume- and retry-functionality are not working as expected. They might in the future, but for now we decided to hide those actions in the UI.

PR #9374, opened 2023-07-10
feat: hide upload info actions for cloud imports

JammingBen:
## Description
Actions like pausing, resuming and retrying uploads is not (yet) supported by cloud imports, hence hide these.

Fixes https://github.com/owncloud/web/issues/9361

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/components/UploadInfo.vue
+++ b/packages/web-runtime/src/components/UploadInfo.vue
@@ -43,7 +43,7 @@
           v-text="infoExpanded ? $gettext('Hide details') : $gettext('Show details')"
         ></oc-button>
         <oc-button
-          v-if="!runningUploads && Object.keys(errors).length"
+          v-if="!runningUploads && Object.keys(errors).length && !disableActions"
           v-oc-tooltip="$gettext('Retry all failed uploads')"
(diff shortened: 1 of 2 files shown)


packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -265,6 +272,10 @@ export default defineComponent({
       this.filesInProgressCount += files.filter((f) => !f.isFolder).length
 
       for (const file of files) {
+        if (!this.disableActions && file.source && file.source !== 'DropTarget') {
```
dschmidt, line 275:
```suggestion
        if (!this.disableActions && file?.source !== 'DropTarget') {
```

JammingBen: That's actually a different behaviour. `file?.source !== 'DropTarget'` would evaluate to `true` if `file.source` is not given, which I don't want.

dschmidt: Yeah, would only be equivalent for `===`

dschmidt (approved):

PR #9389, opened 2023-07-11
Only allow space managers to empty trash bin

JanAckermann:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/9385

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] ...

PR #9395, opened 2023-07-12
[test-only] Change purge trashbin test

ScharfViktor:
after https://github.com/owncloud/web/pull/9389 I changed e2e test where I check that:
- viewer and editor cannot delete resource from space and manager can
- I cleaned up some unused code


JanAckermann (approved):
🥰

PR #9396, opened 2023-07-12
Theme switcher for more themes

elizavetaRa:
## Description
The pr adds the possibility to switch between themes when there are more than 2.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Improving the accessibility by providing high contrast themes in addition to classical default and default dark 

## Screenshots (if appropriate):
Screenshot from 2023-07-12 12-17-20

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #9401, opened 2023-07-12
Theme names

elizavetaRa:
## Description
Adding the "theme name" property that should be visible for switcher of more than 2 themes 
Relates to https://github.com/owncloud/web/pull/9396

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/themes/owncloud/theme.json
+++ b/packages/web-runtime/themes/owncloud/theme.json
@@ -7,6 +7,7 @@
   "ios": {},
   "web": {
     "default": {
+      "theme-name": "Default light",
       "general": {
         "name": "ownCloud",
         "slogan": "ownCloud – A safe home for all your data",
@@ -78,6 +79,7 @@
       }
     },
     "default-dark": {
+      "theme-name": "Default dark",
       "general": {
         "name": "ownCloud",
         "slogan": "ownCloud – A safe home for all your data",


elizavetaRa:
@tbsbdr @pascalwengerter 
To discuss: in the current implementation the keys "default" and "default-dark" for themes are handled as names in `packages/web-runtime/src/composables/theme/useDefaultThemeName.ts`. They are certainly not fitting as names to be shown for user.
- Could they be renamed to themeId or
- Or could we give more user-friendly strings for these keys like "Default light" and "Default dark"? Then there would be no need to add an extra-property

pascalwengerter:
> I think it's reasonable to have a technical and a user visible name - we just need to make sure the user visible strings are translated

Translations are tough-to-impossible for custom themes though (unless I'm missing something)?
Agreed in general and happy to work on it in the upcoming weeks (but after the CERN bucket list?)

dschmidt:
> > I think it's reasonable to have a technical and a user visible name - we just need to make sure the user visible strings are translated
> 
> Translations are tough-to-impossible for custom themes though (unless I'm missing something)? Agreed in general and happy to work on it in the upcoming weeks (but after the CERN bucket list?)

I especially meant the themes we are shipping, if someone ships custom themes they can ship custom translations for theme names as well 😎

Issue #8496, opened 2023-02-23
"Shared on" filter in "Shared with others"  not correct for multi-shared resources


elizavetaRa:
### Steps to reproduce
1. Share file A with Einstein (share 1)
2. Share file B with someone (share 2)
3. Share file A with a Marie (share 3)

### Expected behaviour
- In "Shared with others" share of file A should  appear as more recent on "Shared on" sort, so the time of share 3

### Actual behaviour
- share of file A has "Shared on" time of share 1

PR #9410, opened 2023-07-13
"Shared on" set the latest time

jacob-nv:
## Description

When combining shares, it takes the first time that came from the server, which doesn't have to be the latest. 
Condition checking for a later time has been added to replace the previous time.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/8496

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/helpers/resources.ts
+++ b/packages/web-app-files/src/helpers/resources.ts
@@ -89,6 +89,9 @@ function addSharedWithToShares(shares) {
       previousShare?.file_source === share.file_source
     ) {
       if (ShareTypes.containsAnyValue(ShareTypes.authenticated, [parseInt(share.share_type)])) {
+        if (share.stime > previousShare.stime) {
+          previousShare.stime = share.stime
+        }
         previousShare.sharedWith.push({
           username: share.share_with,
           name: share.share_with_displayname,
@@ -128,6 +131,7 @@ function addSharedWithToShares(shares) {
     }
 
     previousShare = share
+    console.log(share.stime)
     resources.push(share)
   }
   return resources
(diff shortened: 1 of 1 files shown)


packages/web-app-files/src/helpers/resources.ts
```diff
@@ -89,6 +89,9 @@ function addSharedWithToShares(shares) {
       previousShare?.file_source === share.file_source
     ) {
       if (ShareTypes.containsAnyValue(ShareTypes.authenticated, [parseInt(share.share_type)])) {
+        if (share.stime > previousShare.stime) {
+          previousShare.stime = share.stime
+        }
```
JammingBen, line 94: Hmm I'm not the biggest fan of altering the `stime` of the share here. Can't we achieve the same by simply sorting the spaces before iterating them? Something like: `for (const share of shares.sort((a: any, b: any) => b.stime - a.stime))`

jacob-nv: Realistically it has to be this
```
  shares.sort((a: any, b: any) =>
    a.storageId === b.storageId && a.file_source === b.file_source ? b.stime - a.stime : 0
  )
```
Those shares are not grouped by file, so only sorting by stime would display duplicates. I'm not sure it's that performant using sort while the cycle has to run anyway, I would say it's less readable. 
Maybe the whole function might benefit from a little rewrite, group the individual shares by file, then sort inside with simpler comparison, it should be little more cleaner and readable

JammingBen: > Maybe the whole function might benefit from a little rewrite, group the individual shares by file, then sort inside with simpler comparison, it should be little more cleaner and readable

I agree, but let's not make this part of the bug fix 🙂 Let's go with your proposed solution then.

JammingBen (approved):

Issue #9300, opened 2023-06-27
Space permission change flaky


hurradieweltgehtunter:
### Steps to reproduce
1. Create a space as person X
2. Add Person Y to space as manager
3. Set own privileges (Person X) to "Can read" (= remove space manager role)

<img width="1535" alt="image" src="https://github.com/owncloud/web/assets/50302941/eac037ae-a90e-4236-bc4b-790113d07cc1">

### Expected behaviour
- I am not longer able to change Person Y's permission.
- I am not longer able to change my own permission. 

### Actual behaviour
- I am not able to change Person Y permission -> Ok
- I am able to change my own permission -> Not Ok

Request return 200OK but with content: 
```
<ocs><meta><status>error</status><statuscode>404</statuscode><message>No share permission</message></meta></ocs>
```

Reloading the page shows that Person Xs` permission was not reverted -> backend works.

Therefore:
- Frontend needs to understand the error message from backend.
- Frontend should disable the permission changer if permission to change permission is not granted.

PR #9415, opened 2023-07-14
Space permissions, forcing refresh

jacob-nv:
## Description

The solution is similar to the [Avatar issue](https://github.com/owncloud/web/pull/9409), refreshing resources, which will refresh the sidebar under new permissions.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9300

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


JammingBen (commented):
Reactivity should definitely work because spaces are stored globally in the store. I feel like we're not updating the store correctly. Could you please try if changing https://github.com/owncloud/web/blob/master/packages/web-runtime/src/store/spaces.ts#L73 to `Object.assign(spaces[index], space)` works? We did that for normal files a while ago and it worked, I assume it's the same here.

jacob-nv:
> Reactivity should definitely work because spaces are stored globally in the store. I feel like we're not updating the store correctly. Could you please try if changing https://github.com/owncloud/web/blob/master/packages/web-runtime/src/store/spaces.ts#L73 to `Object.assign(spaces[index], space)` works? We did that for normal files a while ago and it worked, I assume it's the same here.

Yeah, you are right, I have expected same thing as with other issue. Fixed now

JammingBen (approved):

PR #9427, opened 2023-07-17
e2e test. search using current folder filter

ScharfViktor:
- added test for using `current folder` filter

### DIFF
--- a/tests/e2e/cucumber/features/smoke/fullTextSearch.ocis.feature
+++ b/tests/e2e/cucumber/features/smoke/fullTextSearch.ocis.feature
@@ -94,12 +94,14 @@ Feature: Search
       | resource     |
       | textfile.txt |
     And "Brian" closes the file viewer
-    Then following resources should be displayed in the files list for user "Brian"
-      | resource                      |
-      | textfile.txt                  |
-      | testFolder/innerTextfile.txt  |
-      | fileToShare.txt               |
-      | fileWithTag.txt               |
-      | withTag.txt                   |
-      | spaceFolder/spaceTextfile.txt |
+    # TODO: uncomment the following step after the bug is fixed
+    # issue: https://github.com/owncloud/web/issues/9428
+    # Then following resources should be displayed in the files list for user "Brian"
+    #   | resource                      |
+    #   | textfile.txt                  |
+    #   | testFolder/innerTextfile.txt  |
+    #   | fileToShare.txt               |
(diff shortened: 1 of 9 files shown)


ScharfViktor:
@JammingBen e2e tests in CI are green. But I have doubt about `fullTextSearch.feature` because locally this test doesn't work for me but green in CI

JammingBen:
> @JammingBen e2e tests in CI are green. But I have doubt about `fullTextSearch.feature` because locally this test doesn't work for me but green in CI

What is the issue locally for you? I tried and had to add `SEARCH_EXTRACTOR_CS3SOURCE_INSECURE: 'true'` like we did in the ocis pipeline a few days ago. Now it succeeds. Maybe it helps for you as well? We should consider adding it to our Docker setup then.

ScharfViktor:
> What is the issue locally for you? I tried and had to add `SEARCH_EXTRACTOR_CS3SOURCE_INSECURE: 'true'` like we did in the ocis pipeline a few days ago. Now it succeeds. Maybe it helps for you as well? We should consider adding it to our Docker setup then.

I added `SEARCH_EXTRACTOR_CS3SOURCE_INSECURE: 'true'` to docker-compose and drone.star to avoid any flaky

The problem of red CI was the old `ocis_commit`  :sweat_smile: , which did not have a location filter

tests/e2e/cucumber/features/smoke/fullTextSearch.ocis.feature
```diff
@@ -91,15 +91,17 @@ Feature: Search
...
+    # TODO: uncomment the following step after the bug is fixed
+    # issue: https://github.com/owncloud/web/issues/9428
+    # Then following resources should be displayed in the files list for user "Brian"
+    #   | resource                      |
+    #   | textfile.txt                  |
+    #   | testFolder/innerTextfile.txt  |
+    #   | fileToShare.txt               |
+    #   | fileWithTag.txt               |
+    #   | withTag.txt                   |
+    #   | spaceFolder/spaceTextfile.txt |
```
saw-jan, line 106: Since, we have issue https://github.com/owncloud/web/issues/9428, this step fails so I have commented out this step. And should be uncommented after the issue is fixed.

CC @ScharfViktor

saw-jan:
One failure related to search in acceptance test: https://drone.owncloud.com/owncloud/web/37451/55/14
I will push a fix

PR #9429, opened 2023-07-17
Port e2e tests #9427 #9398

ScharfViktor:
port #9427 #9398 to stable-7.1


phil-davis:
> OK if CI passes. Do we also need to port this to `stable-7.1` ?

Ohhh - this is targeted to `stable-7.1`

PR #9447, opened 2023-07-19
Keyboard Actions composables in Search Results

jacob-nv:
## Description

Core logic of keyboard actions moved to **useKeyboardActions** composable to web-pkg (this seems to make sense based on [repo-structure.md](https://github.com/owncloud/web/blob/master/docs/development/repo-structure.md)).

Single Keyboard Actions mount to a single element, if no element is filled, it's bound to the document.

Provides 2 functions. 
**bindKeyAction** which accepts the modifier key (based on ModifierKey enum), key (based on Key enum) and callback. 
These actions are saved and called if the event matches the modifier and key. The function returns the index key of action.

**removeKeyAction** allows to remove binded action allowing for dynamic adding / removing key actions.

**useKeyboardActionsSearchTable** is composable utilizing instance of useKeyboardActions, allowing for contextually setting the desired keyActions specifcially for Search Table. 

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9417

## Motivation and Context

Keyboard Actions reusability and extensibility

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -195,9 +195,9 @@ export default defineComponent({
     const scopeQuery = useRouteQuery('scope')
     const doUseScope = useRouteQuery('useScope')
 
-    const { paginatedResources } = useResourcesViewDefaults<Resource, any, any[]>()
+    const resourcesView = useResourcesViewDefaults<Resource, any, any[]>()
     const keyActions = useKeyboardActions('files-view')
-    useKeyboardActionsSearchTable(keyActions, paginatedResources)
+    useKeyboardActionsSearchTable(keyActions, resourcesView.paginatedResources)
 
     const searchTerm = computed(() => {
       return queryItemAsString(unref(searchTermQuery))
@@ -300,7 +300,7 @@ export default defineComponent({
 
     return {
       ...useFileActions({ store }),
-      ...useResourcesViewDefaults<Resource, any, any[]>(),
+      ...resourcesView,
       loadAvailableTagsTask,
       fileListHeaderY,
       fullTextSearchEnabled,
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -144,6 +144,8 @@ import { eventBus, useCapabilityFilesFullTextSearch } from 'web-pkg'
 import ItemFilter from 'web-pkg/src/components/ItemFilter.vue'
 import { isLocationCommonActive } from 'web-app-files/src/router'
 import ItemFilterToggle from 'web-pkg/src/components/ItemFilterToggle.vue'
+import { useKeyboardActions } from 'web-pkg/src/composables/keyboardActions'
+import { useKeyboardActionsSearchTable } from 'web-app-files/src/composables/keyboardActions'
 
 const visibilityObserver = new VisibilityObserver()
 
@@ -192,6 +194,10 @@ export default defineComponent({
     const searchTermQuery = useRouteQuery('term')
     const scopeQuery = useRouteQuery('scope')
     const doUseScope = useRouteQuery('useScope')
+
+    const { paginatedResources } = useResourcesViewDefaults<Resource, any, any[]>()
+    const keyActions = useKeyboardActions('files-view')
+    useKeyboardActionsSearchTable(keyActions, paginatedResources)
 
     const searchTerm = computed(() => {
       return queryItemAsString(unref(searchTermQuery))
--- a/packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
+++ b/packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
@@ -16,26 +16,32 @@ export const useKeyboardActionsSearchTable = (keyActions, paginatedResources) =>
   let fileListClickedMetaEvent
   let fileListClickedShiftEvent
 
-  keyActions.bindKeyAction(null, Key.ArrowUp, () => handleNavigateAction(true))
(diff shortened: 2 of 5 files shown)


jacob-nv:
Please let me know how do you feel about this approach.

packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
```diff
@@ -0,0 +1,162 @@
+import { computed, onBeforeUnmount, onMounted, ref, unref } from 'vue'
+import { useStore } from 'web-pkg/src/composables'
+import { useScrollTo } from 'web-app-files/src/composables/scrollTo'
+import { Key, ModifierKey } from 'web-pkg/src/composables/keyboardActions'
+import { eventBus } from 'web-pkg'
+import { useGettext } from 'vue3-gettext'
+
+export const useKeyboardActionsSearchTable = (keyActions, paginatedResources) => {
```
dschmidt, line 8: Is this really SearchTable specific?

JammingBen (commented):
Awesome stuff, I really like it! Only a few ideas for improvement from my side.

packages/web-app-files/src/composables/keyboardActions/useKeyboardActionsSearchTable.ts
jacob-nv: So here's a thing. At this point, the goal is to provide a search table with actions in a new way.
Technically the search table is the same as a normal table, except you don't have defined space and have fewer actions you can perform. 

This means we can later get rid of `KeyboardAction.vue`, and split it into 2 composables or some abstraction that would combine `paginatedResources` action only and `space` related actions only. 
Then it all depends on personal preference if you want to create a composable that would define each table or you want to define different combinations in the setup of component

JammingBen: > So here's a thing. At this point, the goal is to provide a search table with actions in a new way.

Yes 👍  The scope of this PR is fine as it is, I wouldn't change that. Let's get it merged first.

The next step after that IMO is to refactor `KeyboardAction.vue` into a separate composable like this one (it would be something like `useKeyboardActionsGenericSpaceTable` then) and use in in `GenericSpace.vue`.

After that we can determine the exact differences between both composables and decide if we want to try to combine those two, or if we need to abstract some more logic.

packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
```diff
@@ -0,0 +1,75 @@
...
+        return action.key === key && action.modifier === modifier
+      })
+    if (action) {
+      action.callback(event)
+    }
+  }
+  const bindKeyAction = (
+    modifier: ModifierKey | null = null,
+    key: Key,
+    callback: () => void
```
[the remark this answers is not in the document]
jacob-nv: Yeah, that seems fine, it will keep readability of actions since you can change order.

packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
```diff
@@ -0,0 +1,75 @@
...
+  Space = ' ',
+  ArrowUp = 'ArrowUp',
+  ArrowDown = 'ArrowDown',
+  Esc = 'Escape'
+}
+
+export enum ModifierKey {
+  Ctrl = 'Control',
+  Shift = 'Shift'
+}
```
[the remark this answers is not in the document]
jacob-nv: Here's and issue, keyCode is considered [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).
Since this is a re-imagination of keyboard actions, why not just use the `key` property in terms of longevity?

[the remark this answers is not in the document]
JammingBen: As I already wrote you in chat - you are right, `keyCode` is indeed deprecated. Hence it doesn't make sense for us to keep that alive.

packages/web-app-files/src/components/Search/List.vue
```diff
@@ -193,6 +195,10 @@ export default defineComponent({
     const scopeQuery = useRouteQuery('scope')
     const doUseScope = useRouteQuery('useScope')
 
+    const { paginatedResources } = useResourcesViewDefaults<Resource, any, any[]>()
```
dschmidt, line 198: You're instantiating this twice now

packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const useKeyboardActions = (keyBindOnElementId: string | null = null) => {
+  const actions = ref([])
+  const listener = (event: KeyboardEvent): void => {
+    event.preventDefault()
+    const { key, ctrlKey, metaKey, shiftKey } = event
+    const modifier = metaKey || ctrlKey ? ModifierKey.Ctrl : shiftKey ? ModifierKey.Shift : null
```
dschmidt, line 23: Can you add parentheses here? I think this is rather hard to read

dschmidt (commented):
really nice 🎉

packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
jacob-nv: Parentheses would get removed with linting, but I've just switched it classic if / else if, this should be more readable

dschmidt (approved):
👏🏻

JammingBen (approved):
Awesome! 🚀

PR #9451, opened 2023-07-20
Implement SSE notifications

lookacat:
## Description
We implemented basic SSE to get notifications instantly from the server without requesting all the time.
See https://github.com/owncloud/web/issues/9434

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9434

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


package.json
```diff
@@ -33,6 +33,7 @@
     "not OperaMobile > 0"
   ],
   "dependencies": {
+    "@microsoft/fetch-event-source": "^2.0.1",
```
JammingBen, line 36: Please don't add this to the overall `package.json`, but to the package(s) where this lib is needed.

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -291,11 +292,32 @@ export default {
       setAdditionalData()
     }
 
-    onMounted(() => {
+    const setupSSE = async () => {
```
JammingBen, line 295: Considering we will use that more often in the future, I think it would be great to have this as a composable in `web-pkg`. You'd then pass the URL and 2 (optional) callbacks for `onopen` and `onmessage`.

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -291,11 +292,32 @@ export default {
...
+      await fetchEventSource('/ocs/v2.php/apps/notifications/api/v1/notifications/sse', {
+        headers: {
+          Authorization: `Bearer ${accessToken.value}`
```
JammingBen, line 298: Please add 2 more headers: `Accept-Language` and `X-Request-ID`. See here e.g.: https://github.com/owncloud/web/blob/sse-notifications/packages/web-pkg/src/services/client/client.ts#L110

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -291,11 +292,32 @@ export default {
...
+        },
+        async onopen(response) {
+          if (response.ok && response.headers.get('content-type') === EventStreamContentType) {
+            return
+          } else {
+            console.error(`SSE notifications couldn't be set up ${response.status}`)
+          }
```
JammingBen, line 305: Simplify a bit?

```suggestion
          if (!response.ok || response.headers.get('content-type') !== EventStreamContentType) {
            console.error(`SSE notifications couldn't be set up ${response.status}`)
          }
```

JammingBen (changes requested):
Could you please elaborate a bit how it works in detail? 🙂 I'm especially interested in the token handling, what happens during token renewal? I got a `401` error from the SEE endpoint after a short while, not sure though if that was related to token renewal.

Also it doesn't work with `pnpm vite` for me because it always tries to connect to the wrong port (`9201` instead of `9200`).

packages/web-runtime/src/components/Topbar/Notifications.vue
lookacat: alright

lookacat:
@JammingBen i talked to julian, he told me the connection stays open even if the token is no longer valid. Its weird that you get an 401, i will try to replicate it

lookacat:
@JammingBen i don't get the 401 anymore at least with vite, also port `9201` should work as well as `9200`

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -0,0 +1,35 @@
...
+  onOpen?: (response: Response) => void
+  onMessage?: (msg: EventSourceMessage) => void
+}
+
+export const useServerSentEvents = (options: ServerSentEventsOptions) => {
+  const store = useStore()
+  const language = useGettext()
+  const accessToken = useAccessToken({ store })
+  const setupServerSentEvents = async () => {
+    await fetchEventSource(`${configurationManager.serverUrl}${options.url}`, {
```
lookacat, line 19: `path.join` doesn't work here the output is something like "https:/host.docker..." which is wrong

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -0,0 +1,49 @@
+import { EventSourceMessage, fetchEventSource } from '@microsoft/fetch-event-source'
+import { unref } from 'vue'
+import { v4 as uuidV4 } from 'uuid'
+import { useGettext } from 'vue3-gettext'
+import { configurationManager, useAccessToken, useStore } from 'web-pkg/src'
+export * from './useServerSentEvents'
```
JammingBen, line 6: Looks like something your IDE added 😅

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+      })
+    }
+    try {
+      setupSSE().then(() => {
+        setupServerSentEvents()
+      })
+    } catch (e) {
+      setupServerSentEvents()
+    }
```
JammingBen, line 45: AFAIK the try-catch-block is useless if you don't await `setupSSE` because an error will never be catched.

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -0,0 +1,49 @@
...
+  onOpen?: (response: Response) => void
+  onMessage?: (msg: EventSourceMessage) => void
+}
+
+export const useServerSentEvents = (options: ServerSentEventsOptions) => {
+  const store = useStore()
+  const language = useGettext()
+  const accessToken = useAccessToken({ store })
+  const ctrl = new AbortController()
+  const setupServerSentEvents = async () => {
```
JammingBen, line 19:
```suggestion
  const setupServerSentEvents = () => {
```

JammingBen (commented):
I still have 2 concerns:

* What happens if a user login is actually invalid and the server correctly returns `401`? This solution will constantly try the re-establish the connection, right? Can we add some kind of retry-limit?
* An established connection won't change when the user changes their language. I assume the messages from the server come based on the `Accept-Language` header? In that case we also need to re-establish the connection with the updated language.

JammingBen (changes requested):
I still have 2 issues:

* When changing the language, existing SSE connections don't get terminated. You can see this by changing the language multiple times and then triggering a notification -> there will be multiple notifications.
* Notifications are not being fetched when the tab is inactive, even if I switch back to the tab. I guess we need to call the regular notifications endpoint each time we open a connection. Maybe you can do this in `onOpen`?

lookacat:
alright good findings

JammingBen (approved):
Nice, feels very good now! 👍

Issue #9431, opened 2023-07-18
User allowed to perform delete resources action on empty trashbin 


amrita-shrestha:
### Steps to reproduce
1. user `admin` navigate to `deleted files 
2. click `Empty trash bin` button
[Screencast from 18-7-23 11:26:56 पूर्वाह्न +0545.webm](https://github.com/owncloud/web/assets/54478846/07166731-1bf1-40d9-bab0-b723568f80e5)

3. when hovering over the `Empty trashbin` button got faded on the extreme level
![Screenshot from 2023-07-18 11-30-37](https://github.com/owncloud/web/assets/54478846/3969b444-e187-4370-adce-1925ea76f01a)

### Expected behaviour
When the trash bin is empty user should not be able to click `Empty trash bin` button

### Actual behaviour
Users can click the `Empty trash bin` button multiple times even when trash bin is empty 

### Environment general
**Operating system**: ubuntu 22.04

**Backend (ownCloud Core or Infinite Scale)**: infinite scale

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

![Screenshot from 2023-07-18 11-16-48](https://github.com/owncloud/web/assets/54478846/53dd91e2-1a74-46b4-98cf-bb4561dff07c)

PR #9457, opened 2023-07-20
Display empty trash bin if deleted files

jacob-nv:
## Description

Check if files exist to display "Empty trash bin" action.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9431

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/composables/actions/files/useFileActionsEmptyTrashBin.ts
+++ b/packages/web-app-files/src/composables/actions/files/useFileActionsEmptyTrashBin.ts
@@ -89,7 +89,7 @@ export const useFileActionsEmptyTrashBin = ({ store }: { store?: Store<any> } = 
           return false
         }
 
-        return store.getters['Files/activeFiles'].length > 0
+        return true
       },
       componentType: 'button',
       class: 'oc-files-actions-empty-trash-bin-trigger',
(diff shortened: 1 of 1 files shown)


packages/web-app-files/src/composables/actions/files/useFileActionsEmptyTrashBin.ts
```diff
@@ -89,7 +89,7 @@ export const useFileActionsEmptyTrashBin = ({ store }: { store?: Store<any> } =
           return false
         }
 
-        return true
+        return store.getters['Files/activeFiles'].length > 0
```
JammingBen, line 92: This will hide the button, we prefer to just disable it. You should be able to use the `isDisabled` property for this instead of `isEnabled` (a bit confusing, I know 😅 )

JammingBen (changes requested):

packages/web-app-files/src/composables/actions/files/useFileActionsEmptyTrashBin.ts
jacob-nv: Got it :smile:. Fixed

JammingBen (approved):
👍

PR #9465, opened 2023-07-21
SearchTable composable refactor, KeyboardActions component replaced

jacob-nv:
## Description
Key actions for table search composable split into smaller chunks, adding "space" related actions.
Replacing KeyboardActions components on the GenericSpace page.

Since all keyboard actions are split, it makes it easy to combine to deliver desired actions. Currently, it could be also added to space and other tables, since these need only navigation and mouse actions.

It could be possibly enhanced with keyActionTableWrapper that would initialize individual composables based on some sort of options object (making it "single liner")

## Motivation and Context
Replacing Keyboard Actions component

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Summary
 * Bugfix - Media controls overflow on mobile screens: [#9351](https://github.com/owncloud/web/pull/9351)
 * Bugfix - Space editors see empty trashbin and delete actions in space trashbin: [#9389](https://github.com/owncloud/web/pull/9389)
 * Bugfix - Merging folders: [#9461](https://github.com/owncloud/web/issues/9461)
+* Enhancement - Scroll to newly created folder: [#7600](https://github.com/owncloud/web/issues/7600)
 * Enhancement - Add hasPriority property for editors per extension: [#7967](https://github.com/owncloud/web/pull/7967)
 * Enhancement - Improve extension app topbar: [#8422](https://github.com/owncloud/web/issues/8422)
 * Enhancement - Open individually shared file in dedicated view: [#8445](https://github.com/owncloud/web/issues/8445)
@@ -138,6 +139,15 @@ Details
 
    https://github.com/owncloud/web/issues/9461
    https://github.com/owncloud/web/pull/9477
+
+* Enhancement - Scroll to newly created folder: [#7600](https://github.com/owncloud/web/issues/7600)
+
+   After creating a new folder that gets sorted into the currently displayed resources but
+   outside of the current viewport, we now scroll to the new folder.
+
+   https://github.com/owncloud/web/issues/7600
+   https://github.com/owncloud/web/issues/7601
+   https://github.com/owncloud/web/pulls/8145
 
 * Enhancement - Add hasPriority property for editors per extension: [#7967](https://github.com/owncloud/web/pull/7967)
 


packages/web-app-files/src/composables/keyboardActions/useKeyboardTableActions.ts
```diff
@@ -0,0 +1,15 @@
+import { Key, ModifierKey } from 'web-pkg/src/composables/keyboardActions'
+import { useStore } from 'web-pkg'
+import { useGettext } from 'vue3-gettext'
+
+export const useKeyboardTableActions = (keyActions) => {
```
JammingBen, line 5:
```suggestion
export const useKeyboardTableActions = (keyActions: KeyboardActions) => {
```

packages/web-app-files/src/composables/keyboardActions/useKeyboardTableMouseActions.ts
```diff
@@ -0,0 +1,61 @@
+import { onBeforeUnmount, onMounted, unref, computed } from 'vue'
+import { eventBus, useStore } from 'web-pkg'
+
+export const useKeyboardTableMouseActions = (keyActions) => {
```
JammingBen, line 4:
```suggestion
export const useKeyboardTableMouseActions = (keyActions: KeyboardActions) => {
```

jacob-nv:
Test added for the main composable, for the wrappers that encapsulate specific functionality, I'm getting into issue with useStore.
Added some additional interfaces and computed filtered actions (removes codes repetition for tests as well)

packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
```diff
@@ -15,8 +16,25 @@ export enum ModifierKey {
...
+
+export interface KeyboardAction {
+  primary: Key
+  modifier: ModifierKey | null
+  callback: (event: KeyboardEvent) => void
+}
+
+export const useKeyboardActions = (keyBindOnElementId: string | null = null): KeyboardActions => {
+  const actions = ref<Array<KeyboardAction | null>>([])
+  const actionsFiltered = computed(() => actions.value.filter((action) => action !== null))
```
JammingBen, line 36: I'm a bit confused by `actions` and `actionsFiltered`. So `actions` are all actions that have been registered, including those that might have been removed again? And `actionsFiltered` are only the actions that are currently registered?

JammingBen: And you need the index to successfully remove actions again, right? What do you think of generating an id when registering a new action und use that instead of the index? We have a `uuid` lib that could do that.

I would feel a bit more robust IMO. Also you could use slice and we would get rid of `actionsFiltered`.

JammingBen: It just feels more robust and also we would only have one action property that holds all the currently registered actions. So I prefer working with ids here, yes 👍

jacob-nv: Pushed should be good now.

JammingBen (commented):
One small finding, rest LGTM!

Unit tests in `GenericSpace.spec.ts` still fail though, you should be able to fix them by adding `jest.mock('web-app-files/src/composables/keyboardActions')`.

JammingBen:
@jacob-nv Now our e2e tests are failing: https://drone.owncloud.com/owncloud/web/37716/11/15 😬 

You can run those tests locally via `OCIS=true pnpm test:e2e:cucumber path/to/your/feature/file`. To run the failing test it would be: `OCIS=true pnpm test:e2e:cucumber tests/e2e/cucumber/features/smoke/copyMove.feature:6`. (Note: You can either run all tests of a feature file like `copyMove.feature` or a dedicated test by specifying a line number: `copyMove.feature:6`).

I assume the tests fail because we use capital letters in our test definitions: https://github.com/owncloud/web/blob/master/tests/e2e/support/objects/app-files/resource/actions.ts#L567 (see C, X, V). Changing this should fix the tests.

I'm not sure, do you think we should add `toLowerCase()` in the composable before comparing the strings (https://github.com/owncloud/web/blob/master/packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts#L32)?

jacob-nv:
> 

You are right, toLowerCase makes more sense. I don't see options where person is gonna be specifically using capital letters.

JammingBen (approved):
Nice work! 🙂

Issue #9454, opened 2023-07-20
[cloud import] Starting a second import with one being active already breaks


JammingBen:
### Steps to reproduce
1. Start importing 100+ files from a cloud
2. While the upload is running, perform another import -> The modal will be almost empty, closing it will stop the currently running import.

Solution (for now at least) is to prohibit running 2 cloud imports in parallel.

JammingBen:
Fixed via https://github.com/owncloud/web/pull/9470

PR #9470, opened 2023-07-24
fix: prevent uploads when a cloud import is running

JammingBen:
## Description
Running other uploads or imports when a cloud import is running causes several issues, hence prevent it from happening.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9454

## Screenshots
Personal

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
@@ -96,6 +96,7 @@
       class="oc-width-auto"
       close-on-click
       padding-size="small"
+      @show-drop="showDrop"
     >
       <oc-list id="upload-list" :class="{ 'oc-pb-s': extensionActions.length }">
         <li class="oc-menu-item-hover">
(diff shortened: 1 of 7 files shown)


JanAckermann (approved):

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -274,6 +294,15 @@ export default defineComponent({
       return unref(currentFolder)?.canUpload({ user: store.getters.user })
     })
 
+    const actionKeySuffix = ref(uuidv4())
+    const showDrop = () => {
+      // force actions to be re-rendered when the drop is being opened
+      actionKeySuffix.value = uuidv4()
+    }
```
dschmidt, line 301: meeeh - good workaround for now, but we should figure out how to make everything properly reactive

packages/web-runtime/src/services/uppyService.ts
```diff
@@ -276,6 +276,16 @@ export class UppyService {
     return this.uppy.cancelAll()
   }
 
+  getCurrentUploads(): Record<string, unknown> {
+    return this.uppy.getState().currentUploads
+  }
+
+  isRemoteUploadInProgress(): boolean {
+    // only one remote upload at a time means it's sufficient to check the first file
```
dschmidt, line 284: What if a local upload was started before?

JammingBen: There is no scenario where we have a local and a remote upload running at the same time. However, when a file of an import failed before, that could be an issue here indeed. Let me change the condition a bit.

dschmidt (approved):

PR #9474, opened 2023-07-24
errorLog: Handle multiple errors (graph)

JanAckermann:
## Related Issue

- Part of https://github.com/owncloud/web/issues/9449

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-runtime/src/store/app.ts
```diff
@@ -7,23 +7,24 @@ const state = {
...
+    message.errors = message.error ? [message.error] : message.errors || []
 
-    if (message.error) {
-      const xRequestID = getXRequestID(message.error)
-      if (xRequestID) {
-        message.errorLogContent = `X-Request-ID: ${xRequestID}`
-      }
-    }
+    const xRequestIds = message.errors
+      .map((error) => getXRequestID(error.reason || error))
```
JanAckermann, line 22: Reason comes from Promise.allSettled, this can be done in showErrorMessage fnc call as well. 

Opinions ?

packages/web-runtime/src/store/app.ts
```diff
@@ -7,23 +7,24 @@ const state = {
...
+    const getXRequestID = (error: AxiosError): string | null => {
       if (error.response?.headers?.['x-request-id']) {
         return error.response.headers['x-request-id']
       }
+      return null
     }
+
     message.status = message.status || 'danger'
     message.timeout = message.timeout || 0
+    message.errors = message.error ? [message.error] : message.errors || []
```
JanAckermann, line 19: @dschmidt you can pass errors or error to showErrorMessage. 
Should we purge error and rely on the developer passes errors:[e] to the function

packages/web-pkg/src/components/Spaces/QuotaModal.vue
```diff
@@ -181,15 +182,20 @@ export default defineComponent({
...
       if (succeeded.length) {
         store.dispatch('showMessage', { title: getSuccessMessage(succeeded.length) })
       }
       const errors = results.filter((r) => r.status === 'rejected')
       if (errors.length) {
         errors.forEach(console.error)
-        store.dispatch('showErrorMessage', { title: getErrorMessage(errors.length) })
+        store.dispatch('showErrorMessage', {
+          title: getErrorMessage(errors.length),
+          errors
```
JammingBen, line 197: In the other files you add `.map((f) => f.reason)` to `errors`. Why is it not necessary here?

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -500,70 +502,157 @@ export default defineComponent({
...
+      const succeeded = results.filter((r) => r.status === 'fulfilled')
+      if (succeeded.length) {
+        const title =
+          succeeded.length === 1 && groupsToAdd.length === 1 && affectedUsers.length === 1
+            ? $gettext('Group assignment "%{group}" was added successfully', {
+                group: groupsToAdd[0].displayName
+              })
+            : $ngettext(
+                '%{groupAssignmentCount} group assignment was added successfully',
+                '%{groupAssignmentCount} group assignments were added successfully',
```
JammingBen, line 542: I find the term "group assignment" to be quite technical. The action is labeled "Add to groups", I'd prefer to stay in that terminology, so e.g.  "Users have been added to the selected groups". Ultimately @tbsbdr decides of course.

JanAckermann: Problem is the success/error count if we want group assignment consist of n groups * n users if we want to keep the n users had been added to group, n users failed added to group we need to analyse the requests and add extra logic

packages/web-app-admin-settings/src/composables/actions/groups/useGroupActionsDelete.ts
```diff
@@ -45,7 +48,7 @@ export const useGroupActionsDelete = ({ store }: { store?: Store<any> }) => {
               { groupCount: failed.length.toString() },
               true
             )
-      store.dispatch('showErrorMessage', { title })
+      store.dispatch('showErrorMessage', { title, errors: (failed as any).map((f) => f.reason) })
```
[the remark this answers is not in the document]
JanAckermann: Idk ide is complaining 🤷‍♀️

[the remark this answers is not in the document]
JammingBen: It should tell you why as well 😄 You need to cast `failed` to `PromiseRejectedResult[]`.

packages/web-app-admin-settings/src/views/Users.vue
JammingBen: Okay I understand. What about simply "n groups were added successfully"?

JanAckermann: So with n is users * groups ? That's fine for me

JammingBen: Arghs no, now I fully understand, sorry! That might be confusing as well for the user. Hm. Then I don't have a better idea than "group assignment". That would be "Gruppenzugehörigkeit" in German? 😂

JanAckermann: Or ze very well "Gruppenzuweisung"

JammingBen (approved):

PR #9438, opened 2023-07-18
feat: add collapsible table and enable custom vite config for cern

JammingBen:
## Description
Adds a custom vite config for CERN. It overwrites some of our code paths to implement custom logic that only applies to CERN. Web can be built and run using this config via `pnpm build:w -c vite.cern.config.ts` or `pnpm vite -c vite.cern.config.ts`.

This PR already includes 2 things that have been overwritten:

* Collapsible table for the Shared with me/others views
* A custom composable for the `tokenInfo` request to avoid `404` errors.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/9424
- Fixes https://github.com/owncloud/web/issues/9310
- Part of https://github.com/owncloud/web/issues/9278

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

PR #9504, opened 2023-07-31
Filter sidebar shares for displayName

### LINKED ISSUE #9257: crash on bad egroups
Badly formed response was causing frontend to crash

### Steps to reproduce
1. (tbd)
2.
3.

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

relates to: 808848f25a51f5af7fe251aea616d5e2f06ab710

pascalwengerter:
## Related Issue
- Fixes #9257

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
@@ -198,17 +198,20 @@ export default defineComponent({
     },
 
     collaborators() {
-      return [...this.outgoingCollaborators].sort(this.collaboratorsComparator).map((c) => {
-        const collaborator: typeof c & { key?: string; resharers?: User[] } = { ...c }
-        collaborator.key = 'collaborator-' + collaborator.id
-        if (
-          collaborator.owner.name !== collaborator.fileOwner.name &&
-          collaborator.owner.name !== this.user.id
-        ) {
-          collaborator.resharers = [collaborator.owner]
-        }
-        return collaborator
-      })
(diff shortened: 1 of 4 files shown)


packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
```diff
@@ -198,17 +198,20 @@ export default defineComponent({
...
+          const collaborator: typeof c & { key?: string; resharers?: User[] } = { ...c }
+          collaborator.key = 'collaborator-' + collaborator.id
+          if (
+            collaborator.owner.name !== collaborator.fileOwner.name &&
+            collaborator.owner.name !== this.user.id
+          ) {
+            collaborator.resharers = [collaborator.owner]
+          }
+          return collaborator
+        })
```
JammingBen, line 214: I'd prefer adding the filter to `outgoingCollaborators` in our `useShare` composable.

I guess this is CERN specific? We can also think about overwriting the composable via cern config [like we did here](https://github.com/owncloud/web/pull/9438/commits/4bd2760355a0fa6e9b055a21518b45ef68376e4f). Although it's such a small change, it's probably not worth the effort 🤔

pascalwengerter: I'm open to doing that, the benefit of doing it here is that one could also add some sort of hint to the UI about further shares existing but the webUI not being able to correctly render these (whcih technically could also be done inside the composable and then exported, ofc). Let me discuss this with @tbsbdr

JammingBen (approved):

PR #9508, opened 2023-08-01
init md highlight

JanAckermann:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-text-editor/package.json
+++ b/packages/web-app-text-editor/package.json
@@ -12,5 +12,9 @@
     "vuex": "4.1.0",
     "web-client": "workspace:@ownclouders/web-client@*",
     "web-pkg": "workspace:@ownclouders/web-pkg@*"
+  },
+  "dependencies": {
+    "highlight.js": "^11.8.0",
+    "marked-highlight": "^2.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/web-app-text-editor/package.json
```diff
@@ -12,5 +12,9 @@
     "vuex": "4.1.0",
     "web-client": "workspace:@ownclouders/web-client@*",
     "web-pkg": "workspace:@ownclouders/web-pkg@*"
+  },
+  "dependencies": {
+    "highlight.js": "^11.8.0",
```
pascalwengerter, line 17: Why are these "real" dependencies while e.g. `marked` is a peerDep? 🤔

JanAckermann: Just testing, this pr probably won't make it. 
But if, it must land in pkg or runtime (bc spaces also render md in description)

pascalwengerter: Thanks for clarification! In my opinion it'd be better to put this in a dedicated code editor extension to not blow up the default app bundle :)

JanAckermann: We are in decision to bundle CodiMD with the app instead. 
That is supporting syntax highlighting by default.

JammingBen:
Since we'll be using CodiMD in the future, can we close here? Or still relevant?

Issue #9444, opened 2023-07-19
Search location filter labels do not update on language change


JammingBen:
### Steps to reproduce
1. Go to your account page (avatar on the top right, click on your username)
2. Change the language -> the location filter labels in the search bar ("Current folder" and "All files") do not update

<img width="220" alt="image" src="https://github.com/owncloud/web/assets/50302941/25291f70-882c-44b8-a6ee-f0dbed51a491">

Simply wrapping them via `$gettext()` like we usually do should do the trick.

PR #9528, opened 2023-08-02
Search Bar filter translation

jacob-nv:
## Description

Moving translation to template keeps language and doesn't break "current folder"

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9444

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-pkg/src/components/SearchBarFilter.vue
+++ b/packages/web-pkg/src/components/SearchBarFilter.vue
@@ -4,7 +4,7 @@
       <oc-filter-chip
         :is-toggle="false"
         :is-toggle-active="false"
-        :filter-label="$gettext(currentSelection.title)"
+        :filter-label="currentSelectionTitle"
         :selected-item-names="[]"
         class="oc-search-bar-filter"
         raw
@@ -23,7 +23,7 @@
             :data-test-id="option.id"
             @click="onOptionSelected(option)"
           >
-            <span>{{ $gettext(option.title) }}</span>
+            <span>{{ option.title }}</span>
             <div v-if="option.id === currentSelection.id" class="oc-flex">
               <oc-icon name="check" />
             </div>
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/components/SearchBarFilter.vue
```diff
@@ -64,12 +64,12 @@ export default defineComponent({
     const locationOptions = ref<LocationOption[]>([
       {
         id: SearchLocationFilterConstants.currentFolder,
-        title: $gettext('Current Folder'),
+        title: 'Current Folder',
```
JammingBen, line 67: Please don't remove `$gettext` here. This method (and its variations) are an indicator for our language string extractor to sync strings to Transifex, where they eventually get translated.

packages/web-pkg/src/components/SearchBarFilter.vue
```diff
@@ -122,7 +122,7 @@ export default defineComponent({
       emit('update:modelValue', { value: option })
     }
 
-    return { currentSelection, isIndexGreaterZero, onOptionSelected, locationOptions }
+    return { currentSelection, isIndexGreaterZero, onOptionSelected, locationOptions, $gettext }
```
JammingBen, line 125: No need to return it explicitly.

```suggestion
    return { currentSelection, isIndexGreaterZero, onOptionSelected, locationOptions}
```

JammingBen (changes requested):

packages/web-pkg/src/components/SearchBarFilter.vue
jacob-nv: Right, looks like my previous approach was better. And it seems that moving "currentFolderEnabled" prop from computed into locationOptions (as computed) fixed the issue

JammingBen (approved):

Issue #9494, opened 2023-07-27
Scroll to new folder when being created on another page


JammingBen:
With https://github.com/owncloud/web/pull/9145, we implemented a feature that scrolls to a newly created folder, given the folder has been added to the current page. This does not yet work with pagination. Meaning: when the folder is being created on another page, Web should jump to that page and scroll to the folder.

PR #9548, opened 2023-08-07
Create new folder, navigate to page

jacob-nv:
## Description

Once folder is created it's id is matched with sorted items to get page on which is located

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9494

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


dschmidt (commented):
Can you move the logic tojthe scrollToResource composable, so it can be reused?

pascalwengerter:
> Can you move the logic tojthe scrollToResource composable, so it can be reused?

Also my first thought :) within the `scrollToResource` there's a check if the resourceId is on the current page, if not it currently returns but you could check if it exists within the paginated resources and then trigger the page transition

pascalwengerter:
@jacob-nv I think this PR lacks a changelog item linking this PR and the issue it closes :)

packages/web-app-files/src/composables/scrollTo/useScrollTo.ts
```diff
@@ -22,7 +34,38 @@ export const useScrollTo = (): ScrollToResult => {
...
+    if (unref(viewMode) === ViewModeConstants.tilesView.name) {
+      return determineResourceTilesSortFields(unref(files)[0])
+    }
+    return determineResourceTableSortFields(unref(files)[0])
+  })
+  const { items: sortedItems } = useSort({ items: files, fields: sortFields })
+
+  const scrollToResource = async (resourceId: Resource['id'], options = { forceScroll: false }) => {
+    if (unref(sortedItems)) {
+      const index = findIndex(unref(sortedItems), (item: Resource) => item.id === resourceId)
```
pascalwengerter, line 60: I'd prefer doing this lookup only in the `if (!resourceElement) {...}` and check the "current page DOM" for the fileId first?

dschmidt: I wouldn't prefer that tbh 😬

pascalwengerter: I guess I have an idea why, but would you mind elaborating?

dschmidt: yep, that's what I meant

dschmidt (changes requested):
I don't really like, that we're now sorting and paginating all files again, when it also happened in the view already.

What I thought about: make `useScrollTo`  publish an event over the `eventBus`, that `usePagination` can consume to switch to the right page. Afterwards the actual scrolling can happen


edit:
My initial feeling was that we should just in the sorted/paginated resources from the `useResourceViewDefaults` composable - but that doesn't work when we want to switch to a newly created file/folder, where the action was triggered from `CreateAndUpload` - that's why I think using the eventBus might be the best option

packages/web-app-files/src/composables/scrollTo/useScrollTo.ts
dschmidt: ... but I would solve the whole issue a little differently, wrote it down in detail in my review

jacob-nv:
> I don't really like, that we're now sorting and paginating all files again, when it also happened in the view already.
> 
> What I thought about: make `useScrollTo` publish an event over the `eventBus`, that `usePagination` can consume to switch to the right page. Afterwards the actual scrolling can happen
> 
> edit: My initial feeling was that we should just pass in the sorted/paginated resources from the `useResourceViewDefaults` composable - but that doesn't work when we want to switch to a newly created file/folder, where the action was triggered from `CreateAndUpload` - that's why I think using the eventBus might be the best option

Rewrote, but still had to move part of the useScroll login to usePagination. After publishing event and routing, useScroll is not able to query the DOM element. Or do you have better idea how to deal with? Wasn't able to sync it in a way it would execute without moving that part of logic

dschmidt:
Thanks for taking care!

Yeah, we don't have a proper way to "ack" an event - maybe you could send a "navigated to page" event ... but it's probably not bad like this either and you could leave it like this. I'm about to head out, can look later or tomorrow again

packages/web-pkg/src/composables/pagination/usePagination.ts
```diff
@@ -40,6 +45,32 @@ export function usePagination<T>(options: PaginationOptions<T>): PaginationResul
...
   })
 
+  eventBus.subscribe(
+    'app.files.navigate.page',
+    async ({ resourceId, forceScroll }: { resourceId: string; forceScroll: boolean }) => {
+      const index = findIndex(unref(options.items), (item: any) => item.id === resourceId)
+      if (index >= 0) {
+        const page = Math.ceil((index + 1) / Number(unref(perPage)))
+        await router.push({ ...unref(route), query: { ...unref(route).query, page } })
+
```
dschmidt, line 55: Looking at this again, I'm thinking maybe we should actually send a `app.files.navigate.scroll` event (or similar) over the `eventBus`here. Otherwise it feels too tightly coupled to the dom structure.

```suggestion
    }
    eventBus.publish('app.files.navigate.scrollTo', { resourceId, forceScroll })
  }
```


I think I like that better than using another instance of the useScrollTo composable with an `doActualScroll` method...


@JammingBen thoughts?

JammingBen: I agree with not handling the scroll logic here. But why not call `scrollToResource()` from the `useScroll` composable? IMO it would be de-coupled quite well and we would not need to add more complexity by adding more global events.

jacob-nv: @dschmidt Pushed commit, it seems to work quite well this way.

@JammingBen You mean calling useScroll  in usePagination? This one is not possible because useScroll comes from web-app-files and usePagination comes from web-pkg

JammingBen: Ahh sorry, I thought it has been moved to `web-pkg`. Then this is fine by me!

packages/web-pkg/src/composables/pagination/usePagination.ts
```diff
@@ -40,6 +45,19 @@ export function usePagination<T>(options: PaginationOptions<T>): PaginationResul
...
 
+  eventBus.subscribe(
+    'app.files.navigate.page',
+    ({ resourceId, forceScroll }: { resourceId: string; forceScroll: boolean }) => {
+      const index = findIndex(unref(options.items), (item: any) => item.id === resourceId)
+      if (index >= 0) {
+        const page = Math.ceil((index + 1) / Number(unref(perPage)))
+        router.push({ ...unref(route), query: { ...unref(route).query, page } }).then(() => {
+          eventBus.publish('app.files.navigate.scrollTo', { resourceId, forceScroll })
+        })
```
dschmidt, line 56: We're getting somewhere! 

I'm not completely happy with this, but I played around with it and it's tricky to solve it directly through the page ref. I would like to try a few more things, but have to focus on other things for now. 
If that's okay for you, I will try again tonight and if I don't find anything useful, I'm fine to have this merged tomorrow.

dschmidt: I obviously didn't  get to it, feel free to merge

JammingBen (approved):
Code LGTM, unit tests are failing though: https://drone.owncloud.com/owncloud/web/38188/9/5

packages/web-pkg/src/composables/pagination/usePagination.ts
```diff
@@ -6,6 +6,9 @@ import {
   useRouteQueryPersisted,
   PaginationConstants
 } from 'web-pkg/src/composables'
+import { eventBus } from 'web-pkg'
```
JammingBen, line 9: To avoid weird round trips in our unit tests. This should make them work again.

```suggestion
import { eventBus } from 'web-pkg/src/services/eventBus'
```

jacob-nv: Done

PR #9559, opened 2023-08-08
keyboardActions: Use event.preventDefault only when an action has been found

### LINKED ISSUE #9547: Paste file keyboard action doesn't work (Cmd + V | Ctrl + V)
### Steps to reproduce
1. Go to files
2. Copy File
3. Paste file with keyboard actions

### Expected behaviour
File should be pasted

### Actual behaviour
Nothing happens, no console error

### Environment general
**Operating system**:
Mac

### LINKED ISSUE #9554: Make some keyboard actions global to improve UX
Related issue / Fixes issue: https://github.com/owncloud/web/issues/9547

The refactoring of the keyboard actions looks pretty good i just have one small thing which we need to change :)
With the current implementation we bind `ctrl+c` and `ctrl+v` on the `files-view`. The problem with that is that the user is not always focused on the `files-view` and we need a way the user can paste a file even so he lost the focus on this element. (e.g. try out dropbox, you can paste even when navigating in the left sidebar)
If we don't do that the paste action doesn't work randomly at some point of your workflow which can confuse the user ^^

If you take a look at the old `KeyboardActions.vue` (https://github.com/owncloud/web/pull/9465/files) we did bind the paste action via `document.addEventListener('keydown', handleGlobalShortcuts)` and we use `areCustomKeyBindingsEnabled` to check if we really should do our custom paste event (or if the user e.g. pastes text)

dschmidt:
## Description
... and handle multiple registered actions for bound keys

## Related Issue

- Fixes #9547
- Fixes #9554

## Motivation and Context
With `event.preventDefault()` at the top of the handler all text input is blocked

cc @jacob-nv 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -201,7 +201,7 @@ export default defineComponent({
 
     const resourcesView = useResourcesViewDefaults<Resource, any, any[]>()
 
-    const keyActions = useKeyboardActions('files-view')
+    const keyActions = useKeyboardActions()
     useKeyboardTableNavigation(keyActions, resourcesView.paginatedResources)
     useKeyboardTableMouseActions(keyActions)
     useKeyboardTableActions(keyActions)
--- a/packages/web-app-files/src/components/SideBar/SideBar.vue
+++ b/packages/web-app-files/src/components/SideBar/SideBar.vue
@@ -12,7 +12,7 @@
     :loading="loading"
     :is-header-compact="!!loadedResource"
     v-bind="$attrs"
-    data-custom-key-bindings="true"
+    data-custom-key-bindings-disabled="true"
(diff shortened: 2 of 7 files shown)


dschmidt:
We could further simplify this by using https://vueuse.org/core/useEventListener/

JammingBen (approved):
LGTM, unit tests are still failing though. @jacob-nv That solves the issue we talked about yesterday regarding the blocked browser shortcuts 🙂 

> We could further simplify this by using https://vueuse.org/core/useEventListener/

That would be nice, I think there are a lot of other use-cases for this as well.

PR #9585, opened 2023-08-16
Ease use of disabled in OcTable & show status icon for processing items 

JanAckermann:
## Related Issue

- Part of https://github.com/owncloud/web/issues/9558

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/design-system/src/components/OcResource/OcResource.vue
```diff
@@ -210,6 +210,28 @@ export default defineComponent({
 
     thumbnail() {
       return this.resource.thumbnail
+    },
+
+    showStatusIcon() {
```
JanAckermann, line 215: @lookacat service function :) <3

packages/design-system/src/components/OcTable/OcTable.vue
```diff
@@ -220,11 +220,11 @@ export default defineComponent({
...
     },
     /**
-     * The ids of disabled data items. Null or an empty string/array for no disabled items.
+     * The array of disabled data items. Empty array for no disabled items.
      */
     disabled: {
-      type: [String, Array],
-      default: null
+      type: Array as PropType<any[]>,
+      default: () => []
```
JammingBen, line 227: Since we can't type it properly, could you add a comment explaining that an item needs to have one property that matches `idKey` (default = `id`).

JammingBen (commented):
Seeing it now I'm asking myself: did we gain anything by using full resources instead of ids? 🤔 It adds quite a bit of complexity, especially since we already have a full list of resources in OcTable. Now we additionally pass a subset of them to the same component.

JanAckermann:
> Seeing it now I'm asking myself: did we gain anything by using full resources instead of ids? 🤔 It adds quite a bit of complexity, especially since we already have a full list of resources in OcTable. Now we additionally pass a subset of them to the same component.

IMHO it's more handy, developers can just use a subset of the original array and don't need to map those additionally.  I don't see more complexity

JammingBen:
I mean lines like these are more complex than before: `filter((resource) => !this.disabledResources.find((r) => r.id === resource.id))`. Plus you can only use and rely on `id` anyways because items in `disabled` can literally be anything. So using actual resources from `disabled` for anything other than `id` doesn't feel safe 🤷

JanAckermann:
Well I don't get it, Do you want me to revert it?

JammingBen:
Nono not completely, but I thought we keep working id based and make the `disabled` prop in `OcTable` simply hold an array of ids. And of course remove the `disabled` prop from `ResourceTable` like you already did.

JanAckermann:
Got it :+1:

Issue #9580, opened 2023-08-16
Run `app-provider`  `e2e office-suites` test for `ocis` in nightly


SagarGi:
### Description
Currently the `external-app-provider` e2e tests for office suites only run for `full-ci`. And also it does not run in daily cron `nightly`. It is good that we run that test with only `full-ci` tag but its better to run it on daily `nightly`. 

Build Link not ran in nightly: https://drone.owncloud.com/owncloud/web/38196
Build link ran on `full-ci` tag: https://drone.owncloud.com/owncloud/web/38167/12/1

### TODO
- [ ] make the `app-propvider` e2e officesuites test run in nightly

PR #9600, opened 2023-08-21
[tests-only][full-ci]Run office-suites app-provider tests in nightly as well

SagarGi:
### Description
The recently added `office-suites` app provider tests is not ran in nightly. Its better to run it at least in nightly since it is only ran with `full-ci` tag for the pull request. So this PR runs the test in every nightly.

### Related Issue:
Fixes https://github.com/owncloud/web/issues/9580

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -721,7 +721,7 @@ def e2eTests(ctx):
         for item in default:
             params[item] = matrix[item] if item in matrix else default[item]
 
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
Logic looks good now - we run the `oCIS-app-provider` suite when full-ci is specified, and in all cron jobs. So if it is not full-ci and not cron then do not run this suite.

saw-jan (approved):

Issue #9267, opened 2023-06-16
Backups App


tbsbdr:
- relates to: https://github.com/owncloud/web/pull/9159/commits/27146023f4ca5793a16e16382b353aaf67bb37d4
- enable to run the app as extension
- provide an option to reuse components

kulmann:
We discussed in a call that ownCloud will not implement the `Backups` app, but will provide component reusability as needed. @labkode @diocas @elizavetaRa please confirm.

@elizavetaRa @diocas please provide a full list of components and composables from the files app that are needed in a more generic way (= web-pkg).

labkode:
Hi @kulman,

The list of components are available as `imports` in the following commit:
https://github.com/cernbox/web/commit/27146023f4ca5793a16e16382b353aaf67bb37d4

pascalwengerter:
> Hi @kulman,
> 
> The list of components are available as `imports` in the following commit: [cernbox@2714602](https://github.com/cernbox/web/commit/27146023f4ca5793a16e16382b353aaf67bb37d4)

Just want to point out that this goes beyond the scope of "just four components imported" since
- it imports helpers, composables and more (search for `'web-app-files/` within `packages/web-app-backups/src/components/BackupsSection.vue`
- `AppBar` and `ResourceTable` both use further helpers/composables/components from the files app and interact with the store, which needs to be taken into consideration when estimating the complexity behind an extraction.

Not saying it's not possible though, I suppose an extension system would actually greatly benefit from having these available

labkode:
@pascalwengerter in case it helps: https://github.com/owncloud/core/issues/17610

pascalwengerter:
> @pascalwengerter in case it helps: [owncloud/core#17610](https://github.com/owncloud/core/issues/17610)

Are you sure the link you provided is the right one? It doesn't fit to how the current project is set up or a modern approach to JS. What we will have to do to resolve #8281 (and thereby this issue) is to extract the used functionality into e.g. the `web-pkg` submodule and therefore make it available to extension developers

PR #9602, opened 2023-08-21
Backup App migration

jacob-nv:
## Description

Moved back up app, moved all the dependencies from web-app-files.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9267

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-pkg/src/composables/resourcesViewDefaults/useResourcesViewDefaults.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  sideBarOpen: Ref<boolean>
+  sideBarActivePanel: Ref<string>
+} & SelectedResourcesResult &
+  ScrollToResult
+
+export const useResourcesViewDefaults = <T, TT, TU extends any[]>(
+  options: ResourcesViewDefaultsOptions<TT, TU> = {}
+): ResourcesViewDefaultsResult<T, TT, TU> => {
+  const loadResourcesTask = options.loadResourcesTask || folderService.getTask()
```
dschmidt, line 59: I don't think it's a really good idea to move the whole FolderLoaderService and all of the loaders into web-pkg at this point.

I would suggest to make `options.loadResourcesTask` of type: ` Task<TT, TU> || () =>  Task<TT, TU>`
and then pass it in everywhere where useResourcesViewDefaults is used: `{ loadResourcesTask: () => folderService.getTask() } `

jacob-nv: Looked at it, the main issue is that the folderService is located in web-app-files. Which shouldn't be reachable from web-app-backups

packages/web-pkg/src/composables/scrollTo/useScrollTo.ts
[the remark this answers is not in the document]
jacob-nv: Well, there's actually different id for topbar. So I could modify it to pass the id with default being 'files-app-bar' and specifically for admin settings pass 'admin-settings-app-bar'.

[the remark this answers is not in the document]
dschmidt: or no default at all and require it

packages/web-pkg/src/composables/resourcesViewDefaults/useResourcesViewDefaults.ts
dschmidt: Yes, that's why I said you could inject a getTask function...

dschmidt: My main issue is that there are so many route checks in the loaders, so they are actually pretty files app specific...

packages/web-pkg/src/composables/scrollTo/useScrollTo.ts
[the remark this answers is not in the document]
JammingBen: > or no default at all and require it

Yes that sounds like the best solution IMO. The composable shouldn't make any assumption about the selector here.

packages/web-pkg/src/composables/resourcesViewDefaults/useResourcesViewDefaults.ts
jacob-nv: I might have explained poorly.
`packages/web-app-backups/src/components/BackupsSection.vue
` line: 141

uses `useResourcesViewDefaults`, which then should get the injected `{ loadResourcesTask: () => folderService.getTask() }`, the issue here is that for `folderService` to be injected, it would have to have access to _web-app-files_, which goes against the goal of this issue, no?

packages/web-pkg/src/components/AppBar/AppBar.vue
```diff
@@ -0,0 +1,342 @@
...
+          @item-dropped-breadcrumb="fileDroppedBreadcrumb"
+        >
+          <template #contextMenu>
+            <context-actions
+              :action-options="{ space, resources: breadcrumbsContextActionsItems.filter(Boolean) }"
+            />
+          </template>
+        </oc-breadcrumb>
+        <portal-target v-if="showMobileNav" name="app.runtime.mobile.nav" />
+        <shares-navigation v-if="hasSharesNavigation" />
```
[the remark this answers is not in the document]
jacob-nv: Done

packages/web-pkg/src/composables/scrollTo/useScrollTo.ts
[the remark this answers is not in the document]
jacob-nv: Done. 
Then it could be replaced everywhere no?

packages/web-pkg/src/components/FilesViewWrapper.vue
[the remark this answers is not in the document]
jacob-nv: Done

packages/web-pkg/src/composables/scrollTo/useScrollTo.ts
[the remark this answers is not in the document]
JammingBen: Yes, all other usages of this composable should consume it from `web-pkg` now.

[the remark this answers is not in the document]
jacob-nv: Done

packages/web-pkg/src/composables/resourcesViewDefaults/useResourcesViewDefaults.ts
dschmidt: Aaah, the app itself needs the loaders? Grmpfmhl

JammingBen: Okay so turns out the backups app only uses the `useResourcesViewDefaults` composable to get `selectedResourcesIds`. Hence we suggest to leave the `useResourcesViewDefaults` and all the loaders in the files-app and only move the `useSelectedResources` composable to web-pkg.

jacob-nv: Done, but I could have missed unnecessary files that could be removed as well.

packages/web-pkg/src/components/AppBar/AppBar.vue
```diff
@@ -0,0 +1,341 @@
...
+          @item-dropped-breadcrumb="fileDroppedBreadcrumb"
+        >
+          <template #contextMenu>
+            <context-actions
+              :action-options="{ space, resources: breadcrumbsContextActionsItems.filter(Boolean) }"
+            />
+          </template>
+        </oc-breadcrumb>
+        <portal-target v-if="showMobileNav" name="app.runtime.mobile.nav" />
+        <slot v-if="hasSharesNavigation" name="navigation" />
```
JammingBen, line 32: I think we can remove `hasSharesNavigation` as property and replace it with a `computed` that checks, if the slot `navigation` is given.

packages/web-pkg/src/helpers/resources.ts
JammingBen, line 1: Hmm maybe it makes sense to have that file in `web-client`, alongside with the other files that include our `build...` methods. @dschmidt What do you think?

packages/web-pkg/src/helpers/resource/ancestorMetaData.ts
JammingBen, line 1: Please remove that file, we already have the types in `packages/web-pkg/src/helpers/resource/ancestorMetaData.ts`. Some imports need to be adjusted then of course 🙂

JammingBen (commented):
Getting there, nice 🥳 Some open tasks:

* Remove the moved files from the files-app
* Adjust all affected imports
* Move test `*.spec.ts` files accordingly
* Linter seems to be unhappy
* At the very end: remove the `web-app-backup` package from this PR. After it has been merged, we need to publish a new `web-pkg` package and then create a PR with the `web-app-backup` package in the CERN repo.

jacob-nv:
@JammingBen 
- moved all the relevant tests to web-pkg
- removed all the moved files from web-app-files
- changed imports to web-pkg

Some unit tests are still not happy.
Not really sure what to do with _web-app-files/tests/unit/components/AppBar/CreateAndUpload.spec.ts_
Getting error of _Cannot read properties of undefined (reading 'mockImplementation')_

Issue #9646, opened 2023-09-04
Search: scope select has wrong initial state


fschade:
### Steps to reproduce
1. open web
2. create some files and folders (folder: aa)
3. search for "aa" (everywhere is selected)

### Expected behaviour
the search should not send a scope 

### Actual behaviour
the search should not sends a scope as part of the request, you have to select 'in here' first and then switch back to 'everywhere' to get it working

PR #9653, opened 2023-09-05
[full-ci] - use KQL as default search query language

fschade:
## Description
We've introduced [KQL](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference) as our default query language.
Previously we used our own simple language for queries which is now replaced by kql.

`sample.tx* Tags:important Tags:report Content:annual*`

becomes

`name:"sample.tx*" AND tag:important AND tag:report AND content:"annual*"`

by default KQL uses `AND` as property restriction and the query described above can also be formulated as follows

`name:"sample.tx*" tag:important tag:report content:"annual*"`

More advanced syntax like grouping combined with boolean property restriction is supported too

`(name:"sample*" name:"*txt") tag:important OR tag:report content:"annual*"`

## Related Issue
* https://github.com/owncloud/ocis/pull/7212
* https://github.com/owncloud/ocis/issues/7042
* https://github.com/owncloud/ocis/issues/7179
* https://github.com/owncloud/ocis/issues/7114
* https://github.com/owncloud/ocis/pull/7043
* https://github.com/owncloud/web/issues/9636
* https://github.com/owncloud/web/issues/9646

## Motivation and Context
Defining a own query language is error prone and leads to a lot of documentation, the oCIS team decided in that [ADR](https://github.com/owncloud/ocis/blob/docs/ocis/adr/0020-file-search-query-language.md) to use KQL as our standard. A clear set of rules is set by MS and the dictionary fits well in our usecase.

## How Has This Been Tested?

- unit tests have been adjusted
- acceptance tests have been adjusted 

## Types of changes

- [ ] Technical debt
- [X] Tests

## Checklist:

- [X] Code changes
- [X] Unit tests added
- [X] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -264,8 +264,6 @@ export default defineComponent({
 
       const humanTagsParams = queryItemAsString(unref(tagParam))
       if (humanTagsParams) {
-        // in the legacy implementation there was the option to query by multiple tags, e.g. string.split("+")
-        // this was not in use(?) and is removed.
         add('tag', `"${humanTagsParams}"`)
 
         if (manuallyUpdateFilterChip && unref(tagFilter)) {
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -242,24 +242,31 @@ export default defineComponent({
     }
 
     const buildSearchTerm = (manuallyUpdateFilterChip = false) => {
-      let term = ''
-      if (unref(searchTerm)) {
-        term = `+Name:*${unref(searchTerm)}*`
+      let query = ''
+      const add = (k: string, v: string) => {
+        query = (query + ` ${k}:${v}`).trimStart()
       }
 
-      const fullTextQuery = queryItemAsString(unref(fullTextParam))
-      if (fullTextQuery) {
-        term = `+Content:"${unref(searchTerm)}"`
-      }
-      if (unref(scopeQuery) && unref(doUseScope) === 'true') {
-        term += ` scope:${unref(scopeQuery)}`
+      const humanSearchTerm = unref(searchTerm)
+      const isContentOnlySearch = queryItemAsString(unref(fullTextParam)) == 'true'
+
+      if (isContentOnlySearch && !!humanSearchTerm) {
+        add('content', `"${humanSearchTerm}"`)
+      } else if (!!humanSearchTerm) {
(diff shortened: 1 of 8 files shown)


packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -128,7 +127,7 @@ export default defineComponent({
     const isMobileWidth = inject<Ref<boolean>>('isMobileWidth')
     const scopeQueryValue = useRouteQuery('scope')
     const shareId = useRouteQuery('shareId')
-    const locationFilterId = ref(SearchLocationFilterConstants.inHere)
+    const locationFilterId = ref(SearchLocationFilterConstants.everywhere)
```
JanAckermann, line 130: 👍

packages/web-pkg/src/composables/search/constants.ts
```diff
@@ -1,5 +1,4 @@
 export abstract class SearchLocationFilterConstants {
-  static readonly defaultModeName: string = 'everywhere'
```
JanAckermann, line 2: 👍

packages/web-app-files/src/components/Search/List.vue
```diff
@@ -242,24 +242,31 @@ export default defineComponent({
...
       }
 
-      const tagsQuery = queryItemAsString(unref(tagParam))
-      if (tagsQuery) {
-        tagsQuery.split('+')?.forEach((tag) => {
-          term += ` +Tags:"${unref(tag)}"`
-        })
+      const humanTagsParams = queryItemAsString(unref(tagParam))
+      if (humanTagsParams) {
+        // in the legacy implementation there was the option to query by multiple tags, e.g. string.split("+")
```
JanAckermann, line 267: I don't think we need to comment legacy code removal in the source code, a comment in the pr would be sufficient

JanAckermann (changes requested):

packages/web-app-files/src/components/Search/List.vue
JammingBen: ^ Agreed. The reason it was coded this way is because initially, we wanted to enable filtering via multiple tags. But we need the conjunctive normal form for that to work.

fschade: conjunctive normal form works now, we can add multi tag in a follow up pr. needs slight ui changes too.

JanAckermann (approved):

PR #9669, opened 2023-09-06
SSE: Only retry on 401

lookacat:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/composables/sse/useServerSentEvents.ts
+++ b/packages/web-pkg/src/composables/sse/useServerSentEvents.ts
@@ -28,6 +28,7 @@ export const useServerSentEvents = (options: ServerSentEventsOptions) => {
       unref(ctrl).abort()
       throw new Error('Too many retries')
     }
+    let doRetry = false
     const setupSSE = async () => {
       retryCounter.value++
       try {
@@ -42,6 +43,7 @@ export const useServerSentEvents = (options: ServerSentEventsOptions) => {
           },
           async onopen(response) {
             if (response.status === 401) {
+              doRetry = true
               unref(ctrl).abort()
               return
             }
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -44,9 +46,17 @@ export const useServerSentEvents = (options: ServerSentEventsOptions) => {
             if (response.status === 401) {
               unref(ctrl).abort()
               return
+            } else if (response.status >= 500 || response.status === 404) {
+              throw new FatalError()
```
JammingBen, line 50: It still retries 5 times for me 🤔  That is coming from our implementation I guess? You could set the `retryCounter` to the max limit before throwing the error.

And we should reduce the amounts of retries IMO, 3 should be sufficient.

lookacat: sure

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -24,7 +26,7 @@ export const useServerSentEvents = (options: ServerSentEventsOptions) => {
     }
   )
   const setupServerSentEvents = () => {
-    if (unref(retryCounter) >= 5) {
+    if (unref(retryCounter) >= 3) {
```
JammingBen, line 29: Last nitpick: Please use a variable for the maximum amount of retries, since we're now using it here and down below 🙂

JammingBen (approved):

Issue #9665, opened 2023-09-06
Pagination Footer: Current Page is Highlighted Twice


tbsbdr:
### Steps to reproduce
1. Upload ex. 500 files, set pagination size to 20
2. click on "1" then "2" then "3" ... "5"
3. Current page is highlighted twice: "5" and "6" are highlighted

![screenshot_001635](https://github.com/owncloud/web/assets/26610733/83634198-cee3-4926-b887-fee8e253da0e)

### Expected behaviour
Only "5" should be highlighted as current page

### Actual behaviour
- Current Page is Highlighted Twice

lookacat:
it seems like an issue with :active or sth, if i want to inspect the wrongly highlighted page the highlighting is gone

PR #9678, opened 2023-09-08
Pagination, force unique indexes

jacob-nv:
## Description

This bug comes from indexes which are used as keys in the cycle to render pagination links. 
For example from issue, navigating through clicking on number pages will eventually keep the highlight stuck on 6 in 4,5 (current),6.
This will go on and happen if you navigate to the previous page for further previous pages. 
The second highlight comes from click/hover. If you experience this issue and click outside, the second highlight disappears. This happens because the "new" next page is on the same index as the previous one on which you clicked.

Solution: Mapping pages into objects with "value" (page text) and unique "key" for cycle

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9665

## Motivation and Context

Current Page is Highlighted Twice

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/bugfix-pagination-highlight
+++ b/changelog/unreleased/bugfix-pagination-highlight
@@ -1,0 +1,7 @@
+Bugfix: Pagination highlight
+
+An issue was found where pagination highlight was displayed after navigation in number of pages.
+Resolved by ensuring unique index for each page.
+
+https://github.com/owncloud/web/pull/9678
+https://github.com/owncloud/web/issues/9665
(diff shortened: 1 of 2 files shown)


JammingBen (commented):
Fix LGTM, could you please add a changelog item for this one? You can have a look how the other changelog items in the `changelog/unreleased` folder look like.

jacob-nv:
> `packages/design-system/src/components/OcPagination/OcPagination.spec.ts` fails because of the changes. I wonder why it didn't fail in the pipeline run before 🤔

This was bug on my part. Those pages were return in map only if conditioned passed.

JammingBen (approved):

JanAckermann (approved):

JammingBen:
Ah, I didn't know this, but it seems to have been fixed via CSS only here: https://github.com/owncloud/web/pull/9684 - means this is not necessary.

Please remember to add a comment in the issue when you're working on one @jacob-nv

PR #9680, opened 2023-09-08
[full-ci] Using auth token by default

ScharfViktor:
We've switched to using an api token, and now I make using it by default

### DIFF
--- a/tests/e2e/support/api/http.ts
+++ b/tests/e2e/support/api/http.ts
@@ -22,25 +22,14 @@ export const request = async ({
   const tokenEnvironment = TokenEnvironmentFactory(isKeycloakRequest ? 'keycloak' : null)
 
   const authHeader = {
-    Authorization: 'Basic ' + Buffer.from(user.id + ':' + user.password).toString('base64')
-  }
-
-  if (config.apiToken) {
-    authHeader.Authorization = 'Bearer ' + tokenEnvironment.getToken({ user }).accessToken
-  }
-
-  const basicHeader = {
-    'OCS-APIREQUEST': true as any,
-    ...(user.id && authHeader),
-    ...header
+    Authorization: 'Bearer ' + tokenEnvironment.getToken({ user }).accessToken
   }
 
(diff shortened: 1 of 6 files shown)


saw-jan (commented):
So, we don't support basic auth in e2e now?

saw-jan:
> Yes, I was doubting whether to keep it and decided to delete. Do you suggest to leave?

we are token or basic auths to setup the server (users/groups/resource) so how we do it should not matter. Token auth is needed for some servers where we cannot enable basic auth.

Using basic auth is straightforward and should be less overhead to the execution time than token auth. It's because, to access the token we do admin login via webUI which we don't have to do if we use basic auth.
So for this, even though how we do it doesn't matter, I would prefer token auth to be an option not mandatory.

I see that we are running tests with token auth in CI as well. But any reasons why we switched to token auth?

tests/e2e/support/api/http.ts
```diff
@@ -25,7 +25,7 @@ export const request = async ({
     Authorization: 'Basic ' + Buffer.from(user.id + ':' + user.password).toString('base64')
   }
 
-  if (config.apiToken) {
+  if (!config.basicAuth) {
     authHeader.Authorization = 'Bearer ' + tokenEnvironment.getToken({ user }).accessToken
   }
```
saw-jan, line 30: swap it like:
```js
const authHeader = {
    Authorization: 'Bearer ' + tokenEnvironment.getToken({ user }).accessToken
  }

  if (config.basicAuth) {
    authHeader.Authorization = 'Basic ' + Buffer.from(user.id + ':' + user.password).toString('base64')
  }
```

ScharfViktor: it doesn't work. if `BASIC_AUTH=true` token is undefined

saw-jan: Ah, when `BASIC_AUTH=true`, we don't set `accessToken`

saw-jan (approved):
LGTM :+1:

PR #9698, opened 2023-09-17
Add new ThemeSwitcher logic

pascalwengerter:
## Description
Overhaul of theme loading (used to be object key-value instead of an array of themes, which seems more desireable)-

## Motivation and Context
Supersede && finish some draft PRs, take advantage of upcoming major release

## Open tasks:
- Implement
- Test
- Document


packages/web-runtime/themes/owncloud/theme.json
```diff
@@ -5,12 +5,14 @@
...
-  "web": {
-    "default": {
+  "web": [
+    {
       "general": {
         "name": "ownCloud",
+        "title": "Light Theme",
         "slogan": "ownCloud – A safe home for all your data",
         "privacyUrl": "",
+        "isDark": false,
```
JanAckermann, line 15: can we name it darkMode or isDarkMode? that sounds better to me and gives more context

JanAckermann (dismissed):

packages/web-runtime/themes/owncloud/theme.json
dschmidt: Thinking about it again, it seems a bit odd because the theme *isn't* dark mode .. it is used *for* dark mode. So to me isDark seems actually more appropriate.

pascalwengerter: Yeah, not to be confused with e.g. https://vueuse.org/core/usePreferredDark 's `isDark`... - my `"isDark"` attribute should note whether a theme is created to be "used with dark mode". We could change it to `isDarkModeTheme`? Open for suggestions!

packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -301,3 +301,4 @@ export default {
   }
 }
 </style>
+@ownclouders/web-pkg/src/composables/piniaStores/theme
```
dschmidt, line 304: ?  😋

pascalwengerter: I knew that I had some weird miss-typing somewhere but couldn't find it anymore, thanks :D

packages/web-pkg/src/composables/piniaStores/theme.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    // TODO: Fix this by passing full theme?
+    // TODO: Discuss handling (former) default scenario
+    setAndApplyTheme(availableThemes.value.find((t) => t.general.name === currentThemeName.value))
+  }
+
+  const setAndApplyTheme = (theme: WebTheme) => {
+    currentTheme.value = theme
+
+    for (const param in currentTheme.value.designTokens.colorPalette) {
```
pascalwengerter, line 75: Not sure if modifying the HTML-document's stylesheet in a composable that's technically a store is something to be proud of, but having all the logic in one place sure should have its upside?

dschmidt: Fair enough 🤷‍♂️

packages/web-app-importer/package.json
```diff
@@ -16,6 +16,7 @@
   "peerDependencies": {
     "@ownclouders/design-system": "workspace:*",
     "@ownclouders/web-client": "workspace:*",
-    "@ownclouders/web-pkg": "workspace:*"
+    "@ownclouders/web-pkg": "workspace:*",
+    "pinia": "^2.1.7"
```
pascalwengerter, line 20: Perhaps we should re-export the required part (storeToRefs) from web-pkg?

packages/web-runtime/src/store/config.ts
```diff
@@ -70,9 +42,7 @@ const state = {
     openLinksWithDefaultApp: true,
     tokenStorageLocal: true,
     loginUrl: '',
-    privacyUrl: '',
-    imprintUrl: '',
-    accessDeniedHelpUrl: '',
+    logoutUrl: '', // Fixme: why was this missing before?
```
pascalwengerter, line 45: Was this a conscious omission?

packages/web-pkg/src/composables/piniaStores/theme.ts
```diff
@@ -0,0 +1,130 @@
...
+        availableThemes.value[0]
+    )
+  }
+
+  const setAndApplyTheme = (theme: WebTheme) => {
+    currentTheme.value = theme
+    const currentLocalStorageThemeName = useLocalStorage(themeStorageKey, theme.name)
+    currentLocalStorageThemeName.value = currentTheme.value.name
+
+    // TODO: Shouldn't we loop over all designTokens and set them?
```
pascalwengerter, line 107: @kulmann do you remember if we only applied the colorPalette on purpose instead of applying _all_ designTokens?

kulmann: don't remember it. but yes, we should loop over all design tokens... otherwise one theme would leak into the other, newly selected theme.

packages/web-runtime/src/App.vue
```diff
@@ -241,3 +245,4 @@ iframe {
   border: 0;
 }
 </style>
+@ownclouders/web-pkg/src/composables/piniaStores/theme
```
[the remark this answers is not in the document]
pascalwengerter: Howwwww?

changelog/unreleased/change-theme-handling
```diff
@@ -0,0 +1,19 @@
...
+If two themes are provided and one is a dark and one is a light mode one, there is a light/dark mode switch button in the AppBar.
+If more than two themes are provided or the two provided themes are not a dark and a light mode one respectively,
+the UI features a theme-switch button featuring a dropdown of the human-readable names of the available themes
+
+https://github.com/owncloud/web/issues/2404
+https://github.com/owncloud/web/issues/8424
+https://github.com/owncloud/web/issues/9403
+https://github.com/owncloud/web/issues/9885
+https://github.com/owncloud/web/issues/9939
+
```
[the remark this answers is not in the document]
pascalwengerter: Done

packages/web-pkg/src/composables/piniaStores/theme.ts
pascalwengerter: It technically still could since we're not resetting everything, but at least with the change I just pushed one isn't left wondering why the e.g. fontsize never changes even though a theme might re-define it 😋

packages/web-pkg/src/composables/piniaStores/theme.ts
```diff
@@ -0,0 +1,139 @@
...
+  isDark: boolean
+  name: string
+  loginPage?: LoginPage
+  logo?: Logo
+}
+
+interface WebThemeConfig {
+  defaults: ThemeDefaults
+  themes: WebTheme[]
+}
```
dschmidt, line 68: Why no zod schema? :trollface:

pascalwengerter: My plan was to parse it through zod in web-runtime/helpers/bootstrap.ts, but I guess reusing this schema throughout the place makes sense 😉 next step

packages/web-runtime/src/helpers/theme.ts
```diff
@@ -1,28 +1,40 @@
...
     const response = await fetch(location, { headers: { 'X-Request-ID': uuidV4() } })
     if (!response.ok) {
-      return defaults
+      console.error(`Failed to load theme '${location}', invalid response. Using default theme.`)
+      return defaultOwnCloudTheme
     }
     const theme = await response.json()
-    return { web: theme.web || theme, common: theme.common || {} }
+
+    if (WebThemeConfigSchema.safeParse(theme).success) {
```
pascalwengerter, line 30: @dschmidt is this roughly what you had in mind?

docs/getting-started.md
```diff
@@ -79,8 +79,6 @@ Depending on the backend you are using, there are sample config files provided i
 - `options.tokenStorageLocal` Specifies whether the access token will be stored in the local storage when set to `true` or in the session storage when set to `false`. If stored in the local storage, login state will be persisted across multiple browser tabs, means no additional logins are required. Defaults to `true`.
 - `options.loginUrl` Specifies the target URL to the login page. This is helpful when an external IdP is used. This option is disabled by default. Example URL like: 'https://www.myidp.com/login'.
 - `options.logoutUrl` Adds a link to the user's profile page to point him to an external page, where he can manage his session and devices. This is helpful when an external IdP is used. This option is disabled by default.
-- `options.imprintUrl` Specifies the target URL for the imprint link valid for the ocis instance in the account menu.
-- `options.privacyUrl` Specifies the target URL for the privacy link valid for the ocis instance in the account menu.
```
JammingBen, line 83: I guess the config for this has been removed intentionally? Because then we need to remove it in oCIS as well.

kulmann: Yes, we should. The way to go is the theme, so that all clients can use the URLs. Config is only for web.

packages/web-runtime/src/composables/head/useHead.ts
```diff
@@ -1,20 +1,26 @@
...
+import { storeToRefs } from 'pinia'
 
 export const useHead = ({ store }: { store: Store<any> }) => {
+  const themeStore = useThemeStore()
+  const { currentTheme } = storeToRefs(themeStore)
+
   _useHead(
     computed(() => {
-      const favicon = store.getters['configuration']?.currentTheme?.logo?.favicon
+      const favicon = computed(() => currentTheme.value.logo.favicon)
```
JammingBen, line 14: Computed inside computed? 🤔 Not sure how it behaves, but we should clean this probably.

packages/web-pkg/src/composables/piniaStores/theme.ts
```diff
@@ -0,0 +1,149 @@
+import merge from 'deepmerge'
```
kulmann, line 1: what's the benefit of using `deepmerge` instead of lodash `merge`?

packages/web-runtime/themes/owncloud/theme.json
```diff
@@ -2,198 +2,201 @@
...
-          "large-default": "",
-          "large-max": "",
-          "xlarge": ""
+  "clients": {
+    "android": {},
+    "desktop": {},
+    "ios": {},
+    "web": {
+      "defaults": {
+        "appBanner": {
```
kulmann, line 18: Please kill the `appBanner` in the default theme. Was intentionally disabled for the time being (feature is not production ready).

packages/web-pkg/src/composables/piniaStores/theme.ts
JammingBen: As discussed: let's leave it for now because we already ship the `deepmerge` dependency.

packages/web-app-files/src/views/FilesDrop.vue
```diff
@@ -89,6 +91,9 @@ export default defineComponent({
     const { getInternalSpace } = useGetMatchingSpace()
     useUpload({ uppyService })
 
+    const { currentTheme } = storeToRefs(themeStore)
+    const themeSlogan = computed(() => currentTheme.value.common.slogan)
```
JanAckermann, line 95: Does this need to be computed, as storeToRefs builds already a reactive object, or does this not apply to nested objects?

JammingBen: As far as I understand it: It's not necessary in this case. `currentTheme` is already a `Ref` because of `storeToRefs`.  Meaning the consuming code could go directly via `unref(currentTheme).common.slogan`.

It doesn't hurt to have a computed here though, I guess it's used as a "shortcut".

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -136,25 +136,18 @@ export default defineComponent({
...
-      )
-    })
-    const privacyUrl = computed(() => {
-      return (
-        store.getters.configuration.currentTheme.general?.privacyUrl ||
-        store.getters.configuration.options.privacyUrl
-      )
-    })
+
+    const imprintUrl = computed(() => themeStore.currentTheme.common.urls.imprint)
```
kulmann, line 148: For the record, we might need to re-introduce setting the imprint and privacy urls via config.json for short term deployment reasons.

kulmann (approved):
nice work 💪

PR #9702, opened 2023-09-18
Redesign ApplicationMenu

lookacat:
## Description

Redesign Application menu, make it look like actual apps and less like a dropdown item.

Screenshot (76)

Screenshot (77)

How it could look like if we have more apps:
Screenshot (78)

## Types of changes

- [x] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] Add unittests


packages/design-system/src/components/OcApplicationIcon/OcApplicationIcon.vue
```diff
@@ -0,0 +1,120 @@
...
+  setup(props) {
+    const updateComputed = ref(false)
+    const iconColor = computed(() => {
+      return 'rgba(255,255,255,0.7)'
+    })
+    const getGradient = (primary: string, secondary: string): string => {
+      return `linear-gradient(90deg, ${primary} 0%, ${secondary} 100%)`
+    }
+    const primaryColor = computed(() => {
+      unref(updateComputed)
```
lookacat, line 50: is needed for reactivity, because the OcDrop doesn't destroy and rebuild the applicationMenu (like an v-if would), so when there is a theme switch the primaryColor stays the same

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -21,49 +21,80 @@
...
+  setup() {
+    const appIcons = ref([])
+    const clientService = useClientService()
+    const { $gettext } = useGettext()
+
+    const applicationSwitcherLabel = computed(() => {
+      return $gettext('Application Switcher')
+    })
+    const updateAppIcons = () => {
+      unref(appIcons).forEach((appIcon) => {
```
lookacat, line 79: same as mentioned before, icons need to be updated when the drop opens (because there could be a theme switch)

tbsbdr:
@lookacat could you give me a hint how I can see your screenshot above via the PR in my browser? (dummy data) If I checkout the PR, everything looks like before - nothing changed.

grimmoc:
looks good to me :-)

lookacat:
@tbsbdr 
For dummy entries add in **line 183** in `packages/web-runtime/src/layouts/Application.vue` the following:
```
list.push({
  name: 'OnlyOffice',
  id: 'admin-settings',
  icon: 'file-ppt-2',
  color: '#FF3F09',
  isFileEditor: false,
  type: 'extension'
})
list.push({
  name: 'Calendar',
  id: 'admin-settings',
  icon: 'calendar-event',
  isFileEditor: false,
  type: 'extension'
})
```

packages/design-system/src/components/OcApplicationIcon/OcApplicationIcon.vue
JammingBen: Hmm that looks fishy. Also, I don't understand it 😄 `updateComputed` is being updated in `update()`, but this method is never beding called actually?

Edit: ahhhh now I see it, you're calling that method from the outside. Please avoid that 🙈  You could set a key on `oc-application-icon` which updates as soon as the drop opens. That way the component should be re-rendered.

packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -297,6 +297,9 @@ export default defineComponent({
   li a:focus {
     outline: auto 1px !important;
   }
+  .oc-card {
+    border-radius: 8px !important;
```
JammingBen, line 301:
```suggestion
    border-radius: var(--oc-space-small) !important;
```

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -21,49 +21,80 @@
       toggle="#_appSwitcherButton"
       mode="click"
       padding-size="small"
+      v-on:show-drop="updateAppIcons"
```
JammingBen, line 24: For consistency across the project:

```suggestion
      @show-drop="updateAppIcons"
```

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -21,49 +21,80 @@
...
-          >
-            <span class="icon-box">
-              <oc-icon :name="n.icon" variation="inherit" />
-            </span>
-            <span v-text="$gettext(n.title)" />
-            <oc-icon v-if="n.active" name="check" class="active-check" variation="inherit" />
-          </oc-button>
-        </li>
-      </oc-list>
+      <div style="display: block; position: relative">
```
JammingBen, line 27: Use proper classes please

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -21,49 +21,80 @@
...
+    })
+    const updateAppIcons = () => {
+      unref(appIcons).forEach((appIcon) => {
+        unref(appIcon).update()
+      })
+    }
+    const setClassicUIDefault = () => {
+      const url = urlJoin(configurationManager.serverUrl, '/index.php/apps/web/settings/default')
+      return clientService.httpAuthenticated.post(url, { isDefault: false })
+    }
```
JammingBen, line 86: While you're at it, the classic app stuff can be removed as we dropped oC10 support.

tbsbdr:
@lookacat known issue as you told me - is this going to be fixed?
screenshot_001706

lookacat:
@tbsbdr this is due to the dummy entries (if you look closley you will see they have the same ID) its not an issue, if we add new apps they will have an unique id and the selection will be fine :)

packages/design-system/src/helpers/colors.spec.ts
JammingBen, line 1: Unit tests, nice 😍

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -22,48 +22,62 @@
...
       required: false,
       default: () => []
     }
   },
-  computed: {
-    applicationSwitcherLabel() {
-      return this.$gettext('Application Switcher')
+  setup() {
+    const { $gettext } = useGettext()
+    const appIconKey = ref(false)
```
JammingBen, line 69: According to the Vue docs, child keys should be unique. You can have a look here e.g.: https://github.com/owncloud/web/blob/redesign-apps-menu/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue#L297

It simply creates a random prefix and uses it with the loop index in https://github.com/owncloud/web/blob/redesign-apps-menu/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue#L112.

packages/design-system/src/components/OcApplicationIcon/OcApplicationIcon.vue
```diff
@@ -0,0 +1,112 @@
...
+      max-width: 18px !important;
+      width: 18px !important;
+    }
+  }
+}
+</style>
+
+<docs>
+```js
+<oc-application-icon :count="21" />
```
JanAckermann, line 110: What is count ? 😯

packages/design-system/src/helpers/colors.spec.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  hexToRgb,
+  rgbToHex,
+  shadeColor,
```
JanAckermann, line 4: Can we name this to calculateShadeColor or simple getShadeColor ?

JanAckermann (changes requested):

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -22,48 +22,63 @@
...
-      return this.$gettext('Application Switcher')
+  setup() {
+    const { $gettext } = useGettext()
+    const appIconKey = ref('')
+
+    const applicationSwitcherLabel = computed(() => {
+      return $gettext('Application Switcher')
+    })
+    const updateAppIcons = () => {
+      appIconKey.value = uuid.v4().replaceAll('-', '')
```
JammingBen, line 76: It's not quite what I meant, now the key still isn't unique 😅 The idea was to give `oc-application-icon` a `key` like `${nid}-${actionKeySuffix}`. The `actionKeySuffix` will then be initialized with an empty ref and re-generated in this method.

lookacat: aah yes right

JammingBen (approved):

JanAckermann (approved):

PR #9724, opened 2023-09-22
[docs] Donot generate docs on stable branch

ScharfViktor:
trying to fix https://github.com/owncloud/ocis/issues/7323
we generate documentation twice: in the stable branch and in the master branch. This can lead to duplication 

- [x] don't generate docs twice in the stable and master branch. Same behavior in the `ocis` repo
- [x] set delete flag

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1197,6 +1197,7 @@ def documentation(ctx):
                         "pages_directory": "docs/",
                         "copy_contents": "true",
                         "target_branch": "docs",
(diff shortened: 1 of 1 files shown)


micbar (commented):
This is ok. we generate docs on the stable branch with the .drone.star from the stable branch.

micbar:
Not needed. We need to fix the docs pipelines in the source repos. See https://github.com/owncloud/file-picker/pull/257

Issue #9713, opened 2023-09-20
Sidebar preview gets confused for folders


JammingBen:
### Steps to reproduce
1. Upload an image to your personal space
2. Create a folder in your personal space
3. Open the details panels for the image
4. Select the folder resource while keeping the sidebar open -> the folder will wrongfully show the image preview

**Note:** Fix needs to target `stable-7.1`.

jacob-nv:
@JammingBen Could you please record short video reproducing the bug? Somehow I'm not able to reproduce it. Sorry for inconvenience

JammingBen:
Sure, here it is: 

https://github.com/owncloud/web/assets/50302941/c124e0b0-65ac-42d5-b1ba-4af242c7f659

JammingBen:
Fixed via https://github.com/owncloud/web/pull/9732.

PR #9732, opened 2023-09-26
Remove image preview when switching to folder

jacob-nv:
## Description

Image preview visible in detail after selecting folder.
Fixed by setting preview to undefined (default value for preview) if recognized as folder

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9713

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -208,6 +208,7 @@ import { formatDateFromHTTP, formatFileSize } from 'web-pkg/src/helpers'
 import { eventBus } from 'web-pkg/src/services/eventBus'
 import { SideBarEventTopics } from 'web-pkg/src/composables/sideBar'
 import { Resource, SpaceResource } from 'web-client'
+import { useTask } from 'vue-concurrency'
 import { useGettext } from 'vue3-gettext'
 import { getSharedAncestorRoute } from 'web-app-files/src/helpers/share'
 import { AncestorMetaData } from 'web-app-files/src/helpers/resource/ancestorMetaData'
@@ -233,7 +234,6 @@ export default defineComponent({
     const clientService = useClientService()
     const previewService = usePreviewService()
     const preview = ref(undefined)
-    const isPreviewLoading = ref(true)
 
     const directLink = computed(() => {
       return !unref(isPublicLinkContext)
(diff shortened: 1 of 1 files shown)


JanAckermann (changes requested):
would that also work with a watcher? like this code pattern more

jacob-nv:
> would that also work with a watcher? like this code pattern more

Pushed commit, did you mean something like that?

JanAckermann (approved):
👍

JammingBen (commented):
Hmm tbh I don't like that we lost the task now. We used a task here specifically for 2 reasons:

1. We don't need to care about setting a loading state manually like you now do in the PR.
2. The task is restartable. That means when a preview request is ongoing and a preview for another image is being requested, it automatically cancels ongoing tasks and only performs the latest one.

jacob-nv:
@jacob-nv @JanAckermann What is the final decision on this then? Keeping this or going back one commit?

JanAckermann (commented):
Hmmm couldn't you just watch the new resource and decide in the watcher then how to handle the preview? 

I just try to avoid having to much business logic in the tasks but it's rather a loose opinion

jacob-nv:
> Hmmm couldn't you just watch the new resource and decide in the watcher then how to handle the preview?
> 
> I just try to avoid having to much business logic in the tasks but it's rather a loose opinion

So.. if I revert commit I can do something like this.
Remove the check for folder
image
Use already existing watch and check if resource is folder. If it's folder it can just skip the whole task
image

JanAckermann:
Wonderful 👋

JammingBen:
That's still an issue, sorry, we need to set `preview.value` in the task. Imagine this scenario: You select an image, preview starts to load. While loading, you select a folder -> preview image will be set to `undefined`. After the image preview has finished loading, it will set the image for the folder. You can easily reproduce that by simulating a slow network connection in your browser.

jacob-nv:
> That's still an issue, sorry, we need to set `preview.value` in the task. Imagine this scenario: You select an image, preview starts to load. While loading, you select a folder -> preview image will be set to `undefined`. After the image preview has finished loading, it will set the image for the folder. You can easily reproduce that by simulating a slow network connection in your browser. Your first solution handled that perfectly.

Alright then, reverted commit.

JammingBen (approved):

PR #9749, opened 2023-09-28
web-pkg imports from `@ownclouders` namespace

dschmidt:
## Description
Changes all imports from `web-pkg/...` to `@ownclouders/web-pkg` (besides in the package itself where we now use relative imports).

## Types of changes

- [x] Technical debt
- [ ] Tests


JammingBen (approved):
I didn't go through all of it, but since this only changes imports and test files and the pipeline is green, let's go 🚀

Issue #9689, opened 2023-09-13
Notifications are not shown in Firefox


individual-it:
### Steps to reproduce
1. as Einstein create a folder
2. as Einstein share the folder with admin
3. as admin login into the webUI and look if you see any notifications

### Expected behaviour
Notifications are shown in all browsers

### Actual behaviour
Notifications are seen in the Chrome browser but not in Firefox

Chrome requests:
![grafik](https://github.com/owncloud/web/assets/2425577/b7bf6d6f-676b-40be-ab46-f7d0bd484ac9)
Chrome shows notifications:
![grafik](https://github.com/owncloud/web/assets/2425577/7bb4a153-7562-4880-b1fe-44320e406c03)

Firefox requests:
![grafik](https://github.com/owncloud/web/assets/2425577/d04506ca-14a4-456f-a97a-9f76d7c3a995)
Firefox does not show notifications:
![grafik](https://github.com/owncloud/web/assets/2425577/a47d0623-5c7e-4722-8def-9ec5d3c45314)

### Environment general
**Operating system**:
Ubuntu 22.04
**Backend (ownCloud Core or Infinite Scale)**:
ocis

### Environment ownCloud Infinite Scale
**Version:**
 4.0.0+f6593411f7
Web: 8.0.0-alpha.2

**Updated from an older Infinite Scale or fresh install:**
fresh

**Where did you install Infinite Scale from:**
github

### Client configuration
**Browser:**
FF 117.0 (64-Bit)
Chrome  116.0.5845.96
**Operating system:**
Ubuntu 22.04

#### Web server error log
```
{"level":"error","service":"proxy","error":"failed to verify access token: token is expired by 51.287000635s","authenticator":"oidc","path":"/ocs/v2.php/apps/notifications/api/v1/notifications/sse","time":"2023-09-13T11:42:25.288814301+05:45","message":"failed to authenticate the request"}
{"level":"error","service":"gateway","pkg":"rgrpc","traceid":"00000000000000000000000000000000","error":"internal error: gateway: error finding an auth provider for type: bearer","type":"bearer","time":"2023-09-13T11:42:31.274696868+05:45","message":"error getting auth provider client"}
2023/09/13 11:42:31 http: proxy error: context canceled
```

jacob-nv:
This seems to be backend issue
https://stackoverflow.com/questions/18800540/javascript-sse-not-working-with-firefox

JammingBen:
> This seems to be backend issue

[...]

PR #9771, opened 2023-10-06
Postprocessing state via SSE

JanAckermann:
## Description

For testing please add this to your config:
```
      OCIS_ASYNC_UPLOADS: "true"
      POSTPROCESSING_DELAY: "5s"
```

Otherwise the postprocessing will be done before the webdav request announces the resource with processing=false 

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9769
- Fixes https://github.com/owncloud/web/issues/9689
- 
## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -314,6 +285,16 @@ export default {
       }
     }
 
+    const onSSENotificationEvent = (event) => {
+      const notification = JSON.parse(event.data) as Notification
+      if (!notification) {
```
JanAckermann, line 290: || !notification.notification_id

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -323,19 +304,29 @@ export default {
...
     onUnmounted(() => {
-      if (unref(notificationsInterval)) {
-        clearInterval(unref(notificationsInterval))
+      if (unref(sseEnabled)) {
+        clientService.sseAuthenticated.removeEventListener(
+          MESSAGE_TYPE.NOTIFICATION,
+          onSSENotificationEvent
+        )
+      } else {
+        if (unref(notificationsInterval)) {
```
JanAckermann, line 327: Else if

pnpm-lock.yaml
```diff
@@ -947,12 +947,18 @@ importers:
...
         version: 6.3.3
+      '@microsoft/fetch-event-source':
+        specifier: ^2.0.1
+        version: 2.0.1
       '@ownclouders/web-client':
         specifier: workspace:*
         version: 'link:'
       axios:
         specifier: ^0.27.2 || ^1.0.0
         version: 1.4.0
```
JanAckermann, line 958: Rem

packages/web-client/src/sse.ts
```diff
@@ -5,8 +5,17 @@ export enum MESSAGE_TYPE {
   POSTPROCESSING_FINISHED = 'postprocessing-finished'
 }
 
+const myFetch = async (...args) => {
+  let [resource, config] = args
+  config.headers['Authorization'] = 'Bearer Pimmel'
```
dschmidt: 😂

changelog/unreleased/enhancement-handle-postprocessing-state-via-sse
```diff
@@ -0,0 +1,7 @@
+Enhancement: Handle postprocessing state via Server Sent Events
+
+We've added the functionality to hear for events from the server that announces the postprocessing state,
```
dschmidt, line 3:
```suggestion
We've added the functionality to listen for events from the server that update the postprocessing state,
```

packages/web-client/src/sse.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.onerror?.bind(this)(event)
+
+        /*
+         * Try to reconnect after 30 seconds plus random time in seconds.
+         * This prevents all clients try to reconnect concurrent on server error, to reduce load.
+         */
+        return 30000 + Math.floor(Math.random() * RECONNECT_RANDOM_OFFSET)
+      }
+    })
+  }
```
dschmidt, line 71: Not strictly necessary for this feature, but I would really like, if we tried to stick a bit closer to the behavior of the standard `EventSource`.
It should be pretty easy to update the `readyState` when events occur, i.e. set it to `CONNECTING` initially, to `OPEN` in openopen,  to `CLOSED` on close and error, etc.

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
```diff
@@ -1,79 +0,0 @@
...
-export const useServerSentEvents = (options: ServerSentEventsOptions) => {
-  const store = useStore()
-  const language = useGettext()
-  const accessToken = useAccessToken({ store })
-  const ctrl = ref(new AbortController())
-  const maxRetries = 3
-  const retryCounter = ref(0)
-
-  watch(
-    () => language.current,
```
dschmidt, line 26: Reconnecting with the updated language has been lost in this PR, no?

packages/web-runtime/src/services/auth/userManager.ts
```diff
@@ -156,6 +158,10 @@ export class UserManager extends OidcUserManager {
       })
       this.initializeOwnCloudSdk(accessToken)
 
+      if (this.store.getters.capabilities?.core?.['support-sse']) {
+        ;(this.clientService.sseAuthenticated as SSEAdapter).updateAccessToken(accessToken)
```
dschmidt, line 162: I still don't like this too much, but we tried to refactor it and it got complicated ... I'm ok as it seems to be the lesser evil for now

dschmidt (commented):
sexy af

JanAckermann:
> sexy af

You too !

packages/web-pkg/src/composables/sse/useServerSentEvents.ts
JanAckermann: Yep, I think we can do this as well. Will be another external call but ok for now.

packages/web-runtime/src/services/auth/userManager.ts
JanAckermann: Same here. I think we will need a tokenProvider and languageProvider in the client soon.

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -557,3 +558,34 @@ export const announceCustomStyles = ({
...
+export const registerSSEEventListeners = ({
+  store,
+  clientService
+}: {
+  store: Store<unknown>
+  clientService: ClientService
+}): void => {
+  clientService.sseAuthenticated.addEventListener(MESSAGE_TYPE.POSTPROCESSING_FINISHED, (msg) =>
+    onSSEProcessingFinishedEvent({ store, msg })
+  )
```
JammingBen, line 590: Is there a reason for the post-processing SSE registration + handling to live in the runtime? Isn't that logic bound to the files app?

JammingBen (approved):
Awesome stuff 😍 The re-connect after language change would be important, the rest looks super good!

packages/web-runtime/src/container/bootstrap.ts
JanAckermann: Yes we talked about this. We want to make this as global as possible. For example if you open a file in a editor (which is locked) it might be disabled to edit and will be indicated trough the app and indicates as editable as soon the locking-released event was fired.

Issue #9788, opened 2023-10-10
Check for `.space` folder when updating image or readme for a space


JammingBen:
When updating the image or readme file of a space, Web currently expects the `.space` folder to be present. Since the user can delete that folder we should make sure it exists and create it if not.

PR #9793, opened 2023-10-10
create .space folder if removed

jacob-nv:
## Description

If image is being updated, files are checked for '.space' existing. If the folder is missing, it will be created with default readme.md

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9788

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-app-files/src/composables/actions/spaces/useSpaceActionsUploadImage.ts
```diff
@@ -56,6 +56,14 @@ export const useSpaceActionsUploadImage = ({
       extraHeaders['X-OC-Mtime'] = '' + file.lastModified / 1000
     }
 
+    if (!store.getters['Files/files'].find((file) => file.name === '.space')) {
```
JammingBen, line 59: You need to query for the folder via API because it's not guaranteed that files are loaded in the store. That's the case e.g. on the Spaces overview (not inside a specific space).

It seems we already have such logic in other actions, see here for example: https://github.com/owncloud/web/blob/master/packages/web-pkg/src/composables/actions/files/useFileActionsSetImage.ts#L30C1-L30C1

JammingBen (changes requested):

packages/web-app-files/src/composables/actions/spaces/useSpaceActionsUploadImage.ts
jacob-nv: Got it. Updated, added update to drive, thus readme is displaying properly in space as well.

JammingBen (commented):
Nice 👍 Please add a changelog item (type enhancement), then this is good to go.

packages/web-app-files/src/composables/actions/spaces/useSpaceActionsUploadImage.ts
```diff
@@ -56,6 +57,38 @@ export const useSpaceActionsUploadImage = ({
...
+    try {
+      await clientService.webdav.getFileInfo(selectedSpace, { path: '.space' })
+    } catch (_) {
+      await clientService.webdav.createFolder(selectedSpace, { path: '.space' })
+      await clientService.webdav.putFileContents(selectedSpace, {
+        path: '.space/readme.md',
+        content: $gettext('Here you can add a description for this Space.')
+      })
+      const file = await clientService.webdav.getFileInfo(selectedSpace, {
+        path: '.space/readme.md'
```
JanAckermann, line 69: I wonder if we can outsource the code to a helper func, like: InitializeSpaceStructure and call it anywhere we need it, for example in the space overview? (while createing a space)

JanAckermann (changes requested):

packages/web-app-files/src/composables/actions/spaces/useSpaceActionsUploadImage.ts
jacob-nv: @JanAckermann Check last commit please. Do you mean something like that?

packages/web-pkg/src/composables/spaces/useCreateSpace.ts
```diff
@@ -10,36 +10,41 @@ export const useCreateSpace = () => {
...
       ...createdSpace,
       serverUrl: configurationManager.serverUrl
     })
 
-    await webdav.createFolder(spaceResource, { path: '.space' })
-    const markdown = await webdav.putFileContents(spaceResource, {
+    return await createDefaultConfigFolder(spaceResource)
+  }
+
+  const createDefaultConfigFolder = async (space: SpaceResource) => {
```
JanAckermann, line 23: I'd like to name it createDefaultMetaFolder, otherwise perfect 👍

JanAckermann (approved):

JammingBen:
Code LGTM, one unit test is failing though: https://drone.owncloud.com/owncloud/web/39514/9/5

JammingBen (approved):

Issue #9768, opened 2023-10-05
Provide Embed Mode for Location and File Picking


tbsbdr:
# Description

## User Stories

* > As a Chat-user I want to select files from within ownCloud in a familiar way so that I can use owncloud files in my chat conversations.
## Definition of ready
[ ] everybody needs to understand the value written in the user story
[ ] acceptance criteria has to be defined
[ ] all dependencies of the user story need to be identified
[ ] feature should be seen from an end user perspective
[ ] user story has to be estimated
[ ] story points need to be less then 20

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] code review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

tbsbdr:
cc @dj4oC @JammingBen @hodyroff

dschmidt:
I assume multi select is only supposed to be supported from a single folder, right?

dj4oC:
@AversaFTW please provide feedback

LukasHirt:
A technical question coming up (maybe @dschmidt can answer this).

> No themeswitcher (dark/light etc.)

Is there only the manual switch or is there also auto detection of system settings that would need to be taken care of?

JammingBen:
> Is there only the manual switch or is there also auto detection of system settings that would need to be taken care of?

Currently there is no auto detection although we already thought about implementing it. But that's not relevant for this iteration I'd say.

dschmidt:
> > Is there only the manual switch or is there also auto detection of system settings that would need to be taken care of?
> 
> Currently there is no auto detection although we already thought about implementing it. But that's not relevant for this iteration I'd say.

full ack. Not part of this iteration, but in the long run: 
We want auto detection via vueuse. #1373 introduces the concept of dark and light themes

LukasHirt:
Some notes regarding the messenger integration:

- upload needs to be hidden as well as that will be handled in the messenger
- in the create dropdown, only the new folder option should be available (having it as a direct action instead of dropdown makes more sense then)
- we need to be able to easily send a link to the resource into the chat (we thought about adding extra button to the bottom section next to select action @tbsbdr I would love to get your input on this)
- we need to be able to handle dynamic config due to different schools using the file picker (@JammingBen I would love to get your input on this)

[...]

PR #9818, opened 2023-10-16
feat: introduce mode config option

LukasHirt:
## Description

Introduce new config option to assert in which mode the UI should be rendered. There are two ways to set the mode, either by adding it into the `options` object in the config or via `mode` query param. The config option takes precedence when both are defined. To showcase the functionality of the config, hide elements in the embed mode in the top bar component.

## Related Issue

- refs #9768

## Motivation and Context

This is needed to have functioning embed mode. Providing both config and query approaches makes us future proof. Same applies to having the option as `mode?: string` instead of e.g. `embedded?: boolean` so that any other mode could be introduced in the future...

## How Has This Been Tested?

- test environment: local dev env
- test case 1: open https://localhost:9200?mode=embed
- ~test case 2: set `options.mode` to `embed` in config.json~ - does not work because of custom config not working (but unit tests are passing)

localhost_9200__mode=embed

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -5,7 +5,10 @@
     :aria-label="$gettext('Top bar')"
   >
     <div class="oc-topbar-left oc-flex oc-flex-middle oc-flex-between">
-      <applications-menu v-if="appMenuItems.length" :applications-list="appMenuItems" />
+      <applications-menu
+        v-if="appMenuItems.length && mode !== 'embed'"
```
dschmidt, line 9: Please add a composable that provides a `isEmbeddedMode` (or something like that) - I really don't want this magical untyped string comparison all over the code base

LukasHirt: Done, added `useEmbedMode` composable into pkg package.

dschmidt (approved):

JammingBen (approved):
> test case 2: set options.mode to embed in config.json

Does this work for you without setting the query param? Normally it shouldn't because the config is still missing in oCIS 🤔

LukasHirt:
> Does this work for you without setting the query param? Normally it shouldn't because the config is still missing in oCIS 🤔

Strange... I tried it once and it worked. I must have screwed up somehow my local env. Trying it again does not work indeed.

Anyway, I guess we can merge this still? The unit test for the config value is passing so once config is working in oCIS - it should work :shrug:

Issue #9799, opened 2023-10-12
Private link does not open in default app


tbsbdr:
### Steps to reproduce
1. Create `for-Albert.docx` and create a private link "Only invite ppl can view"
2. send the link via email to albert
3. Albert accepts the share
4. Albert clicks on the link in the email
5. Private Link does not open in default app

### Expected behaviour
- The private link should open in the default app (eg. Collabora - same behaviour as for public links)

### Actual behaviour
- Private Link does not open in default app

tbsbdr:
Labeling as bug as enduser experienced it as "broken". debatable, but I decided to take the enduser perspective. (also shows that the public link experience is quite well received)

tbsbdr:
@JammingBen wondering if there will occur any unwanted sideeffects if we change the behaviour of the private link to the behaviour of the public link. seeing the file-tree (location of the shared file) will still be possible if you select X-Close in the file control bar. so that should still work out...

JammingBen:
@tbsbdr Turns out this is a bit messy as it is currently... Internal and public links behave differently when it comes to opening files. When solving this issue, private links will resolve into the default app, whereas public links resolve into the file list and highlight the file. Should we align this and always resolve into the default app?

Issue #9776, opened 2023-10-06
No redirect to shares with me page while opening internal link in case auto_accept_share


ScharfViktor:
Precondition: 
- run ocis using `FRONTEND_AUTO_ACCEPT_SHARES: "true"`
- user `katherine` disables auto sharing: `curl --insecure -X POST -u katherine:gemini https://localhost:9200/api/v0/settings/values-save -d '{"value":{"accountUuid":"me","bundleId":"2a506de7-99bd-4f0d-994e-c38e72c28fd9","settingId":"ec3ed4a3-3946-4efc-8f9f-76d38b12d3a9","resource":{"type":"TYPE_USER"},"boolValue":false}}'`

Steps: 
- admin creates folder and shares to katherine and richard
- admin creates internal link to folder
- users open link

Expected: redirect to shares->shares with me 
Actual: if users have  `FRONTEND_AUTO_ACCEPT_SHARES: "true"` (in the feature -> default) redirect doesn't work

https://github.com/owncloud/web/assets/84779829/029dde6f-c3d2-4d89-9a20-bde150bfac8a

NOTE: please enablecode after fixing: 
https://github.com/owncloud/web/pull/9775/files#diff-51d1aa97f08184d1170eefe0765f26efa2b7f0f92bfa8079115e30eaf959218eR105-R112

and https://github.com/owncloud/web/pull/9775/files#diff-613727c41c6d01f3377cea44044287cc7f6948c87f3e2a2c4160405fbf14066eR22

PR #9821, opened 2023-10-17
fix: resolving internal links into share spaces and shared files

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/9776
- Fixes https://github.com/owncloud/web/issues/9799

## Types of changes

- [ ] Technical debt
- [ ] Tests


tests/e2e/cucumber/features/smoke/internalLink.feature
```diff
@@ -17,9 +17,6 @@ Feature: internal link share
     When "Alice" edits the public link named "Link" of resource "myfolder" changing role to "Invited people"
     And "Brian" opens the public link "Link"
     And "Brian" logs in from the internal link
-    
-    # comment out after fixing #9776
-    # And "Brian" opens shared-with-me page from the internal link
```
JammingBen, line 22: @ScharfViktor Just fyi I removed this because with this PR, the internal link directly opens the file editor. So the following steps are actually correct.

tests/e2e/cucumber/features/smoke/link.feature
```diff
@@ -103,11 +103,11 @@ Feature: link
       | resource  |
       | lorem.txt |
     When "Brian" opens the public link "textLink"
-    Then for "Brian" file "shareToBrian.txt" should be selected
+    And "Brian" closes the file viewer
```
kulmann, line 106: Hm, I get that this is somehow an implicit verification step of `being in a file viewer`... but reads super weird. What do you think about introducing a generic step for `Then "Brian" is in a file viewer` or the like? @ScharfViktor any hints here?

kulmann (changes requested):

tests/e2e/cucumber/features/smoke/link.feature
ScharfViktor: You are right, the `open public link` step does not contain any checking, and `close file viewer` just clicks the close button. 
need implement step that we can ensure that text editor was opened. A step should be implemented to ensure that the text editor has been opened.

added to my TODO list

packages/web-app-files/src/views/shares/SharedWithMe.vue
```diff
@@ -55,6 +56,7 @@ import { useGroupingSettings } from '@ownclouders/web-pkg'
 import SharesNavigation from 'web-app-files/src/components/AppBar/SharesNavigation.vue'
 import { useGettext } from 'vue3-gettext'
 import { useStore } from '@ownclouders/web-pkg'
+import { useOpenWithDefaultApp } from '../../composables/openWithDefaultApp'
```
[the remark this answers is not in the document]
JammingBen: Good point, fixed here for the `useOpenWithDefaultApp` composable. We need to be careful with the imports still, sometimes unit tests require these specific imports to be able to mock stuff and to avoid circular imports.

tests/e2e/cucumber/features/smoke/link.feature
kulmann: Not necessarily the text editor. I think for a first step it's already sufficient to have a check for any editor / viewer. That's already fulfilled if the close button in the top bar is present. So with the implementation that @JammingBen chose the tests are doing what they should... but they are hard to understand from just reading the Gherkin.

packages/web-app-files/src/composables/openWithDefaultApp/useOpenWithDefaultApp.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!resource || resource.isFolder) {
+      return
+    }
+
+    const fileActionsOptions = {
+      resources: [resource],
+      space: space
+    }
+
+    const defaultEditorAction = getDefaultEditorAction(fileActionsOptions)
```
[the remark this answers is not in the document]
dschmidt: apparently "composables" weren't chosens as name by accident ;-)

dschmidt (approved):

kulmann (approved):

PR #9826, opened 2023-10-19
[tests-only][full-ci] Bump ocis commit-id

saw-jan:
bump ocis commit id for tests

Fixes ocis crash

Part of https://github.com/owncloud/QA/issues/834


SwikritiT (approved):
LGTM if CI is also happy

Issue #9166, opened 2023-06-06
Link-copy file-list-action not working in Safari 16.5 (18615.2.9.11.4)


michaelstingl:
![CleanShot 2023-06-06 at 13 55 55](https://github.com/owncloud/web/assets/214010/2db4d9b8-1068-40af-b578-181c1b9a3758) | ![CleanShot 2023-06-06 at 13 59 43](https://github.com/owncloud/web/assets/214010/5032df65-4b82-4bb3-baf6-05588137386b)
-- | --

```
[Log]  ownCloud Web UI 7.0.0-rc.39  (index.html-12559239.mjs, line 7)
[Log]  Infinite Scale 3.0.0-rc.4+e0840b899 Community  (index.html-12559239.mjs, line 7)
```

@kulmann @tbsbdr

fschade:
@michaelstingl, fixed https://github.com/owncloud/web/pull/9832

JanAckermann:
Gonna check that

LukasHirt:
The quick link seems to be working now fine in the latest versions of Safari. There is still an issue when clicking on the "Copy link" button in the create link modal.

DeepDiver1975:
Closing as fixed. Copy-link now works in Safari.

The Safari `NotAllowedError` came from calling `navigator.clipboard.writeText()` after an async link-create request, which loses the user-gesture context WebKit requires. Current source routes all copy-link actions through the `useClipboard` composable, which uses the documented WebKit workaround — `navigator.clipboard.write([new ClipboardItem(...)])` — with a `@vueuse/core` legacy fallback:

- `packages/web-pkg/src/composables/clipboard/useClipboard.ts`
- consumers: `useFileActionsCreateLink.ts`, `useFileActionsCopyPermanentLink.ts`

The `CreateShortcutModal.vue` path that threw in the 2023-11 trace no longer makes a direct clipboard call. If you still hit this on a current release, please reopen with the Safari version and a fresh trace.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9832, opened 2023-10-19
fix: webkit quicklinks copy for webkit

fschade:
## Description
copying quicklings was broken on webkit based browsers, this is fixed now.

* https://developer.apple.com/forums/thread/691873
* https://webkit.org/blog/10855/

## Related Issue

- Fixeshttps://github.com/owncloud/web/issues/9166

## Motivation and Context
... hmmmm, no bugs, happy dev 🤣
ezgif-2-09128abecb

## How Has This Been Tested?
- unit tests

Screenshot 2023-10-19 at 15 25 10

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-pkg/src/helpers/share/link.ts
+++ b/packages/web-pkg/src/helpers/share/link.ts
@@ -22,6 +22,49 @@ export interface CreateQuicklink {
   ability: Ability
 }
 
+export const copyQuicklink = async (args: CreateQuicklink) => {
+  const { store, language } = args
+  const { $gettext } = language
+
+  // doCopy creates the requested link and copies the url to the clipboard,
+  // the copy action uses the clipboard // clipboardItem api to work around the webkit limitations.
+  //
+  // https://developer.apple.com/forums/thread/691873
+  //
+  // if those apis not available (or like in firefox behind dom.events.asyncClipboard.clipboardItem)
+  // it has a fallback to the vue-use implementation.
+  //
+  // https://webkit.org/blog/10855/
+  const doCopy = async () => {
+    if (typeof ClipboardItem && navigator && navigator.clipboard && navigator.clipboard.write) {
+      await navigator.clipboard.write([
+        new ClipboardItem({
+          'text/plain': createQuicklink(args).then(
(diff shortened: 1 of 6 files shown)


packages/web-pkg/src/helpers/share/link.ts
```diff
@@ -22,6 +22,49 @@ export interface CreateQuicklink {
...
+  // the copy action uses the clipboard // clipboardItem api to work around the webkit limitations.
+  //
+  // https://developer.apple.com/forums/thread/691873
+  //
+  // if those apis not available (or like in firefox behind dom.events.asyncClipboard.clipboardItem)
+  // it has a fallback to the vue-use implementation.
+  //
+  // https://webkit.org/blog/10855/
+  const doCopy = async () => {
+    if (typeof ClipboardItem && navigator && navigator.clipboard && navigator.clipboard.write) {
```
dschmidt, line 39:
```suggestion
    if (typeof ClipboardItem && navigator?.clipboard?.write) {
```

packages/web-pkg/src/helpers/share/link.ts
```diff
@@ -22,6 +22,49 @@ export interface CreateQuicklink {
...
+      await navigator.clipboard.write([
+        new ClipboardItem({
+          'text/plain': createQuicklink(args).then(
+            (link) => new Blob([link.url], { type: 'text/plain' })
+          )
```
dschmidt, line 44: Does this really accept a promise? seems weird

dschmidt (commented):
really nice!

packages/web-pkg/src/helpers/share/link.ts
fschade: https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/ClipboardItem

An Object with the MIME type as the key and data as the value. The data can be represented as a Blob, a String or a Promise which resolves to either a blob or string.

dschmidt (approved):

PR #9841, opened 2023-10-23
feat: add actions into the embed mode

LukasHirt:
## Description

Add Cancel, Share and Select action into the files app available only when the mode is set to embed. All actions are ported below the app container.

## Related Issue

- refs #9768

## Motivation and Context

These new actions allow external developers to consume resources or links from within the embedded web ui.

## How Has This Been Tested?

- test environment: run the embed mode locally
- test case 1: check for cancel event
- test case 2: select resources and check for select event
- test case 3: select resources and check for share event

localhost_9200_oidc-callback_code=2tkxvNouXXnbP5XMPktTt3BFC-bE3Z4O scope=profile%20email%20openid session_state=8840316b1888469a5680edb38092be53e07838ea1626d2291817adba119b8e1e gX01Vko_u01UXklokhxBaizwML7Ukr70bJ5vROXRycU state=64d3cfa3ba804

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/EmbedActions/EmbedActions.vue
+++ b/packages/web-app-files/src/components/EmbedActions/EmbedActions.vue
@@ -22,7 +22,7 @@
   </section>
 </template>
 
-<script setup lang="ts">
+<script lang="ts">
 import { computed } from 'vue'
 import {
   createQuicklink,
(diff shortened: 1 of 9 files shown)


LukasHirt:
@JammingBen @dschmidt ping for review 🙏

LukasHirt:
I had to update snapshots due to the change in files wrapper... the change is kept in a separate commit to spot it more easily

packages/web-app-files/src/components/EmbedActions/EmbedActions.vue
```diff
@@ -0,0 +1,142 @@
...
+      variation="inverse"
+      appearance="filled"
+      :disabled="areSelectActionsDisabled"
+      @click="emitSelect"
+      >{{ $gettext('Attach as copy') }}</oc-button
+    >
+  </section>
+</template>
+
+<script setup lang="ts">
```
JammingBen, line 25: Even though I really like the Vue setup script approach, we're not using it in Web (currently). I'm unsure how to feel about it in terms of consistency across the project. @dschmidt What are your thoughts?

LukasHirt:
The test failure is strange... the test passes when the portal in files wrapper is commented out but looking in the code, I do not see anything that would be causing recurring calls.

LukasHirt:
> The test failure is strange... the test passes when the portal in files wrapper is commented out but looking in the code, I do not see anything that would be causing recurring calls.

Found it together with Jannik... we need to stub out the portal component as it seems to cause troubles in tests. We are not testing portal itself in those tests so it is fine to stub it out.

packages/web-app-files/src/components/EmbedActions/EmbedActions.vue
LukasHirt: Rewrote it to use the setup function instead.

JammingBen (approved):
Nice 👍 

You probably need to stub `portal` in `GenericTrash.spec.ts` as well for the unit tests to go green.

LukasHirt:
> You probably need to stub portal in GenericTrash.spec.ts as well for the unit tests to go green.

You're right, thanks!... it should pass now

Issue #9842, opened 2023-10-23
Search text overlaid on the search location filter


ScharfViktor:
Step: 
- open https://ocis.ocis-wopi.latest.owncloud.works 
- put a large content to search line

Expect: search text doesn't overlaid to the location filter 

<img width="511" alt="Screenshot 2023-10-23 at 16 09 24" src="https://github.com/owncloud/web/assets/84779829/6987b666-bb6e-4358-9b0e-d19bb37336ad">

Actual:

https://github.com/owncloud/web/assets/84779829/f67d9ac0-c296-47fb-bfd0-e7ed292179bc

PR #9856, opened 2023-10-24
Search text overlaid on the search location filter fix

jacob-nv:
## Description

Background color set to SearchBarFilter component

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9842

image
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/design-system/src/components/OcSearchBar/OcSearchBar.vue
+++ b/packages/design-system/src/components/OcSearchBar/OcSearchBar.vue
@@ -62,7 +62,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, ref, watch } from 'vue'
+import { computed, defineComponent, ref, useSlots, watch } from 'vue'
 
 import OcButton from '../OcButton/OcButton.vue'
 import OcGrid from '../OcGrid/OcGrid.vue'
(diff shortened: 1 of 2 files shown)


JammingBen (changes requested):
While it works, I'm not the biggest fan of solving it like this. Reason being that text will be hidden behind the background color. So when you type in a very long search term, eventually the end of your input string is being hidden. I'd rather expect the text to "scroll" with your input, just like here:

https://github.com/owncloud/web/assets/50302941/7678d977-cd8b-4dca-915a-2b03145d645a

jacob-nv:
> While it works, I'm not the biggest fan of solving it like this. Reason being that text will be hidden behind the background color. So when you type in a very long search term, eventually the end of your input string is being hidden. I'd rather expect the text to "scroll" with your input, just like here:
> 
>  Bildschirmaufnahme.2023-10-25.um.07.26.39.mov

Got it, went with different padding from right side inside input based on slots

JammingBen (approved):

Issue #9245, opened 2023-06-15
Use reva token for auth


tbsbdr:
# Description

## User Stories

* > As an admin I want to control the session lifetime in the client so that I'm not dependent on the IAM.

## Definition of ready
[ ] everybody needs to understand the value written in the user story
[ ] acceptance criteria has to be defined
[ ] all dependencies of the user story need to be identified
[ ] feature should be seen from an end user perspective
[ ] user story has to be estimated
[ ] story points need to be less then 20

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] code review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

realtes to https://github.com/owncloud/web/pull/9159/commits/9c37d8cfa80a7fd3dafc50e14df8647ea1cf3680

dragotin:
Proposal how to implement: 

The extension system allows to register custom code in a post-sign-in-hook. There, the CERN code could be added to an extension that pulls the reva token and stores it in the browser.

Benefit for the product would for example be that an extension could be written that lists all logged in users on a dashboard. For that, the custom extension code would call an dashboard endpoint with the user  name that just signed in.

tbsbdr:
see same example at mattermost

tbsbdr:
@DeepDiver1975 this patch https://github.com/owncloud/web/commit/9c37d8cfa80a7fd3dafc50e14df8647ea1cf3680 from CERN uses the REVA token so that the session lifetime can be controlled via the client. Can you assess if this option would be a security issue for the product?  (Generally we wonder if this would be an option for the product.). thank you!

tbsbdr:
Thx! Fyi @hodyroff @labkode

kulmann:
Let me rephrase @tbsbdr question:

The web patch uses a kind of post login hook to use the just freshly created access token from the IdP to make an authenticated request to a reva endpoint which then generates a custom token (session life time is still controlled by the backend). This custom token is then used instead of the original access token  in all following requests from ownCloud Web. The original access token is being sent into nirvana.

Please check if you need to reevaluate based on this @DeepDiver1975 :) 

Regardless of the outcome, the possibility of a post login hook via extension system is a good idea.

DeepDiver1975:
[...]

PR #9871, opened 2023-10-26
[cern] feat: add cern reva auth token handling

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/9245

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/services/auth/userManager.ts
+++ b/packages/web-runtime/src/services/auth/userManager.ts
@@ -287,9 +287,9 @@ export class UserManager extends OidcUserManager {
   }
 
   // copied from upstream oidc-client-ts UserManager with CERN customization
-  protected async _signinEnd(url: string, verifySub?: string): Promise<User> {
+  protected async _signinEnd(url: string, verifySub?: string, ...args): Promise<User> {
     if (!this.configurationManager.options.isRunningOnEos) {
-      return super._signinEnd(url, verifySub)
+      return (super._signinEnd as any)(url, verifySub, ...args)
     }
 
     const logger = this._logger.create('_signinEnd')
@@ -313,7 +313,7 @@ export class UserManager extends OidcUserManager {
       Use that longer token in all calls to the backend (so, replace the default store token)
     */
     try {
-      console.log('CERNBox: loging successful, exchange sso token with reva token')
+      console.log('CERNBox: login successful, exchange sso token with reva token')
       const httpClient = this.clientService.httpAuthenticated
       const revaTokenReq = await httpClient.get('/ocs/v1.php/cloud/user')
       const revaToken = revaTokenReq.headers['x-access-token']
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/services/auth/userManager.ts
```diff
@@ -284,6 +286,52 @@ export class UserManager extends OidcUserManager {
     this.store.commit('SET_CAPABILITIES', capabilities)
   }
 
+  // copied from upstream oidc-client-ts UserManager with CERN customization
+  protected async _signinEnd(url: string, verifySub?: string): Promise<User> {
```
dschmidt, line 290: maybe we should use `...args` instead to be more resilient against api changes

dschmidt (approved):
Not that I like it... but it seems like the least annoying approach

packages/web-runtime/src/services/auth/userManager.ts
dschmidt: ... or rather: additionally

kulmann: good idea

packages/web-runtime/src/services/auth/userManager.ts
```diff
@@ -284,6 +286,52 @@ export class UserManager extends OidcUserManager {
...
+      }
+      logger.debug('current user matches user returned from signin')
+    }
+
+    /* CERNBox customization
+      Do a call to the backend, as this will reply with the internal reva token.
+      Use that longer token in all calls to the backend (so, replace the default store token)
+    */
+    try {
+      console.log('CERNBox: loging successful, exchange sso token with reva token')
```
kulmann, line 316:
```suggestion
      console.log('CERNBox: login successful, exchange sso token with reva token')
```

packages/web-runtime/src/services/auth/userManager.ts
```diff
@@ -284,6 +286,52 @@ export class UserManager extends OidcUserManager {
...
+        logger.debug(
+          'current user does not match user returned from signin. sub from signin:',
+          user.profile.sub
+        )
+        throw new ErrorResponse({ ...signinResponse, error: 'login_required' })
+      }
+      logger.debug('current user matches user returned from signin')
+    }
+
+    /* CERNBox customization
```
kulmann, line 311: Did you discuss adding a kind of `post login hook` extension entrypoint? It could receive the loaded user as param and then `do stuff` ™️  / modify the user, etc. I can think of other useful scenarios for this... but at the same time my assumption is, that it's too early for such an extension entrypoint.

dschmidt: As discussed: let's not do this. It's super-non-standard and I don't want an "official extension point". If you don't need to mess with the token itself or its expiration time, you can just use events that are triggered by the userManager...

Issue #9255, opened 2023-06-16
O365: Open Non "Personal" Documents in View Mode per Default


tbsbdr:
# Description

## User Stories

* > As a user who clicks on a presentation, document or spreadsheet outside my Personal folder, I want that the file gets opened in view-only-mode in O365 so that I don't apply changes accidentally.

## Value
- error prevention
- improve collaboration

## Acceptance Criteria
- applies only to documents that are stored outside the users Home Folder (= "Personal"), eg. Shares & Spaces
- click on presentation, document or spreadsheet opens O365 in view mode per default
- if the user wants to edit the file, he/she needs an extra click on "Edit" to switch to editing Mode 

**In view mode, switch to edit mode:**
![screenshot_001232](https://github.com/owncloud/web/assets/26610733/dcf3e0c4-f92f-4032-885c-2f4485b7afa8)

**In edit mode, switch to other modes:**
![246333658-7cfec590-de88-4904-a71b-1299e819d59c](https://github.com/owncloud/web/assets/26610733/d0f8f9d7-53bb-4c9e-a3ed-2cbb51d4adf5)

## Definition of ready
[ ] everybody needs to understand the value written in the user story
[ ] acceptance criteria has to be defined
[ ] all dependencies of the user story need to be identified
[ ] feature should be seen from an end user perspective
[ ] user story has to be estimated
[ ] story points need to be less then 20

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] code review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

relates to https://github.com/owncloud/web/pull/9159/commits/2894f0cbe9a2c3515be87eb9fb4ed679f52e400a

tbsbdr:
@kulmann makes sense for the product I'd say, do you agree?

kulmann:
> @kulmann makes sense for the product I'd say, do you agree?

For Shares, including public links, I agree. But for Spaces I'm not fully convinced. At least it's more complex than that. I believe the wish for exceptions like "immediate edit mode if I'm the file owner" or "immediate edit mode if I'm the space manager" would arise quickly...

JammingBen:
[...]

PR #9879, opened 2023-10-27
web-app-external: view mode

dschmidt:
## Description
Open non-personal files in "view" (readonly) mode when enabled via config: `editor.openAsPreview`.

Based on the logic implemented by CERN this currently is the case for all files in shares and public links.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9255

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -38,7 +38,11 @@ import {
   useRouteQuery,
   useStore
 } from '@ownclouders/web-pkg'
-import { isPublicSpaceResource, isShareSpaceResource } from '@ownclouders/web-client/src/helpers'
+import {
+  isProjectSpaceResource,
+  isPublicSpaceResource,
+  isShareSpaceResource
+} from '@ownclouders/web-client/src/helpers'
 
 export default defineComponent({
   name: 'ExternalApp',
@@ -151,7 +155,7 @@ export default defineComponent({
     }).restartable()
 
     const determineOpenAsPreview = (appName: string) => {
-      const openAsPreview = configurationManager.options.editors.openAsPreview
+      const openAsPreview = configurationManager.options.editor.openAsPreview
       return (
         openAsPreview === true || (Array.isArray(openAsPreview) && openAsPreview.includes(appName))
       )
@@ -191,7 +195,9 @@ export default defineComponent({
         let viewMode = props.isReadOnly ? 'view' : 'write'
         if (
           determineOpenAsPreview(unref(applicationName)) &&
-          (isShareSpaceResource(props.space) || isPublicSpaceResource(props.space))
+          (isShareSpaceResource(props.space) ||
+            isPublicSpaceResource(props.space) ||
+            isProjectSpaceResource(props.space))
         ) {
           viewMode = 'view'
         }
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -84,6 +84,13 @@ export default defineComponent({
 
     const loadAppUrl = useTask(function* (signal, viewMode: string) {
       try {
+        if (props.isReadOnly && viewMode === 'write') {
+          store.dispatch('showErrorMessage', {
+            title: $gettext('Cannot open file in edit mode as it is read-only')
+          })
+          return
+        }
+
         const fileId = props.resource.fileId
         const baseUrl = urlJoin(
           configurationManager.serverUrl,
@@ -146,10 +153,6 @@ export default defineComponent({
     const catchClickMicrosoftEdit = (event) => {
       try {
         if (JSON.parse(event.data)?.MessageId === 'UI_Edit') {
-          if (props.isReadOnly) {
-            console.error('Cannot switch to write mode as file is read-only')
-            return
-          }
           loadAppUrl.perform('write')
         }
       } catch (e) {}
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -33,6 +33,7 @@ import { urlJoin } from '@ownclouders/web-client/src/utils'
 import {
   isSameResource,
   queryItemAsString,
+  useConfigurationManager,
   useRequest,
   useRouteQuery,
   useStore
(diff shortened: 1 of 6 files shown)


packages/web-app-external/src/App.vue
```diff
@@ -132,12 +149,49 @@ export default defineComponent({
...
+        if (JSON.parse(event.data)?.MessageId === 'UI_Edit') {
+          loadAppUrl.perform('write')
+        }
+      } catch (e) {}
+    }
+    watch(
+      applicationName,
+      (newAppName, oldAppName) => {
+        console.log('appNames', newAppName, oldAppName)
+        if (newAppName === 'Office365' && newAppName !== oldAppName) {
```
dschmidt, line 164: can this be configured or is this hardcoded in the app provider? I really don't know ... 

@micbar @kulmann ?

packages/web-app-external/src/App.vue
```diff
@@ -132,12 +149,49 @@ export default defineComponent({
...
+          window.addEventListener('message', catchClickMicrosoftEdit)
+        } else {
+          window.removeEventListener('message', catchClickMicrosoftEdit)
+        }
+      },
+      {
+        immediate: true
+      }
+    )
```
dschmidt, line 173: It's not so nice to have this kind of specific handling here, but handling this as an extension seems overkill right now... it would be nice if all applications would emit the same kind of event, but that might be impossible especially with proprietary third party applications...

packages/web-app-external/src/App.vue
```diff
@@ -132,12 +149,49 @@ export default defineComponent({
...
+
     watch(
-      props.resource,
-      () => {
-        loadAppUrl.perform()
+      [props.resource],
+      ([newResource], [oldResource]) => {
+        let viewMode = 'write'
+        // FIXME: introduce config option (?)
+        const featureFlag = true
```
dschmidt, line 180: What should we name this option?

micbar:
Our product supports editing in a public link share. We must do this based on the share permissions

dschmidt:
> Our product supports editing in a public link share. We must do this based on the share permissions

This is not about forbidding it completely, it's just about opening files in shares/public links in read only mode by default - the user can then click edit and edit the file given they have the permission to do so.

micbar:
Hmm i would expect only one smart open action which always uses the highest possible permission.

dschmidt:
Yeah, so config option to make it opt in. 

Why I pinged you tho: is Office365 as wopi app name fixed or is that configurable?

micbar:
Configurable in the app provider

packages/web-app-external/src/App.vue
dschmidt: we need another config option for the wopi app name to check for

packages/web-app-external/src/App.vue
```diff
@@ -73,8 +83,15 @@ export default defineComponent({
       })
     }
 
-    const loadAppUrl = useTask(function* () {
+    const loadAppUrl = useTask(function* (signal, viewMode: string) {
       try {
+        if (props.isReadOnly && viewMode === 'write') {
```
lookacat, line 88: would it make sense to create an enum for the viewMode?

dschmidt: seems overkill to me tbh. In theory we could also support more modes which are defined by the backend... this is just some specific handling for an edge case

packages/web-app-external/src/App.vue
```diff
@@ -132,12 +150,54 @@ export default defineComponent({
...
+    )
+
+    watch(
+      [props.resource],
+      ([newResource], [oldResource]) => {
+        if (isSameResource(newResource, oldResource)) {
+          return
+        }
+
+        debugger
```
JanAckermann, line 189: Maybe not

JanAckermann (changes requested):

packages/web-app-external/src/App.vue
dschmidt: Hehe I fixed that already, maybe I forgot to push

packages/web-app-external/src/App.vue
kulmann: For the record (you already know long time, but I'd like to write it down): it's configurable in the app provider.

packages/web-app-external/src/App.vue
```diff
@@ -132,12 +150,54 @@ export default defineComponent({
...
+
+        let viewMode = props.isReadOnly ? 'view' : 'write'
+        if (
+          determineOpenAsPreview(unref(applicationName)) &&
+          (isShareSpaceResource(props.space) || isPublicSpaceResource(props.space))
```
kulmann, line 194: didn't we want to have this behaviour for project spaces as well?

packages/web-pkg/src/configuration/manager.ts
```diff
@@ -101,6 +101,16 @@ export class ConfigurationManager {
       'openLinksWithDefaultApp',
       get(options, 'openLinksWithDefaultApp', true)
     )
+
+    // when this setting is enabled, non-personal files i.e. files in shares, spaces or public links
+    // are opened in read only mode and the user needs another click to switch to edit mode
+    // it can be set to true/false or an array of wopi app names
```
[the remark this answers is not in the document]
JammingBen: I adjusted the text and added it to the dev docs. Apparently we already have editor configs available under the `editor` namespace, hence I decided to put it here as well.

kulmann (approved):

JanAckermann (approved):

packages/web-app-external/src/App.vue
```diff
@@ -132,12 +154,56 @@ export default defineComponent({
...
+    )
+
+    watch(
+      [props.resource],
+      ([newResource], [oldResource]) => {
+        if (isSameResource(newResource, oldResource)) {
+          return
+        }
+
+        debugger
```
ScharfViktor, line 193: @dschmidt you forgot to delete it 
Screenshot 2023-11-16 at 15 55 50

ScharfViktor: I'll drop it in my PR

dschmidt: doh 😂

Issue #9452, opened 2023-07-20
[cloud import] Dark theme does not apply to uppy dashboard


JammingBen:
The dark theme does not apply to the uppy dashboard which is used for the cloud import. We should use our color tokens here.

<img width="472" alt="image" src="https://github.com/owncloud/web/assets/50302941/09b66917-f79a-4c7e-acc2-4e6d6e4c4e6e">

dschmidt:
Especially the WebDav Formulars look pretty bad, we need to check whether we need to handle that in our upstream PR as well - somehow...

dschmidt:
... oh and uppy has a dark mode as well, maybe we can just use that. We would need to tie it to our own dark mode switch then

PR #9884, opened 2023-10-30
enable dark theme on uppy importer

JanAckermann:
## Description

image

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9452

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-importer/src/extensions.ts
+++ b/packages/web-app-importer/src/extensions.ts
@@ -52,7 +52,7 @@ export const extensions = ({ applicationConfig }: ApplicationSetupOptions) => {
 
   const handler = async () => {
     const currentThemeName = window.localStorage.getItem('oc_currentThemeName')
-    const renderDarkTheme = unref(currentThemeName) === 'default-dark'
+    const renderDarkTheme = currentThemeName === 'default-dark'
 
     const modal = {
       variation: 'passive',
--- a/packages/web-app-importer/src/extensions.ts
+++ b/packages/web-app-importer/src/extensions.ts
@@ -51,6 +51,9 @@ export const extensions = ({ applicationConfig }: ApplicationSetupOptions) => {
   })
 
   const handler = async () => {
+    const currentThemeName = window.localStorage.getItem('oc_currentThemeName')
+    const renderDarkTheme = unref(currentThemeName) === 'default-dark'
+
     const modal = {
       variation: 'passive',
       title: $gettext('Import files'),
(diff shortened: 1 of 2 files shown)


packages/web-app-importer/src/extensions.ts
```diff
@@ -51,6 +52,9 @@ export const extensions = ({ applicationConfig }: ApplicationSetupOptions) => {
   })
 
   const handler = async () => {
+    const currentThemeName = useLocalStorage('oc_currentThemeName', '')
+    const renderDarkTheme = unref(currentThemeName) === themeNameDark
```
JanAckermann, line 56: unable to make a computable out of that, not reactive

packages/web-app-importer/package.json
```diff
@@ -16,6 +16,7 @@
   "peerDependencies": {
     "@ownclouders/design-system": "workspace:*",
     "@ownclouders/web-client": "workspace:*",
-    "@ownclouders/web-pkg": "workspace:*"
+    "@ownclouders/web-pkg": "workspace:*",
+    "web-runtime": "workspace:*"
```
JammingBen, line 20: Please don't add the runtime as a dependency to other packages.

packages/web-app-importer/src/extensions.ts
JammingBen, line 56: I feel like we could add a composable in `web-pkg` for the light/dark-mode related things? It could return `currentTheme` as a `computed` and also export the theme names. Then you don't need to import from runtime.

JanAckermann: not possible currently due to fuckups

JammingBen (approved):

dschmidt (commented):
I would really prefer if we could


a) not hardcode the localStorage key in an app
I suggest a composable in web-pkg
b) not hardcode a specific theme name to check against
I suggest to add a property to the theme that says whether it's light or dark, as I commented in the original issue

JanAckermann:
@dschmidt I merge this and we could check together if we can build something together, ok? 

Because Jannik and me agreed that we don't want to kick up a shindy

dschmidt:
Yeah, go ahead. I didn't request changes for a reason 👍


But let's do it soon before we forget...

PR #9908, opened 2023-11-02
Shortcut url followup 

JanAckermann:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/tests/unit/composables/actions/files/useFileActionsOpenShortcut.spec.ts
+++ b/packages/web-pkg/tests/unit/composables/actions/files/useFileActionsOpenShortcut.spec.ts
@@ -108,7 +108,6 @@ function getWrapper({
     })
   }
 
-  // url contains xss code to test xss protection
   mocks.$clientService.webdav.getFileContents.mockResolvedValue(
     mock<GetFileContentsResponse>({
       body: getFileContentsValue


packages/web-pkg/src/components/CreateShortcutModal.vue
```diff
@@ -48,14 +68,48 @@ export default defineComponent({
...
     const inputUrl = ref('')
     const inputFilename = ref('')
+
+    const showDrop = computed(() => unref(inputUrl).trim())
+    const dropItemUrl = computed(() => {
+      let url = unref(inputUrl).trim()
+
+      if (
+        !(
+          ['h', 'ht', 'htt', 'http', 'http:', 'http:/', 'https', 'http:', 'https:/'].includes(
```
kulmann, line 81: what on earth is this? 🙈 :trollface:

JanAckermann: Please enter code below ⬇️

JanAckermann:
TBD:
- unit tests for createShortCutModal
- Fix create on enter not working
- Follow-up for search

packages/web-pkg/src/components/CreateShortcutModal.vue
```diff
@@ -48,14 +79,61 @@ export default defineComponent({
...
+
+    const dropItemUrlClicked = () => {
+      inputUrl.value = unref(dropItemUrl)
+      try {
+        let fileName = new URL(unref(dropItemUrl)).host
+        if (unref(files).some((f) => f.name === `${fileName}.url`)) {
+          fileName = resolveFileNameDuplicate(`${fileName}.url`, 'url', unref(files)).slice(0, -4)
+        }
+        inputFilename.value = fileName
+      } catch (e) {
```
JanAckermann, line 127: Unable to extract host, let the user do that

packages/web-pkg/src/composables/actions/files/useFileActionsOpenShortcut.ts
```diff
@@ -34,12 +36,20 @@ export const useFileActionsOpenShortcut = ({ store }: { store?: Store<any> } = {
...
+      let url = extractUrl(fileContents)
+
+      // Add protocol if missing
+      url = url.match(/^http[s]?:\/\//) ? url : `https://${url}`
 
       // Omit possible xss code
-      const sanitizedUrl = DOMPurify.sanitize(url, { USE_PROFILES: { html: true } })
+      url = DOMPurify.sanitize(url, { USE_PROFILES: { html: true } })
+
+      if (url.startsWith(configurationManger.serverUrl)) {
```
lookacat, line 47: will the serverUrl always be defined, i ask because if the serverUrl is an empty string it would enable any url to pass :)

lookacat (changes requested):

packages/web-pkg/src/composables/actions/files/useFileActionsOpenShortcut.ts
JanAckermann: always there

lookacat (approved):
LGTM 👍🏼

JanAckermann:
Converting to draft, adding search as well as it seems not to be that demanding as I thought first

JanAckermann:
- Search should not search url finds 
- Implement preview

packages/web-pkg/src/composables/actions/files/useFileActionsOpenShortcut.ts
JanAckermann: But it's not sever url 😑

JanAckermann:
@kulmann refactored search with @dschmidt, have phun with review 🤡

packages/web-pkg/src/components/CreateShortcutModal.vue
```diff
@@ -39,24 +39,14 @@
...
-                  :is-resource-clickable="false"
-                  :path-prefix="getPathPrefix(resource)"
-                  :is-path-displayed="true"
-                  :folder-link="getFolderLink(resource)"
-                  :parent-folder-link-icon-additional-attributes="
-                    getParentFolderLinkIconAdditionalAttributes(resource)
-                  "
-                  :parent-folder-name="getParentFolderName(resource)"
-                />
+                <ResourcePreview :search-result="value" :is-clickable="false"></ResourcePreview>
```
JammingBen, line 49: `resource-preview` (for some reason I can't use the code-suggestion feature)

packages/web-pkg/src/components/CreateShortcutModal.vue
```diff
@@ -210,11 +182,11 @@ export default defineComponent({
       } catch (_) {}
     }
 
-    const dropItemResourceClicked = (resource: Resource) => {
+    const dropItemResourceClicked = ({ data }) => {
```
JammingBen, line 185: Can you type the arg here, what does `data` look like?

packages/web-pkg/src/composables/search/useSearch.ts
```diff
@@ -0,0 +1,66 @@
...
+import { useConfigurationManager } from '../configuration'
+import { useStore } from '../store'
+import { useClientService } from '../clientService'
+import { isProjectSpaceResource } from '@ownclouders/web-client/src/helpers'
+
+export const useSearch = () => {
+  const store = useStore()
+  const configurationManager = useConfigurationManager()
+  const clientService = useClientService()
+  const areHiddenFilesShown = store.state.Files?.areHiddenFilesShown
```
JammingBen, line 14: This needs to be a computed to be reactive.

packages/web-pkg/src/composables/search/useSearch.ts
JammingBen, line 1: Nice! 🤩 

You can probably re-use most of the test cases from `List.spec.ts` for this composable.

packages/web-pkg/src/components/CreateShortcutModal.vue
dschmidt:
```suggestion
    const dropItemResourceClicked = ({ data }: SearchResultValue) => {
```

packages/web-pkg/src/components/CreateShortcutModal.vue
```diff
@@ -6,34 +6,94 @@
...
+                class="oc-width-1-1"
+                appearance="raw"
+                justify-content="left"
+                @click="dropItemUrlClicked"
+              >
+                <oc-icon name="external-link" />
+                <span v-text="dropItemUrl" />
+              </oc-button>
+            </li>
+            <li v-if="searchTask.isRunning" class="oc-p-xs oc-flex oc-flex-center">
```
kulmann, line 39: This and the next `<li>` should be wrapped into a `template` block with a `v-if` that checks if we're inside an authenticated context. Search doesn't work from within public links.

kulmann (changes requested):
Woah dude. The UX feels so flawless. 💪  Found one issue: we didn't think about public links during concepting... 🤦 at least the search part should be disabled (don't even make the requests!) if in a public link...

kulmann (commented):
One more UX thing... when you do a right click on a `.url` file, the action for following the url is sorted below the sharing options. I would have expected it at the very top (like every other file handler, e.g. `Open in TextEditor`). Not sure though, if that's easily possible in the current state of web. For me also ok if you just open a followup issue.

kulmann:
Aaaand another conceptual issue, but that's definitely something for a followup: https://github.com/owncloud/web/issues/9917

JanAckermann:
> One more UX thing... when you do a right click on a `.url` file, the action for following the url is sorted below the sharing options. I would have expected it at the very top (like every other file handler, e.g. `Open in TextEditor`). Not sure though, if that's easily possible in the current state of web. For me also ok if you just open a followup issue.

Not an issue to implement that ;)

kulmann (approved):
🥳

packages/web-pkg/src/components/Search/ResourcePreview.vue
```diff
@@ -89,9 +79,22 @@ export default defineComponent({
...
 
+    const additionalAttrs = computed(() => {
+      if (!props.isClickable) {
+        return {}
+      }
+
+      return {
+        isResourceClickable: true,
+        parentFolderLink: getParentFolderLink(unref(resource)),
+        onClick: resourceClicked,
```
lookacat, line 90: does this work? i think `v-bind` is only for attributes and `v-on` is for events

lookacat (commented):
one thing found

packages/web-pkg/src/components/Search/ResourcePreview.vue
JanAckermann: It works

PR #9918, opened 2023-11-07
fix: closing editors on search result page

JammingBen:
## Description
We noticed this error in the ocis pipeline: https://drone.owncloud.com/owncloud/ocis/28475/40/12. For some reason it didn't fail in Web...

## Related Issue

- Regression from https://github.com/owncloud/web/pull/9891.

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
+++ b/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
@@ -172,8 +172,8 @@ export default defineComponent({
             driveAliasAndItem
           },
           query: {
+            ...unref(currentRoute).query,
             fileId: id,
-            ...(unref(currentRoute).query?.app && { app: unref(currentRoute).query?.app }),
             contextRouteName: 'files-spaces-generic',
             contextRouteParams: { driveAliasAndItem: dirname(driveAliasAndItem) } as any
           }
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -186,6 +187,7 @@ export default defineComponent({
           driveAliasAndItem
         },
         query: {
+          ...unref(currentRoute).query,
           fileId: id,
           shareId: space.shareId,
           ...(unref(currentRoute).query?.app && { app: unref(currentRoute).query?.app }),
```
dschmidt, line 193: This here and above becomes unnecessary with your change, I'm wondering if this is really what we want or if this might cause issues somewhere, e.g. with the context route query params...

JammingBen: Good catch!

> I'm wondering if this is really what we want or if this might cause issues somewhere

You mean when re-setting all of the current route's query params? I decided against this initially. But apparently now other things break because params will get lost :/

dschmidt: Yeah, maybe we can filter out some... iirc we had issues  in some bizzare place when we had implemented it like this inititially, maybe login/logout...

dschmidt: Ah, misreading the diff.. issue happened in closeApp iirc. So if this is feally only overriding a few params but stays on the same page l otherwise, it might be fine.

JammingBen: It seems so, at least I didn't find any other issues for now.

dschmidt (approved):

PR #9926, opened 2023-11-09
Vite serve cern

diocas:
This allows us to run in debug mode in our config behind nginx (removes ssl which is handled by nginx, basically).

I've also deleted some code from your main config file... This code was not being used anywhere, but I don't know if you use it to check if accessing the config file succeeds while you build. Either way, you would notice it as soon as you launch web, as it will try to load the config either way, no?
If not, I can download our config by passing the env var with a diff url.

### DIFF
--- a/vite.cern.config.ts
+++ b/vite.cern.config.ts
@@ -19,6 +19,11 @@ export default defineConfig(async (args) => {
     config = _defineConfig
   }
 
+  config.server = {
+    port: 9201,
+    strictPort: true
+  }
+
   // collapsible table
   config.resolve.alias['design-system/src/components/OcTable/OcTable.vue'] = join(
     projectRootDir,
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -163,13 +163,6 @@ export default defineConfig(async ({ mode, command }) => {
   const configUrl =
     process.env.OWNCLOUD_WEB_CONFIG_URL || 'https://host.docker.internal:9200/config.json'
 
-  let proxiedServerUrl
-  if (command === 'serve') {
-    // determine server url
-    const configJson = await getJson(configUrl)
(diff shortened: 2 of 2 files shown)


vite.cern.config.ts
```diff
@@ -19,6 +19,13 @@ export default defineConfig(async (args) => {
     config = _defineConfig
   }
 
+  if (args.mode !== 'production') {
```
dschmidt, line 22: Why don't you check the `command` to be `serve`? 
Just as in the code that your rightfully removed

In the end I don't mind what you do in your config file :)

vite.config.ts
```diff
@@ -163,13 +163,6 @@ export default defineConfig(async ({ mode, command }) => {
   const configUrl =
     process.env.OWNCLOUD_WEB_CONFIG_URL || 'https://host.docker.internal:9200/config.json'
 
-  let proxiedServerUrl
-  if (command === 'serve') {
-    // determine server url
-    const configJson = await getJson(configUrl)
-    proxiedServerUrl = configJson.server
-  }
-
```
dschmidt, line 172: Good catch, it's a leftover from when we had to use the built in Vite proxy - but we got rid of that by now

dschmidt (approved):
cool, let me know if you still want to change something or if I should just merge it

vite.cern.config.ts
dschmidt: Ah, I see. You copied our logic and the server settings, just without the https settings. You can possibly get away without the mode check even. I think we only have it to avoid the dependency on the certs when running in build mode e.g. in CI. As you remove that you can probably apply your settings unconditionally

diocas: Removed the check. Feel free to merge.

Issue #9922, opened 2023-11-09
Enabling the password requirement for the link breaks the "invite people via link" feature


ScharfViktor:
related https://github.com/owncloud/ocis/issues/7682

## Steps: 
- run ocis with `OCIS_SHARING_PUBLIC_SHARE_MUST_HAVE_PASSWORD=true` (in my case It was as default)
- create folder and click by `copy link` button and set the required password for the link (by default we create link with viewer permission). That means user with role `user` cannot create link with 
- share folder to einstein 
- change link permission to `invited people`

## Expected
- `invited people` is on the list of roles for the link
-  If the link role gets downgraded from "anyone can..." to "Only Invited ...", password and exp. date gets removed from the link

## Current behaviour
no `invited people` in the list. You can only get this permission if you delete password for link (admin feature)
Note: using api we can create public link with permission `0` (invite people) without password: - is correct
![image](https://github.com/owncloud/web/assets/84779829/72b167dc-a0a5-4a3e-b255-90f5f7b1f087)

https://github.com/owncloud/web/assets/84779829/923f6974-5ec4-42eb-a1e8-6850fdcd95c7

kulmann:
@ScharfViktor is it possible for a non-admin to update the link to `permission: 0` and removing the password on an API level? Then we could solve it web-only.

tbsbdr:
update: added more details to expected behaviour in initial post of yours, @ScharfViktor

ScharfViktor:
> @ScharfViktor is it possible for a non-admin to update the link to `permission: 0` and removing the password on an API level? Then we could solve it web-only.

yes, non-admin can update link 
![Screenshot 2023-11-09 at 11 45 43](https://github.com/owncloud/web/assets/84779829/60bf73dc-7909-47f0-9813-b94793cce5ab)

kulmann:
@jacob-nv is this something you could take over?

jacob-nv:
> @jacob-nv is this something you could take over?

For sure

PR #9931, opened 2023-11-10
enabling "invite people" for password-protected folder/file

jacob-nv:
## Description
Enabling users to switch link permission to `invited only` for password protected folders/files.
While switching to `invited only`, the password and expiration date is removed

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9922

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -639,7 +646,7 @@ export default defineComponent({
         this.hasPublicLinkEditing,
         this.hasPublicLinkContribute,
         this.hasPublicLinkAliasSupport,
-        !!link.password,
+        false,
```
JanAckermann, line 649: does this always work? for example if you have password enforced for editable links only and switch permissions from existing link that only has view permissions to editable permissions, will the password required still pop up as it did before?

JanAckermann: Sure, please remove all the other env settings and set 

SHARING_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD
SHARING_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD

To true 

https://owncloud.dev/services/sharing/configuration/


This will let you create a standard readonly link but as soon you want to change to edit permissions the popup opens and asks you to set a password

jacob-nv: Got popup asking to setup password in case of "can upload", "can edit" and "secret file drop". Still able to change to "invited people" which will drop password. After dropping password, I'm only able to set to "can view" without popup

kulmann: If I read the code correctly you can remove the `hasPassword` param from the `LinkShareRoles.list` and `LinkShareRoles.filterByBitmask` functions entirely, right? If that's correct then please do so.

jacob-nv: @kulmann 
I was attempting to remove it from the start, but wasn't successful, quickest fix was setting false for it in specific instances.
If I would to test again removing it completely, how would this condition would be resolved?
`
      ...(hasAliasLinks && !hasPassword ? [linkRoleInternalFile, linkRoleInternalFolder] : []),
`
_packages/web-client/src/helpers/share/role.ts LinkShareRoles.list_

Don't have enough understanding of what and how alias link works and what it could affect

kulmann: > [...] how would this condition would be resolved? `...(hasAliasLinks && !hasPassword ? [linkRoleInternalFile, linkRoleInternalFolder] : []),` _packages/web-client/src/helpers/share/role.ts LinkShareRoles.list_

Alias links are what we now call `only invited people` or `internal links` in the ui. They don't hold permissions on their own, they just serve as a pointer to a file or folder (but in the same data structure like a public link). Meaning: the one who clicks on such a link needs something else which grants permission to the file/folder, i.e. a share or a space membership with certain privileges.

That being said, a password doesn't make sense at all for an alias link, because the receiver of the link needs to hold permissions to access the linked file/folder anyway. Don't know why we would ever hide the alias link role when a password has been set before. Removing the `&& !hasPassword` in fact seems to be the underlying bugfix, that you addressed by always setting it to `false`. Just remove the `&& !hasPassword` condition entirely please. And with that there are no other references to the `hasPassword` param anymore => you can get rid of the param.

kulmann (commented):
Code looks good to me now. Please add a `bugfix` scoped changelog item :-)

JammingBen (approved):

PR #9937, opened 2023-11-14
[test-only ]upgrade playwright

ScharfViktor:
I upgraded `@playwright/test` from `1.33.0` to `1.39.0`
we don't need more `playwright`  see here https://playwright.dev/docs/release-notes#%EF%B8%8F-breaking-changes-1

previously, browsers were automatically loaded into CI. Now we have to use a specific script. We only use chromium, so I'm loading only chromium into CI to reduce cache size
image
https://playwright.dev/docs/browsers#managing-browser-binaries

@d7oc I upgraded playwright what I hope solve your issue 
https://playwright.dev/docs/release-notes#-debian-12-bookworm-support


saw-jan (approved):
:+1:

Issue #5741, opened 2021-08-27
writing to locked files/folders give only a generic error message


individual-it:
### Steps to reproduce
1. lock a folder e.g. `curl -u admin:admin http://localhost/owncloud-core/remote.php/webdav/folder1 -X LOCK -d '<?xml version="1.0" encoding="UTF-8"?><d:lockinfo xmlns:d="DAV:"><d:lockscope><d:shared/></d:lockscope> </d:lockinfo>'`
2. try to upload a file into that locked folder via web

### Expected behaviour
Error message should say `Upload failed because the current folder is locked.`

### Actual behaviour
Message only says: `File upload failed…`

See also discussion in https://github.com/owncloud/web/pull/5728#pullrequestreview-739288168

PR #9940, opened 2023-11-14
Added locked error message

jacob-nv:
## Description

Added error message in failed attempt of uploading to locked folder

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5741

## Screenshots:
image

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


JammingBen (changes requested):
We already have some error handling in place here: https://github.com/owncloud/web/blob/master/packages/web-runtime/src/components/UploadInfo.vue#L630. Could you please add it there?

Also, please add a changelog item 🙂 

Edit: Pipeline seems to fail because of some unrelated change, rebasing with current master should fix it though.

jacob-nv:
> We already have some error handling in place here: https://github.com/owncloud/web/blob/master/packages/web-runtime/src/components/UploadInfo.vue#L630. Could you please add it there?
> 
> Also, please add a changelog item 🙂
> 
> Edit: Pipeline seems to fail because of some unrelated change, rebasing with current master should fix it though.

Not able to move the error message. I'm getting null responseCode there.

JammingBen:
I see. Can you maybe retrieve the error for the file via `this.errors[file.meta.uploadId]` and then check for status code `423`? I'd really like to have to error handling in one place here.

kulmann:
> I see. Can you maybe retrieve the error for the file via `this.errors[file.meta.uploadId]` and then check for status code `423`? I'd really like to have to error handling in one place here.

Fully agree UX-wise. The error in the title section of the upload info widget needs to stay short.

jacob-nv:
> > I see. Can you maybe retrieve the error for the file via `this.errors[file.meta.uploadId]` and then check for status code `423`? I'd really like to have to error handling in one place here.
> 
> Fully agree UX-wise. The error in the title section of the upload info widget needs to stay short.

What would convey the message properly?
```
Upload failed: Locked
Upload failed: Folder is locked
```

jacob-nv:
> I see. Can you maybe retrieve the error for the file via `this.errors[file.meta.uploadId]` and then check for status code `423`? I'd really like to have to error handling in one place here.

But you still have 2 places where you check for errors. Title of the window and then individual items.
I added error message for that "window" title, you are asking about error message for individual items. So should it be applied to both or only the items while window title has generic "Upload failed"?

jacob-nv:
Something like this? @JammingBen @kulmann 
image

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -226,9 +226,9 @@ export default defineComponent({
         return this.$gettext('Upload cancelled')
       }
       if (Object.keys(this.errors).length) {
-        const isLocked = Object.values(this.errors).some((error: any) => error.statusCode === 423)
+        const isLocked = Object.values(this.errors).some((error: any) => error?.statusCode === 423)
         if (isLocked) {
-          return this.$gettext('Upload failed because the current folder is locked.')
+          return this.$gettext('Upload failed: Locked')
```
JammingBen, line 231: I'd say it's sufficient to have the detailed error message down below, no need to handle the lock information here as well.

packages/web-runtime/src/components/UploadInfo.vue
```diff
@@ -626,6 +630,9 @@ export default defineComponent({
       }
 
       const errorObject = formatErrorMessageToObject(error.message)
+      if (this.errors[item.meta.uploadId]?.statusCode === 423) {
+        return this.$gettext('Upload folder is locked')
```
JammingBen, line 634: @kulmann @tbsbdr Suggestions for text? How about `The folder you're uploading to is locked`?

packages/web-runtime/src/components/UploadInfo.vue
jacob-nv: Got it

packages/web-runtime/src/components/UploadInfo.vue
kulmann: yep, works for me 👍

JammingBen (approved):
Nice 👍 

Needs another rebase with current master though, there have been quite a few pipeline changes recently.

PR #9967, opened 2023-11-16
Add support for custom components to modals

dschmidt:
## Description
This allows setting `customComponent` and `customComponentAttrs` on dispatched modals.
It's basically the same as `customContent` just more powerful and more idiomatic in Vue.

Moreover, with this change we don't need a "hosting" component for a custom dialog anymore (see `QuotaModal` or `CreateShortcutModal` which always need to be added to a template and cannot be triggered just "from code")

A file action handler can now look like this:

```typescript
  const handler = async ({ resources }: FileActionOptions) => {
    const resource = resources[0]
    const modal = {
      variation: 'danger',
      title: $gettext('Export'),
      cancelText: $gettext('Cancel'),
      confirmText: $gettext('Do it!'),
      message: $gettext('Are you sure you want to do it?'),
      hasInput: false,

      customComponent: DoDialog,
      customComponentAttrs: { resource },


      onCancel: () => store.dispatch('hideModal'),
      onConfirm: () =>
        loadingService.addTask(async () => {
          try {
            const { data } = await clientService.httpAuthenticated.post(`/what/ever`, { id: resource.id })
            console.log(data)
          } catch (error) {
            console.log(error)
            return store.dispatch('showErrorMessage', {
              title: $gettext('An error occurred'),
              desc: $gettext("Couldn't export file"),
              error
            })
          } finally {
            store.dispatch('hideModal')
          }
        })
    }

    store.dispatch('createModal', modal)
  }
```

with `DoDialog` looking like this:

```vue
<template>
  <div>
    <h1>File: {{ resource?.name }}</h1>
  </div>
</template>

<script lang="ts">
import { defineComponent, PropType } from 'vue'

  props: {
    // not used for now, but might come in handy later... 
    modal: {
      type: Object as PropType<any>,
      required: true
    },
    resource: {
      type: Object as PropType<Resource>,
      required: true
    }
  }
})
</script>
```

## Motivation and Context
I think  currently it's really cumbersome to handle custom content or custom modals in our application, as the components need to be "hosted" somewhere. For external applications it would need quite some trickery to make it work at all.
This approach makes it very easy for external applications to render arbitrary components in a modal (and for us as well).

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


JanAckermann (approved):
LGTM

JammingBen (approved):
Nice 👍

PR #9974, opened 2023-11-17
refactor: remove keyCode dependency

JammingBen:
## Description
Removes the keyCode dependency because the lib seems to be unmaintained and we need to patch it for that reason. Also, the `keyCode` property in the js event API is deprecated (see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). We only need it for the `vue-select` component (hopefully this will change soon).

## Types of changes

- [x] Technical debt
- [ ] Tests


dschmidt (approved):
Love it, thanks

kulmann (approved):
👍

PR #9993, opened 2023-11-17
Fix Vite warnings

dschmidt:
## Description
Switch our package type to module to avoid using vite's deprecated commonjs api. 
Required a bunch of other changes to enforce commonjs mode for other tools

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Our cucumber-js is very old, doesn't support esm configs and so needs to use a .cjs extension for the config (when package type is module), but it doesnt lookup that path. It is also so old, that it doesn't support --config cli option. 
We can get rid of the patch as soon as we update it. 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


kulmann (approved):
👌🏼

PR #10005, opened 2023-11-18
Do not open OcSelect Dropdown on focus

dschmidt:
## Description
Do not open the oc-select dropdown when oc-select gets focus, only on enter or mouse click.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
When vue-select gets focus, it immediately opens its dropdown. This is very odd in a formular, where the first item is a oc-select and the dropdown opens on page load (or when a modal opens). Apart from that it behaves differently from native select boxes and usually it's a good idea to stick as closely as possible to native behaviors if native components can't be used for whatever reason.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcSelect/OcSelect.vue
+++ b/packages/design-system/src/components/OcSelect/OcSelect.vue
@@ -311,8 +311,7 @@ export default defineComponent({
       dropdownEnabled.value = enabled
     }
 
-    const selectDropdownShouldOpen = (vueSelect) => {
-      const { noDrop, open, mutableLoading } = vueSelect
+    const selectDropdownShouldOpen = ({ noDrop, open, mutableLoading }) => {
       return !noDrop && open && !mutableLoading && unref(dropdownEnabled)
     }
 
--- a/packages/design-system/src/components/OcSelect/OcSelect.vue
+++ b/packages/design-system/src/components/OcSelect/OcSelect.vue
@@ -16,8 +16,13 @@
       :multiple="multiple"
       class="oc-select"
       style="background: transparent"
+      :dropdown-should-open="selectDropdownShouldOpen"
+      :map-keydown="selectMapKeydown"
       v-bind="additionalAttributes"
       @update:model-value="$emit('update:modelValue', $event)"
+      @click="onSelectClick()"
+      @search:blur="onSelectBlur()"
+      @keydown="onSelectKeyDown($event)"
     >
       <template #search="{ attributes, events }">
         <input class="vs__search" v-bind="attributes" @input="userInput" v-on="events" />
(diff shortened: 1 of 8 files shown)


packages/design-system/src/components/OcSelect/OcSelect.vue
```diff
@@ -296,7 +301,55 @@ export default defineComponent({
       setComboBoxAriaLabel()
     })
 
-    return { select, userInput }
+    const dropdownEnabled = ref(false)
+    const setDropdownEnabled = (enabled) => {
```
JanAckermann, line 305: Is this helper function really needed ?

JanAckermann (changes requested):

packages/design-system/src/components/OcSelect/OcSelect.vue
dschmidt: No, obviously not - but it makes it easier to debug issues because you can set breakpoints or log here.

JanAckermann (approved):

tests/e2e/support/objects/app-admin-settings/users/actions.ts
```diff
@@ -101,7 +101,7 @@ export const changeQuota = async (args: {
   value: string
 }): Promise<void> => {
   const { page, value, uuid } = args
-  await page.locator(quotaInput).fill(value)
+  await page.locator(quotaInput).pressSequentially(value)
   await page.locator(util.format(quotaValueDropDown, `${value} GB`)).click()
```
JanAckermann, line 105: Maybe keypress enter helps

dschmidt:
@ScharfViktor can you help with the e2e test failures? I've spent quite some time already, but I can't make the oc-select in the quotamodal in the batch edit mode work, no matter what I do.

It's in a superweird state where the text is light grey (as for a placeholder), no matter how often I click in playwright nothing happens and the dropdown doesn't show, that's why the test hangs...

ScharfViktor:
> @ScharfViktor can you help with the e2e test failures? I've spent quite some time already, but I can't make the oc-select in the quotamodal in the batch edit mode work, no matter what I do.
> 
> It's in a superweird state where the text is light grey (as for a placeholder), no matter how often I click in playwright nothing happens and the dropdown doesn't show, that's why the test hangs... when I click in the browser the test immediately continues, I'm clueless

Sure. I have look

JammingBen (approved):
Nice 👍

PR #10026, opened 2023-11-21
feat: handle FileReadyEvent for new files

dschmidt:
## Description

Handle new files and insert them into the current file listing, if they are added to the currently viewed folder. 

## Related Issue

- Kind of fixes: https://github.com/owncloud/web/issues/9782
This uses the FileReadyEvent emitted after post-processing. This is a nice start, but it's not fired for empty or copied files, probably not for restored from trash files, etc... 
- Depends on https://github.com/owncloud/ocis/pull/7770
## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

https://github.com/owncloud/web/assets/448487/07efb99c-014f-4a18-af5f-6c0ab2dcc437

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-client/src/webdav/listFilesById.ts
+++ b/packages/web-client/src/webdav/listFilesById.ts
@@ -1,0 +1,33 @@
+import { buildResource } from '../helpers/resource'
+import { DavProperties, DavPropertyValue } from './constants'
+
+import { urlJoin } from '../utils'
+import { DAV, buildAuthHeader } from './client'
+import { WebDavOptions } from './types'
+import { unref } from 'vue'
+import { ListFilesResult } from './listFiles'
+
+export type ListFilesByIdOptions = {
+  depth?: number
+  davProperties?: DavPropertyValue[]
+}
+
+export const ListFilesByIdFactory = (dav: DAV, { accessToken }: WebDavOptions) => {
+  return {
+    async listFilesById(
+      { fileId }: { fileId?: string },
+      { depth = 1, davProperties }: ListFilesByIdOptions = {}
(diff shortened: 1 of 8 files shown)


packages/web-client/src/webdav/listFilesById.ts
```diff
@@ -0,0 +1,33 @@
...
+      const webDavResources = await dav.propfind(webDavPath, {
+        depth,
+        headers,
+        properties: davProperties || DavProperties.Default
+      })
+      const resources = webDavResources.map(buildResource)
+      return { resource: resources[0], children: resources.slice(1) } as ListFilesResult
+    }
+  }
+}
```
dschmidt, line 33: `ListFiles` is already some of the worst spaghetti code in the whole project to handle different types of spaces, to autocorrect paths based on ids etc. 
Also I think it's semantically bad to have it work with a null space passed to it, as that usually might very well be a mistake.. that's why I went for a dedicated factory/method for this

JanAckermann:
LGTM and works, I'd still be happy about meaningfully Event types. Like "fileCreated" "fileRestored" and so on.

JanAckermann (approved):

kulmann (commented):
A successful upload does a propfind (and then internally recreates the file list) when the upload is finished. Can't we kill that code path as well now with your implementation? Would make sense in this PR, but also ok as a followup... pick what you like better.

kulmann (approved):

Issue #10027, opened 2023-11-21
Share type labels in filter are not showing


JammingBen:
## Steps to reproduce
1. as `admin`, share one file with `einstein` and one file with the group `users`
2. as `einstein`, go to the "Shared with me" page and open the share types filter -> the options don't have a label

<img width="540" alt="image" src="https://github.com/owncloud/web/assets/50302941/3f6c47c5-0716-4c88-88a2-ab1b5a094f9e">

dschmidt:
Maybe I broke that in #10003 ...?

tbsbdr:
I missed, that we have a share type-filter 😄
Trying to guess the userstory - what would it be?
"As a user I want to see only shares that had been shared with a group so that .." @kulmann @JammingBen what do you think? My assumption is that a share type filter is seldom helpful, do you think different?

JammingBen:
We primarily built it to filter for federated shares and thought "well, might be useful to filter group-shares as well" 😄

kulmann:
And federated shares are on the horizon, so let's keep it... 

But, I noticed that the share type filter is searchable. That's not needed, as it will only show 1-3 filter options (user/group/federated).

tbsbdr:
hm, I don't get why filtering by share-type is useful for regular users but ok, lets try

Issue #10013, opened 2023-11-20
[web] Add "Shared by" filter to "Shared with me"


tbsbdr:
# Description

## User Stories

* > As a user who is searching for a share I want to filter by "Shared by" so that I don't need to recall filenames. 

## Value
- usability
- improve file retrieval

## Acceptance Criteria
- Add "Shared by" filter to "Shared with me"
- The filter is a multiselect filter (same behaviour like in admin settinigs > users > groups / roles)
- filter items are people (first and last name)
- filtering happens client-side as that's not supported on the API level, yet

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

### Initial
![screenshot_001966](https://github.com/owncloud/web/assets/26610733/d21d675d-dedd-4fc5-98fa-7184849edcbf)

### Multiple selected items
(screenie shows groups - should be user names)
![screenshot_001963](https://github.com/owncloud/web/assets/26610733/062c2634-2154-4f1d-8d6c-b33ea90eb413)

PR #10029, opened 2023-11-21
feat: add shared by filter on 'Shared with me'-page

JammingBen:
## Description
Adds a "Shared by"-filter to the received shares on the "Shared with me"-page. Also fixes a bug where the share type labels were empty inside the "Share type"-filter and ensures to always show it.

## Screenshot
image

## Related Issue
- Fixes https://github.com/owncloud/web/issues/10013
- Fixes https://github.com/owncloud/web/issues/10027

## Types of changes

- [ ] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-app-files/src/views/shares/SharedWithMe.vue
+++ b/packages/web-app-files/src/views/shares/SharedWithMe.vue
@@ -20,7 +20,6 @@
             @toggle-filter="setAreHiddenFilesShown"
           />
           <item-filter
-            v-if="shareTypes.length > 1"
             :allow-multiple="true"
             :filter-label="$gettext('Share Type')"
             :filterable-attributes="['label']"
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/views/shares/SharedWithMe.vue
```diff
@@ -20,22 +20,40 @@
             @toggle-filter="setAreHiddenFilesShown"
           />
           <item-filter
-            v-if="shareTypes.length > 1"
```
kulmann, line 23: dev leftover that you removed the `v-if`?

JammingBen: No, that's a decision we (@tbsbdr and I) made on purpose because it's confusing that the filter just sometimes shows.

kulmann: Ok, thank you for clarifying 👍 makes sense!

JanAckermann (approved):

Issue #10015, opened 2023-11-20
[web] Disable App Banner


tbsbdr:
## Issue Description
The app banner is not in shape for a public release. It would need further improvements, which we can't schedule right now, as the priority for the feature is too low.
- detection for when to show the app banner does not work reliably, see https://github.com/owncloud/web/issues/9139#issuecomment-1779810169
- install link for users who don't have the app installed is missing, see https://github.com/owncloud/web/issues/9139#issuecomment-1779819629

**Please track times**

## Requirement / acceptance criterion
- Don't show the app banner if the `appBanner` key is missing or empty in the theme.json.
- adjust in dev docs how to enable/disable the feature

tbsbdr:
cc @michaelstingl

michaelstingl:
@tbsbdr aren't current config options sufficient? Isn't there an option to disable it?

- https://github.com/owncloud/web/issues/9836
- https://github.com/owncloud/web/pull/9872

tbsbdr:
naa, there is no option to disable it 😢

JanAckermann:
We could add an option alternatively?

michaelstingl:
Yes, config option please, and keep it enabled on our [continuous deployments](https://owncloud.dev/ocis/deployment/continuous_deployment/). This way, the teams can continue iterate on it and improve.

JanAckermann:
Let's discuss this in spring planning today

kulmann:
Thanks for feedback, I updated the issue description accordingly. We need to decide if we want to put disabling the app banner into the config.json where we would usually do something like that or into the theme.json where all the app banner info lives.

PR #10034, opened 2023-11-22
Disable AppBanner

lookacat:
## Description
See https://github.com/owncloud/web/issues/10015

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10015

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-pkg/src/components/AppBanner.vue
+++ b/packages/web-pkg/src/components/AppBanner.vue
@@ -1,6 +1,6 @@
 <template>
   <portal to="app.app-banner">
-    <div v-if="useAppBanner" class="app-banner hide-desktop" :hidden="isVisible === false">
+    <div v-if="isAppBannerAvailable" class="app-banner hide-desktop" :hidden="isVisible === false">
       <oc-button
         variation="brand"
         appearance="raw"
@@ -55,7 +55,7 @@ export default defineComponent({
     const store = useStore()
     const router = useRouter()
     const appBannerSettings = unref(store.getters.configuration.currentTheme.appBanner)
-    const useAppBanner = computed(
+    const isAppBannerAvailable = computed(
       () => appBannerSettings && Object.keys(appBannerSettings).length != 0
     )
     const appUrl = computed(() => {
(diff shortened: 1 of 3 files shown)


JanAckermann (approved):

JammingBen (changes requested):
Please add a sentence to the dev docs to explain that not setting this in the `theme.json` will result in disabling the banner. We already have a section somewhere on the theme page I believe.

packages/web-pkg/src/components/AppBanner.vue
```diff
@@ -55,6 +55,9 @@ export default defineComponent({
     const store = useStore()
     const router = useRouter()
     const appBannerSettings = unref(store.getters.configuration.currentTheme.appBanner)
+    const useAppBanner = computed(
```
pascalwengerter, line 58:
```suggestion
    const isAppBannerAvailable = computed(
```

Not sure if such a convention is used in this repo, but I sometimes find it helpful to have a hint of the variable being a boolean in the name already - and  `use[...]` usually hints at a composable?

packages/web-pkg/src/components/AppBanner.vue
```diff
@@ -55,6 +55,9 @@ export default defineComponent({
...
+      () => appBannerSettings && Object.keys(appBannerSettings).length != 0
```
pascalwengerter, line 59: Only checking for length could still mean someone gets a broken app banner since they configure one of the keys (e.g. title) but none of the others?

packages/web-pkg/src/components/AppBanner.vue
kulmann: Yep, fully agree on this

packages/web-pkg/src/components/AppBanner.vue
kulmann: For now we just want a way to disable the feature, because the current state of the implementation is not ready for a public release. Heavy changes expected afterwards. No need to go any extra miles here

JammingBen (approved):

PR #10046, opened 2023-11-23
fix: do not shallow copy items in ItemFilter

dschmidt:
## Description
Keep the original objects, so we can do identity checks against them and don't need to copy them for getters, so we can copy them and then copy them.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
This was a workaround to keep item order stable on search, but we can achieve the same without copying by just filtering the original array by the results.

Side effect: this should be more efficient

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


lookacat (approved):
LGTM, works as intended 👍🏼

Issue #10045, opened 2023-11-23
Files duplication on paste action


amrita-shrestha:
Build:https://drone.owncloud.com/owncloud/web/40558/9/15
```
1) Scenario: Copy and move resources in personal space (attempt 2) # tests/e2e/cucumber/features/smoke/copyMove.feature:6
   ✔ Before # tests/e2e/cucumber/environment/index.ts:43
   ✔ Given "Admin" creates following user using API # tests/e2e/cucumber/steps/api.ts:7
       | id    |
       | Alice |
   ✔ And "Alice" logs in # tests/e2e/cucumber/steps/ui/session.ts:30
   ✔ And "Alice" creates the following folders in personal space using API # tests/e2e/cucumber/steps/api.ts:56
       | name                  |
       | PARENTCopy1           |
       | PARENTCopy2           |
       | PARENTMove            |
       | PARENTCopy3           |
       | PARENTCopy4/Sub1/Sub2 |
       | PARENT                |
       | PARENT/Sub1/Sub       |
       | PARENT/Sub2           |
       | PARENT/Sub3           |
       | PARENT/Sub4           |
       | PARENT/Sub5           |
   ✔ And "Alice" creates the following files into personal space using API # tests/e2e/cucumber/steps/api.ts:101
       | pathToFile               | content                             |
       | PARENTCopy3/example1.txt | example text                        |
       | PARENTCopy3/example2.txt | example text                        |
       | KeyboardExample.txt      | copy with the help of keyboard      |
       | dragDrop.txt             | copy with the help of drag-drop     |
       | sidebar.txt              | copy with the help of sidebar panel |
       | PARENT/fileToCopy1.txt   | some content                        |
       | PARENT/fileToCopy2.txt   | some content                        |
       | PARENT/fileToCopy3.txt   | some content                        |
       | PARENT/fileToCopy4.txt   | some content                        |
       | PARENT/fileToCopy5.txt   | some content                        |
   ✔ And "Alice" opens the "files" app # tests/e2e/cucumber/steps/ui/application.ts:14
   ✔ When "Alice" copies the following resource using sidebar-panel # tests/e2e/cucumber/steps/ui/resources.ts:117
       | resource    | to          |
       | sidebar.txt | PARENTCopy2 |
   ✔ And "Alice" copies the following resource using dropdown-menu # tests/e2e/cucumber/steps/ui/resources.ts:117
       | resource                 | to          |
       | PARENTCopy3/example1.txt | PARENTCopy1 |
[...]

PR #10053, opened 2023-11-23
fix: sse resource upsert

JammingBen:
## Description
Resources that were upserted via the SSE post-processing event always had `/` as path, which is incorrect. It should be the current folder's path and the resource name concatenated.

This lead to some issues with file duplicates in the files list.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10045

## Types of changes

- [ ] Technical debt
- [ ] Tests


dschmidt (approved):
Thanks!

Issue #9817, opened 2023-10-15
Uploading an entire folder over the Web Interface freezes the frontend


marciofeld:
## Describe the bug

After dragging or choosing a local folder with lots of files to the Infinite Scale Web Interface to be uploaded, the frontend freezes. Sometimes, after waiting a big amount of minutes, it unfreezes the frontend and starts uploading the files, but most of the time it fails the upload.

![image](https://github.com/owncloud/ocis/assets/85581321/b315ec36-768a-493a-8285-90e8c7ecf2a6)

## Steps to reproduce

1. Log in to Infinite Scale using the Web Browser
2. Drag a local folder with lots of files and subfolders to your personal folder

## Expected behavior
The web frontend freezes. The browser may report it as an Unresponsive Page.

## Actual behavior
There are no server logs generated or frontend logs. When opening the browser task manager, it shows the Infinite Scale Tab Task consuming a large amount of memory.
No outbound network traffic was reported from the client or inbound network traffic to the server.

## Setup

Infinite Scale version 4.0.2 Docker Container running in a M5.Large AWS EC2 Instance, using AWS S3 Bucket as storage.
Using AWS ALB (Application Load Balancer) as the Reverse Proxy.

Please find the `docker-compose.yml` file below:

<details>
<p>

```
version: "3.7"

services:
  ocis:
    image: owncloud/ocis:4.0.2
    entrypoint:
      - /bin/sh
    # run ocis init to initialize a configuration file with random secrets
    # it will fail on subsequent runs, because the config file already exists
    # therefore we ignore the error and then start the ocis server
    command: ["-c", "ocis init || true; ocis server"]
    environment:
      OCIS_URL: <MY INFINITE SCALE URL>
      OCIS_LOG_LEVEL: "info"
      OCIS_LOG_COLOR: "false"
      PROXY_TLS: "false" # do not use SSL between Traefik and oCIS
      # activate s3ng storage driver
      STORAGE_USERS_DRIVER: s3ng
      STORAGE_SYSTEM_DRIVER: ocis # keep system data on ocis storage since this are only small files atm
      # s3ng specific settings
      STORAGE_USERS_S3NG_ENDPOINT: "<MY S3 ENDPOINT>"
      STORAGE_USERS_S3NG_REGION: "<MY AWS REGION>"
      STORAGE_USERS_S3NG_ACCESS_KEY: ${S3_ACCESS_KEY}
[...]

PR #10057, opened 2023-11-23
perf: speed up folder tree creation during upload

JammingBen:
## Description
Speeds up the folder tree creation during upload by running the process asynchronously and by saving up PROPFIND requests on nested folders.

In my test with a folder including 155 sub folders the time for creating those decreased from 20-30 seconds to 5-7. That's still quite a bit, but most likely as good as it can get when doing this on the client-side.

## Related Issue

- refs https://github.com/owncloud/web/issues/9817

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/changelog/unreleased/enhancement-upload-folder-tree-creation
+++ b/changelog/unreleased/enhancement-upload-folder-tree-creation
@@ -1,0 +1,6 @@
+Enhancement: Folder tree creation during upload
+
+The performance of the folder tree creation during upload has been improved.
+
+https://github.com/owncloud/web/pull/10057
+https://github.com/owncloud/web/issues/9817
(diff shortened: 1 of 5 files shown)


kulmann (approved):

JammingBen:
@kulmann I pinged you for a re-review since I adjusted the folder tree creation again to make it run asynchronously (thanks to help from @dschmidt ). In my tests we're now down from 20-30 seconds to 5-7 seconds for ~150 folders.

dschmidt (approved):
Love it! Great idea and impressive result.

Happy I could help a little :)

kulmann (approved):
🥇

Issue #9910, opened 2023-11-03
[web] Disable browser translations on content


kulmann:
# Description

## User Stories

* > As someone who uses the auto translate feature in a browser, I want content to not be translated, so that I don't see invalid data in my ownCloud.

## Value

- Auto translation doesn't mess with user generated content.
- Prevent errors caused via autotranslations 

## Acceptance Criteria
files app:
- file and folder names don't get auto translated
- space titles don't get auto translated
- space descriptions don't get auto translated
- user and group names (e.g. in sharing) don't get auto translated
- tags
- search results

user settings app:
- user, group and space names don't get auto translated

[probably more to be added] @tbsbdr 

## References
This issue has been around for quite some time in the oc10 classic ui and was fixed recently, see https://github.com/owncloud/core/issues/39946
The solution that was chosen was to forbid auto translation on the entire html document. For web this is not a good solution, as we don't have a good translation coverage, yet, and we will probably never have the same language support that auto translation offers. I'd be in favour of a solution where only the content gets protected, i.e. file names.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

dschmidt:
As mentioned in the core ticket, there is https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate

tbsbdr:
👍 updated to your initial comment

kulmann:
Lessons learned:
- doing this "right" (i.e. only disable browser translations on user generated content) requires quite the amount of code changes for little value and would need to be strictly followed by any developer (core and community), which is very unlikely to happen.
- at the same time, doing this "right" (as described above) does not go well with the interpolation feature on our translations (i.e. defining translation templates and inject content at runtime, which is a mix of translatable and non-translatable content)

[...]

PR #10069, opened 2023-11-27
disable browser translations for various elements

JanAckermann:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
+++ b/packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
@@ -7,6 +7,7 @@
         :data-key="index"
         :data-item-id="item.id"
         :aria-hidden="item.isTruncationPlaceholder"
+        :translate="index === 0 ? 'yes' : 'no'"
         :class="[
           'oc-breadcrumb-list-item',
           'oc-flex',
@@ -327,6 +328,7 @@ export default defineComponent({
 <style lang="scss">
 .oc-breadcrumb {
   overflow: visible;
+
   &-item-dragover {
     transition:
       background 0.06s,
@@ -337,6 +339,7 @@ export default defineComponent({
     box-shadow: 0 0 0 5px var(--oc-color-background-highlight);
     border-radius: 5px;
   }
+
   &-item-text {
     max-width: 200px;
     white-space: nowrap;
@@ -398,9 +401,8 @@ export default defineComponent({
     span:first-of-type {
       font-size: var(--oc-font-size-medium);
       color: var(--oc-color-text-default);
-      display: inline-block;
-      vertical-align: sub;
       line-height: normal;
+      display: flex;
     }
   }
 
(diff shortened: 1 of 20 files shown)


JanAckermann:
TBD:
- Tooltips (e.g shared with others)
- Settings app

packages/web-pkg/src/components/SideBar/Spaces/Details/SpaceDetails.vue
```diff
@@ -247,8 +246,12 @@ export default defineComponent({
     this.loadImageTask.perform(this)
   },
   methods: {
+    $gettext,
```
JanAckermann, line 249: Arghhhh 🤬

packages/web-pkg/src/components/SideBar/Spaces/Details/SpaceDetails.vue
```diff
@@ -52,12 +52,16 @@
...
         <th scope="col" class="oc-pr-s oc-font-semibold" v-text="$gettext('Subtitle')" />
-        <td v-text="resource.description" />
+        <td v-oc-browser-translate-off v-text="resource.description" />
       </tr>
       <tr>
         <th scope="col" class="oc-pr-s oc-font-semibold" v-text="$gettext('Manager')" />
         <td>
-          <span v-text="ownerUsernames" />
+          <span v-for="(owner, index) in owners" :key="index">
+            <span v-text="owner.displayName" />
```
kulmann, line 61: this one needs the `v-oc-browser-translate-off` directive as well, no?

JanAckermann: 🙈

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -255,6 +263,13 @@ export default defineComponent({
       hiddenItems.value.length >= 1 ? unref(hiddenItems)[unref(hiddenItems).length - 1] : { to: {} }
     )
 
+    const translate = (breadcrumbIndex: Number) => {
+      if (!props.representsFolderStructure) {
```
kulmann, line 267: Could you use the `isStaticNav` prop on the breadcrumb item instead so that you don't need to introduce the new component prop? Or check for absence of an `id` on the breadcrumb item?

packages/design-system/src/directives/OcTooltip.ts
```diff
@@ -58,8 +60,13 @@ const initOrUpdate = (el, { value = {} }: any) => {
     return
   }
 
+  if (modifiers['browser-translate-off']) {
+    value.content = `<span translate="no">${sanitizeHtml(value.content)}</span>`
```
kulmann, line 64: 👌

packages/web-app-admin-settings/src/components/AppTemplate.vue
```diff
@@ -10,6 +10,7 @@
               id="admin-settings-breadcrumb"
               class="oc-flex oc-flex-middle"
               :items="breadcrumbs"
+              :represents-folder-structure="false"
```
kulmann, line 13: I really hope that this prop is not needed, see other comment 🙈  But if we need it, I just realized here, that `folder-structure` doesn't fit as wording. Breadcrumbs can represent any hierarchy, e.g. `Users > Horst` if we'd decide to have a view for single users in the future.

kulmann (commented):
Could you have a look where we could add a dev docs section for this? Only makes sense to introduce the directive if we're strict with it in the future...

packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
JanAckermann: I'll check that :) is is unfortunately not sufficient

JanAckermann:
> Could you have a look where we could add a dev docs section for this? Only makes sense to introduce the directive if we're strict with it in the future...

Sure I also had the idea to give a short introduction in the team, because we need to obey this in future development

kulmann:
Collectively decided to not do this. Thank you @JanAckermann for the work and brain you put into this! ❤️  See https://github.com/owncloud/web/issues/9910#issuecomment-1832027449 for the rationale.

Issue #10074, opened 2023-11-27
[web] update permissions check on share as link embed action


LukasHirt:
## Is your feature request related to a problem? Please describe.

The share as link embed action uses old permissions check which needs to be changed to reflect more up to date one.

## Describe the solution you'd like

There is a `canCreateLinks` computed prop which needs to be used. While we're at it, we can hide the button when it is false instead of disabling the button.

PR #10076, opened 2023-11-27
feat: show modal when creating links in embed mode

JammingBen:
## Description
Adds a modal that pops up after clicking "Share links" in embed mode where the user can specify the props of the link(s) they want to create: password, role, name. It also changes the behavior so there will be always new links created instead of re-using existing links.

A few things had to be refactored course of this.

* moved the link role dropdown to a re-usable component `LinkRoleDropdown` (https://github.com/owncloud/web/pull/10079) (actually this was not needed in the end, but doesn't hurt)
* moved the expiration rules (former "expirationDate", although they define rules IMO, hence the rename) to a helper method `getExpirationRules` (https://github.com/owncloud/web/pull/10080)
* created new type and capability-composable for `files_sharing.public.password.enforced_for` (https://github.com/owncloud/web/pull/10081)

Minor other changes:

* Renamed `Create a new folder` to `New folder`
* Improved look of `OcRadio` component, added types for the `modelValue`

## Related Issue

- refs https://github.com/owncloud/web/issues/9768
- fixes https://github.com/owncloud/web/issues/10074

## How to test

- enable embed mode via query param `embed=true`
- select one or more resources
- Click on "Share links" at the bottom right of your screen

## Screenshot
image

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -0,0 +1,427 @@
...
+            succeeded.map(({ value }) => value.url)
+          )
+        }
+      }
+
+      const failed = result.filter(({ status }) => status === 'rejected')
+      if (failed.length) {
+        failed.forEach((error) => {
+          console.error(error)
+          store.dispatch('showErrorMessage', {
```
kulmann, line 326: a bit weird to have a generic, pluralized message for each and every error individually. 😅

PR #10078, opened 2023-11-28
feat: decouple creating links from the highlighted file

JammingBen:
## Description
Previously, creating links was coupled to the highlighted file in some ways. As a result, share types were only updated for the highlighted file and the determination whether the link is direct or indirect was done by checking the highlighted file. This is bad because the highlighted file is always the first selected file, though we want to be able to create links for multiple selected files in the future.

This change retrieves the actual file from the store instead of using the highlighted file.

## Types of changes

- [x] Technical debt
- [ ] Tests


dschmidt (approved):

JanAckermann (approved):
😅💪

PR #10079, opened 2023-11-28
refactor: move link role dropdown to a re-usable component

JammingBen:
## Description
Works towards https://github.com/owncloud/web/pull/10076

## Types of changes

- [x] Technical debt
- [ ] Tests


packages/web-app-files/src/components/SideBar/Shares/Links/DetailsAndEdit.vue
```diff
@@ -1,66 +1,12 @@
...
-              </span>
-            </oc-button>
-          </li>
-        </oc-list>
-      </oc-drop>
+      <link-role-dropdown
+        v-model="currentLinkRole"
+        :available-role-options="availableRoleOptions"
+        drop-offset="0"
+        @update:model-value="updateSelectedRole"
```
dschmidt, line 8: I know we do this sometimes, but I think it's a bit ugly to use v-model and update handlers together as you never know which one fires first.. so maybe set the new value in updateSelectedRole and use `:model-value`?

dschmidt (approved):

packages/web-app-files/src/components/SideBar/Shares/Links/DetailsAndEdit.vue
JammingBen: Yep I agree, that's cleaner 👍

PR #10080, opened 2023-11-28
refactor: move link expiration rules to a helper method

JammingBen:
## Description
This is necessary because we need the method in more than one place in the future. Also, previously this was named "expiration date", although it defines rules for setting an expiration date for a link. Hence the rename to "expiration rules".

Works towards https://github.com/owncloud/web/pull/10076

## Types of changes

- [x] Technical debt
- [ ] Tests


packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -219,6 +222,8 @@ export default defineComponent({
       )
     }
 
+    const expirationRules = computed(() => getExpirationRules({ store, currentLanguage }))
```
kulmann, line 225: Why did you decide to create a helper instead of a composable if you're only ever using it in this `computed` here?

JammingBen: Mainly to keep all link helpers in one place. IMO in the long term all these helpers should be one (or more) composable(s)... but it can't be done yet because of our pesky quick actions 😅 

But I could start creating a composable for this now and we can transition with time?

kulmann: Yep, that would be cool. Start the composable and basically just move this one `computed` from here into the composable.

kulmann (approved):
👍

Issue #9970, opened 2023-11-16
Search filter "in here" only applies together with search term


kulmann:
## Describe the bug
The search filter "in here" from the search input field only applies together with a search term, not with empty search term but other filters.

## Steps to reproduce
1. Have two "test.md" files, one in your personal space and one in a project space
2. Navigate into the project space
3. Select "in here" in the search input field and trigger the search (enter / click on the magnifier icon)
4. Select a media type filter for "Documents" (which includes markdown)

## Expected behavior
Search result only shows documents from the project space

## Actual behavior
Search result also shows the "test.md" file from your personal space. Adding "test" as a search term successfully adds the "in here" location filter as well.

jacob-nv:
Any ideas how to go about this?
Having empty query removes "scope", which is why empty search with "in here" searches everywhere.
Tried forcing the search with empty query, but that throws error from WebDAV

lookacat:
@kulmann i think then we either can't fix it or we need a backend ticket

kulmann:
Correct, I'd say this is a backend bug. The only thing to do in the frontend would be to remove this line:
https://github.com/owncloud/web/blob/22ea1b4484b28302a542ccc08403e2570c3e350b/packages/web-app-search/src/portals/SearchBar.vue#L223


But then this backend issue happens:
https://github.com/owncloud/ocis/issues/7812

PR #10092, opened 2023-11-29
Allowing in-here search with empty query

jacob-nv:
## Description

Requirement for "term" has been removed from forcing scope.
Ensuring that "scope" filter is moved to the end of search query, which removes possibility of query starting with "AND"
https://github.com/owncloud/ocis/issues/7812#issuecomment-1828252212

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9970

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -397,31 +397,36 @@ export default defineComponent({
         query['mediatype'] = mediaTypeParams.split('+').map((t) => `"${t}"`)
         updateFilter(mediaTypeFilter)
       }
+      // By definition (KQL spec) OR, AND or (GROUP) is implicit for simple cases where
+      // different or identical keys are part of the query.
+      //
+      // We list these operators for the following reasons nevertheless explicit:
+      // * request readability
+      // * code readability
+      // * complex cases readability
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/components/Search/List.vue
```diff
@@ -397,31 +397,36 @@ export default defineComponent({
...
-            if (!isArrayValue) {
-              acc.push(`${prop}:${query[prop]}`)
-            }
-
-            if (isArrayValue) {
-              acc.push(`${prop}:(${query[prop].join(' OR ')})`)
-            }
-
-            return acc
-          }, [])
```
kulmann, line 422: Please use the `sort` array helper before the `join` to make sure that the scope is at the end (just specify a sort callback function that does nothing else than sorting items starting with `scope:` to the end). Can be called in all cases, even on an empty result after the reducer finished, so you can keep the old structure of the code and only add the sort call.

kulmann (changes requested):

packages/web-app-files/src/components/Search/List.vue
jacob-nv: Updated

Issue #10010, opened 2023-11-20
Bug while playing vertical videos


raphaelben55:
## Describe the bug
Video player doesn't display 9/16, vertical video correctly. It shows the video extremely zoomed in.

## Steps to reproduce
1. Open a vertical video with for example the resolution of 1080x1920.

## Setup
Default setup of OCIS on Unraid

kulmann:
@jacob-nv could you look into this?

PR #10101, opened 2023-11-30
Updated video dimensions to fit view horizontal/vertical

jacob-nv:
## Description

Video preview is automatically settings dimensions to eliminate cropping for video, mostly benefical for vertical videos.
Dimensions are updated once video metadata are loaded and further on window resize

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10010

Before:
Horizontal
Screenshot from 2023-11-30 20-26-42
Vertical
Screenshot from 2023-11-30 20-26-56

Now:
Horizontal
Screenshot from 2023-11-30 20-25-32

Vertical
Screenshot from 2023-11-30 20-25-24

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


packages/web-app-preview/src/components/Sources/MediaVideo.vue
```diff
@@ -18,6 +25,23 @@ export default defineComponent({
...
       default: true
     }
+  },
+  beforeUnmount() {
+    window.removeEventListener('resize', this.resizeVideoDimensions)
+  },
+  methods: {
+    updateDimensions() {
+      this.resizeVideoDimensions()
+      window.addEventListener('resize', this.resizeVideoDimensions)
```
JammingBen, line 35: Each time this method gets called you register an event listener, is this done intentionally? Wouldn't it make more sense to register the event listener once on mount?

packages/web-app-preview/src/components/Sources/MediaVideo.vue
```diff
@@ -18,6 +25,23 @@ export default defineComponent({
       type: Boolean,
       default: true
     }
+  },
+  beforeUnmount() {
+    window.removeEventListener('resize', this.resizeVideoDimensions)
+  },
+  methods: {
```
JammingBen, line 32: Please use the composition API by adding the methods and `onBeforeUnmount` in `setup`.

packages/web-app-preview/src/components/Sources/MediaVideo.vue
jacob-nv: Well this method is called only once for each video. It's triggered on loadedmetadata, which is triggered only once and it ensures that videoWidth and videoHeight are available (loadeddata runs before that and at that point those data are not available yet).

jacob-nv: This condition might be misleading. The height, width and video have to be checked for not being nullable because when you go to next video it triggers, but it's already nullable. So it's here mostly because of file navigation.
```
 if (maxHeight && maxWidth && this.$refs.video) {
```

packages/web-app-preview/src/components/Sources/MediaVideo.vue
```diff
@@ -18,6 +24,30 @@ export default defineComponent({
       type: Boolean,
       default: true
     }
+  },
+  setup() {
+    const video = ref<Ref<HTMLVideoElement | null>>(null)
```
kulmann, line 29: Issue with the type:
```
packages/web-app-preview/src/components/Sources/MediaVideo.vue(31,65): error TS2339: Property 'offsetHeight' does not exist on type 'Element'.
packages/web-app-preview/src/components/Sources/MediaVideo.vue(32,64): error TS2339: Property 'offsetWidth' does not exist on type 'Element'.
```

kulmann (changes requested):
Please add a bugfix changelog item.

JanAckermann (approved):

changelog/unreleased/bugfix-video-dimensions
```diff
@@ -0,0 +1,7 @@
+Bugfix: Displaying full video in their dimensions
+
+Fixed bug which set video width and height to 100% of the screen size instead of the video size.
```
kulmann, line 3:
```suggestion
We've fixed a bug which set the video width and height to 100% of the screen size instead of the actual video size.
```

kulmann (approved):

PR #10104, opened 2023-12-01
feat: show link modal when password is enforced 

JammingBen:
## Description
Show the full expanded link modal for creating links when a password is enforced.

I needed for refactor creating link for this because it was kinda messy and spread all over the place. The main changes:

* The `useFileActionsCreateLinks` composable is now being used to create links, no matter how the link is being created (sidebar, quick action, context menu). Previous implementations like in `link.ts` or `FileLinks.vue` are now using that action.
  * The composable holds 2 actions, one for normal links, one for quick links. That's the only difference between both.
  * The action(s) can either enforce a modal (important for embed mode) or determine if a modal is necessary by checking if links are enforced.
  * The action(s) have an optional callback. This is needed e.g. when copying quick links, because then we need to copy the created link afterwards and show a different message.
* The `useFileActionsCreateQuicklink` composable has been renamed to `useFileActionsCopyQuicklink`. Its purpose is to check for an existing quick link and either copy that, or let `useFileActionsCreateLinks` create a new one and then copy that.
* The messages have been removed from `CreateLinkModal`. Instead, you can now pass a callback to handle the messages in the parent as this might be highly individual.
* Our old password modal is now only being used when updating an existing link.

Note that I will add and adjust unit tests in a follow up.

## Types of changes

- [ ] Technical debt
- [ ] Tests


packages/web-pkg/src/composables/actions/files/useFileActionsCreateLink.ts
```diff
@@ -1,60 +1,142 @@
...
-      hideActions: true
+  const proceedResult = (result: PromiseSettledResult<Share>[]) => {
+    const succeeded = result.filter(
+      (val): val is PromiseFulfilledResult<Share> => val.status === 'fulfilled'
+    )
+    if (succeeded.length && showMessages) {
+      store.dispatch('showMessage', {
+        title:
+          succeeded.length > 1
+            ? $gettext('Links have been created successfully')
```
JanAckermann, line 42: Couldn't you just use nGettext?

JanAckermann (changes requested):
Some minor stuff, otherwise it's wonderful 💪

JanAckermann (changes requested):
Some minor stuff, otherwise it's wonderful 💪

packages/web-pkg/src/composables/actions/files/useFileActionsCreateLink.ts
```diff
@@ -1,60 +1,142 @@
...
+        errors: (failed as PromiseRejectedResult[]).map(({ reason }) => new Error(reason)),
+        title:
+          succeeded.length > 1
+            ? $gettext('Failed to create links')
+            : $gettext('Failed to create link')
+      })
+    }
+
+    if (callback) {
+      callback(result)
```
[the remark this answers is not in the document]
JammingBen: I don't quite understand, sorry 😅  The callback is completely new, so it wasn't used before.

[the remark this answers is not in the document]
JanAckermann: Got it, I was asking because I removed former callbacks. Since callback is such a generic name could you rename it with onBlablabla 👍

packages/web-app-files/tests/unit/components/SideBar/Shares/FileLinks.spec.ts
```diff
@@ -87,12 +90,11 @@ describe('FileLinks', () => {
     })
 
     describe('when the add-new-link button is clicked', () => {
-      it('should call addNewLink', async () => {
-        const spyAddNewLink = jest.spyOn((FileLinks as any).methods, 'addNewLink')
-        const { wrapper } = getWrapper()
-        expect(spyAddNewLink).toHaveBeenCalledTimes(0)
+      // TODO: fix and add tests
```
JanAckermann, line 93: Is this still a todo ? 🙈

packages/web-pkg/src/composables/clipboard/useClipboard.ts
```diff
@@ -0,0 +1,27 @@
...
+  // doCopy creates the requested link and copies the url to the clipboard,
+  // the copy action uses the clipboard // clipboardItem api to work around the webkit limitations.
+  //
+  // https://developer.apple.com/forums/thread/691873
+  //
+  // if those apis not available (or like in firefox behind dom.events.asyncClipboard.clipboardItem)
+  // it has a fallback to the vue-use implementation.
+  //
+  // https://webkit.org/blog/10855/
+  const copyToClipboard = (quickLinkUrl: string) => {
```
JanAckermann, line 13: Could you please check if we still need this, we have so many code paths where we just use the browser native code

JanAckermann (changes requested):

packages/web-app-files/tests/unit/components/SideBar/Shares/FileLinks.spec.ts
JammingBen: Yes, unit tests will follow in a separate PR because of the strict release timeline.

packages/web-pkg/src/composables/clipboard/useClipboard.ts
JammingBen: @fschade recently added this via https://github.com/owncloud/web/pull/9832 because of some Safari-foo.

JanAckermann: I mean the else case. Just checked that and navigation.clipboard was introduced 5years ago to all recent browsers. We don't use the useClipboard from vue-use elsewhere. I think we could kick this legacy else case

JammingBen: But the else-case is not legacy, FF e.g. needs it according to Flo's comment.

Either way, if changes are needed here let's do it in another PR because it's just moved code and nothing I want to be touching here.

JanAckermann: I will investigate if I have some free time

Issue #7930, opened 2022-11-07
Epic: Extension system


kulmann:
We're preparing to improve/build our extension system and become more approachable by community developers.

## Preparation

The following preparation steps need to be done:
- [x] publish `web-pkg` and `web-client` to npmjs. They are not in perfect shape, yet, but we already cleaned them up a lot. https://github.com/owncloud/web/issues/8380
- [x] develop strategy for versioning of packages published to npmjs.
- [x] pull the owncloud design system into the web monorepo. Not strictly needed for development (design system is bundled with the web runtime), but having types would improve the DX of extension development a lot.
- [x] clean up config.json / proper config management 

## Ideas

We have a few ideas of "good first extensions" to gain more insights regarding requirements. These include:
- [x] file listing view modes should be their own extensions (file list, tiles view, single file, ...)
- [x] editors/viewers for various file types (like we already have for text files, images, ...)
- [x] panels for the right sidebar (e.g. a panel for EXIF data of photos)
- [ ] any kind of file related action (to be included in the context menu in the file list / right sidebar "Actions" panel, like we already have "copy/move", "delete", "rename", ...)
- [ ] audio/video player for the left sidebar + context panel for the right sidebar, e.g. "add to playlist"
- [x] https://github.com/owncloud/web/issues/5891#event-7938818793
- [ ] ...
- [ ] request interceptors to e.g. add custom headers

# Related 
- [ ] https://github.com/owncloud/web/issues/2958
- [ ] https://github.com/owncloud/web/issues/5135
- [x] https://github.com/owncloud/web/issues/2585
- [x] https://github.com/owncloud/web/issues/5460
- [x] https://github.com/owncloud/web/issues/2423
- [x] https://github.com/owncloud/web/issues/9892
- [ ] https://github.com/owncloud/web/pull/7096#issuecomment-1357372690
- [x] https://github.com/owncloud/web/issues/8281
- [x] https://github.com/owncloud/web/issues/8698
- [x] #9088
- [x] https://github.com/owncloud/web/issues/9210
- [ ] https://github.com/owncloud/web/issues/9217
- [x] https://github.com/owncloud/web/issues/9215
- [x] https://github.com/owncloud/web/issues/9239
- [x] #9302
- [ ] https://github.com/fschade/ocis-arcade/issues/3
- [x] https://github.com/owncloud/web/issues/10256
- [ ] https://github.com/owncloud/web/issues/10387

ChrisEdS:
[...]

PR #10111, opened 2023-12-04
Sidebar panel extension

kulmann:
## Description
This PR introduces a new extension point `sidebarPanel`, which allows apps to register additional panels for the right sidebar. It also extracts the right sidebar that was specific to the `files` app into a generic `FileSideBar` component which lives in the `web-pkg` package. The `FileSideBar` is then used in all viewer and editor apps so far to display a right sidebar with the `files` sidebar panels which were registered by the `files` app. I.e. panels live in `web-app-files`, but are used in any viewer/editor app as well.

## Related Issue
- Works towards https://github.com/owncloud/web/issues/7930

## Motivation and Context
Provide files app functionality like viewing file details, sharing, etc, directly in viewers and editors. Make it possible for viewer/editor apps to bring their own right sidebar panels and display them along the other files app sidebar panels.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] either use `AppWrapper` in `preview` app or embed the `FileSideBar` directly in the preview app
- [x] debug file info panel in apps (missing information, e.g. share indicators and last modified date)
- [ ] double-/triple-check sidebar edge cases in files app regarding shares (e.g. sub-folder inside a share): is the share info loaded correctly?
- [x] find a place for the right sidebar toggle button in apps
- [x] details panel: disable previews in file details, the editor/app is BETTER than the preview ;-) only display the icon.
- [ ] update file size and last modified date on save in editors :-) 
- [ ] fix `panzoom`-bug in preview app when navigating through images with open sidebar
- [ ] fix sidebar width in apps
- [ ] fix sidebar toggle in embed mode


kulmann:
@JanAckermann @lookacat @JammingBen @dschmidt this PR is not quite done, but I'd like to ask for a critical review of three specific things (+ discussion together in a video call probably). This PR changes the `SideBarPanel` interface and exposes a new extension type for the panel. Hence I'd like to give it some extra brain power so that we don't publicly release a new extension point with maybe obvious issues I just didn't see...

The new `SideBarPanelExtension` interface lives in `packages/web-pkg/src/composables/piniaStores/extensionRegistry.ts`. Pretty minimalistic, should be ok, but just for the sake of completeness. This is the interface that needs to be implemented in order to register side bar panels for the right sidebar from any app/extension.

The `Panel` interface has been renamed to `SideBarPanel` in the course of making this an official extension point. It lives in  `packages/web-pkg/src/components/SideBar/types.ts`. This is the interface where I'd like to ask you guys to look very closely if this has rough edges or even major issues. I decided to go for callbacks instead of getters or props with primitive types because that way we can get rid of the old style of having a generator function. 

You can have a look at the new interface in action in `packages/web-app-files/src/fileSideBars.ts` where all right sidebar panels of the files app are defined. With one major difference: these right sidebar panels are now **registered** by the files app to be used in **every other app** which has a file context. That's already working for all the editors we have. It's somewhat broken in the preview app, working on that today. Just wanted you to see how a panel definition and registration looks like with the new interfaces.

There is one new component, which is the `FileSideBar.vue`. That is the one dynamically querying registered panels.

You can check out the PR and run it, it is already in quite a good shape...
- within the files app the goal is that you don't recognize any difference ;-)
- e.g. open a text file in the text editor. For the WIP state of the PR there is a dummy button in the topbar for opening and closing the right sidebar. Still need to figure out a good place for that... you should be able to do really just anything that also works in the files app (sharing, tagging, version management, etc).

packages/web-app-preview/src/App.vue
```diff
@@ -64,14 +64,21 @@
           @toggle-next="next"
         />
       </div>
-    </template>
+      <file-side-bar :open="sideBarOpen" :active-panel="sideBarActivePanel" :space="space" />
```
JanAckermann, line 67: Hence you already rename vars, I'd  love to rename this to isSideBarOpen

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -9,8 +9,9 @@
     />
     <loading-screen v-if="loading" />
     <error-screen v-else-if="loadingError" :message="loadingError.message" />
-    <div v-else class="oc-height-1-1">
+    <div v-else class="oc-flex oc-width-1-1 oc-height-1-1">
       <slot v-bind="slotAttrs" />
+      <file-side-bar :open="sideBarOpen" :active-panel="sideBarActivePanel" :space="space" />
```
JanAckermann, line 14: Same here

packages/web-client/src/helpers/resource/types.ts
```diff
@@ -40,8 +41,7 @@ export interface SpaceRole {
 }
 
 // TODO: add more fields to the resource interface. Extend into different resource types: FileResource, FolderResource, ShareResource, IncomingShareResource, OutgoingShareResource, ...
-export interface Resource {
-  id: number | string
+export interface Resource extends Item {
```
JammingBen, line 44: Ahh nice to finally have an item interface 😍

packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -0,0 +1,286 @@
...
+    }
+  },
+  setup(props) {
+    const store = useStore()
+    const router = useRouter()
+    const clientService = useClientService()
+    const extensionRegistry = useExtensionRegistry()
+    const eventBus = useEventBus()
+
+    const loadedResource = ref()
```
JammingBen, line 88:
```suggestion
    const loadedResource = ref<Resource>()
```

packages/web-pkg/src/composables/piniaStores/extensionRegistry.ts
```diff
@@ -1,15 +1,18 @@
 import { Action } from '../actions'
```
JanAckermann, line 1: I'd be a little more generic with the naming here, for example, if we replace Pinia store with something new in the future. Just in case

JanAckermann:
Do you plan to have something similar like FileSideBar.vue and fileSideBars.ts for web-app-admin-settings as well?
so we will  only have one or two files for the Sidebar here as well? I think that could make sense as the current Views are getting tooo big :(

packages/web-pkg/src/components/SideBar/types.ts
```diff
@@ -1,13 +1,28 @@
 import { defineComponent } from 'vue'
 import { IconFillType } from 'design-system/src/helpers'
+import { Item } from '@ownclouders/web-client/src/helpers'
 
-export interface Panel {
-  app: string
+export interface SideBarPanelContext<P, T extends Item> {
+  parent?: P
+  items?: T[]
+}
```
JammingBen, line 8: Just some unfiltered thought process:

Do we need the space in the future? I'm a bit torn. On one hand, it's quite files-specific since we don't have a space in the admin-settings e.g. But then again, you could say the same about `parent`.

The only reason to add it here would be to have it available in `isVisible` and `isRoot`, right? Might be helpful if I want my panel to be visible in project spaces only. It could also save us from some of the not-so-nice route checks.

JanAckermann:
About the correct place for the sidebar toggle in apps: 
I'd go with a light transparent toggle button, that is located at the same position as the x button in the sidebar. If that is not an option, because it might look bad with OO or Collabroa, I think it's time to refactor the app-top-bar and give it a 3 dot menu

packages/web-client/src/helpers/index.ts
```diff
@@ -1,5 +1,8 @@
+export * from './group'
 export * from './publicLink'
 export * from './resource'
 export * from './share'
 export * from './space'
 export * from './user'
+
```
JanAckermann, line 7: new line by accident?

packages/web-pkg/src/composables/piniaStores/extensionRegistry.ts
JammingBen: We intentionally named it like this to have a clear separation between our current vuex store implementation and the pinia one. Once we moved to pinia I agree though 👍

packages/web-client/src/helpers/item.ts
```diff
@@ -0,0 +1,3 @@
+export interface Item {
```
JanAckermann, line 1: Not a fan of the naming because it's so generic, even if its just there to extend existing interfaces

packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -0,0 +1,286 @@
...
+  SpaceResource,
+  Resource
+} from '@ownclouders/web-client/src/helpers'
+import { WebDAV } from '@ownclouders/web-client/src/webdav'
+
+export default defineComponent({
+  name: 'FileSideBar',
+  components: { FileInfo, SpaceInfo, InnerSideBar },
+  props: {
+    open: {
```
JanAckermann, line 66: isOpen, open gives me the idea of passing a function

JanAckermann (changes requested):

JammingBen:
I love where this is going!

A few bugs I noticed:

* Opening the sidebar for the own personal space without selecting a resource will show the space as resource. While technically correct, we previously showed the No-Selection-Panel
* Opening the sidebar on pages without a space does not work when having no items selected (e.g. projects view, trash bin, Shared with me, ...)
* On the projects page, select a space, open the sidebar, then go to your personal space -> console error `TypeError: Cannot read properties of null (reading 'id')`

packages/web-client/src/helpers/index.ts
kulmann: yes 🙈

packages/web-pkg/src/components/SideBar/types.ts
kulmann: Trying to keep the generic scope, I added a `root` field to the context interface and set the space as `root` wherever possible. Not using it, yet, but I agree that it makes a whole lot of sense to have the space root available.

packages/web-app-preview/src/App.vue
kulmann: done 👍

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
kulmann: done here and in all the other places ;-)

packages/web-app-files/src/fileSideBars.ts
```diff
@@ -5,233 +5,311 @@ import FileVersions from './components/SideBar/Versions/FileVersions.vue'
...
-    componentAttrs: {
-      showSpaceMembers: true,
-      get showLinks() {
-        if (capabilities.files_sharing) {
-          return capabilities.files_sharing.public.enabled
+  return computed(
+    () =>
+      [
+        {
+          id: 'com.github.owncloud.web.files.sidebarPanels.no-selection',
```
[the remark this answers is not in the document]
kulmann: no, brain fart :-) thanks for the pointer

dschmidt:
> I love where this is going!
> 
> A few bugs I noticed:
> 
> * Opening the sidebar for the own personal space without selecting a resource will show the space as resource. While technically correct, we previously showed the No-Selection-Panel
> * Opening the sidebar on pages without a space does not work when having no items selected (e.g. projects view, trash bin, Shared with me, ...)
> * On the projects page, select a space, open the sidebar, then go to your personal space -> console error `TypeError: Cannot read properties of null (reading 'id')`

Search behaves weirdly as well now, maybe because of my latest changes

JanAckermann (approved):

JammingBen (approved):
LGTM 👍 

There are still some occurrences of `fileSideBars` in the code base, we should deprecate/remove them in a follow-up.

dschmidt:
😍 😍 😍

PR #10113, opened 2023-12-04
fix: skip reusing old token in embed mode with delegated auth

LukasHirt:
## Description

When Web is included in an iframe in embed mode with delegated authentication, skip reusing the old token. The parent application is responsible for handling the token and we do not really care about reloading there which the reuse seems to be intended to handle. If the old token is reused, it can lead to an error when fetching user with expired token. That will then lead to a redirect to access denied page once the delegated access_token is received.

## Motivation and Context

Fixing a bug which leads to access denied page.

## How Has This Been Tested?

- test environment: local
- test case 1: embed web in an iframe with delegated auth, postMessage with valid token, wait for the token to expire, remove the iframe, build the iframe again with expired token, postMessage with new valid token

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] add tests
- [x] add changelog

### DIFF
--- a/changelog/unreleased/bugfix-skip-token-reuse-in-embed-mode
+++ b/changelog/unreleased/bugfix-skip-token-reuse-in-embed-mode
@@ -1,0 +1,6 @@
+Bugfix: Skip reusing saved token in Embed mode with delegated authentication
+
+We've stopped reusing saved token in the AuthService when Web is included in Embed mode with delegated authentication.
+The parent application is responsible for handling the access_token.
+
+https://github.com/owncloud/web/pull/10113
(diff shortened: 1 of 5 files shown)


LukasHirt:
I have also added a few logs into the code so that it is easier to debug the embed mode.

kulmann (approved):
I added two commits myself to fast-track this...
1: no extra changelog item needed as the delegated authentication was never publicly released. I just added the PR url to an existing changelog item
2: I reduced the log level of your new log messages to `debug` to not spam the browser console.

Otherwise perfect, thanks for jumping on this quickly 👍

Issue #10064, opened 2023-11-24
Login when coming from the "Logged out" page doesn't redirect correctly


kulmann:
## Describe the bug
When I have an **expired** session in the browser, I need to follow a private link from my desktop client **twice** for reaching the correct location.

## Steps to reproduce
1. Login in to your oCIS instance
2. Close the browser instance and grab a coffee (more than 5 minutes)
3. Follow a private link from your desktop client

## Expected behavior
Log in, then reach the location encoded in the private link.

## Actual behavior
I see the "logged out"-page (due to https://github.com/owncloud/web/issues/10063 ) and when I follow the login button from there, I get redirected to my personal space afterwards. Or in other words: following the login from the "logged out" page loses the information about the desired location.

JammingBen:
Seems to be fixed via https://github.com/owncloud/web/pull/10881, hence closing here.

Issue #10063, opened 2023-11-24
"Logged out" page is showing when re-visiting the instance


kulmann:
## Describe the bug
When I re-visit my oCIS instance after a longer time it greets me with the `Logged out` page being shown.

## Steps to reproduce
1. Login in to oCIS
2. Close the browser and grab a coffee
3. Open the browser and re-visit oCIS

## Expected behavior
Getting greeted by the login page.

## Actual behavior
Getting greeted by the "logged out" page.

## Additional context
While technically correct (my old access token in the localStorage has in fact expired in the meantime), the UX is awful. When the access token has expired for a longer period of time (basically anything more than a few seconds), the expectation is to just get to a login page. Seeing the "logged out" page is useless information and having to click on login again is just one extra-click.

cc @tbsbdr

kulmann:
@tbsbdr thinking about making this a p2 because it's really annoying... what do you think?

Issue #4677, opened 2021-02-02
Exit page re-appears in loop when logged in user is deleted


swoichha:
### Steps to reproduce
1. create a new user and login
2. delete the logged-in user
3. Reload the page as logged in user
4. The user should be redirected to an error page
5. The user press exit to log in with another account

### Expected behavior
- The user should be redirected to the login page

### Actual behavior
- Instead of login page, the exit page re-appears. 

![Screenshot from 2021-02-02 12-01-59](https://user-images.githubusercontent.com/17021686/106560294-e80db900-654e-11eb-86dd-1bf62754b803.png)

JanAckermann:
Actual behavior differs a little:
You will be redirected to the Login Error page, clicking on exit, doesn't trigger any action.
After reloading you will be stuck in the Authentication failed page.

Unfortunately, https://github.com/owncloud/web/pull/7072 does not fix the problem.

Reopening...

JanAckermann:
Might be related: https://github.com/owncloud/ocis/issues/3521

rhafer:
I believe this is not fully fixed. I can't get back to the login page. Always get redirected to an error page:

https://github.com/owncloud/web/assets/373399/aa4d08d6-8af6-4c70-9ae3-1a08a6527854

JammingBen:
Hitting a dead end here. There seems to be some issue when trying to unload the user in case of an error. For some reason, the state/storage doesn't get cleared properly.

I digged deep into the oidc lib, but the only way I found to make it clean properly is to call `signoutRedirect`, which eventually redirects the user to the login page. Hence we would lose the `accessDenied` page. You can specify another URL to route to after logout, but that doesn't work either. I'm out of ideas honestly...

kobergj:
Assuming this fixed meanwhile.

PR #10165, opened 2023-12-13
Redirect after token expiration

jacob-nv:
## Description

Trying to navigate to specific link with token being expired will force redirect to ``` /files/spaces/personal ``` which overrides original redirect. 

Found one solution which is setting redirect in UserManager constructor, while checking for existence of stored redirect in combination with ``` /files/spaces/personal ``` to prevent override of this storage. Other part of condition is to check for url not being oidc redirect or callback, since this would be stored (in this case setting url from constructor) and the app would get stuck on callback pages.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10064

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-runtime/package.json
+++ b/packages/web-runtime/package.json
@@ -25,7 +25,7 @@
     "lodash-es": "^4.17.21",
     "luxon": "3.0.1",
     "marked": "^4.0.12",
-    "oidc-client-ts": "^2.1.0",
+    "oidc-client-ts": "^2.3.0",
     "owncloud-sdk": "3.1.0-alpha.10",
     "p-queue": "^6.6.2",
     "pinia": "2.1.7",
--- a/packages/web-runtime/src/services/auth/authService.ts
+++ b/packages/web-runtime/src/services/auth/authService.ts
@@ -80,7 +80,8 @@ export class AuthService {
         configurationManager: this.configurationManager,
         store: this.store,
         ability: this.ability,
-        language: this.language
+        language: this.language,
+        redirectUrl: to.path
       })
     }
 
--- a/packages/web-runtime/src/services/auth/authService.ts
+++ b/packages/web-runtime/src/services/auth/authService.ts
@@ -80,8 +80,7 @@ export class AuthService {
         configurationManager: this.configurationManager,
         store: this.store,
         ability: this.ability,
-        language: this.language,
-        redirectUrl: to.path
+        language: this.language
       })
     }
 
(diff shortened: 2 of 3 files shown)


JammingBen (commented):
Tested it and it works great. I'm not 100% sold on the way how it's implemented though 😅  Initializing the `userManager` with a redirect URL feels a bit weird. Also the edge case for `/files/spaces/personal`.

After a token has expired and the user opens a page, they will get logged out: https://github.com/owncloud/web/blob/master/packages/web-runtime/src/services/auth/authService.ts#L262C1-L262C1. After the logout the user is automatically being redirected to the login page again, but it loses every context such as the `redirectUri` during that process.

One way to possibly solve this would be to do a silent sign out instead and then manually redirect to the login page while also passing the correct `redirectUri`. I couldn't find a way to make a silent logout work in the past though... It would be also beneficial for this issue: https://github.com/owncloud/web/issues/4677. Maybe you could check if you can make it work? We basically want what `signoutRedirect` does but without the redirect. [The lib](https://github.com/authts/oidc-client-ts) has a method `signoutSilent` (although not exported to the outside for some reason), maybe that would work somehow. Our `userManager` is extending the one from the lib.

Or maybe @kulmann / @dschmidt have an idea (or are fine with the current approach).

jacob-nv:
> Tested it and it works great. I'm not 100% sold on the way how it's implemented though 😅 Initializing the `userManager` with a redirect URL feels a bit weird. Also the edge case for `/files/spaces/personal`.
> 
> After a token has expired and the user opens a page, they will get logged out: https://github.com/owncloud/web/blob/master/packages/web-runtime/src/services/auth/authService.ts#L262C1-L262C1. After the logout the user is automatically being redirected to the login page again, but it loses every context such as the `redirectUri` during that process.
> 
> One way to possibly solve this would be to do a silent sign out instead and then manually redirect to the login page while also passing the correct `redirectUri`. I couldn't find a way to make a silent logout work in the past though... It would be also beneficial for this issue: #4677. Maybe you could check if you can make it work? We basically want what `signoutRedirect` does but without the redirect. [The lib](https://github.com/authts/oidc-client-ts) has a method `signoutSilent` (although not exported to the outside for some reason), maybe that would work somehow. Our `userManager` is extending the one from the lib.
> 
> Or maybe @kulmann / @dschmidt have an idea (or are fine with the current approach).

signoutSilent method is exported but in 2.3 version, current version is 2.1
this way it seems to work, cannot await signoutSilent method, it throws error `Refused to frame 'https://host.docker.internal:9200/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'none'"`

But somehow this seems to work well, it signouts user, sets the proper redirect url, redirects to login and from login you get to stored redirect url

JammingBen:
Needs to be tested and refined (probably), hence I converted it back to a draft. The last time I checked it worked, however, it broke https://github.com/owncloud/web/issues/10063.

JammingBen:
Superseded by https://github.com/owncloud/web/pull/10881.

Issue #10095, opened 2023-11-30
Refactor modal


JammingBen:
The current implementation of our modal has quite a few downsides, such as holding logic that it is not supposed to hold (e.g. password enforcement rules) and being limited to one active modal.

### To-Dos

- [ ] Decouple modal from store and use a wrapper component that handles the modals (e.g. via event bus)
- [ ] Allow stacking multiple modals that will be processed in a queue
- [x] Remove custom logic such as password enforcement rules and use dedicated components instead
- [x] Clean up modal implementations where the logic is currently spread all over the place. These things should be file actions that create a modal using a custom component. Logic can then sit either in the file action, or in the component (see e.g. how it's done in https://github.com/owncloud/web/pull/10076).
  - [x] Quota modal
  - [x] Create user modal
  - [x] Edit user login modal
  - [x] Add/remove users to/from groups modal
  - [x] Create group modal
  - [x] Create shortcut modal
  - [x] Readme content modal
- [ ] Establish pattern for onConfirm/onCancel handling
  - [ ] Establish clean pattern to replace the buttons while still leveraging the onConfirm handling
  - [ ] Close modal automatically when onConfirm successfully returns/resolves
  - [ ] Show progress indication while promise has not returned (and disable button)

PR #10180, opened 2023-12-14
refactor: user action modals

JammingBen:
## Description
Move several modal action handlers from `Users.vue` into their own components where they belong. This a) cleans up the component nicely and b) prevents the logic from being spread all over the place.

Affected actions:

* "Create user"-action moved into `useUserActionsCreateUser.ts`. It opens the `CreateUserModal.vue` which holds all the logic
* "Edit login"-action handler move into `EditLoginModal.vue`
* "Add to groups"-action handler move into `AddToGroupsModal.vue`
* "Remove from groups"-action handler move into `RemoveFromGroupsModal.vue`

Also adds unit test for all these things.

This PR is best reviewed by its individual commits.

## Related Issue

- refs https://github.com/owncloud/web/issues/10095

## Types of changes

- [x] Technical debt
- [ ] Tests


dschmidt (commented):
Great!

It can be another iteration, but I would like to get rid of the store actions triggering in the dialog components, I think the modal should handle that automatically based on promises returned from onCancel and onConfirm.

JammingBen:
> Great!
> 
> It can be another iteration, but I would like to get rid of the store actions triggering in the dialog components, I think the modal should handle that automatically based on promises returned from onCancel and onConfirm.

You mean `store.dispatch('hideModal')` right? I agree that this would be nice, probably easy to do as well. I'd rather do in in a follow-up though.

dschmidt:
yep, as I said - can be in another iteration

JanAckermann (approved):
🤩

PR #10182, opened 2023-12-14
Readonly versions panel

kulmann:
## Description
Make versions panel readonly when the right sidebar is opened from within a viewer or editor.
Reason for that is that it would require complex user interaction to make reverting versions from within an editor view (= the file content is already displayed) possible. Let's avoid that by only allowing the download action, not the revert action.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
+++ b/packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="oc-file-versions-sidebar" class="-oc-mt-s">
-    <oc-loader v-if="loading" />
-    <ul v-if="!loading && hasVersion" class="oc-m-rm oc-position-relative">
+    <oc-loader v-if="areVersionsLoading" />
+    <ul v-else-if="versions.length" class="oc-m-rm oc-position-relative">
       <li class="spacer oc-pb-l" aria-hidden="true"></li>
       <li
         v-for="(item, index) in versions"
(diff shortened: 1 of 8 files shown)


JanAckermann (approved):
Nice 💪

packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
```diff
@@ -54,93 +54,126 @@
...
+    const fetchVersionsTask = useTask(function* () {
+      yield store.dispatch('Files/loadVersions', {
+        client: clientService.webdav,
+        fileId: unref(resource).fileId
+      })
+    })
+    const areVersionsLoading = computed(() => {
+      return !fetchVersionsTask.last || fetchVersionsTask.isRunning
+    })
+    watch(
```
JanAckermann, line 103: Unmounted wasn't possible ? :(

kulmann: Can either do a: `onMounted` AND a `watch` (without `immediate = true`) or b: only the `watch` with `immediate = true`. I usually decide for b in such a case. Happy about suggestions

JanAckermann: No that's fine, if we can't omit the watcher. Possible could also be a :key=resource.id on the respective panel and a onMounted, I prefer this over a watcher. But only if possible, otherwise looks nice to me

kulmann: Oh no, the watcher actually does more than watching the id. If you have the right sidebar in versions panel open from within an editor (e.g. text editor), the panel updates the versions list when you save the file. :-) but you actually gave me a different idea now, I switched over to only watching the `etag` of the resource, not the whole resource.

PR #10183, opened 2023-12-14
folder view extensions

dschmidt:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...


packages/web-app-files/src/components/Shares/SharedWithMeSection.vue
```diff
@@ -18,8 +18,7 @@
       v-else
       v-model:selectedIds="selectedResourcesIds"
       :data-test-share-status="shareStatus"
-      class="files-table"
```
kulmann, line 21: I bet this is being used as a selector in the nightwatch acceptance tests 🙈

packages/web-app-files/src/extensions.ts
```diff
@@ -19,11 +20,13 @@ export const extensions = ({ applicationConfig }: ApplicationSetupOptions) => {
...
   const { actions: quickLinkActions } = useFileActionsCopyQuickLink()
 
-  const panels = sideBarPanels()
+  const panels = useSideBarPanels()
+  const folderViews = useFolderViews()
 
   return computed(
     () =>
       [
+        ...unref(folderViews),
```
kulmann, line 29: folderViews are not a ref, but just an array... looks like that's ok, but you don't need to `unref` the var then. The sidebar panels are built as a `computed`, but looking at the code I think we could omit that and also just return an array. And also not `unref` it here then.

packages/web-pkg/src/composables/viewMode/constants.ts
```diff
@@ -1,36 +1,12 @@
...
-      name: 'apps-2',
-      fillType: 'line'
-    }
-  }
-  static readonly queryName: string = 'view-mode'
-  static readonly tilesSizeDefault: number = 1
-  static readonly tilesSizeQueryName: string = 'tiles-size'
-}
+export const FolderViewModeConstants = {
+  // FIXME: we have a few places where we still match against hardcoded names, get rid of that and this constants
```
kulmann, line 2: 👍

packages/web-pkg/src/ui/types.ts
```diff
@@ -1,10 +1,14 @@
...
+export type FolderView = {
   name: string
   label: string
   icon: {
     name: string
     fillType: IconFillType
   }
+  isScrollable?: boolean
+  component: ReturnType<typeof defineComponent>
+  componentAttrs?: () => Record<string, unknown>
```
kulmann, line 13: In the case of FolderViews it would be super useful to be able to define the minimal set of required props any folder view component needs... but not sure if that would be over-engineered. :-/

packages/web-app-files/src/components/Shares/SharedWithMeSection.vue
dschmidt: Yes, that's why I added it in the component... and you realize the tests are already green? 😇

kulmann: duuuuude! nvm then 😇

packages/web-pkg/src/ui/types.ts
dschmidt: What are you thinking of?

dschmidt: Ah, you mean some kind of interface/default props like selectedIds, resources etc?

dschmidt: ... and events that are handled

packages/web-app-files/src/extensions.ts
dschmidt: okay, I'm fixing the new code, but won't touch the other code for now to avoid breaking anything while thoughtlessly changing something 😅

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -63,89 +64,60 @@
...
-              />
-            </template>
-          </resource-table>
+          <template v-else>
+            <resource-details
+              v-if="displayResourceAsSingleResource"
+              :single-resource="paginatedResources[0]"
+              :space="space"
+            />
+            <component
```
JammingBen, line 73: I think here is a `v-else` missing, no? Or do we want `resource-details` to be displayed together with the table/tiles view?

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -347,6 +355,14 @@ export default defineComponent({
       required: false,
       default: true
     },
+    /**
+     * TODO
```
JammingBen, line 359: 👀

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -420,6 +445,10 @@ export default defineComponent({
       required: false,
       default: null
     },
+    resourceType: {
+      type: String,
+      default: 'file'
+    },
```
JammingBen, line 451: Maybe you can add a validator with all the supported strings here?

packages/web-pkg/src/ui/types.ts
```diff
@@ -1,10 +1,14 @@
...
-export type ViewMode = {
+export type FolderView = {
   name: string
   label: string
   icon: {
     name: string
     fillType: IconFillType
   }
+  isScrollable?: boolean
+  component: ReturnType<typeof defineComponent>
```
JammingBen, line 12: This will basically type `component` as `any`. I still haven't figured out component typing in Vue, documentation is a bit lacklustre (or I simply didn't find the right place 😄 ). But I guess you could just use `Component` as type here. It also lets you define required props if you have some (`resources` maybe?).

packages/web-app-files/src/components/FilesList/ResourceTiles.vue
```diff
@@ -34,8 +34,12 @@
         </oc-list>
       </oc-drop>
     </div>
-    <oc-list class="oc-tiles oc-flex" :class="resizable ? 'resizableTiles' : ''">
-      <li v-for="resource in data" :key="resource.id" class="oc-tiles-item has-item-context-menu">
+    <oc-list class="oc-tiles oc-flex">
+      <li
+        v-for="resource in resources"
+        :key="resource.id"
```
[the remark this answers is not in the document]
dschmidt: Yeah, let's keep it in mind for the follow up

JammingBen (approved):
😍

PR #10185, opened 2023-12-15
refactor: quota modal usage

JammingBen:
## Description
Before, the state of the quota modal was handled by the outside where it was being used. This caused quite some boilerplate code because it's effectively always the same.

Refactors the quota modal usage so that the modal state handling is done via the quota action instead, meaning the consuming party doesn't need to care about it anymore.

## Related Issue

- refs https://github.com/owncloud/web/issues/10095

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/packages/web-runtime/src/App.vue
+++ b/packages/web-runtime/src/App.vue
@@ -46,7 +46,7 @@
           v-else
           ref="modalComponent"
           :modal="modal"
-          v-bind="modal.customComponentAttrs"
+          v-bind="modal.customComponentAttrs?.() || {}"
         />
       </template>
     </oc-modal>
--- a/packages/web-app-admin-settings/src/components/Spaces/ContextActions.vue
+++ b/packages/web-app-admin-settings/src/components/Spaces/ContextActions.vue
@@ -1,20 +1,14 @@
 <template>
   <div>
     <context-action-menu :menu-sections="menuSections" :action-options="{ resources: items }" />
-    <quota-modal
-      v-if="quotaModalIsOpen"
-      :cancel="closeQuotaModal"
-      :spaces="items"
-      :max-quota="maxQuota"
-    />
   </div>
 </template>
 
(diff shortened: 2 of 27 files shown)


packages/web-runtime/src/App.vue
```diff
@@ -50,7 +50,11 @@
           v-else
           ref="modalComponent"
           :modal="modal"
-          v-bind="modal.customComponentAttrs"
+          v-bind="
+            typeof modal.customComponentAttrs === 'function'
+              ? modal.customComponentAttrs()
+              : modal.customComponentAttrs
+          "
```
dschmidt, line 57: We will have this logic in a lot of places in the future I assume - maybe worth adding a helper function for this?

dschmidt (approved):

packages/web-runtime/src/App.vue
kulmann: Hate it (the `if function then call else use value`), that was exactly the reason why I went for `always the callback` when I adjusted the `SideBarPanel` interface...

Helper would be ok-ish, but the helper might also advocates such type definitions. I'm also fine without the helper. 🤷

kulmann (approved):

packages/web-runtime/src/App.vue
JammingBen: Then let's make it always a function, I need to rebase anyway.

packages/web-runtime/src/App.vue
```diff
@@ -46,7 +46,7 @@
           v-else
           ref="modalComponent"
           :modal="modal"
-          v-bind="modal.customComponentAttrs"
+          v-bind="modal.customComponentAttrs ? modal.customComponentAttrs() : {}"
```
kulmann, line 49:
```suggestion
          v-bind="modal.customComponentAttrs?.() || {}"
```

Issue #10181, opened 2023-12-14
Move Theme switcher to Account Settings Page


tbsbdr:
## Improvement proposal
- Move theme switcher drop to account settings
- Remove theme icon from top bar to declutter top bar as it has now the new sidebar toggle button 
- Bonus: Provide an Auto-Option (default) to adjust the dark/light theme to system setting

![image](https://github.com/owncloud/web/assets/26610733/e83a0a90-e7d8-477f-9edd-9020ab6a5415)

JanAckermann:
We will allow anon users (public link) to access a minimal version of the preferences.

PR #10207, opened 2023-12-18
Preferences for not logged in users (public link context)

JanAckermann:
## Description
Enhancement: Enable user preferences in public links

We've enabled user preferences in public links, so any user even without an account can open
preferences in a public link context and for example change the current language.

## Related Issue

- Part of https://github.com/owncloud/web/issues/10181

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -246,7 +246,11 @@ export default defineComponent({
       })
     })
     const showWebDavDetails = computed(() => {
-      return store.getters['Files/areWebDavDetailsShown']
+      /**
+       * webDavPath might not be set when user is navigating on public link,
+       * even if the user is authenticated and the file owner.
+       */
+      return store.getters['Files/areWebDavDetailsShown'] && unref(resource).webDavPath
     })
     const formatDateRelative = (date) => {
       return formatRelativeDateFromJSDate(new Date(date), language.current)
--- a/packages/web-app-files/src/index.ts
+++ b/packages/web-app-files/src/index.ts
@@ -34,12 +34,9 @@ const appInfo = {
   icon: 'resource-type-folder',
   color: 'var(--oc-color-swatch-primary-muted)',
   isFileEditor: false,
-  extensions: [],
-  applicationMenu: {
-    enabled: () => true,
-    priority: 10
-  }
+  extensions: []
 }
+
 export const navItems = (context): AppNavigationItem[] => {
   const spacesStores = useSpacesStore()
   const userStore = useUserStore()
@@ -127,8 +124,18 @@ export const navItems = (context): AppNavigationItem[] => {
 
 export default defineWebApplication({
   setup() {
+    const userStore = useUserStore()
+
     return {
-      appInfo,
+      appInfo: {
+        ...appInfo,
+        applicationMenu: {
+          enabled: () => {
+            return !!userStore.user
+          },
+          priority: 10
+        }
+      },
       store,
       routes: buildRoutes({
         App,
--- a/packages/web-app-ocm/src/index.ts
+++ b/packages/web-app-ocm/src/index.ts
@@ -1,5 +1,5 @@
 import App from './views/App.vue'
-import { defineWebApplication, useRouter } from '@ownclouders/web-pkg'
+import { defineWebApplication, useRouter, useUserStore } from '@ownclouders/web-pkg'
 import translations from '../l10n/translations.json'
 import { extensions } from './extensions'
 import { RouteRecordRaw } from 'vue-router'
@@ -27,13 +27,19 @@ export default defineWebApplication({
   setup() {
     const { $gettext } = useGettext()
     const router = useRouter()
+    const userStore = useUserStore()
 
     const appInfo = {
       name: $gettext('ScienceMesh'),
       id: 'ocm',
       color: '#AE291D',
       icon: 'contacts-book',
-      isFileEditor: false
+      isFileEditor: false,
+      applicationMenu: {
+        enabled: () => {
+          return !!userStore.user
+        }
+      }
     }
 
     router.addRoute({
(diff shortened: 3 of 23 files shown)


packages/web-runtime/tests/unit/pages/account.spec.ts
```diff
@@ -43,7 +43,7 @@ jest.mock('@ownclouders/web-pkg', () => ({
     })
 }))
 
-describe('account page', () => {
+describe.skip('account page', () => {
```
JanAckermann, line 46: 😈

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -257,7 +258,7 @@ export default defineComponent({
 
     const capabilityFilesTags = useCapabilityFilesTags()
     const hasTags = computed(() => {
-      return props.tagsEnabled && unref(capabilityFilesTags)
+      return props.tagsEnabled && unref(capabilityFilesTags) && unref(isUserContext)
```
JanAckermann, line 261: We need do check if userContext otherwise, showing tags will result in a 401 error in case user is on a public link and not authenticated

packages/web-app-files/src/index.ts
```diff
@@ -35,7 +36,9 @@ const appInfo = {
   isFileEditor: false,
   extensions: [],
   applicationMenu: {
-    enabled: () => true,
+    enabled: (store: Store<unknown>) => {
+      return !!store.getters?.user?.id
```
JanAckermann, line 40: Files app doesn't need to be shown is unauthenticated and on a public link

packages/web-app-files/src/index.ts
```diff
@@ -128,7 +131,7 @@ export const navItems = (context): AppNavigationItem[] => {
 export default defineWebApplication({
   setup(args) {
     return {
-      appInfo,
+      appInfo: appInfo as any,
```
JanAckermann, line 134: taked with @JammingBen it is what it is for now

packages/web-app-ocm/src/index.ts
```diff
@@ -27,12 +27,18 @@ export default defineWebApplication({
...
 
     const appInfo = {
       name: $gettext('ScienceMesh'),
       id: 'ocm',
       icon: 'contacts-book',
-      isFileEditor: false
+      isFileEditor: false,
+      applicationMenu: {
+        enabled: () => {
+          return !!store.getters?.user?.id
```
JanAckermann, line 39: was missing and ocm app was shown in unauth public link context

packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -125,14 +138,10 @@ export default {
...
         .filter((app) => {
           if (app.type === 'extension') {
-            // check if the extension has at least one navItem with a matching menuId
             return (
-              store.getters
-                .getNavItemsByExtension(app.id)
-                .filter((navItem) => isNavItemPermitted(permittedMenus, navItem)).length > 0 ||
-              (app.applicationMenu.enabled instanceof Function &&
-                app.applicationMenu.enabled(store, ability) &&
-                !permittedMenus.includes('user'))
```
JanAckermann, line 135: legacy code that was breaking the enabled in app applicationMenu configuration

packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -20,7 +20,7 @@
         class="oc-topbar-unauthenticated-avatar oc-flex-inline oc-flex-center oc-flex-middle"
         :name="$gettext('User Menu login')"
         :width="32"
-        icon="user-add"
+        icon="user"
```
JanAckermann, line 23: now user can login, go to preferences so its not only create account (which also doesnt exist :D)

packages/web-pkg/src/apps/types.ts
```diff
@@ -51,8 +51,8 @@ export interface ApplicationQuickActions {
 export type AppConfigObject = Record<string, any>
 
 export interface ApplicationMenuItem {
-  enabled: () => boolean
-  priority: number
+  enabled?: () => boolean
```
lookacat, line 54: shouldnt enabled be mandatory?

JanAckermann, line 54:
```suggestion
  enabled: () => boolean
```

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
JammingBen: Not something for now necessarily, but IMO we should move `unref(capabilityFilesTags) && unref(isUserContext)` to the panel and only use the `tagsEnabled` property here.

packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -125,14 +138,10 @@ export default {
...
+              app.applicationMenu.enabled instanceof Function &&
```
JammingBen, line 142: Hmm do we still need the check if it's a function? It's correctly typed as a function, so I guess we can assume that here.

JanAckermann: unfortunately yes, applicationMenu will  be always present, see `packages/web-runtime/src/store/apps.ts`
so unfortantly optional chaining also doesn't work, as you will call object.undefinted(x,y)

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
JanAckermann: Unfortunately not possible without refactoring, as FileDetails will also be called on a single shared resource and doesn't have the tagsEnabled prop accessible (defaults to true)

JammingBen (approved):

Issue #10208, opened 2023-12-19
Gap size in "copy instead" modal is not aligned with other modals


kulmann:
## Describe the bug
When cutting files/folders and then trying to paste them into a different space, we show a modal that informs the user about moving files not being possible and asking if the user wants to copy the files instead.
The gap size between a: header separator line and the first block of text and b: last block of text and the footer row with the buttons is bigger than usual and looks off.

## Steps to reproduce
1. Select files in your personal space
2. Use batch action "cut" (or cmd+x keyboard shortcut)
3. Navigate into a different space
4. Use "paste" action (or cmd+v keyboard shortcut)

## Expected behavior
Gap sizes in modal look like in every other modal.

## Actual behavior
Gap sizes are not aligned with other modals, see Screenshot.

## Screenshots
<img width="508" alt="Screenshot 2023-12-19 at 06 41 48" src="https://github.com/owncloud/web/assets/3532843/a1208f9f-c8d8-420f-b97f-c34586905530">
<img width="505" alt="Screenshot 2023-12-19 at 06 45 41" src="https://github.com/owncloud/web/assets/3532843/e4c224a7-82ce-461b-943b-bfa77d6192de">

## Additional context
The "copy instead" modal holds two `<p>` tags, which bring a margin of 14px top and bottom. Those margins (top on first `<p>` and bottom on last `<p>`) are unnecessary.

Issue #10210, opened 2023-12-19
Migrate from vuex store to pinia


JammingBen:
### Description

In the long-term we want to migrate from vuex to pinia since it gives us e.g. better typing and makes unit testing easier. Also, pinia is the official recommended way for Vue store implementations.

There's a long way to go, meaning this is an ongoing process and won't be done via one single PR. Stores that need to be refactored:

### Stores to migrate

- [x] Files & AncestorMetaData (https://github.com/owncloud/web/pull/10368)
- [x] Spaces (https://github.com/owncloud/web/pull/10316)
- [x] Messages (https://github.com/owncloud/web/pull/10309
- [x] User (https://github.com/owncloud/web/pull/10240)
- [x] Capabilities (https://github.com/owncloud/web/pull/10296)
- [x] Navigation (https://github.com/owncloud/web/pull/10326 & https://github.com/owncloud/web/pull/10329)
- [x] Modal (https://github.com/owncloud/web/pull/10212)
- [x] Config (https://github.com/owncloud/web/pull/10336)
- [x] Auth (https://github.com/owncloud/web/pull/10323)
- [x] Apps (https://github.com/owncloud/web/pull/10331)
- [x] External (https://github.com/owncloud/web/pull/10346)

### To clarify

- ~In general: how much buisness logic do we want to have inside our store composables?~ Depends on the store. For now we keep it mostly the same as before to ease the migration process.
- ~How do we want to organize the file store? Some parts like e.g. the current folder shouldn't live exclusively in the files app. Some other things though are only needed in the files app. Split it up into 2 stores, one living in `web-pkg`, one in the files app?~ See PR description here: https://github.com/owncloud/web/pull/10368
- Split the files store in files and shares?
- ~What parts from the navigation store are still needed? Looks a bit like old/deprecated extension code to me 🤷~ -> removed the deprecated stuff
- ~Do we want to keep storing left nav items in a separated store, or make it part of the extension system?~ -> handle via extension registry
- ~Do we want the integrate our message queue in the store again?~ -> yes!
- ~The config store as well as the configuration manager currently share logic such as defining defaults. There should be only one place to do that, which means we need to discuss responsibilities here.~ -> one composable to rule them all!

dschmidt:
We could try to identify stores with the least dependencies on other stores to determine a reasonable order. I remember when we tried to port ancestorMetadata it was basically impossible to do that in isolation because of dependencies....

JammingBen:
[...]

Issue #9183, opened 2023-06-07
Provide all dialogs with loading spinners


hurradieweltgehtunter:
**All** buttons, which confirm a form or some sort of dialog, should receive a spinner, which indicates that the desired action is running. This applies to:

— Modals
— Forms, for example sharing dialogue
— … 

Current state leaves the user often with an unresponsive UI, which seems to be doing nothing or broken.
The user is tempted to reload the page, which cancels the action running in the background and leads to unwanted results.

Related: https://github.com/owncloud/web/issues/9182

kulmann:
On top of that we also really do need to prevent submitting the respective action multiple times, see e.g. https://github.com/owncloud/web/issues/9189

hurradieweltgehtunter:
> On top of that we also really do need to prevent submitting the respective action multiple times, see e.g. #9189

I think this is also connected to #9018

PR #10212, opened 2023-12-19
refactor: modal implementation

JammingBen:
## Description
Remove the modal from vuex store and integrates it into pinia instead. This change completely reworks how to work with modals. The main benefits and changes are:

* Proper typings
* Allow stacking multiple modals
* Modals close automatically on cancel/confirm
* Modals automatically use the loading service on confirm
* Streamline modal usage to have one way how to create modals

### How to use it

Modals can be dispatched via the `dispatchModal` method like so:

```
const { dispatchModal } = useModals()

dispatchModal({
  title: 'Create folder',
  onConfirm: () => createFolder()
})
```

The modal will automatically be removed when either the cancel or the confirm button has been clicked and the resulting promise resolved successfully. If needed, a modal can be removed manually via the provided method `removeModal`.

It is also possible to register another modal while having a modal open. This will add the new modal to the top of the queue and immediately display it. After it closes, the previous modal will show again.

A modal which is present in the queue but not showing can be displayed via the method `setModalActive`.

### Custom components

Modals can have custom components with custom properties:

```
const { dispatchModal } = useModals()

dispatchModal({
  title: 'Create folder',
  onConfirm: () => createFolder(),
  customComponent: CreateFolderComponent,
  customComponentAttrs: () => ({ space })
})
```

This will render the `CreateFolderComponent` inside the modal while still showing the cancel and confirm buttons at the bottom.

It is also possible to define the `onConfirm` and `onCancel` methods inside the component. They must be named `onConfirm` / `onCancel` and be exposed via `expose({ onConfirm, onCancel })`. Note that if these methods are defined in `registerModal`, they will take priority over defined methods in the component.

Modal actions can be hidden if you need to implement custom actions inside the component:

```
const { dispatchModal } = useModals()

dispatchModal({
  title: 'Create folder',
  onConfirm: () => createFolder(),
  customComponent: CreateFolderComponent,
  customComponentAttrs: () => ({ space }),
  hideActions: true
})
```

Now you can implement these actions on your own. Note that you should still expose `onConfirm` / `onCancel` via `expose` and not call them directly with your actions. Register additional event handlers on the buttons instead: `emit('confirm')` and `emit('cancel')`. This ensures that the methods get called via the wrapping component which provides the benefits such as automatically closing the modal after confirm or cancel.

## Open questions / follow-ups

* Move modal components in respective `Modals` folders
* Rename modal components from `...Modal` to `...ModalContent`?
* Add dev docs on how to use the modal?

## Related Issue

- fixes https://github.com/owncloud/web/issues/10095
- fixes https://github.com/owncloud/web/issues/10208
- refs https://github.com/owncloud/web/issues/9183

## Types of changes

- [x] Technical debt
- [ ] Tests


packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -271,31 +270,29 @@ export default defineComponent({
...
       )
     }
 
-    const confirm = async () => {
+    const onConfirm = async () => {
       if (!unref(selectedRoleIsInternal)) {
         if (unref(passwordEnforced) && !unref(password).value) {
           password.error = $gettext('Password must not be empty')
-          return
+          return Promise.reject(new Promise((reject) => reject('')))
```
dschmidt, line 291: see above

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -271,31 +270,29 @@ export default defineComponent({
...
         }
 
         if (!passwordPolicy.check(unref(password).value)) {
-          return
+          return Promise.reject(new Promise((reject) => reject('')))
```
dschmidt, line 295: see above

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -325,19 +322,15 @@ export default defineComponent({
 
       if (userFacingErrors.length) {
         password.error = $gettext(userFacingErrors[0].message)
-        return
+        return Promise.reject(new Promise((reject) => reject('')))
```
dschmidt, line 325: see above

packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFile.ts
```diff
@@ -191,21 +184,15 @@ export const useFileActionsCreateNewFile = ({
 
     const inputSelectionRange = !areFileExtensionsShown.value
       ? null
-      : [0, defaultName.length - (extension.length + 1)]
+      : ([0, defaultName.length - (extension.length + 1)] as [number, number])
```
dschmidt, line 187: maybe as const would work as well?

packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
```diff
@@ -200,23 +183,21 @@ export const useFileActionsRename = ({ store }: { store?: Store<any> } = {}) =>
       !resources[0].isFolder && !areFileExtensionsShown ? nameWithoutExtension : resources[0].name
 
     const inputSelectionRange =
-      resources[0].isFolder || !areFileExtensionsShown ? null : [0, nameWithoutExtension.length]
+      resources[0].isFolder || !areFileExtensionsShown
+        ? null
+        : ([0, nameWithoutExtension.length] as [number, number])
```
dschmidt, line 188: as const?

packages/web-pkg/src/composables/piniaStores/modals.ts
```diff
@@ -0,0 +1,90 @@
...
+import { v4 as uuidV4 } from 'uuid'
+import { Component, ComponentPublicInstance, computed, ref, unref } from 'vue'
+
+export type CustomModalComponent = Component<{ modal: Modal } & Record<string, any>>
+
+export type CustomModalComponentEmits = {
+  confirm?: () => Promise<void>
+  cancel?: () => void
+  'update:confirmDisabled': (value: boolean) => void
+} & Record<string, any>
```
dschmidt, line 12: Can't we pass this to `defineComponent` when defining our components? Could give us type safety for emitting events etc

packages/web-pkg/tests/unit/composables/piniaStores/useModals.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+          const data = { title: 'test' }
+          const modal = instance.dispatchModal(data)
+
+          expect(modal.id).toBeDefined()
+          expect(modal.title).toEqual(data.title)
+          expect(instance.activeModal).toEqual(modal)
+        }
+      })
+    })
+  })
```
dschmidt: Adding a second modal and checking the right one is active would be nice as well

packages/web-pkg/tests/unit/composables/piniaStores/modals.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  describe('method "dispatchModal"', () => {
+    it('adds a modal to the stack of modals', () => {
+      getWrapper({
+        setup: (instance) => {
+          const data = { title: 'test' }
+          const modal = instance.dispatchModal(data)
+
+          expect(modal.id).toBeDefined()
+          expect(modal.title).toEqual(data.title)
+          expect(instance.activeModal).toEqual(modal)
```
dschmidt, line 19: Would be nice to add a second one and then check again the right modals are (in)active.

Same for removeModal

packages/web-runtime/src/App.vue
```diff
@@ -162,12 +103,6 @@ export default defineComponent({
     )
   },
   methods: {
-    focusModal(component, event) {
-      ;(this as any).focus({
-        revert: event === 'beforeUnmount'
-      })
-    },
-
```
dschmidt, line 170: Is this now handled anywhere else?

packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -298,10 +298,7 @@ export class AuthService {
     // TODO: create UserUnloadTask interface and allow registering unload-tasks in the authService
     await this.store.dispatch('runtime/auth/clearUserContext')
     await this.store.dispatch('resetUserState')
-    await Promise.all([
-      this.store.dispatch('clearDynamicNavItems'),
-      this.store.dispatch('hideModal')
-    ])
```
dschmidt, line 304: What did this even do?

packages/web-runtime/src/store/modal.ts
```diff
@@ -1,106 +0,0 @@
...
-  SET_MODAL_CONFIRM_BUTTON_DISABLED(state, status) {
-    state.confirmDisabled = status
-  }
-}
-
-export default {
-  state,
-  actions,
-  mutations
-}
```
dschmidt, line 106: bye bye we won't miss you

dschmidt (approved):

packages/web-runtime/src/services/auth/authService.ts
JammingBen: Closing a potentially open modal I guess 😄

packages/web-runtime/src/App.vue
JammingBen: No, but I didn't find a reason to keep it. Modals are being focused correctly due to the focus trap in `OcModal`.

packages/web-pkg/src/composables/piniaStores/modals.ts
JammingBen: I tried that without success. As we talked about earlier, the typing of these components can definitely be improved in the future. Maybe we should go on a hunt for good docs about this whole topic 👀

packages/web-app-admin-settings/tests/unit/components/Groups/CreateGroupModal.spec.ts
```diff
@@ -59,10 +59,13 @@ describe('CreateGroupModal', () => {
...
   describe('method "onConfirm"', () => {
     it('should not create group if form is invalid', async () => {
+      jest.spyOn(console, 'error').mockImplementation(() => undefined)
       const { wrapper, storeOptions } = getWrapper()
 
       const eventSpy = jest.spyOn(eventBus, 'publish')
-      await wrapper.vm.onConfirm()
+      try {
+        await wrapper.vm.onConfirm()
+      } catch (error) {}
```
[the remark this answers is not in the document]
JammingBen: Since we test for an error here the test would fail otherwise. So we wouldn't be able to set any expectations afterwards.

packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFile.ts
JammingBen: As discussed this does not work the `const` type seems to be immutable.

JammingBen:
I'll create a follow-up PR which addresses all the resolved comments, coming right up.

PR #10238, opened 2023-12-22
chore: use the project's typescript version in vscode

JammingBen:
At least for me this causes much less issues in vscode. Otherwise it would use typescript `5.3` which is not yet compatible with Web.

## Types of changes

- [x] Technical debt
- [ ] Tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,7 +13,6 @@
   },
   "editor.formatOnSave": true,
   "eslint.format.enable": true,
-  "eslint.packageManager": "pnpm",
   "jest.autoRun": {
     "watch": false,
     "onStartup": true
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -13,7 +13,6 @@
   },
   "editor.formatOnSave": true,
   "eslint.format.enable": true,
-  "eslint.packageManager": "pnpm",
```
JammingBen, line 16: Fyi this is deprecated and is now being detected automatically.

.vscode/settings.json
```diff
@@ -27,5 +26,5 @@
   "cucumberautocomplete.steps": [
     "tests/e2e/cucumber/steps/**/*.ts"
   ],
-  "cucumberautocomplete.strictGherkinCompletion": true
```
dschmidt, line 30: Why was this removed?

JammingBen: Because I messed up 🤡

dschmidt (approved):

PR #10240, opened 2023-12-22
refactor: move vuex user to pinia

JammingBen:
## Description
Moves the user from the vuex store to pinia.

Also removes loading the old OCS user since it now lives in the Graph API. This means that the user sitting in the store now looks a bit different. The changes are:

* `uuid` -> `id` 
* `id` -> `onPremisesSamAccountName`
* `email` -> `mail`
* `displayname` -> `displayName`
* `groups` -> `memberOf`

This also means that the quota is now being handled on the user's personal drive instead of on the user itself. On a drive item this looks like this:

```
{
    [...]
    quota: {
        remaining: 4658892800,
        state: "normal",
        total: 0,
        used: 15091902
    }
}
```

### How to

The user can be retrieved via the `useUserStore` composable like so:

```
const userStore = useUserStore()
const user = userStore.user
```

There is no need to unref `userStore.user` since it behaves like Vue's `reactive`. You only need to be careful when destructuring the user since it would lose reactivity. To compensate for that you can use pinia's `storeToRefs` method:

```
const userStore = useUserStore()
const { user } = storeToRefs(userStore)
```

Note that you need to unref the user then though.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests


changelog/unreleased/change-remove-ocs-user
```diff
@@ -0,0 +1,8 @@
+Change: Remove ocs user
+
+BREAKING CHANGE for developers: The user from the ocs api has been removed in favor of the graph user.
```
kulmann, line 3: Please add a hint to the breaking change notice, that the underlying data structure of the `user` has changed.

kulmann (approved):
Awesome 😍

Issue #10250, opened 2024-01-02
Convert remaining JS unit tests to TS


JammingBen:
We have a few unit tests left in the design system that still need to be converted to TypeScript. After doing that we probably can (and should) remove the `babel-jest` dependency.

JammingBen:
It seems that the remaining JS tests are for the ODS docs... let's wait for https://github.com/owncloud/web/pull/10243 and then decide what to do with them. Maybe we can remove them, I'm not sure how valuable they actually are.

dschmidt:
Yeah, unit tests for the docs seems a bit pointless, I would love e2e tests on the built artifacts tho

PR #10253, opened 2024-01-02
chore: remove babel-jest dependency

JammingBen:
## Description

Converts the ODS doc tests to TS files and removes the `babel-jest` dependency as we don't have JS files anymore. Note that some tests we simply commented out as we currently don't run them anyways. Need fixing (or a decision if the need them) in the future.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10250

## Types of changes

- [x] Technical debt
- [x] Tests


packages/design-system/docs/components/tokens/AllTokens.spec.ts
```diff
@@ -1,7 +1,7 @@
 import Vue from 'vue'
 import AllTokens from './AllTokens.vue'
 
-const Constructor = Vue.extend(AllTokens)
+const Constructor = (Vue as any).extend(AllTokens)
```
dschmidt, line 4: Eeeh... it's probably missing from the type because it doesn't exist anymore?!

JammingBen: Yep.. but I'm closing this PR again because we found out that we can't simply remove the `babel-jest` dependency (for now) 😬

dschmidt: Can you elaborate for later reference?

PR #10263, opened 2024-01-03
chore: remove depcheck from tooling chain

JammingBen:
## Description
Removes `depcheck` from out tooling chain because we're not using it.

Follow-up for https://github.com/owncloud/web/pull/10262.

## Types of changes

- [x] Technical debt
- [ ] Tests


dschmidt:
Is this actually used? How and where?

JammingBen:
> Is this actually used? How and where?

Good point, we checked it and since we don't use it (and the output looks not useful/wrong) we decided to kick the dep.

kulmann (approved):

dschmidt (approved):

Issue #10259, opened 2024-01-03
Stop support for old acceptance tests


ScharfViktor:
2 years ago, we stopped implement old acceptance tests and started using e2e with playwright. 
Now we want to: 
- delete all unworkable, repetitive, irrelevant tests
- rewrite useful tests
- close all issues related acceptance tests

Steps: 
- [x] delete all tests labeled @skipOnOCIS
  - https://github.com/owncloud/web/pull/10416
- [x] Remove repetitive test scenarios (compare with e2e) 
  - https://github.com/owncloud/web/pull/10267
- [x] check the expected-failures-with-ocis-server-ocis-storage.md file and remove the string and irrelevant tests (@saw-jan @PrajwolAmatya)
  - https://github.com/owncloud/web/pull/10427
  - https://github.com/owncloud/web/pull/10428
  - https://github.com/owncloud/web/pull/10449
  - https://github.com/owncloud/web/pull/10480
  - https://github.com/owncloud/web/pull/10488
  - https://github.com/owncloud/web/pull/10550
  - https://github.com/owncloud/web/pull/10574
  - https://github.com/owncloud/web/pull/10575
- [x] find and close all the related issues in GH @ScharfViktor 
- [x] rewrite existing acceptance tests
   - [x] api tests
     - https://github.com/owncloud/ocis/pull/9015
     - https://github.com/owncloud/ocis/pull/9018
     - https://github.com/owncloud/ocis/pull/9083
     - https://github.com/owncloud/ocis/pull/9121
     - https://github.com/owncloud/ocis/pull/9126
     - https://github.com/owncloud/ocis/pull/9150
   - [x] e2e tests (Updated from comment https://github.com/owncloud/web/issues/10259#issuecomment-2006653617)
     - [x] breadcrumb navigation on normal folders https://github.com/owncloud/web/pull/10664
     - [x] preview different files (.png file) @Salipa-Gurung (https://github.com/owncloud/web/pull/10670)
     - [x] pubilc user preview files: https://github.com/owncloud/web/pull/10565
     - [x] edit password of public link and public user refresh the public link page @grgprarup https://github.com/owncloud/web/pull/10826
     - [x] update content of file (Assigned to pradip) https://github.com/owncloud/web/pull/10838
   - [ ] unit tests
- [x] delete acceptance tests suite
  - https://github.com/owncloud/web/pull/10858
  - https://github.com/owncloud/web/pull/10861
  - https://github.com/owncloud/web/pull/10878
[...]

PR #10267, opened 2024-01-03
[test-only] Deleted acceptance tests. part1

ScharfViktor:
part of https://github.com/owncloud/web/issues/10259

part 1
- [x] removed duplicated with e2e tests scenarios and skipped tests (NOT all)
- [x] deleted webUIFavorites - in agreement with @kulmann 
- [x] deleted disable sharing. it is not implemented ocis feature, tests were slkipped. We'll create e2e test when it will be available


kulmann (approved):
Nice 😍

PR #10293, opened 2024-01-04
refactor: remove opinionated components from ODS

JammingBen:
## Description
Removes the following opinionated components from the ODS and moves them to `web-pkg` instead:

- `OcResource` (now `ResourceListItem`)
- `OcTile` (now `ResourceTile`)
- `OcResourceIcon` (now `ResourceIcon`)
- `OcGhostElement` (now `ResourceGhostElement`)
- `OcResourceLink` (now `ResourceLink`)
- `OcResourceName` (now `ResourceName`)
- `OcResourceSize` (now `ResourceSize`)

Those components hold logic about resources, which is something that is not supposed to live in the ODS. The ODS should only contain design- and layout-specific components.

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/design-system/changelog/unreleased/change-remove-opinionated-components
```diff
@@ -0,0 +1,12 @@
+Change: Remove opinionated components
+
+We've removed the following opinionated components from the ODS and moved them to `web-pkg` instead:
```
kulmann, line 3: Please add a breaking change for developers notice, like we usually do.

packages/design-system/changelog/unreleased/change-remove-opinionated-components
```diff
@@ -0,0 +1,12 @@
...
+
+- `OcResource` (now `ResourceListItem`)
```
kulmann, line 5: There are three more components about resources:
- `OcResourceLink`
- `OcResourceName`
- `OcResourceSize`

Could you please move and rename those as well?

packages/web-pkg/src/components/DragGhostElement.vue
kulmann, line 1: move into subfolder `FilesList`?

packages/web-pkg/src/components/ResourceListItem.vue
kulmann, line 1: Move into subfolder `FilesList`?

packages/web-pkg/src/components/ResourceIcon.vue
kulmann, line 1: Move into a new subfolder `Resource`?

packages/web-pkg/src/components/ResourceTile.vue
kulmann, line 1: Move into subfolder `FilesList`? The `ResourceTiles.vue` component still lives in the files app, but it would make sense to move it here into the `FilesList` folder anyway... if you think the scope of this PR can handle that move as well, then go for it 😇

packages/web-pkg/src/helpers/resource/icon.ts
```diff
@@ -1,4 +1,12 @@
-import { IconType } from './types'
+export type IconFillType = 'fill' | 'line' | 'none'
+export type IconType = {
+  name: string
+  color?: string
+  fillType?: IconFillType
+}
+
+export type OcResourceIconMapping = Record<'mimeType' | 'extension', Record<string, IconType>>
```
kulmann, line 8: Small nitpick: If we keep the provide/inject of icon mappings, these two lines should also not contain `oc`-prefixes anymore.

packages/web-pkg/src/components/ResourceIcon.vue
kulmann, line 1: Since this component now moved from the (potentially standalone) design system into `web-pkg` we don't need to provide/inject the icon mappings anymore. Instead the component could query that from the runtime / a composable / whatever. But maybe something for a followup... could you create an issue? Or just add a todo in your PR description, so that I can create the issue before we merge and after we agreed that we want to change it. 😅

kulmann (changes requested):

packages/web-pkg/src/components/ResourceTile.vue
JammingBen: Moving `ResourceTiles.vue` is a little more work to do, I created an issue for this: https://github.com/owncloud/web/issues/10304

dschmidt:
`DragGhostElement` - I think this should *somehow* also contain `Resource` in the name

packages/web-pkg/src/components/ResourceIcon.vue
JammingBen: Good catch, I've created https://github.com/owncloud/web/issues/10305 for this.

JammingBen:
> `DragGhostElement` - I think this should _somehow_ also contain `Resource` in the name

I decided for `ResourceGhostElement`, IMO we don't necessarily need the information "drag" here.

kulmann (changes requested):
Nice! Now my last request for change: remove `web-client` from the dependency list please 😅

kulmann (approved):
Hooray, thank you 🥳

Issue #9748, opened 2023-09-28
Define default capabilities in in one place


JammingBen:
The `useCapability...` composables in `web-pkg` currently set default values, which is bad design because other services that access the capabilities via the store directly don't know anything about these defaults.

The goal is to implement one access point to define and provide capabilities including their defaults. This is probably done best via a pinia store.

PR #10296, opened 2024-01-05
refactor: move capabilities to pinia store

JammingBen:
## Description
Moves the capabilities from vuex store to pinia and removes the `use...Capability` composables in favor of the pinia store composable. As a side effect this guarantees that we now have one single location that defines the capabilities and their default values. Before this was happening in the vuex store as well as in the `use...Capability` composable.

Furthermore, the `enabled` callback as well as the `name` callback of the `AppNavigationItem` no longer have the `capabilities` parameter. Instead, you should use `defineWebApplication` in your app's `index.ts`, which gives you a proper setup-context where you can use `useCapabilities`.

Note that this is a breaking change for developers!

## Related Issue

- fixes https://github.com/owncloud/web/issues/9748

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-pkg/src/composables/capability/useCapability.ts
dschmidt, line 1: this was a nice finger excise to get warmer with typescript ... but I won't miss this code :)

dschmidt (approved):
<3 <3 <3

Issue #10288, opened 2024-01-04
Clean up eslint warnings


JammingBen:
Running `pnpm lint --fix` currently returns about 200 warnings, which should be fixed. The only warning that is okay-ish for now is `Component name [...] should always be multi-word`.

PR #10300, opened 2024-01-05
eslint cleanup

jacob-nv:
## Description

Took down warning to 11 ("Component name should be multi-word"), disabled eslint in some parts like v-html, some of the mock functions which should accept argument to simulate the right function without using it

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10288

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -45,7 +45,8 @@ export default defineComponent({
      */
     popperOptions: {
       type: Object,
-      required: false
+      required: false,
+      default: () => {}
```
JammingBen, line 49: I'd prefer if we would return an empty object here instead of an empty method.

```suggestion
      default: () => ({})
```

packages/web-app-files/src/index.ts
```diff
@@ -126,7 +126,7 @@ export const navItems = (context): AppNavigationItem[] => {
 }
 
 export default defineWebApplication({
-  setup(args) {
+  setup() {
```
JammingBen, line 129: Was the linter really complaining about `args` here? Because we passed it down to `extensions()` down below 🤔

packages/web-client/src/webdav/index.ts
```diff
@@ -37,7 +37,7 @@ export const webdav = (options: WebDavOptions): WebDAV => {
   const listFilesByIdFactory = ListFilesByIdFactory(dav, options)
   const { listFilesById } = listFilesByIdFactory
 
-  const getFileInfoFactory = GetFileInfoFactory(listFilesFactory, options)
+  const getFileInfoFactory = GetFileInfoFactory(listFilesFactory)
```
JammingBen, line 40: Please keep passing the `options` to the factories. We don't always need them now, but we might in the future. That way, all factories have the same options.

packages/web-client/src/webdav/index.ts
```diff
@@ -46,7 +46,7 @@ export const webdav = (options: WebDavOptions): WebDAV => {
   const { putFileContents } = PutFileContentsFactory(dav, getFileInfoFactory, options)
 
   const { getFileUrl, revokeUrl } = GetFileUrlFactory(dav, getFileContentsFactory, options)
-  const { getPublicFileUrl } = GetPublicFileUrlFactory(dav, options)
+  const { getPublicFileUrl } = GetPublicFileUrlFactory(dav)
```
JammingBen, line 49: Here as well.

packages/web-client/tests/unit/sse.spec.ts
```diff
@@ -58,6 +58,7 @@ describe('SSEAdapter', () => {
 
   test('it should handle onclose events and throw RetriableError', () => {
     const fetchOptions = { method: 'GET' }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const sseAdapter = new SSEAdapter(url, fetchOptions)
```
JammingBen, line 62: Simply this?

```suggestion
    new SSEAdapter(url, fetchOptions)
```

JammingBen (commented):
A few annotation from my side, most of it looks good to me 👍 

Could you please target this to master, since this doesn't fix a user-facing or critical bug? Sorry for not clarifying earlier!

packages/web-app-files/src/index.ts
jacob-nv: Removed them, because the extension function doesn't do anything with it. I'm not sure if this is something that should be available so it can be modified by other developers and just give it exception with disabling eslint.

JammingBen: Ahh okay, now I get it. It's okay then I'd say. If the `extensions` method needs it one day we can re-add it.

packages/design-system/src/components/OcDrop/OcDrop.vue
jacob-nv: Got it, this is in few other components. Gonna fix it

JammingBen (approved):
Nice 👍

PR #10316, opened 2024-01-10
refactor: move spaces to pinia store

JammingBen:
## Description
Removes the vuex spaces store module and adds a pinia store for spaces instead. The pinia store can be accessed via the new `useSpacesStore()` composable.

The spaces store could still be refactored a little bit, but I wanted to keep the scope here as small as possible. What I don't like is that the store doesn't feel consistent to me. E.g. The method `addSpaces` simply adds given spaces to the store, whereas `addSpaceMembers` does not only add space members to the store, but does also requests to the server.

Note that this is a breaking change for developers!

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-runtime/src/store/index.ts
```diff
@@ -4,16 +4,12 @@ import auth from './auth'
 import config from './config'
 import user from './user'
 import navigation from './navigation'
-import spaces from './spaces'
-
-const strict = process.env.NODE_ENV === 'development'
```
JammingBen, line 9: This causes issues because spaces are now sitting in pinia store (spaces store) as well as in vuex store (files store). Since this only affects dev environments and we will get rid of vuex eventually, I'd propose to disable strict mode.

dschmidt (approved):

dschmidt:
💖💖💖

PR #10329, opened 2024-01-12
refactor: register nav items via extension registry

JammingBen:
## Description
Removes the nav items from the vuex navigation store module. It now uses the extension registry with the already existing `SidebarNavExtension` type for registering and retrieving  nav items.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-runtime/src/container/api.ts
```diff
@@ -61,22 +61,25 @@ const announceRoutes = (applicationId: string, router: Router, routes: RouteReco
...
   }
 
-  store.commit('SET_NAV_ITEMS_FROM_CONFIG', {
-    extension: applicationId,
-    navItems: navigationItems
-  })
+  const navExtensions = navigationItems.map((navItem) => ({
+    type: 'sidebarNav',
+    navItem,
+    scopes: [applicationId, `app.${applicationId}`]
```
dschmidt, line 79: Hmm why are both scopes needed?

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -555,7 +570,8 @@ export const announceDefaults = ({
     return r.path.startsWith(`/${defaultExtensionId}`) && r.meta?.entryPoint === true
   })
   if (!route) {
-    route = store.getters.getNavItemsByExtension(defaultExtensionId)[0]?.route
+    route = getExtensionNavItems({ extensionRegistry, appId: defaultExtensionId })[0]
+      ?.route as RouteRecordNormalized
```
dschmidt, line 574: Why do we need the cast here? Can't we fix the type?

dschmidt (approved):

packages/web-runtime/src/container/api.ts
JammingBen: Not necessarily, but it is requested with those 2 types. If we decide for one, I'd go with `app.${applicationId}`. What do you think?

packages/web-runtime/src/container/bootstrap.ts
JammingBen: I knew you were going to remark this 😂 I guess so, but then we need to change it in the appInfo type and in all apps as well.

packages/web-runtime/src/container/api.ts
dschmidt: Yeah, let's use only one. with prefix sounds good

packages/web-runtime/src/container/bootstrap.ts
dschmidt: not sure, but I figure there might be a better place for this cast at least

dschmidt: ok, the place is good - but I'm happier with widening the type of the route var instead of casting the `.route` prop to something it isn't

PR #10331, opened 2024-01-12
refactor: move app vuex store to pinia

JammingBen:
## Description
Removes the vuex app store module in favour of a pinia store instead.

I also tried cleaning up and improving the naming and types of app related stuff. It's not all perfect yet, but it's definitely an improvement. The most notable change here is that `extensions` inside the `appInfo` object are now of type `ApplicationFileExtension[]`, since they have nothing to do with our extension library, but are file extensions.

Also, there is no such thing as a `newFileHandler`. This is basically a `ApplicationFileExtension` with an added `newFileMenu` that includes a `menuTitle`.

### Breaking changes for developers

* Removes the `announceExtensions` method inside the app's `ready` hook, which could be used to register file editors. Apparently it was not used by any app because there is already another way to register file editors, which is via the `extensions` property inside the `appInfo` object.
* Removes the global `$store` variable. The recommended way to use the vuex store is via `useStore`.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-app-preview/src/App.vue
```diff
@@ -153,6 +139,10 @@ export default defineComponent({
     const cachedFiles = ref<CachedFile[]>([])
     const folderLoaded = ref(false)
 
+    const mimeTypes = computed(() => {
+      return getMimeTypes(appsStore.externalAppConfig[appId]?.mimeTypes)
```
dschmidt, line 143: Can't we use the applicationConfig prop for this?

packages/web-app-preview/src/index.ts
```diff
@@ -1,41 +1,46 @@
...
-    mimeType,
-    routeName,
-    label: $gettext('Preview')
-  }))
-}
+    const appInfo = {
+      name: $gettext('Preview'),
+      id: appId,
+      icon: 'eye',
+      extensions: getMimeTypes(appsStore.externalAppConfig[appId]?.mimeTypes).map((mimeType) => ({
```
dschmidt, line 33: and can't we use the applicationConfig from the ApplicationSetupOptions here?

packages/web-app-text-editor/src/index.ts
```diff
@@ -50,17 +53,16 @@ export default defineWebApplication({
         }
       ]
 
-      const config = (window as any).__$store.getters.extensionConfigByAppId(appId)
-      extensions.push(...(config.extraExtensions || []).map((ext) => ({ extension: ext })))
+      const config = appsStore.externalAppConfig[appId] || {}
```
dschmidt, line 56: and here

packages/web-pkg/src/apps/types.ts
```diff
@@ -42,18 +42,33 @@ export interface ApplicationQuickAction {
...
+  color?: string
   id?: string
   name?: string
   icon?: string
+  iconFillType?: IconFillType
+  iconColor?: string
+  img?: string
   isFileEditor?: boolean
-  extensions?: any[]
+  extensions?: ApplicationFileExtension[]
```
dschmidt, line 71: 😍

packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFile.ts
```diff
@@ -208,27 +209,25 @@ export const useFileActionsCreateNewFile = ({
...
+        ext: appFileExtension.extension
       })
     }
     for (const mimeType of unref(mimetypesAllowedForCreation) || []) {
-      const openAction = false
       actions.push({
         name: 'create-new-file',
         icon: 'add',
-        handler: (args) => handler(args, mimeType.ext, openAction),
+        handler: (args) => handler(args, mimeType.ext, undefined),
```
dschmidt, line 230: Why the undefined?

packages/web-pkg/src/composables/appDefaults/useAppConfig.ts
```diff
@@ -13,9 +12,8 @@ export interface AppConfigResult {
 }
 
 export function useAppConfig(options: AppConfigOptions): AppConfigResult {
-  const applicationMetaResult = useAppMeta(options)
   const applicationConfig = computed(
-    () => unref(applicationMetaResult.applicationMeta).config || {}
+    () => options.appsStore.externalAppConfig[options.applicationId] || {}
```
dschmidt, line 16: Is this something else or just removing an indirection?

packages/web-pkg/src/composables/appDefaults/useAppMeta.ts
dschmidt, line 1: 😍

packages/web-runtime/src/container/application/classic.ts
```diff
@@ -128,7 +127,8 @@ export const convertClassicApplication = async ({
     supportedLanguages
   })
 
-  await store.dispatch('registerApp', applicationScript.appInfo)
+  const appsStore = useAppsStore()
+  appsStore.registerApp(applicationScript.appInfo)
```
dschmidt, line 131: Exciting - why was this async before?

packages/web-runtime/src/container/application/index.ts
```diff
@@ -122,7 +122,7 @@ export const buildApplication = async ({
     if (!isObject(applicationScript.appInfo) && !applicationScript.setup) {
       throw new RuntimeError('next applications not implemented yet, stay tuned')
```
dschmidt, line 123: We should get rid of Next Applications as a concept... Classic and Next Applications are somewhat converging anyhow

dschmidt (approved):
😍 😍 😍

packages/web-app-preview/src/App.vue
JammingBen: As discussed, we need this to be wrapped in `AppWrapper` first.

packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFile.ts
JammingBen: As discussed, before it was `false`, so this is the same behaviour but correct typing.

packages/web-pkg/src/composables/appDefaults/useAppConfig.ts
JammingBen: As discussed, config was removed from the app meta.

packages/web-runtime/src/container/application/classic.ts
JammingBen: As discussed, because the signature of `store.dispatch` is async.

PR #10334, opened 2024-01-15
Move theme switcher to account settings

lookacat:
## Description
See https://github.com/owncloud/web/issues/10181

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10181

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] Add the auto theming option


JanAckermann (approved):

packages/web-runtime/src/components/Topbar/ThemeSwitcher.vue
JammingBen, line 1: Please move this file away from the `Topbar` folder when it's not part of the topbar anymore.

packages/web-runtime/src/pages/account.vue
```diff
@@ -398,6 +406,10 @@ export default defineComponent({
       }
     }
 
+    const onUpdateTheme = () => {
+      showMessage({ title: $gettext('Preference saved.') })
+    }
```
JammingBen, line 411: Why handle this here via events? I'd expect the `ThemeSwitcher` to handle the message 😅

JanAckermann: Why would you expect that?

JammingBen: Because it seems unnecessarily complicated to me 🤷 It's not that we're updating data or something in the parent component which would justify that. I'd rather have all theme-logic decoupled in the ThemeSwitcher component (same like e.g. the GDPRExport component handles all gdpr related stuff).

PR #10352, opened 2024-01-17
[docs-only] update documentation of visibility classes

koebel:
## Description
It seems like the documentation about the use of `.oc-visible@s`, `.oc-visible@m`, `.oc-visible@l` and `.oc-visible@xl` to display elements only on certain screen sizes is not correct. Working with these classes I noticed that these classes hide the element on screen sizes that are smaller than the given breakpoint and display them on devices with screen sizes bigger or equal to the given breakpoint – as their name also implies (see also css code examples below generated by applying these classes). However the documentation says exactly the opposite which is confusing for developers who want to use these classes.

```
@media (max-width: 959px)
.oc-visible\@m {
    display: none !important;
}

@media (max-width: 639px)
.oc-visible\@s {
    display: none !important;
}
```

## Related Issue
no issue created so far since this is to my understanding no bug but expected behaviour of the given css classes

## Motivation and Context
update documentation to match it with actual code behaviour and make it thus easier for developers :)

## How Has This Been Tested?
applying the css classes mentioned above in the development of [web-app-dicom-viewer](https://github.com/owncloud/web-app-dicom-viewer)

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [x] Documentation

## Open tasks:
none

### DIFF
--- a/packages/design-system/docs/visibility.md
+++ b/packages/design-system/docs/visibility.md
@@ -7,10 +7,10 @@ Add one or more of the following classes to any element to display elements only
 | Class            | Description                                              |
 | ---------------- | -------------------------------------------------------- |
 | .oc-invisible    | Hides element without removing it from the document flow |
-| .oc-visible@s    | Displays element on screens smaller than 640px           |
-| .oc-visible@m    | Displays element on screens smaller than 960px           |
-| .oc-visible@l    | Displays element on screens smaller than 1200px          |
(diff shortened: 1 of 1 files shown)


JammingBen (approved):
Oh yes, makes sense actually. Good catch, thx!

JanAckermann (approved):

PR #10363, opened 2024-01-18
refactor: remove file versions from vuex files store module

JammingBen:
## Description
Removes the file versions from the vuex files store module because there is no need to store them in any store.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/tests/unit/components/SideBar/Versions/FileVersions.spec.ts
+++ b/packages/web-app-files/tests/unit/components/SideBar/Versions/FileVersions.spec.ts
@@ -177,12 +177,6 @@ function getMountedWrapper({
   loadingStateDelay = 0
 } = {}) {
   const storeOptions = defaultStoreMockOptions
-  storeOptions.modules.Files.getters.versions.mockImplementation(() => versions)
-  storeOptions.modules.Files.actions.loadVersions.mockImplementation(() => {
-    if (loadingStateDelay > 0) {
-      return new Promise((res) => setTimeout(res, loadingStateDelay))
-    }
-  })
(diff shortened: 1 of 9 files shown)


packages/web-app-files/tests/unit/components/SideBar/Versions/FileVersions.spec.ts
```diff
@@ -191,6 +185,14 @@ function getMountedWrapper({
     downloadFile
   }
   mocks.$clientService.webdav.getFileInfo.mockResolvedValue(mock<Resource>({ id: '1' }))
+  mocks.$clientService.webdav.listFileVersions.mockImplementation(() => {
+    if (loadingStateDelay > 0) {
+      return new Promise((res) => setTimeout(() => res(versions), loadingStateDelay))
```
dschmidt, line 190: if `loadingStateDelay` defaults to 0 anyway - can't we just always run this branch? What benefit does the check add? a tick less? 🤷🏻

dschmidt (approved):

packages/web-app-files/tests/unit/components/SideBar/Versions/FileVersions.spec.ts
JammingBen: Good catch, fixed. It's necessary for checking the loading spinner, since we still don't have a way to mock tasks properly.

JammingBen: Hmm CI seems unhappy that way, so I reverted it

PR #10389, opened 2024-01-22
chore: replace easygettext with scripts from vue3 gettext

### LINKED ISSUE #10249: Get rid of `easygettext` and use `vue3-gettext` instead
## Is your feature request related to a problem? Please describe.
We currently use `easygettext` for extracting our language strings so they can be imported into Transifex. That's not optimal because we needed to fork the lib and therefore can't easily bump it if needed (like currently because of some security alerts).

## Describe the solution you'd like
Since we're already using `vue3-gettext` we can and should use the extractor from that lib.

LukasHirt:
## Description

- use vue3 gettext to extract and compile translations
- introduce gettext config which configures extraction and compilation
- drop LC_MESSAGES from path of translation files (needed due to configuration of extraction and compilation where we cannot specify path after locale)

## Related Issue

- resolves #10249

## Motivation and Context

Get rid of dep alerts

## How Has This Been Tested?

- test environment: macos x64
- test case 1: makefile

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/design-system/docs/i18n.md
+++ b/packages/design-system/docs/i18n.md
@@ -1,6 +1,6 @@
 ## Writing translations
 
-For translations we are using combination of [vue-gettext](https://github.com/Polyconseil/vue-gettext) and [easygettext](https://github.com/polyconseil/easygettext).
+For translations we are using [Vue 3 Gettext](https://jshmrtn.github.io/vue3-gettext/).
 
 ### Escaping
 
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -57,11 +57,11 @@ export default defineComponent({
   emits: ['update:applicationName'],
   setup(props, { emit }) {
     const language = useGettext()
+    const { $gettext } = language
     const { showErrorMessage } = useMessages()
     const capabilityStore = useCapabilityStore()
     const configStore = useConfigStore()
 
-    const { $gettext } = language
     const { makeRequest } = useRequest()
 
     const appNameQuery = useRouteQuery('app')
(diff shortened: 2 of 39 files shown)


LukasHirt:
@JammingBen @dschmidt please let me know how would you like to approach changelog here

dschmidt:
no change log pretty much irrelevant for users and external developers

packages/design-system/docs/i18n.md
```diff
@@ -1,6 +1,6 @@
 ## Writing translations
 
-For translations we are using combination of [vue-gettext](https://github.com/Polyconseil/vue-gettext) and [easygettext](https://github.com/polyconseil/easygettext).
+For translations we are using combination of [Vue 3 Gettext](https://jshmrtn.github.io/vue3-gettext/).
```
dschmidt, line 3: no combination

packages/web-app-external/src/App.vue
```diff
@@ -105,7 +104,7 @@ export default defineComponent({
 
         const query = stringify({
           file_id: fileId,
-          lang: language.current,
+          lang: currentLanguage,
```
dschmidt, line 107: This was probably the reason why there was a language object and the return value of useGettext() wasn't destructured ... this loses reactivity

packages/web-runtime/src/container/sse/files.ts
```diff
@@ -153,13 +153,14 @@ export const onSSEItemTrashedEvent = ({
     // If initiated by current client (browser tab), action unnecessary. Web manages its own logic, return early.
     return
   }
+  const { $gettext } = language
```
dschmidt, line 156: if this works, you can use this approach above where you need to access `language.current`

dschmidt (changes requested):

packages/web-app-external/src/App.vue
kulmann: yep, sure. Could you check the PR again please?

Issue #10381, opened 2024-01-22
Stop loading versions separately in each sidebar panel


JammingBen:
## Is your feature request related to a problem? Please describe.
Currently 2 sidebar panels are loading all versions for the selected resource: `FileVersions.vue` and `FileDetails.vue`. However, there is no need for this as it creates unnecessary network load.

## Describe the solution you'd like
Loading the versions should happen "globally" in `FileSideBar.vue` and then being provided via Vue's `provide` to the panel components.

PR #10391, opened 2024-01-22
Injecting versions [WIP]

jacob-nv:
## Description
`SideBar.vue` fetches versions which are provided to `FileVersions.vue` and `FileSideBar.vue`

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10381

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -188,7 +188,7 @@ export default defineComponent({
     const authStore = useAuthStore()
     const { publicLinkContextReady } = storeToRefs(authStore)
 
-    const versions: Ref<Resource[]> = inject('versions')
+    const { versions }: { versions: Ref<Resource[]>; fetchVersions: Function } = inject('versions')
 
     const isPreviewEnabled = computed(() => {
       if (unref(resource).isFolder) {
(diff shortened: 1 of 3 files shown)


jacob-nv:
@JammingBen Any recommendation for how to handle tests? Lots of the testing is based around fetching. Here is everything provided except one instance where restoring old version refreshes versions through provided method

JammingBen:
> @JammingBen Any recommendation for how to handle tests? Lots of the testing is based around fetching. Here is everything provided except one instance where restoring old version refreshes versions through provided method

Hmm I didn't think about the re-fetching, not only from a unit test perspective, but the actual implementation. The fact that the versions panel needs to reload versions after reverting kinda makes me question the provide/inject approach... @kulmann @dschmidt what do you think?

kulmann:
> > @JammingBen Any recommendation for how to handle tests? Lots of the testing is based around fetching. Here is everything provided except one instance where restoring old version refreshes versions through provided method
> 
> Hmm I didn't think about the re-fetching, not only from a unit test perspective, but the actual implementation. The fact that the versions panel needs to reload versions after reverting kinda makes me question the provide/inject approach... @kulmann @dschmidt what do you think?

Sounds to me like a pinia store (similar to the `shares` pinia store implementation) would make sense. 🤔

JammingBen:
Closing here since this approach of fixing it is not really feasible because of the reasons mentioned above.

Issue #9893, opened 2023-10-30
[web] Make use of /me/drives/sharedwithme endpoint


kulmann:
# Description

As soon as https://github.com/owncloud/ocis/issues/7436 is done we can switch over from OCS to graph for listing incoming shares.

## User Stories

* > As an Infinite Scale GraphApi consumer i need to collect data about all my incoming shares to make access to my data transparent.

## Value

Make use of NG Api to assist deprecating the existing OCS API

## Acceptance Criteria
- [ ] on the `Shared with me` page, use the `/me/drives/sharedwithme` graph endpoint to list incoming shares (see details about filters in the ocis issue)

## Definition of ready
- [ ] everybody needs to understand the value written in the user story
- [ ] acceptance criteria has to be defined
- [ ] all dependencies of the user story need to be identified
- [ ] feature should be seen from an end user perspective
- [ ] user story has to be estimated
- [ ] story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] functionality described in the user story works
  - [ ] acceptance criteria are fulfilled
- Quality
  - [ ] code review happened
  - [ ] CI is green
  - [ ] critical code received unit tests by the developer
- Non-functional requirements
  - [ ] no sonar cloud issues

PR #10392, opened 2024-01-23
refactor: use sharing NG for share listings

JammingBen:
## Description
Use the new sharing NG API for the share listings (Shared with/by me, Shared via link).

I also removed the `shareType` property on the `ShareResource` because it can have multiple different share types (e.g. when it's shared with another user as well as via link). The already existing `shareTypes` property now holds all the types. In addition to that, each item in the `sharedWith` array now holds information about the share type.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9893

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


tests/acceptance/features/webUISharingAcceptShares/acceptShares.feature
```diff
@@ -11,7 +11,7 @@ Feature: accept/decline shares coming from internal users
       | Brian    |
     And user "Brian" has logged in using the webUI
 
-  @issue-ocis-1950
+  @issue-ocis-1950 @skipOnOCIS
```
JammingBen, line 14: This test doesn't work in that form anymore because oCIS now sends 2 shares when one resource is being shared with a user directly as well as via group.

Could be refactored, but since it's an acceptance test and we won't have the concept of "declining" shares in the future I guess we can just let the test die.

tests/acceptance/features/webUISharingAcceptShares/acceptShares.feature
```diff
@@ -11,21 +11,6 @@ Feature: accept/decline shares coming from internal users
       | Brian    |
     And user "Brian" has logged in using the webUI
 
-  @issue-ocis-1950
-  Scenario: reject a share that you received as user and as group member
```
JammingBen, line 15: This test doesn't work in that form anymore because oCIS now sends 2 shares when one resource is being shared with a user directly as well as via group.

Could be refactored, but since it's an acceptance test and we won't have the concept of "declining" shares in the future I guess we can just let the test die.

packages/web-app-files/src/views/shares/SharedWithMe.vue
```diff
@@ -244,7 +246,11 @@ export default defineComponent({
     }
 
     const shareTypes = computed(() => {
-      const uniqueShareTypes = uniq(unref(paginatedResources).map((i) => i.shareType))
+      const uniqueShareTypes = uniq(
+        unref(paginatedResources)
+          .map((i) => i.shareTypes)
```
kulmann, line 251:
```suggestion
          .flatMap((i) => i.shareTypes)
```
😉

kulmann (changes requested):
Share names in `Shared with others` and `Shared via link` are broken, see screenshot. Folder names from the personal space become a single `.` instead in the outgoing share lists.

Screenshot 2024-01-30 at 14 34 57
Screenshot 2024-01-30 at 14 35 02

JammingBen:
> Share names in Shared with others and Shared via link are broken, see screenshot. Folder names from the personal space become a single . instead in the outgoing share lists.

Yep that was due to some of the server changes (https://github.com/owncloud/ocis/pull/8307) I was mentioning earlier 😄 The server now correctly sends the parent path without the resource name, hence we need to join both on our own.

kulmann (approved):
Don't know if that's an issue of this PR, but whenever I use row actions like hide/unhide/sync/unsync in the shared with me list, the avatar in the `shared with` column disappears. Everything else seems to work fine 💪 😍

JammingBen:
> Don't know if that's an issue of this PR, but whenever I use row actions like hide/unhide/sync/unsync in the shared with me list, the avatar in the `shared with` column disappears. Everything else seems to work fine 💪 😍

That's because changing the status/hidden state re-fetches the resource from the old ocs api, resulting in slightly different data. Gonna be fixed via the follow-up https://github.com/owncloud/web/pull/10421 👍

PR #10393, opened 2024-01-23
refactor: introduce ShareResource type

JammingBen:
## Description
Introduces a new `ShareResource` type that extends the `Resource` type by share relevant information. This allows us to better distinguish between normal and shared resources.

Also clarifies the chaos regarding file- and share-ownership props a bit. A `Resource` now always has an `owner` of the type `Identity` (includes `id` and `displayName`). A `ShareResource` additionally has `sharedBy` (of type `Identity`) and `sharedWith` (of type `Identity[]`).

## Related Issue

- Works towards https://github.com/owncloud/web/issues/9893

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -74,9 +74,6 @@
             <p class="oc-m-rm">
               {{ ownerDisplayName }}
               <span v-if="ownedByCurrentUser" v-translate>(me)</span>
-              <span v-if="!ownedByCurrentUser && ownerAdditionalInfo"
-                >({{ ownerAdditionalInfo }})</span
-              >
```
JammingBen, line 79: There is no additional info in a share resource, hence we can remove it here.

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -996,59 +997,100 @@ export default defineComponent({
...
+    getSharedWithAvatarDescription(resource: Resource) {
+      if (!isShareResource(resource)) {
+        return
+      }
+
       const resourceType =
         resource.type === 'folder' ? this.$gettext('folder') : this.$gettext('file')
-      const shareCount = resource.sharedWith.filter((u) => !u.link).length
-      const linkCount = resource.sharedWith.filter((u) => !!u.link).length
+
```
JammingBen, line 1028: The following lines might be a bit complicated to read. 

The previous implementation basically accumulated people and link shares into one description. But there is no scenario where we need the description for both at the same time because we're either looking at a people share resource, or a link share resource ("Shared by me" <> "Sharing via link"). As a result we can simply this here.

packages/web-app-files/src/components/Shares/SharedWithMeSection.vue
```diff
@@ -284,15 +284,9 @@ export default defineComponent({
         onExit: debounced.cancel
       })
     },
-    getShowSynchedIcon(resource: Resource) {
+    getShowSynchedIcon(resource: ShareResource) {
       return resource.status === ShareStatus.accepted
```
dschmidt, line 288: Maybe now is a good time to rename ` .status` to something semantically more meaningful?

packages/web-app-files/src/composables/resourcesViewDefaults/useResourcesViewDefaults.ts
```diff
@@ -65,7 +64,7 @@ export const useResourcesViewDefaults = <T, TT, TU extends any[]>(
   })
 
   const resourcesStore = useResourcesStore()
-  const { activeResources: storeItems } = storeToRefs(resourcesStore)
+  const storeItems = computed(() => resourcesStore.activeResources) as unknown as Ref<T[]>
```
dschmidt, line 67: if we need this more often, it might make sense to add this a function to the resourcesStore... fine for now

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -996,59 +997,94 @@ export default defineComponent({
...
+    isResourceClickable(resource: Resource) {
       if (!this.areResourcesClickable) {
         return false
       }
 
       // TODO: remove as soon as pending & declined shares are accessible
-      if (status === ShareStatus.pending || status === ShareStatus.declined) {
+      if (
+        isShareResource(resource) &&
+        (resource.status === ShareStatus.pending || resource.status === ShareStatus.declined)
```
dschmidt, line 1008:
```suggestion
        ([ShareStatus.pending, ShareStatus.declined].includes(resource.status)
```

dschmidt (approved):
I love love love love this - thank you so much for going down this rabbit hole

packages/web-app-files/src/components/Shares/SharedWithMeSection.vue
JammingBen: > Maybe now is a good time to rename  .status to something semantically more meaningful?

Yes, actually we should get rid of the `status` completely in favour of the new `syncEnabled` since we won't have statuses anymore 😄

packages/web-client/src/helpers/share/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  readBasic = 'libre.graph/driveItem/basic/read',
+  readPath = 'libre.graph/driveItem/path/read',
+  readQuota = 'libre.graph/driveItem/quota/read',
+  readContent = 'libre.graph/driveItem/content/read',
+  readChildren = 'libre.graph/driveItem/children/read',
+  readDeleted = 'libre.graph/driveItem/deleted/read',
+  updatePath = 'libre.graph/driveItem/path/update',
+  updateDeleted = 'libre.graph/driveItem/deleted/update',
+  deleteStandard = 'libre.graph/driveItem/standard/delete'
+}
```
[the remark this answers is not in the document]
JammingBen: Technically it belongs to the other PR, but since I needed to type `SharePermissions` here anyway, I figured why not do it with this change already.

PR #10067, opened 2023-11-26
Add documentation for extension system

pascalwengerter:
First iteration, let's get it in quickly and then iterate as the extension system evolves

PR #10394, opened 2024-01-23
Allow generic search result items for search providers

pascalwengerter:
## Description
Just tried writing a basic search provider while working on #10067 and realized search results were hardcoded to be webdav-specific. Figured I'd allow for a more generic approach, at least from a type perspective 🙃

### DIFF
--- a/packages/web-pkg/src/components/Search/types.ts
+++ b/packages/web-pkg/src/components/Search/types.ts
@@ -1,8 +1,15 @@
 import { SearchResource } from '@ownclouders/web-client/src/webdav/search'
 
+export interface GenericSearchResultItem {
+  id: string
+  displayName: string
+  icon?: string
+}
+
 export interface SearchResultValue {
-  id: string | number
-  data: SearchResource
+  id: string
+  // FIXME: SearchResource is very specific for webdav search
+  data: GenericSearchResultItem | SearchResource
 }
 
 export interface SearchResult {
@@ -11,7 +18,7 @@ export interface SearchResult {
 }
 
 export interface SearchList {
-  component: unknown
+  component: unknown // Component needs to match SearchResultValue
 
   search(term: string): Promise<SearchResult>
 }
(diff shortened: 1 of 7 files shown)


packages/web-pkg/src/composables/piniaStores/extensionRegistry.ts
```diff
@@ -7,11 +7,12 @@ import { Item } from '@ownclouders/web-client/src/helpers'
 import { useConfigStore } from './config'
 import { FolderView } from '../../ui'
 
+export type AvailableExtensions = 'action' | 'search' | 'sidebarNav' | 'sidebarPanel' | 'folderView'
```
pascalwengerter, line 10: @dschmidt are you fine with this?

dschmidt: If you only allow these as type... you are preventing apps from registering their own extension types 😏

pascalwengerter: Good point, thx!

packages/web-pkg/src/components/Search/types.ts
```diff
@@ -1,8 +1,16 @@
...
+  name: string
+  icon?: string
+}
 
 export interface SearchResultValue {
-  id: string | number
-  data: SearchResource
+  id: string
+  // FIXME: SearchResource is very specific for webdav search
+  data: GenericSearchResultItem | SearchResource
```
JammingBen, line 13: Just thinking, do we really need the `SearchResource` here then? 🤔 Shouldn't it be as generic as possible in `web-pkg`?

Search implementations like the webdav search could still have their own type, as long as they implement the generic type.

pascalwengerter: The purpose of this PR was to widen the current typing a little, I don't have a strong opinion on how generic it ideally _should_ be 😇

JammingBen: Alright, I guess the purpose of this PR is served with this 😄

JammingBen (approved):

kulmann (approved):

PR #10408, opened 2024-01-26
chore: remove lookacat ssh key from cd instances

kulmann:
😢


dschmidt (approved):
I don't want to approve this ...

Issue #10380, opened 2024-01-22
Nightly failures on multiple unit tests


PrajwolAmatya:
## Description
Multiple unit tests failed with the following log:

```
FAIL packages/web-app-preview/tests/unit/components/MediaControls.spec.ts
  ● Test suite failed to run

    A jest worker process (pid=138) was terminated by another process: signal=SIGKILL, exitCode=null. Operating system logs may contain more information on why this occurred.

      at ChildProcessWorker._onExit (node_modules/.pnpm/jest-worker@29.7.0/node_modules/jest-worker/build/workers/ChildProcessWorker.js:370:23)

FAIL packages/web-app-admin-settings/tests/unit/views/Groups.spec.ts
  ● Test suite failed to run

    A jest worker process (pid=125) was terminated by another process: signal=SIGKILL, exitCode=null. Operating system logs may contain more information on why this occurred.

      at ChildProcessWorker._onExit (node_modules/.pnpm/jest-worker@29.7.0/node_modules/jest-worker/build/workers/ChildProcessWorker.js:370:23)
```

Builds:
- https://drone.owncloud.com/owncloud/web/42019/8/4
- https://drone.owncloud.com/owncloud/web/42021/8/4

Salipa-Gurung:
Unit tests failed with log:
```
FAIL packages/web-app-files/tests/unit/views/spaces/GenericTrash.spec.ts
  ● Test suite failed to run

    A jest worker process (pid=125) was terminated by another process: signal=SIGKILL, exitCode=null. Operating system logs may contain more information on why this occurred.

      at ChildProcessWorker._onExit (node_modules/.pnpm/jest-worker@29.7.0/node_modules/jest-worker/build/workers/ChildProcessWorker.js:370:23)
```
Build: https://drone.owncloud.com/owncloud/web/42119/8/4

kulmann:
The actual issue is stated above the failures:

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```

Could you try to run the unit tests with the `--detectOpenHandles` option and see if it detects anything?

JammingBen:
Some more insight from my side (basically copied from chat):

When we first stumbled upon this issue a few months ago, limiting the `maxWorkers` to 50% helped. But the pipeline runs with this setting as well, so apparently it doesn't help anymore.

[...]

PR #10412, opened 2024-01-26
test: switch from jest to vitest

JammingBen:
## Description
Removes `jest` in favour of `vitest` for our unit test suite. The API remains nearly unchanged (expect for that `jest` now became `vi`).

Unfortunately we lose our nice snapshot formatter. We have to see if there is some äquivalent for vitest.

Possibly/hopefully fixes https://github.com/owncloud/web/issues/10380.

## Types of changes

- [ ] Technical debt
- [x] Tests
- [ ] Documentation

# Followup Tasks
- [ ] eslint rules
- [ ] vscode integration
- [ ] happy dom
- [ ] sharing?
- [ ] single config file for vite and vitest
- [ ] remove HttpClient cancel function which makes little sense from our pov


packages/web-pkg/tests/unit/helpers/datetime.spec.ts
```diff
@@ -25,7 +25,8 @@ describe('datetime helper', () => {
     })
   })
   describe('formatDateFromJSDate', () => {
-    it('should give correct output', () => {
+    // FIXME
```
dschmidt, line 28: Is this fixed after we're now setting the timezone via env var maybe?

packages/web-pkg/tests/unit/composables/piniaStores/shares.spec.ts
```diff
@@ -59,7 +59,9 @@ describe('useSharesStore', () => {
             { id: 3, outgoing: false, indirect: false },
             { id: 4, outgoing: false, indirect: true }
           ])
-          expect(clientService.owncloudSdk.shares.getShares).toHaveBeenCalledTimes(4)
+
+          // FIXME: pqueue makes issues?!
+          // expect(clientService.owncloudSdk.shares.getShares).toHaveBeenCalledTimes(4)
```
dschmidt, line 64: uaaaaaaarghs - what do we make out of this?

packages/web-pkg/tests/unit/composables/actions/useOpenWithDefaultApp.spec.ts
```diff
@@ -1,11 +1,11 @@
...
-import { mock } from 'jest-mock-extended'
+import { mock } from 'vitest-mock-extended'
 import { Resource, SpaceResource } from '@ownclouders/web-client'
 import { useFileActions, Action, useOpenWithDefaultApp } from '../../../../src/composables'
 
-jest.mock('../../../../src/composables/actions/files', () => ({
-  ...jest.requireActual('../../../../src/composables/actions/files'),
-  useFileActions: jest.fn()
+vi.mock('../../../../src/composables/actions/files', async (importOriginal) => ({
+  ...(await (importOriginal() as any)),
```
dschmidt, line 7: use any as type argument?

packages/web-pkg/tests/unit/composables/actions/files/useFileActionsCreateLink.spec.ts
```diff
@@ -2,15 +2,15 @@ import { computed, unref } from 'vue'
...
 import { Resource } from '@ownclouders/web-client'
 import { useCreateLink, useDefaultLinkPermissions } from '../../../../../src/composables/links'
 import { SharePermissionBit } from '@ownclouders/web-client/src/helpers'
 
-jest.mock('../../../../../src/composables/links', () => ({
-  ...jest.requireActual('../../../../../src/composables/links'),
-  useCreateLink: jest.fn(),
-  useDefaultLinkPermissions: jest.fn()
+vi.mock('../../../../../src/composables/links', async (importOriginal) => ({
+  ...(await (importOriginal() as any)),
```
dschmidt, line 11: type arg syntax?

packages/web-app-search/tests/unit/portals/SearchBar.spec.ts
dschmidt, line 1: Would it be potentially better to use fake timers than sleeping 600ms?

Not that it matters thaaaat much in this single test but in theory...

dschmidt (approved):

packages/web-runtime/tests/unit/components/Topbar/FeedbackLink.spec.ts
```diff
@@ -1,21 +1,19 @@
...
-        }
-      })
-    ).toHaveNoViolations()
+    // expect(
+    //   await axe(wrapper.html(), {
+    //     rules: {
+    //       region: { enabled: false }
+    //     }
+    //   })
+    // ).toHaveNoViolations()
```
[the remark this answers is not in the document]
JammingBen: There is a comment already, this test has always been flaky and I don't know why and how to fix it 🤷

packages/web-runtime/tests/unit/components/Topbar/__snapshots__/FeedbackLink.spec.ts.snap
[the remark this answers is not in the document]
JammingBen: As you correctly assumned that's because the corresponding test is being skipped and never actually ran once with vitest.

tests/unit/config/jest.overrides.ts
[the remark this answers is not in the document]
JammingBen: Good catch, I don't know to be honest. Definitely something we should check.

packages/web-pkg/tests/unit/composables/piniaStores/shares.spec.ts
JammingBen: We didn't find a solution 😅  That's the only place where this causes issues...

packages/web-runtime/tests/unit/components/Topbar/FeedbackLink.spec.ts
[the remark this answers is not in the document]
kulmann: Oh I thought we don't use `axe` anymore... if I remember correctly this was just a quick test / use case for automated accessibility tests. Since we didn't follow up I'd say completely get rid of it for now (here and in `OcTable`).

dschmidt (approved):

packages/web-app-search/tests/unit/portals/SearchBar.spec.ts
JammingBen: I tried it, but it didn't work for some reason. [The docs](https://vitest.dev/guide/mocking.html#timers) are mentioning only `setTimeout ` and `setInterval`... maybe we can't use it for lodash's `debounce`.

kulmann (approved):
💪

Issue #9897, opened 2023-10-31
e2e tests for url stability


kulmann:
## Description
Clients rely on URLs being correctly resolved in ownCloud Web for some features. We had some regressions recently (see links below), so I'd like to make sure with some new e2e tests that URLs in ownCloud Web resolve reliably.

### Recent regressions
1. https://github.com/owncloud/web/issues/9868
2. https://github.com/owncloud/web/issues/9867
3. https://github.com/owncloud/web/issues/9804

## New e2e tests

I'd like to see a new URL journey in web with at least the following features. Open for suggestions for additional features.

### File details panels (regression no. 1)
The right sidebar within the files app holds a few details panels like `details`, `sharing` or `versions`. Clients can open the web ui in the file list with the `versions` panel opened in the right sidebar by appending `?details=versions` to the private link. Same for the sharing panel by appending `?details=sharing` to the private link. This behaviour must remain stable.

### Resolving all plain private links (regression no. 2)
Clients have a feature for showing a file or folder in the web browser. This opens the private link (without additional url params) in the web browser. Currently the default behaviour of ownCloud Web is, that it shows the file in the file list if there is no editor for the file. And it opens the default viewer/editor for the file if there is one. This is fine like it is. A new e2e test should check that private links open in the web ui for cases: a) a file without viewer/editor, b) a file with viewer/editor, c) a folder, d) the root of a space, e) the (hardcoded!) share jail id. The last one (share jail) is an edge case which is currently broken. It can't resolve in web, because there is no file listing which shows the share jail itself as a child element. So this needs to redirect to the shared with me page (currently broken).

### open-with-web endpoint (regression no. 3)
Clients can make a request to the `open-with-web` ocis endpoint to obtain the browser URL for opening the respective file in an app from the app provider, e.g. opening a file in OnlyOffice. Was recently broken, this is a very important product feature and must remain stable.

kulmann:
@ScharfViktor do you have additional URL related features in mind that should be represented in this journey?

kulmann:
@tbsbdr Fine to add this to one of the next sprints? I'd like to give this into the hands of @ScharfViktor

kulmann:
@michaelstingl cc

SwikritiT:
Hi @kulmann for the 3rd regression
I tried getting the URL using the for collabora with following curl request

```
[...]

PR #10420, opened 2024-01-29
[full-ci] url stability. open file through Collabora

ScharfViktor:
part of #9897 

check that file can be open in Collabora via url. client feature:

Screenshot 2024-01-29 at 21 36 03

### DIFF
--- a/tests/e2e/support/objects/url-navigation/actions.ts
+++ b/tests/e2e/support/objects/url-navigation/actions.ts
@@ -17,6 +17,7 @@ export interface openResourceViaUrlArgs {
   resource?: string
   user: User
   space?: string
+  editorName?: string
 }
 
 export const navigateToDetailsPanelOfResource = async (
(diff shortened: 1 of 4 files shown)


tests/e2e/support/objects/url-navigation/actions.ts
```diff
@@ -29,9 +30,12 @@ export const navigateToDetailsPanelOfResource = async (
 }
 
 export const openResourceViaUrl = async (args: openResourceViaUrlArgs) => {
-  const { page, resource, user, space } = args
+  const { page, resource, user, space, collabora } = args
```
kulmann, line 33: Wouldn't it make sense to pass the editor name as variable instead of a boolean? When want to test the same for OnlyOffice, don't we? 🤔

ScharfViktor: I just didn't find way how to open file in the OnlyOffice using url (wanted to ask about it tomorrow.)
Even if I set the `app=OnlyOffice` parameter, the file still opens in `collabora`

`https://ocis.owncloud.test/external/open-with-web?app=OnlyOffice&fileId=fileuuId`

ScharfViktor: I spoke with @JammingBen. Looks like it's a bug

> Wouldn't it make sense to pass the editor name as variable instead of a boolean? When want to test the same for OnlyOffice, don't we?

sure, I'll change it

kulmann: That would be another regression, right? Super important to have the test for both Collabora and OnlyOffice then 😁

ScharfViktor: created bug for that: https://github.com/owncloud/web/issues/10422

ScharfViktor:
while the tests are like this 
`.odt` I open via Collabora
`.docx` I open via OnlyOffice

after fixing https://github.com/owncloud/web/issues/10422 I'll add tests where I can open one file format in the different editor

saw-jan (approved):
LGTM

Issue #10418, opened 2024-01-29
Sharing NG


JammingBen:
Sharing NG is the next gen sharing API in oCIS that deprecates the old OCS API. The purpose of this issue is to keep track of all the things that need to be done in Web to fully make the switch to the new API.

### Basic functionality

- [x] Implement sharing NG for listing shares (https://github.com/owncloud/web/pull/10392)
- [x] Implement sharing NG for sidebar share loading & CRUD (https://github.com/owncloud/web/pull/10433)
- [ ] Make proper use of permissions (https://github.com/owncloud/web/issues/10770)
- [ ] Denies (currently blocked by the server)

### Bugs

- [x] Password modal is missing when passwords are enforced for links and an internal link is being updated to the type `edit`. (https://github.com/owncloud/web/pull/10690)
- [x] Banned password error message isn't captalized and therefore can't be translated (https://github.com/owncloud/web/pull/10685)
- [x] The role display names and descriptions are not as they were before (server issue)
- [x] Some role display names and descriptions are not being translated yet, they need to be added to the additional translations for Transifex to pick them up. (https://github.com/owncloud/web/pull/10695)
- [x] Share date is missing (https://github.com/owncloud/web/issues/10778)
- [x] rework sharing indicators since there is no more resharing (https://github.com/owncloud/web/pull/10759)

### Hacks to be removed when server is ready

- [x] Tokeninfo for public links (see issue https://github.com/owncloud/web/issues/10689)
- [x] `hidden` property (https://github.com/owncloud/web/issues/10768)
- [x] Suggested roles (https://github.com/owncloud/web/pull/10774)
- [x] Space root endpoint for space roots (https://github.com/owncloud/web/pull/10745)
- [x] Role icon map (https://github.com/owncloud/web/issues/10777)

### Tests

- [x] Add unit tests (https://github.com/owncloud/web/pull/10704 & https://github.com/owncloud/web/pull/10710)
- [x] Adjust e2e and acceptance test role selectors and remove human readable id mappings (https://github.com/owncloud/web/pull/10684)

### Checks

- [ ] Check that federated sharing (OCM) still works (currently blocked by https://github.com/owncloud/ocis/issues/8564)
- [ ] Check that `fullShareOwnerPaths` still work as expected
- [x] Check if we still need to distinguish between space and non-space share types. Or more radical: can we remove the concept of share types as a whole? (https://github.com/owncloud/web/pull/10701)
- [x] Check that the capabilities for passwords are respected correctly
[...]

PR #10421, opened 2024-01-30
refactor: remove share status in favour of syncEnabled

JammingBen:
## Description
Remove share status in favour of the `syncEnabled` property since sharing NG won't have share statuses anymore. Also renames the actions and test steps accordingly.

The remaining places where we still use the status (e.g. `triggerShareAction`) are deprecated and will be removed with one of the next changes.

In addition to that I decided to introduce `OutgoingShareResource` and `IncomingShareResource` interfaces that extend `ShareResoruce`, since some properties are only available on incoming share resources. This makes type checking easier and even allows us to kill some of the pesky route checks (e.g. `IncomingShareResource` == "Shared with me" page).

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -289,7 +293,7 @@ export default defineComponent({
      * - modificationDate: The date of the last modification of the resource
      * - shareDate: The date when the share was created
      * - deletionDate: The date when the resource has been deleted
-     * - status: The status of the share. Contains also actions to accept/decline the share
+     * - status: The sync status of the share
```
dschmidt, line 296: the diff is hard to read, but shouldn't this be `- syncEnabled:`

packages/web-pkg/src/composables/actions/files/useFileActionsCopyQuicklink.ts
```diff
@@ -89,11 +85,10 @@ export const useFileActionsCopyQuickLink = () => {
         if (resources.length !== 1) {
           return false
         }
-        if (isLocationSharesActive(router, 'files-shares-with-me')) {
-          if (isShareResource(resources[0]) && resources[0].status !== ShareStatus.accepted) {
-            return false
-          }
+        if (isIncomingShareResource(resources[0]) && !resources[0].syncEnabled) {
+          return false
```
dschmidt, line 89: sweeeeeeeeeeeeeeeeeeeeeeeeeeeeet

packages/web-pkg/src/composables/actions/files/useFileActionsToggleHideShare.ts
```diff
@@ -41,19 +33,17 @@ export const useFileActionsToggleHideShare = () => {
       triggerPromises.push(
         triggerQueue.add(async () => {
           try {
-            const share = await triggerShareAction({
+            await triggerShareAction({
               resource,
-              status: resource.status,
+              status: resource.syncEnabled ? 0 : 2,
```
dschmidt, line 38: This is probably for the time being until we're moving towards the graph client completely?

packages/web-pkg/src/helpers/share/triggerShareAction.ts
```diff
@@ -1,30 +1,18 @@
...
+/** @deprecated */
 export async function triggerShareAction({
   resource,
   status,
-  hasResharing,
-  hasShareJail,
   client,
-  hidden = undefined,
-  spaces = [],
-  fullShareOwnerPaths = false
```
dschmidt, line 18: Reminds me - have you tried this with fullShareOwnerPaths? 🙈 👀

dschmidt (approved):
Greeeeaaaaaat! 🎉 

Personally I probably wouldn't have gotten rid of the ShareStatus enum/class just yet, because I feel the magic numbers will be harder to review in the upcoming iterations ... but I wouldn't revert the change now either

packages/web-client/src/helpers/share/functions.ts
```diff
@@ -160,12 +167,13 @@ export function buildSharedResource(
...
-    sharedWith: share.sharedWith || []
+    sharedWith: share.sharedWith || [],
+    outgoing: !incomingShares
   }
 
   if (incomingShares) {
-    resource.status = parseInt(share.state)
-    resource.hidden = share.hidden === 'true' || share.hidden === true
+    ;(resource as IncomingShareResource).syncEnabled = parseInt(share.state) === 0
+    ;(resource as IncomingShareResource).hidden = share.hidden === 'true' || share.hidden === true
```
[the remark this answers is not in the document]
JammingBen: This will be removed later today, so I decided to not invest any time 😄

packages/web-pkg/src/composables/actions/files/useFileActionsToggleHideShare.ts
JammingBen: This will be removed as one of the next steps, so again I decided to not invest much time into making it more readable.

packages/web-pkg/src/helpers/share/triggerShareAction.ts
JammingBen: Not yet, it will probably break eventually with the transition to sharing NG (or tbh I think the first PR already broke it 😅 ). But I'd like to get this transition going first, especially because code paths like these will change in a few days anyway.

I added it to the epic though: https://github.com/owncloud/web/issues/10418

JammingBen:
> Personally I probably wouldn't have gotten rid of the ShareStatus enum/class just yet, because I feel the magic numbers will be harder to review in the upcoming iterations ... but I wouldn't revert the change now either

Yeah, I was thinking about the same but decided for the clean cut here because the remaining magic numbers will be removed as one of the next steps anyway (which is basically the `triggerShareAction` method).

PR #10426, opened 2024-01-31
refactor: remove aggregateResourceShares and share prop on share resources

JammingBen:
## Description
* Removes `aggregateResourceShares` because we don't need it anymore. The proper way to build share resources is via `buildIncomingShareResource` and `buildOutgoingShareResource`.
* Removes the `share` property on `ShareResource` because a) there is no need to expose the full share on a resource and b) a `ShareResource` can "hold" multiple shares, rendering that property basically useless.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-app-files/src/components/SideBar/Versions/FileVersions.vue
```diff
@@ -120,10 +119,6 @@ export default defineComponent({
         if (unref(resource).permissions !== undefined) {
           return unref(resource).permissions.includes(DavPermission.Updateable)
         }
-        const res = unref(resource)
-        if (isShareResource(res) && res.share?.role) {
-          return res.share.role.hasPermission(SharePermissions.update)
-        }
```
JammingBen, line 126: File versions currently can't be loaded for shares (including files inside a shared folder), so this check is not needed at the moment. If needed later it should check the graph permissions instead.

dschmidt (approved):
crazy 😍

PR #10430, opened 2024-01-31
refactor: adjust building share resources after server changes

JammingBen:
## Description
Bumps the libre graph client and adjusts the way how share resources are built due to recent API changes (see https://github.com/owncloud/ocis/pull/8322). Multiple shares for the same resource are now being returned as one drive item including multiple permissions. That means no more duplicated resources in the files list.

The `shareId` and the `shareRoot` on a `Resource` are now somewhat misleading, since a resource can hold multiple "shares" (= permissions). We still need those props to be able to resolve into the share jail, but we should remove this in the future.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -68,7 +68,10 @@
             <span v-text="sharedByDisplayNames" />
           </td>
         </tr>
-        <tr v-if="ownerDisplayName" data-testid="ownerDisplayName">
+        <tr
+          v-if="ownerDisplayName && ownerDisplayName !== sharedByDisplayNames"
+          data-testid="ownerDisplayName"
+        >
           <th scope="col" class="oc-pr-s oc-font-semibold" v-text="$gettext('Owner')" />
           <td>
             <p class="oc-m-rm">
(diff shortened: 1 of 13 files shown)


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -65,16 +65,7 @@
         <tr v-if="showSharedBy" data-testid="shared-by">
           <th scope="col" class="oc-pr-s oc-font-semibold" v-text="$gettext('Shared by')" />
           <td>
-            <span v-text="sharedByDisplayName" />
-          </td>
-        </tr>
-        <tr v-if="ownerDisplayName" data-testid="ownerDisplayName">
```
kulmann, line 71: Do we still need to show the owner for CERN? 🤔

JammingBen: > Do we still need to show the owner for CERN?

Probably depends on how they will be using sharing NG 😅 

> Maybe we can keep the owner, but hide it if it's identical with the share owner.

In which scenario would they be different though? Because for project spaces we don't have file owners. And in the personal space we don't show the file owner anyway because it's always the current user.

dschmidt (approved):
LGTM when @kulmann 's remark is resolved

packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
kulmann: You might have an owner without any share. CERN allows access to files of other users without a share (shared root folder without an ownCloud share).

kulmann (approved):
LGTM 👍

PR #8851, opened 2023-04-18
Restructure account page - GDPR export

hurradieweltgehtunter:
## Description
Re-Structure the account page. Gives the GDPR export an own section - it felt misplaced under "Account information".
Also, the request and download buttons look as buttons now and a line of information regarding §20 GDPR is added.

### Before
grafik

### After
grafik

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

PR #10435, opened 2024-02-02
Restructure GDPR export / turn into dedicated section on accounts page

pascalwengerter:
## Description
Supersedes stale #8851 and drops unrelated changes

Colorful borders only to make changes more obvious. As discussed with @kulmann the buttons shouldn't be filled-primary since it's not _the_ main action on the page

Screenshot 2024-02-02 at 08 59 59


JammingBen (approved):
Nice, I like it!

PR #10452, opened 2024-02-07
[full-ci] chore: Vue 3.4

dschmidt:
## Description
This updates Vue to 3.4 and with it `@vue/*` and `vue-tsc`

## Motivation and Context
Stay up to date.

For some reason we're now running into more issues with useRouteQuery race conditions, that's why I've implemented a similar approach to what we tried in #8414. ~~Unfortunately it feels a little sluggish in development builds, but seems to be okay in production. I'm trying to figure out if there is some feature that we can disable for development environments to speed it up.~~ Cannot reproduce this now :-| 

Needs thorough testing ... 

If anyone has a better approach I'm all ears :-\ 

## Types of changes

- [x] Technical debt
- [x] Tests
- [ ] Documentation

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -10,6 +10,9 @@ hoist-pattern[]=sass
 # eslint (public so they are picked up by IDEs)
 public-hoist-pattern[]=*eslint*
 
-
 # Make ODS available to tests, even if the main package does not require it
 public-hoist-pattern[]=@ownclouders/design-system
+
+# c.f. https://github.com/vuejs/test-utils/pull/2197
+hoist-pattern[]=@vue/server-renderer
+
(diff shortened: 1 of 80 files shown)


kulmann:
> Code looks good... however, I found two weird things when smoke testing the UI:
> 
> * the active nav item state for the left sidebar is incorrect on the `Shares` nav item. The `Shares` nav item has a grey background when you navigate to the shared with me page  and is not active at all when you navigate to `Shared with others` or `Shared via link` subsequently. In all of these cases the `Shares` nav item is supposed to be rendered as active element, i.e. blue background. On `master` this works fine, so it seems to be fallout of this PR.
> * When opening an image in the `preview` app and then navigating back/forth within the `preview` app, there is an error on the browser console about failing to remove and event listener.
> 
> Didn't find anything else, good job 👍

Oooh, the second issue is on `master`. So it's only the first one for this PR.

kulmann (approved):
💪

Issue #10464, opened 2024-02-09
epub preview polishing


tbsbdr:
1. optimize for ipad
    - add buttons to change font size (incl. chapter-overview)
    - nice to have: collapse sidebar (+button to do so)
    - nice to have: swiping
2. remember chapters
4. support dark mode
5. polish chapter sidebar
   - geyish and boxed like on https://epub-reader.online/ , and see visuals
   - truncate chapters

## Visual
![image](https://github.com/owncloud/web/assets/26610733/596ef41f-627e-4583-9cc6-d87f098d13ce)
![image](https://github.com/owncloud/web/assets/26610733/93e5de23-41ee-48f5-bb70-dc3a0859fed8)

## visual example from epub-reader.online
![screenshot_002364](https://github.com/owncloud/web/assets/26610733/3a63089f-ff65-432c-bac1-d9d84e1f6b77)

PR #10470, opened 2024-02-10
Make iPad usage comfortable

JanAckermann:
## Description
- Add font size increase/decrease buttons
- Add chapter select for smaller texts 

image

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10464

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


JanAckermann:
I also increased the left side bar chapter navigation paddings and text sizes again, because it looked somewhat messy, please give feedback

packages/web-app-epub-reader/src/App.vue
```diff
@@ -163,13 +222,27 @@ export default defineComponent({
...
+      FONT_SIZE_PERCENTAGE_STEP
     }
   }
 })
 </script>
 <style lang="scss">
 .epub-reader {
+  &-controls {
+    &-chapters-select {
+      @media (min-width: $oc-breakpoint-large-default) {
```
kulmann, line 240: we already have a class for that: `oc-hidden@l`

kulmann (changes requested):
Having a small css nag, see comment 😁 

I had the feeling that I would like to know the current font size percentage if != 100%. Currently it's only possible to deduct the current percentage from the tooltips on the two buttons. Do you have a good idea for that maybe?

packages/web-app-epub-reader/src/App.vue
```diff
@@ -44,6 +84,7 @@ import {
 } from '@ownclouders/web-pkg'
 import ePub, { Book, NavItem, Rendition, Location } from 'epubjs'
 import { DisplayedLocation } from 'epubjs/types/rendition'
+import { $gettext } from '@ownclouders/web-pkg/src/router/utils'
```
JanAckermann, line 87: 🤬🤬🤬

packages/web-app-epub-reader/src/App.vue
```diff
@@ -79,6 +122,7 @@ export default defineComponent({
     const currentChapter = ref<NavItem>()
     const navigateLeftDisabled = ref(true)
     const navigateRightDisabled = ref(false)
+    const fontSizePercentage = ref(100)
```
JanAckermann, line 125: current...

packages/web-app-epub-reader/src/App.vue
```diff
@@ -4,36 +4,76 @@
...
+            class="epub-reader-controls-font-size-reset"
+            gap-size="none"
+            @click="resetFontSize"
+            v-text="`${fontSizePercentage}%`"
+          />
+          <oc-button
+            v-oc-tooltip="`${fontSizePercentage + FONT_SIZE_PERCENTAGE_STEP}%`"
+            class="epub-reader-controls-font-size-increare"
+            :disabled="increaseFontSizeDisabled"
+            gap-size="none"
```
JanAckermann, line 43: Rm ?

packages/web-app-epub-reader/src/App.vue
```diff
@@ -163,9 +237,17 @@ export default defineComponent({
...
+      increaseFontSize,
+      decreaseFontSize,
+      increaseFontSizeDisabled,
+      decreaseFontSizeDisabled,
+      fontSizePercentage,
+      FONT_SIZE_PERCENTAGE_STEP
     }
-  }
+  },
+  methods: { $gettext }
```
kulmann, line 250: No options API please. Why do you need it at all? When I remove it my IDE still shows $gettext correctly in the template...

packages/web-app-epub-reader/src/App.vue
kulmann: looks fine like it is now.

packages/web-app-epub-reader/src/App.vue
```diff
@@ -4,36 +4,76 @@
...
       </li>
     </oc-list>
-    <div class="oc-flex oc-flex-center oc-width-1-1">
-      <div class="oc-flex oc-flex-middle oc-mx-l">
-        <oc-button :disabled="navigateLeftDisabled" appearance="raw" @click="navigateLeft">
-          <oc-icon name="arrow-left-s" fill-type="line" size="xlarge" />
-        </oc-button>
+    <div class="oc-width-1-1 oc-height-1-1">
+      <div class="epub-reader-controls oc-flex oc-flex-middle oc-m-s">
+        <div class="oc-flex oc-button-group">
```
kulmann, line 21: Button group breaks into two lines. Also, is it intended that the dropdown spans the full width?

Screenshot 2024-02-20 at 06 29 59

packages/web-app-epub-reader/src/App.vue
```diff
@@ -4,36 +4,76 @@
...
+            gap-size="none"
+            @click="increaseFontSize"
+          >
+            <oc-icon name="font-family" fill-type="none" size="small" />
+            <oc-icon name="add" size="xsmall" />
+          </oc-button>
+        </div>
+        <oc-select
+          v-model="currentChapter"
+          class="epub-reader-controls-chapters-select oc-width-1-1 oc-px-s oc-hidden@l"
```
kulmann, line 52: Don't know why, but I don't have a `pointer` cursor when hovering over the OcSelect. Could you add the `oc-cursor-pointer` class and check if that helps?

kulmann (changes requested):
I like where this is going, really nice improvements! 👨‍🎨 

Left a few comments, nothing big.

But I do have one wish: Could you please add some unit tests? Buttons for font size reset and navigating back/forth should be easy to do. Chapter selection would make sense as well. And that small and big screen sizes show a different chapter selection. Happy to have that in a followup!

packages/web-app-epub-reader/src/App.vue
JanAckermann: I don't need it at all, I marked a comment already. My ide is always auto importing it 🤬

packages/web-app-epub-reader/src/App.vue
JanAckermann: I think a big select helps on mobile view (iPad) what do you think ? Better on the initial screenshot ?

JanAckermann:
Yes! I planned unit tests and a changelog item in a follow up

packages/web-app-epub-reader/src/App.vue
JanAckermann: Could you compare to our other oc selects ? I don't see a difference (even if cursor is not a pointer but smth else)

packages/web-app-epub-reader/src/App.vue
kulmann: ah, ok, so only a reminder for yourself 👍

packages/web-app-epub-reader/src/App.vue
kulmann: hmm... don't have a hard preference and on even smaller screen it makes sense to use the full width. so maybe best to just keep it like this and wait for feedback.

packages/web-app-epub-reader/src/App.vue
kulmann: Aaaah, found it. The `:searchable="false"` attribute is missing. :-) Please add it, no free text input allowed for this dropdown, only selecting the predefined options is allowed.

kulmann (approved):
Really, really awesome! 😍 

One last thing: please store the font size in the localStorage as well. 😇

packages/web-app-epub-reader/src/App.vue
```diff
@@ -227,6 +232,10 @@ export default defineComponent({
 
     watch(currentFontSizePercentage, () => {
       rendition.themes.fontSize(`${unref(currentFontSizePercentage)}%`)
+      localStorageData.value = {
+        ...localStorageData,
```
kulmann, line 236: Needs to be `...unref(localStorageData)`

packages/web-app-epub-reader/src/App.vue
```diff
@@ -155,6 +230,14 @@ export default defineComponent({
       }
     )
 
+    watch(currentFontSizePercentage, () => {
+      rendition.themes.fontSize(`${unref(currentFontSizePercentage)}%`)
+      localStorageData.value = {
+        ...localStorageData,
```
JanAckermann, line 236:
```suggestion
        ...unref(localStorageData),
```

Issue #10538, opened 2024-03-04
Align lock and postprocessing info with sharing indicators


kulmann:
# Description

## User Stories

* > As a user, I want to see important status information at a glance so that I immediately know which status a file/folder has.

## Value
Make status info more prominent.

## Acceptance Criteria
- Rename the `Shares` column in the file list to `Status`
- Move `lock` and `postprocessing` info into the `Status` column (= share indicators are now status indicators)
- Remove the badge from ResourceIcon and preview image
- Tiles view: try to fit the status indicators in between the file name and the context menu button (might need to increase the base tile size a bit in the theme)

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

PR #10544, opened 2024-03-05
Align lock and postprocessing info with sharing indicators

AlexAndBear:
## Description
image
image

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10538

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -1340,58 +1336,10 @@ export default defineComponent({
...
-
-  .oc-icon {
-    svg {
-      fill: var(--oc-color-background-highlight) !important;
-    }
-  }
-
-  .oc-spinner {
-    color: var(--oc-color-background-highlight) !important;
+// Show tooltip on status indicators without handler
```
AlexAndBear, line 1339: Move to disabled

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -518,6 +522,17 @@ export default defineComponent({
 
     const createNewFolderAction = computed(() => unref(createNewFolder)[0].handler)
 
+    const getIndicators = (resource: Resource) => {
+      if (unref(folderView).name !== FolderViewModeConstants.name.tiles) {
+        return resource.indicators
+      }
+
+      const visibleTilesViewIndicators = ['resource-locked', 'resource-processing']
```
kulmann, line 530: What do you think about introducing a new variable on the indicator interface, named like `critical`, `system`, whatever, and filter for that? I.e. tiles view only shows indicators that are `critical` while other views can decide to show all indicators? Don't know if this is a good solution, but I don't like to filter on the names of the indicators.

packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -92,6 +92,13 @@
               @row-mounted="rowMounted"
               @sort="handleSort"
             >
+              <template #indicators="{ resource }">
```
kulmann, line 95: I'd prefer to have the indicators implemented in the `ResourceTable` and `ResourceTiles` components, respectively. I.e. ResourceTable remains unchanged (shows all indicators) while in the ResourceTiles you can do the filtering. The template part is not so much code duplication. In my opinion having the logic (show all indicators vs. show a reduced set of indicators) in the respective components is more important than avoiding the 5-line code duplication.

packages/web-app-files/src/views/spaces/GenericSpace.vue
kulmann: (this would need to be moved to the `ResourceTiles` component as well if you agree with my other comment on this file).

kulmann (changes requested):
Conceptual question in the comments... visually this is such an improvement! Much easier for the eye! 💯

packages/web-app-files/src/views/spaces/GenericSpace.vue
AlexAndBear: > What do you think about introducing a new variable on the indicator interface, named like `critical`, `system`, whatever, and filter for that? I.e. tiles view only shows indicators that are `critical` while other views can decide to show all indicators? Don't know if this is a good solution, but I don't like to filter on the names of the indicators.

Suggesting variable name "category" and values could be "sharing" and "system"

JammingBen (approved):
LGTM 👍 

I feel like we should add a proper type for resource indicators as a follow-up task.

AlexAndBear:
> LGTM 👍 
> 
> 
> 
> I feel like we should add a proper type for resource indicators as a follow-up task.

Sure 👍 but I think this sprint is quite too tight to do so

kulmann (approved):
loving it!  ❤️‍🔥

PR #10545, opened 2024-03-05
[full-ci]change asset path

ScharfViktor:
changed path in the docker-compose because `WEB_ASSET_PATH` is deprecated 
it was reason why I could not build fresh web

see here: https://github.com/owncloud/ocis/pull/8523#issuecomment-1971017580

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1210,7 +1210,7 @@ def ocisService(type, tika_enabled = False, enforce_password_public_link = False
         "LDAP_GROUP_SUBSTRING_FILTER_TYPE": "any",
         "LDAP_USER_SUBSTRING_FILTER_TYPE": "any",
         "PROXY_ENABLE_BASIC_AUTH": True,
-        "WEB_ASSET_PATH": "%s/dist" % dir["web"],
+        "WEB_ASSET_CORE_PATH": "%s/dist" % dir["web"],
         "FRONTEND_SEARCH_MIN_LENGTH": "2",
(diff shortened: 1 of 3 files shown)


AlexAndBear (approved):

kulmann (approved):
Thank you! 👍

kulmann (changes requested):
Oh wait... no. This is correct for master, but not for `stable-8.0`.

AlexAndBear:
FUUUUUUUCK

kulmann:
> FUUUUUUUCK
> 
> but we use ocis master as image, does this make any difference ?

Oh, you're right. We shouldn't. docker-compose stack in `stable-8.0` should pull `owncloud/ocis-5.0.0-rc.5`

AlexAndBear:
Ahhh ok 😱

ScharfViktor:
> Oh, you're right. We shouldn't. docker-compose stack in `stable-8.0` should pull `owncloud/ocis-5.0.0-rc.5`

in the stable version, we also use `ocis:latest`
https://github.com/owncloud/web/blob/stable-8.0/docker-compose.yml#L2

kulmann:
> in the stable version, we also use `ocis:latest` https://github.com/owncloud/web/blob/stable-8.0/docker-compose.yml#L2

We shouldn't :-) 

@ScharfViktor could you recreate this pull request and point it at the `master` branch?

Issue #10018, opened 2023-11-20
Increase standard tile size


tbsbdr:
# Description

**Please track times**

## User Stories

* > As a user I want to see Tiles in a reasonable size so that I can select the right one real quick.

## Value

- improve navigation

## Acceptance Criteria
- Tile size overview should be ~ 260px
- size-slider should be set to the second-smallest step (currently its set to the smallest step)
- theming variable `base-tile-size` should be slightly bigger

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

### Problem
![screenshot_001979](https://github.com/owncloud/web/assets/26610733/6585e790-54a8-4a67-a6fa-dfd736164760)

### Solution
Set the default tiles view slider to +1 step
Note: the +1 step looks quite big (almost +100%) lets try to adjust the steps to a more decent size eg. +50%
![screenshot_001980](https://github.com/owncloud/web/assets/26610733/8f8943cb-d460-4f94-a5b3-ea7300986014)

PR #10558, opened 2024-03-07
feat: adjust tile sizes

kulmann:
## Description
Played around with tile sizes... stepping from one tile size to the next was too big.

Previously we had 9rem as base tile size and added 9rem for every step in the tile size slider.
Now we have 10rem as base tile size and add 6rem for every step in the tile size slider.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/10018
- Fixes https://github.com/owncloud/web/issues/10564

### new default tile size (the second stepping, slighty smaller than before: now 16rem, before 18rem)
Screenshot 2024-03-07 at 09 17 59
Screenshot 2024-03-07 at 09 25 03

### new smallest tile size (the first stepping, slightly bigger than before: now 10rem, before 9rem)
Screenshot 2024-03-07 at 09 18 03
Screenshot 2024-03-07 at 09 19 20

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/enhancement-tile-sizes
+++ b/changelog/unreleased/enhancement-tile-sizes
@@ -1,0 +1,8 @@
+Enhancement: Tile sizes
+
+We've adjusted the tile sizes to have a bigger base size and smaller stepping from one tile size to the next.
+In addition to that the default tile size has been changed to the second stepping, so that initially both
+space tiles and file/folder tiles have a sufficient size for reading longer names.
+
+https://github.com/owncloud/web/issues/10018
+https://github.com/owncloud/web/pull/10558
--- a/packages/design-system/src/tokens/ods/size.yaml
+++ b/packages/design-system/src/tokens/ods/size.yaml
@@ -24,6 +24,6 @@ size:
     value: 14px
   tiles:
     resize-step:
-      value: 9rem
+      value: 6rem
     default:
-      value: 9rem
+      value: 10rem
--- a/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
@@ -101,11 +101,7 @@
         :key="`ghost-tile-${index}`"
         class="ghost-tile"
         :aria-hidden="true"
-      >
-        <div>
-          {{ viewSize }}
-        </div>
-      </li>
+      />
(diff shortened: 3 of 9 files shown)


AlexAndBear:
IHMO we should also limit the max tile size. It's too big to make sense

kulmann:
> IHMO we should also limit the max tile size. It's too big to make sense

you mean the design tokens / css vars? base size and step size? what do you think are reasonable values? Steps are limited to 6, i.e. with the defaults that we now have in the design tokens the max tile size is `10rem + (5 * 6rem) = 40rem`, which should be ok.

AlexAndBear:
If I have 3 spaces on the board and set the tile size to max, it looks like this:

image

kulmann:
> If I have 3 spaces on the board and set the tile size to max, it looks like this:
> Which I think looks ridiculous :(

As long as that's not your only option to display space tiles at all... I don't see the issue honestly.

AlexAndBear:
Yeah I found the issue. In the spaces view it renders 1 tile in the same display resolution and tile size setting and in the personal/files view it renders 2 tiles. There is also an issue with the tile size stays the same just the inner icon size changes on smaller screens between some tiles view size steps 🤷‍♀️

Feels a little inconsistent overall but maybe scope creep

kulmann:
> Yeah I found the issue. In the spaces view it renders 1 tile in the same display resolution and tile size setting and in the personal/files view it renders 2 tiles. There is also an issue with the tile size stays the same just the inner icon size changes on smaller screens between some tiles view size steps 🤷‍♀️
> 
> Feels a little inconsistent overall but maybe scope creep

uh... sounds worth fixing. can you provide detailed steps to reproduce? then I can tackle it in this PR as well maybe.

JammingBen (approved):
LGTM (aside from the issues Alex was mentioning) 👍

AlexAndBear:
Unfortunately not for me :(



https://github.com/owncloud/web/assets/26169327/6377fdd2-eab2-44ed-a012-2738f4b9b7a2

kulmann:
@AlexAndBear could you please test again? :-) slider max value now gets adjusted according to max available steps calculated in the tiles view.

AlexAndBear (approved):
I love it ! 😮

PR #10565, opened 2024-03-07
[test-only]check that image is open

ScharfViktor:
related https://github.com/owncloud/ocis/issues/8602

we had test for this bug but unfortunately the test didn't expect an error

the image viewer was open, but the image did not appearScreenshot 2024-03-07 at 17 39 25

I'm wait locators now
image

we can check the response, but sometimes that doesn't give us a 100% guarantee that everything works. Example, openning broken pdf file: 
image

for txt or md files we can check content(ctr+c and evaluate clickboard) but it dificult when user viewer

### DIFF
--- a/tests/e2e/support/objects/app-files/utils/editor.ts
+++ b/tests/e2e/support/objects/app-files/utils/editor.ts
@@ -4,7 +4,7 @@ const closeTextEditorOrViewerButton = '#app-top-bar-close'
 const saveTextEditorOrViewerButton = '#app-save-action'
 const texEditor = '#text-editor'
 const pdfViewer = '#pdf-viewer'
-const imageViewer = '.preview-controls-action-bar'
+const imageViewer = '.stage'
 
 export const close = (page: Page): Promise<unknown> => {
   return Promise.all([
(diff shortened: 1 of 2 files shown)


ScharfViktor:
don't merge please before fixing https://github.com/owncloud/ocis/issues/8602

tests/e2e/support/objects/app-files/utils/editor.ts
```diff
@@ -4,7 +4,7 @@ const closeTextEditorOrViewerButton = '#app-top-bar-close'
 const saveTextEditorOrViewerButton = '#app-save-action'
 const texEditor = '#text-editor'
 const pdfViewer = '#pdf-viewer'
-const imageViewer = '#preview'
+const imageViewer = '.preview-controls-action-bar'
```
kulmann, line 7:
```suggestion
const imageViewer = '.stage'
```

The `stage` selector will only appear if loading the content was successful.

ScharfViktor: thanks 😍

SagarGi (commented):
Test failing as the selector was changed. LGTM :+1: if CI is happy.

saw-jan (approved):

nabim777 (approved):

PR #10566, opened 2024-03-08
refactor: make expand on user and group listings optional

JammingBen:
## Description
The endpoints for listing users and groups will be used for fetching potential sharees in the future. Regular users (non-admins) are not allowed to expand these requests though, hence we need to make the expand option optional.

To be honest, our current wrapper of the Graph SDK just feels like useless boilerplate since we're basically re-implementing it 1:1 (with the exception of some parts that were left out or are set to default - which bites us here e.g.). IMO we should either drop this wrapper or make it somewhat useful.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


kulmann (approved):
💪

PR #10582, opened 2024-03-11
Use blacklisted filename character capability throughout project

### LINKED ISSUE #8123: [Bug] Filter special Unicode characters
Some characters in Unicode can be used for bad things. One example is the right to left override character (https://unicode-explorer.com/c/202E) which changes the orientation of text. Such characters should be blocked on input.

### Steps to reproduce
1. Create a folder name with the RLO character

### Expected behaviour
Should be blocked

### Actual behaviour
Folder is created which leads to strange effects. Folder is displayed "correct" in the web:
Screenshot 2022-12-15 at 18 47 07

but the Mac Desktop App notification renders strange text (note the RLO character was entered after "Reverse"):

Screenshot 2022-12-15 at 18 47 59

even though the folder in the file system seems to work.

This is only one example. There might be other "bad" character which might possibly harm system and thus should be prohibited.

### Environment general
All tested on https://web.cloud.owncloud.com

pascalwengerter:
## Description
Goes with https://github.com/cs3org/reva/pull/4558

## Related Issue
- Fixes #8123

### DIFF
--- a/packages/web-client/src/ocs/capabilities.ts
+++ b/packages/web-client/src/ocs/capabilities.ts
@@ -106,6 +106,7 @@ export interface Capabilities {
     files: {
       app_providers?: AppProviderCapability[]
       archivers?: ArchiverCapability[]
+      blacklisted_files_regex?: string
       favorites?: boolean
       full_text_search?: boolean
       permanent_deletion?: boolean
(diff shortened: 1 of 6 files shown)


pascalwengerter:
Closed as unwanted/unrelevant via discussion in https://github.com/owncloud/web/issues/8123

PR #10586, opened 2024-03-12
SSE events for deleted (trashed) and restored files and folders

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/9849

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


changelog/unreleased/enhancement-add-sse-event-for-locking-renaming-deleting-restoring
```diff
@@ -0,0 +1,12 @@
+Enhancement: Add SSE events for locking, renaming, deleting, and restoring
+
+We've added Server-Sent Events (SSE) for locking, renaming, deleting, and restoring actions.
+This enhancements ensures that these actions are instantly propagated to other browsers and devices where the user is logged in.
```
kulmann, line 4:
```suggestion
This enhancement ensures that these actions are instantly propagated to other browsers and devices where the logged in user has access the relevant files and folders.
```
Not 100% happy about my change, but we should make clear that a logged in space member gets the events as well, even when a different space member triggers the action.

packages/web-runtime/src/container/sse.ts
```diff
@@ -203,3 +206,88 @@ export const onSSEProcessingFinishedEvent = async ({
...
+  messageStore: MessageStore
+  msg: MessageEvent
+}) => {
+  try {
+    const sseData = fileReadyEventSchema.parse(JSON.parse(msg.data))
+    const currentFolder = resourcesStore.currentFolder
+    const resourceIsCurrentFolder = currentFolder.id === sseData.itemid
+
+    if (resourceIsCurrentFolder) {
+      return messageStore.showMessage({
```
kulmann, line 229: Found one issue here: When I am in folder `a/b/c` and delete `c` from the breadcrumb menu, I get two messages in the bottom right:
- success message that the folder was deleted (+ a navigation to `a/b` happens automatically because I triggered the delete action myself)
- this warning here that my current location was removed (although I was already navigated up)

Any idea how to solve this?

AlexAndBear: Ugh no other idea then adding a delay :(

kulmann: then let's try with a slight delay on the SSE handling.

AlexAndBear: Sure. Later on we should work with client ids and early return if the original event was initiated by the same client id

kulmann: Yep, nice idea. Do you have an ocis ticket for that already? if not, please create one. 👍

AlexAndBear: Just popped up in my mind 5 minutes ago lol

packages/web-runtime/src/container/sse.ts
```diff
@@ -11,17 +12,14 @@ import PQueue from 'p-queue'
 import { extractNodeId, extractStorageId } from '@ownclouders/web-client/src/helpers'
 import { z } from 'zod'
 import { Router } from 'vue-router'
+import { Language } from 'vue3-gettext'
 
 const fileReadyEventSchema = z.object({
```
AlexAndBear, line 17: Rename to generalEventSchema

packages/web-runtime/src/container/sse.ts
```diff
@@ -11,17 +12,14 @@ import PQueue from 'p-queue'
...
   itemid: z.string(),
-  parentitemid: z.string()
+  parentitemid: z.string(),
+  spaceid: z.string().optional()
```
AlexAndBear, line 20: Check if still needs optional

packages/web-runtime/src/container/sse.ts
AlexAndBear: The issue will be created tomorrow, work around has been implemented. Please re-review

kulmann (approved):
🥳

PR #10596, opened 2024-03-13
Fix DisplayLocation can't be imported

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-epub-reader/src/App.vue
+++ b/packages/web-app-epub-reader/src/App.vue
@@ -94,7 +94,6 @@ import {
   useThemeStore
 } from '@ownclouders/web-pkg'
 import ePub, { Book, NavItem, Rendition, Location } from 'epubjs'
-import { DisplayedLocation } from 'epubjs/types/rendition'
 
 const DARK_THEME_CONFIG = {
   html: {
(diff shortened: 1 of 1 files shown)


packages/web-app-epub-reader/src/App.vue
```diff
@@ -222,7 +221,7 @@ export default defineComponent({
         })
 
         unref(rendition).on('relocated', () => {
-          const currentLocation = unref(rendition).currentLocation() as DisplayedLocation & Location
+          const currentLocation = unref(rendition).currentLocation() as any & Location
```
kulmann, line 224: Oh no :-( aren't they exporting the type anymore or what's the problem?

AlexAndBear: It was never exported in the types.ts but in the the types/redention which seems not to be accessible when built

kulmann: 😢

kulmann (approved):

Issue #10515, opened 2024-02-26
Moving or copying the folder with a replace option does not work


ScharfViktor:
`ocis: 5.1.0-prealpha+44bdea9608`
`web: 8.0.0-rc.5
`

Steps: 
- create `folder1/file1.txt` and `subfolder/folder1/file2.txt`
- copy or move `folder1` to `subfolder/folder1` with replays option

Expect: `subfolder/folder1` contains two files: `file1.txt` and `file2.txt`
Actual: 
<img width="395" alt="image" src="https://github.com/owncloud/web/assets/84779829/7089533c-326c-4d52-bd38-e5c73a708b2e"> 

https://github.com/owncloud/web/assets/84779829/2b00eac5-aa05-4392-a706-553a5c6d76fc

Same case uploading folder with replace works fine

kulmann:
I don't agree with the expected behaviour. The content after replacing should be `file1.txt`. I'd only expect to have both text files if there was a `Merge` option, but there is none.

JammingBen:
Not sure if we can do much about that the moment since uploading and copying/moving are being handled differently.

Uploading a folder goes through each file separately to upload it. That means existing files in a folder are not affected. Whereas copying/moving are WebDAV operations and handle the folder itself as one resource. In case of replacing, the client sends an overwrite header to the server so it will replace the folder with all its contents completely.

Solutions I can think of:

* Improve on our wording: "Replace" becomes "Merge" in case of an upload.
* Delete the existing folder first before uploading a new folder with the "Replace" option.

I prefer the first approach to be honest.

kulmann:
> Solutions I can think of:
> 
> * Improve on our wording: "Replace" becomes "Merge" in case of an upload.
> * Delete the existing folder first before uploading a new folder with the "Replace" option.
> 
> I prefer the first approach to be honest.

Yes, I think the same. The "Replace" option in folder upload conflict handling should be named "Merge" instead. Doing a delete before uploading would open a can of worms for error handling (I'd expect a rollback for the deleted folder in case the upload goes wrong...)

JammingBen:
I created https://github.com/owncloud/web/issues/10640 to keep track of it. Which means I'm closing here as this has been resolved.

PR #10597, opened 2024-03-13
fix: replacing folders

JammingBen:
## Description
Fixes the "Replace" conflict option when copying/moving a folder.

The mechanism of `isOverwritingParentFolder` falsely prevented this though it should only care about the edge case scenario when moving a sub-folder with the same name as the current folder's parent folder into the parent (via breadcrumbs). This needs an additional check if the current folder differs from the target location.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10515

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-folder-replace
+++ b/changelog/unreleased/bugfix-folder-replace
@@ -1,0 +1,6 @@
+Bugfix: Folder replace
+
+The "Replace" conflict option, which previously didn't work at all when trying to copy/move a folder, has been fixed.
+
+https://github.com/owncloud/web/issues/10515
+https://github.com/owncloud/web/pull/10597
--- a/packages/web-app-files/src/helpers/resource/actions/transfer.ts
+++ b/packages/web-app-files/src/helpers/resource/actions/transfer.ts
@@ -11,6 +11,7 @@ import {
 } from '@ownclouders/web-pkg'
 
 import { TransferType } from '.'
+import { Ref, unref } from 'vue'
 
 export class ResourceTransfer extends ConflictDialog {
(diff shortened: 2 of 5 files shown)


AlexAndBear (approved):
🙌

changelog/unreleased/bugfix-folder-replace
```diff
@@ -0,0 +1,6 @@
+Bugfix: Folder replace
+
+The "Replace" conflict option when copying/moving a folder has been fixed.
```
kulmann, line 3: Could you describe in which way it was misbehaving before?

packages/web-app-files/src/helpers/resource/actions/transfer.ts
```diff
@@ -141,10 +143,19 @@ export class ResourceTransfer extends ConflictDialog {
...
+  public isOverwritingParentFolder(
+    resource: Resource,
+    targetFolder: Resource,
+    targetFolderResources: Resource[]
+  ) {
     if (resource.type !== 'folder') {
       return false
     }
+
+    if (targetFolder.path === unref(this.currentFolder)?.path) {
```
kulmann, line 155: Would it make sense to check on the `id` prop instead? Or can that lead to false results?

JammingBen: It would work, I guess, though I feel like checking for `path` is a little more robust in this context, since identical paths are what cause this issue.

kulmann (approved):
Works as expected now 💪

Issue #10397, opened 2024-01-23
No redirection to project space from notification


ScharfViktor:
ocis 5.1.0-prealpha+17b14ee78
web 8.0.0-rc.1

Steps:

- admin creates space
- admin adds member eintein 
- einstein open notification and click by space name

Actual: no redirect to Project space page. only after refreshing page

https://github.com/owncloud/web/assets/84779829/82dcec74-7d8f-4863-be20-f7f1a95abcb5

JammingBen:
Good catch, a redirect to the space will only work after a page reload because then the space is being loaded by the runtime. In order to achieve that without a page reload we need to:

* Check if the notification subject is of type "space shared" (or something like that)
* If so, load the space by its id and add it to the list of loaded spaces (spaces store)

LukasHirt:
Could not reproduce with `ownCloud Web UI 11.3.1`. Must have been solved in the meantime.

PR #10603, opened 2024-03-14
fix: space notification links

JammingBen:
## Description
Fixes links for incoming space notifications if the user has been added to a new space. Also makes sure links are being removed if a user has been removed from a space and the space had been loaded before.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10397

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/components/Topbar/Notifications.vue
+++ b/packages/web-runtime/src/components/Topbar/Notifications.vue
@@ -104,6 +104,7 @@ import { useGettext } from 'vue3-gettext'
 import { useTask } from 'vue-concurrency'
 import { createFileRouteOptions } from '@ownclouders/web-pkg'
 import { MESSAGE_TYPE } from '@ownclouders/web-client/src/sse'
+import { buildSpace } from '@ownclouders/web-client/src/helpers'
 
 const POLLING_INTERVAL = 30000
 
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -175,21 +176,47 @@ export default {
         const space = store.getters['runtime/spaces/spaces'].find(
           (s) => s.fileId === messageRichParameters?.space?.id.split('!')[0] && !s.disabled
         )
-        if (space) {
+
+        if (space && subject === 'Removed from Space') {
```
kulmann, line 180: Isn't the subject translated by the backend? So this would only work for users who have set English as language?

kulmann: also... don't we have a different property to check on? Subject is really ugly for this :-(

JammingBen: Haha yes, totally forgot 🙈 Well, then this is blocked by the server. There is no other way to distinguish between the subjects.

I'll create an issue for it.

kulmann: thank you 💪

JammingBen:
Currently blocked by https://github.com/owncloud/ocis/issues/8645.

JammingBen:
Still blocked, doesn't get attention apparently, so closing here (for now). Can of course be re-opened at any time.

Issue #9709, opened 2023-09-19
Remove owncloud-sdk and re-implement it in web-client


JammingBen:
The `owncloud-sdk` is quite old, does not fullfil our code standards and is hard to maintain because of the linking process. The latter makes future adjustments very time consuming, especially since we're planning to switch to id-based WebDAV requests.

 Hence we figured it makes sense to remove the `owncloud-sdk` and implement its logic in our `web-client` package.

#### To-Do's

- [x] Make sure Web uses the provided `webdav` abstraction instead of the sdk directly for webdav requests
- [x] Port all WebDAV requests to a new abstraction in `web-client`
- [x] Port OCS requests to a new abstraction in `web-client` (expect for the sharing stuff, which will be replace via sharing NG)
- [x] Check if settings API needs to be ported -> no needed
- [x] What about `signUrl` logic? -> https://github.com/owncloud/web/pull/10621
- [x] Add proper (unit-) tests for the new code

kulmann:
Hopefully pays off now that @dschmidt and myself already built an abstraction layer for webdav in web-client. 🙈

PR #10607, opened 2024-03-14
refactor: remove usage of sdk for file version urls

JammingBen:
## Description
Removes the usage of the ownCloud SDK for file versions URLs since it's deprecated and we want to get rid of it. This could be achieved quite simple as nearly all the logic we had in `useDownloadFile` is already present in `getFileUrl` of our webdav implementation.

## Related Issue

- refs https://github.com/owncloud/web/issues/9709

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-client/src/webdav/getFileUrl.ts
+++ b/packages/web-client/src/webdav/getFileUrl.ts
@@ -8,7 +8,7 @@ import { DAV } from './client'
 export const GetFileUrlFactory = (
   dav: DAV,
   getFileContentsFactory: ReturnType<typeof GetFileContentsFactory>,
-  { sdk, capabilities }: WebDavOptions
+  { sdk, capabilities, clientService, user }: WebDavOptions
 ) => {
   return {
     async getFileUrl(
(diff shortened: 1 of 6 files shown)


packages/web-client/src/webdav/getFileUrl.ts
```diff
@@ -8,18 +8,22 @@ import { DAV } from './client'
...
   return {
     async getFileUrl(
       space: SpaceResource,
       resource: Resource,
       {
         disposition = 'attachment',
-        signUrlTimeout = 86400
+        signUrlTimeout = 86400,
+        version = null,
+        preflightHeadRequest = false
```
kulmann, line 21: I find the var name a bit confusing. `preflight` is usually an `OPTIONS` request, not `HEAD`. Maybe `doHeadRequest` or something similar would be more accurate? What do you think?

JammingBen: No hard feelings here, I'm fine with `doHeadRequest` 👍

kulmann (approved):
LGTM 👍

Issue #10534, opened 2024-02-13
[web] Enable Fulltext Search by default


tbsbdr:
# Description

## User Stories

* > As a user I want that the standard search searches in fulltext so that I get maximum searchresults.

## Value

- improve file search behaviour
- consistent behaviour of search filters (all filter narrow down; currently "search only in content" extends the search results)

## Acceptance Criteria
- klick on 🔍 searches in content and title per default (if fulltext search is not available -> title only, like it already is now)
- replace filter "~search only in content~" with "Title only" (if fulltext search is not available -> don't show the filter)
  - click on "Title only" narrows down search results for matches in the title only
- tech cherry on top: don't escape quotes anymore. @fschade claims that search works without escaping the quotes.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] e2e tests are adjusted to new behaviour
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

PR #10626, opened 2024-03-18
feat: make full text search default

JammingBen:
## Description
Enables full text search as default if the server supports it.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10534

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -71,9 +71,9 @@
 
         <item-filter-toggle
           v-if="fullTextSearchEnabled"
-          :filter-label="$gettext('Search only in content')"
-          filter-name="fullText"
-          class="files-search-filter-full-text oc-mr-s"
+          :filter-label="$gettext('Title only')"
+          filter-name="titleOnly"
+          class="files-search-filter-title-only oc-mr-s"
         />
       </div>
       <app-loading-spinner v-if="loading" />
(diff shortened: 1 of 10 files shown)


packages/web-app-files/src/components/Search/List.vue
```diff
@@ -342,12 +344,12 @@ export default defineComponent({
...
+      const isTitleOnlySearch = queryItemAsString(unref(titleOnlyParam)) == 'true'
 
-      if (isContentOnlySearch && !!humanSearchTerm) {
-        query['content'] = `"${humanSearchTerm}"`
-      } else if (!!humanSearchTerm) {
-        query['name'] = `"*${humanSearchTerm}*"`
+      if ((isTitleOnlySearch || !unref(fullTextSearchEnabled)) && !!humanSearchTerm) {
+        query['name'] = `*${humanSearchTerm}*`
+      } else if (!!humanSearchTerm && unref(fullTextSearchEnabled)) {
+        query['content'] = `${humanSearchTerm}`
```
kulmann, line 352: The idea was to have either `title only` or `title OR content`. This implementation doesn't include the file name search anymore by default.

kulmann (changes requested):
Same logic as in the search result page should also apply for the search preview.

packages/web-app-files/src/components/Search/List.vue
JammingBen: Oh I see, then I got it wrong 🙈

packages/web-app-files/src/components/Search/List.vue
```diff
@@ -272,11 +272,13 @@ export default defineComponent({
     const tagParam = useRouteQuery('q_tags')
     const lastModifiedParam = useRouteQuery('q_lastModified')
     const mediaTypeParam = useRouteQuery('q_mediaType')
-    const fullTextParam = useRouteQuery('q_fullText')
+    const titleOnlyParam = useRouteQuery('q_titleOnly')
+
+    const fullTextSearchEnabled = computed(() => capabilityStore.searchContent?.enabled)
```
AlexAndBear, line 277: Why not using storeToRefs for that ?

JammingBen: Because it's nested in the capabilities, meaning even if we'd use `storeToRefs`, we would still need to further narrow it down via a `computed`.

JammingBen:
Okay, now it searches for `name` and `content` at the same time if supported. Also applies to the search bar.

I decided to remove the generic query building we had before because it didn't really allow edge cases like we have here.

PR #10658, opened 2024-03-21
[test-only] more e2e test for space link

ScharfViktor:
added test for project space link where public is doing grud operations like: upload, download, rename, change file content, delete file 

related https://github.com/owncloud/ocis/issues/8699


PrajwolAmatya (approved):
LGTM

PR #10695, opened 2024-04-02
fix: missing translations for share roles

JammingBen:
## Description
Share role labels and descriptions are now coming from the server, hence we need to wrap them manually via `gettext` in our additional translations for them to appear in Transifex (as least for now until we have a better way of handling this).

## Related Issue

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/helpers/additionalTranslations.ts
+++ b/packages/web-runtime/src/helpers/additionalTranslations.ts
@@ -18,5 +18,16 @@ export const additionalTranslations = {
   ocsErrorPasswordOnBannedList: $gettext(
     'Unfortunately, your password is commonly used. please pick a harder-to-guess password for your safety'
   ),
-  openAppFromSmartBanner: $gettext('OPEN')
+  openAppFromSmartBanner: $gettext('OPEN'),
+  shareRoleDescriptionViewer: $gettext('View and download.'),
+  shareRoleDescriptionEditor: $gettext('View, download, upload, edit, add and delete.'),
+  shareRoleDescriptionFileEditor: $gettext('View, download and edit.'),
+  shareRoleDescriptionUploader: $gettext('View, download and upload.'),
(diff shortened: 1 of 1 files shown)


kulmann (commented):
Why are those not translated by the backend already? We're sending the `Accept-Language` header... so I'd expect the translations to be in place in the first iteration already. Maybe just not translated in transifex, yet?

JammingBen:
> Why are those not translated by the backend already? We're sending the `Accept-Language` header... so I'd expect the translations to be in place in the first iteration already. Maybe just not translated in transifex, yet?

There haven't been talks yet about how we want to handle translations, currently the server just sends those strings in English. I already added this to the list of open questions in https://github.com/owncloud/web/issues/10418. For now we need this solution I'm afraid.

packages/web-runtime/src/helpers/additionalTranslations.ts
```diff
@@ -18,5 +18,16 @@ export const additionalTranslations = {
...
     'Unfortunately, your password is commonly used. please pick a harder-to-guess password for your safety'
   ),
-  openAppFromSmartBanner: $gettext('OPEN')
+  openAppFromSmartBanner: $gettext('OPEN'),
+  shareRoleDescriptionViewer: $gettext('View and download.'),
+  shareRoleDescriptionEditor: $gettext('View, download, upload, edit, add and delete.'),
+  shareRoleDescriptionFileEditor: $gettext('View, download and edit.'),
+  shareRoleDescriptionUploader: $gettext('View, download and upload.'),
+  shareRoleDescriptionManager: $gettext(
+    'View, download, upload, edit, add, delete and manage members.'
```
kulmann, line 27: This one sounds like a list of things you can do regarding space members. 😆 but I guess this is not the place to discuss wording.

kulmann (approved):

PR #10699, opened 2024-04-02
refactor: space member loading

JammingBen:
## Description
Refactors space member loading so members are only being loaded when the sidebar for a resource or a space is being opened. It happened in the drive resolver before, which is a bad place to handle this.

## Related Issue

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/Spaces/SpaceHeader.vue
+++ b/packages/web-app-files/src/components/Spaces/SpaceHeader.vue
@@ -94,7 +94,6 @@ import {
   useClientService,
   usePreviewService,
   ProcessorType,
-  useSpacesStore,
   useResourcesStore,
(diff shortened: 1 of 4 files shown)


kulmann (changes requested):
I encountered two issues (didn't check if they are on master as well):
- space member count doesn't update reactively when members are being added
- sidebar space member list is not updated reactively when a new member gets added

Might be that both issues have the same underlying problem. Could you check this?

JammingBen:
Good catch, thx! One issue was introduced with this PR, the other was actually wrong on master as well (although this PR made it then visible to the user).

kulmann (approved):
🚀

Issue #10681, opened 2024-03-27
Improve separation between `Resource` and `SpaceResource` interface


JammingBen:
We already have `Resource` and `SpaceResource` interfaces to distinguish normal resources and space resources, however, the `SpaceResource` interface is poorly maintained at the moment. Most of the space properties are included as optional property in the regular `Resource` interface.

The goal is to improve this separation so that the `Resource` interface doesn't contain any space-only properties. We probably also need to implement a new helper `isSpaceResource` which allows a type predicate.

PR #10715, opened 2024-04-05
refactor: improve SpaceResource interface

JammingBen:
## Description
Improves the `SpaceResource` interface by moving all space relevant properties from the `Resource` interface to there as well as by making them non-optional. This allows us for better distinction between the both.

## Related Issue

- fixes https://github.com/owncloud/web/issues/10681

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-client/src/helpers/space/types.ts
+++ b/packages/web-client/src/helpers/space/types.ts
@@ -6,20 +6,45 @@
 // ```
 // because in the else block resource gets the type never. If this is changed in a later TypeScript version
 // or all types get different members, the underscored props can be removed.
-import { DriveItem } from '@ownclouders/web-client/src/generated'
-import { Resource } from '../resource'
-import { User } from '../../generated'
+import { DriveItem, Identity, Quota, User } from '@ownclouders/web-client/src/generated'
+import { Ability, Resource } from '../resource'
 
 export const SHARE_JAIL_ID = 'a0ca6a90-a365-4782-871e-d44447bbc668'
 export const OCM_PROVIDER_ID = '89f37a33-858b-45fa-8890-a1f2b27d90e1'
 
+export interface SpaceRole extends Identity {
+  kind: 'user' | 'group'
+  isMember(u: User): boolean
+}
+
+export interface SpaceRoles {
+  viewer: SpaceRole[]
+  editor: SpaceRole[]
+  manager: SpaceRole[]
+}
+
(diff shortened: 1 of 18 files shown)


packages/web-client/src/helpers/space/types.ts
```diff
@@ -35,22 +60,22 @@ export interface PersonalSpaceResource extends SpaceResource {
   __personalSpaceResource?: any
 }
 export const isPersonalSpaceResource = (resource: Resource): resource is PersonalSpaceResource => {
-  return resource?.driveType === 'personal'
+  return (resource as SpaceResource)?.driveType === 'personal'
```
JammingBen, line 63: Another (maybe cleaner) solution would be to do a type predicate via `isSpaceResource`. But it would require more effort in unit tests because the method checks for `type === 'space'`. That means each time we mock `SpaceResource` we would need to manually set `type` to `space` for this to evaluate to `true`.

dschmidt: I like to keep things reasonably simple - keeping it as it is right now is fine with me

packages/web-pkg/src/components/Spaces/QuotaModal.vue
```diff
@@ -151,7 +151,11 @@ export default defineComponent({
           })
         }
         spacesStore.updateSpaceField({ id: space.id, field: 'spaceQuota', value: driveData.quota })
-        updateResourceField({ id: space.id, field: 'spaceQuota', value: driveData.quota })
+        updateResourceField<SpaceResource, any>({
```
dschmidt, line 154: Do you really need to specify `any` here? Isn't partial specialization possible?

dschmidt (commented):
love it 😍

packages/web-pkg/src/components/Spaces/QuotaModal.vue
JammingBen: I tried, but couldn't get the types to work. Maybe the types of `updateResourceField` are not quite correct? I'm a bit stuck here.

JammingBen: As discussed, found a solution: https://github.com/owncloud/web/pull/10715/commits/b3f89fa5cb1913a9e345d771edc63bd93798f243

Issue #10712, opened 2024-04-05
User other than admin cannot share resources to other users


nabim777:
## Describe the bug
There is redundant resharing feature in the ocis that makes nightly builds failure. From this issue https://github.com/owncloud/web/issues/10559, we changed ocis caching behaviour on nightly. Currently, web PR builds uses specified ocis commitId and ocis branch from .drone.env file to cache ocis on nightly build. But on nightly, script file runs which update ocis commit ID of .drone.env file to latest commit id of ocis repo.

The solution for this issue is bump latest commit of ocis and remove or update step where reshare-related code is implemented.
https://drone.owncloud.com/owncloud/web/43795/10/12

```feature
Scenario: delete and restore a file inside a received shared folder (attempt 1, retried) # tests/e2e/cucumber/features/smoke/trashbinDelete.feature:45
   ✔ Before # tests/e2e/cucumber/environment/index.ts:43
   ✔ Given "Admin" creates following user using API # tests/e2e/cucumber/steps/api.ts:7
       | id    |
       | Alice |
       | Brian |
   ✔ And "Alice" logs in # tests/e2e/cucumber/steps/ui/session.ts:30
   ✔ Given "Alice" creates the following resources # tests/e2e/cucumber/steps/ui/resources.ts:19
       | resource      | type   |
       | folderToShare | folder |
   ✔ And "Alice" uploads the following resource # tests/e2e/cucumber/steps/ui/resources.ts:35
       | resource  | to            |
       | lorem.txt | folderToShare |
   ✖ And "Alice" shares the following resource using the sidebar panel # tests/e2e/cucumber/steps/ui/shares.ts:30
       | resource      | recipient | type | role     | resourceType |
       | folderToShare | Brian     | user | Can edit | folder       |
       page.waitForResponse: Timeout 30000ms exceeded while waiting for event "response"
           at Function.sendInvitation (/var/www/owncloud/web/tests/e2e/support/objects/app-files/share/collaborator.ts:110:14)
           at Function.inviteCollaborators (/var/www/owncloud/web/tests/e2e/support/objects/app-files/share/collaborator.ts:134:24)
           at async Object.createShare (/var/www/owncloud/web/tests/e2e/support/objects/app-files/share/actions.ts:80:3)
           at async Share.create (/var/www/owncloud/web/tests/e2e/support/objects/app-files/share/index.ts:15:5)
[...]

PR #10703, opened 2024-04-03
fix: fall back to username in share autocomplete items

JammingBen:
## Description
Since user emails are sensitive information and we can't rely on the server providing it anymore, share autocomplete items now fall back to the username (`onPremisesSamAccountName` in graph).

## Related Issue

- refs https://github.com/owncloud/ocis/issues/8726

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation
- [ ] Maintenance (e.g. dependency updates or tooling)

PR #10738, opened 2024-04-05
fix: sharing for non-admin users

JammingBen:
## Description
Since the server doesn't send any information other than `id` and `displayName` for non-admins when querying users, we need to change the way how to distinguish between users and groups. This basically undos the simplified approach done via PR #10703 as it doesn't work anymore.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10712

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


SagarGi:
The nightly build seems to be passing. So this issue can be closed. https://drone.owncloud.com/owncloud/web/43831

Issue #10619, opened 2024-03-15
Opening the right sidebar fetches versions twice


JammingBen:
## Describe the bug
Opening the right sidebar for a single file fetches its versions twice.

## Steps to reproduce
1. Upload a text file
2. Update the text file to create versions
3. Open the network tab of your browser
4. Open the right sidebar for that file

## Expected behavior
Versions are being fetched once.

## Actual behavior
Versions are being fetched twice.

kulmann:
Also somehow relates to https://github.com/owncloud/web/issues/10381

PR #10746, opened 2024-04-08
Centralize version loading

AlexAndBear:
## Description
We've fixed a bug, where the versions endpoint was fetched multiple times when the sidebar was opened
and therefore added unnecessary load to the server.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10381
- Fixes https://github.com/owncloud/web/issues/10619

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -334,6 +346,42 @@ export default defineComponent({
...
+          !resource.isFolder &&
+          !isProjectSpaceResource(resource) &&
+          !isIncomingShareResource(resource) &&
+          !unref(isPublicFilesLocation) &&
+          !unref(isTrashLocation) &&
+          !unref(isSharedWithOthersLocation) &&
+          !unref(isSharedViaLinkLocation)
+        ) {
+          try {
+            isVersionsLoading.value = true
```
kulmann, line 373: Instead of maintaining a `isVersionsLoading` ref manually, you should have a `computed` in place which just returns `loadVersionsTask.isRunning`. Less error-prone and built-in into the `vue-concurrency` lib.

packages/web-app-files/src/composables/extensions/useFileSideBars.ts
```diff
@@ -227,7 +228,10 @@ export const useSideBarPanels = () => {
               }
               if (
                 isLocationTrashActive(router, 'files-trash-generic') ||
-                isLocationPublicActive(router, 'files-public-link')
+                isLocationPublicActive(router, 'files-public-link') ||
+                isLocationSharesActive(router, 'files-shares-with-others') ||
```
kulmann, line 232: Since we don't support resharing anymore we can be certain, that files listed on `shared with others` or `shared via link` are our own files, so those two routes should actually show the versions panel.

packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -334,6 +345,39 @@ export default defineComponent({
...
+
+        const resource = unref(panelContext).items[0]
+
+        if (loadVersionsTask.isRunning) {
+          loadVersionsTask.cancelAll()
+        }
+
+        if (
+          !resource.isFolder &&
+          !isProjectSpaceResource(resource) &&
```
JammingBen, line 364: Better be more restrictive here as it goes for all kinds of spaces.

```suggestion
          !isSpaceResource(resource) &&
```

packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -334,6 +345,39 @@ export default defineComponent({
...
+          !isIncomingShareResource(resource) &&
+          !unref(isPublicFilesLocation) &&
+          !unref(isTrashLocation) &&
+          !unref(isSharedWithOthersLocation) &&
+          !unref(isSharedViaLinkLocation)
```
JammingBen, line 369: I think you can remove those 2, versions should be available here since we don't have resharing anymore.

packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -334,6 +345,39 @@ export default defineComponent({
       sharesStore.setLoading(false)
     }).restartable()
 
+    watch(
+      () => [...unref(panelContext).items, props.isOpen],
+      async () => {
```
JammingBen, line 350: Hmm now we have 2 watchers on the very same properties. Did you check that it doesn't have any negative effects, especially performance-wise? Not sure if we should do this. Maybe @dschmidt knows?

JammingBen (approved):
👍

packages/web-pkg/src/components/SideBar/FileSideBar.vue
dschmidt: Well obviously it's better to have only one watcher, but if it's not in a component that is used hundreds of times in a web page I doubt it has a real performance impact. What I'm usually more worried about (havent checked this case tho) is that it becomes hard to reason about as it might not be clear which handler is executed first

PR #10758, opened 2024-04-09
Introduce more extension points and delete `scopes`

kulmann:
## Description
This PR introduces more extension points, allows registering extension points through app bootstrapping and gets rid of the `scopes` property on the `Extension` type.

All of the changes are "under the hood", if anything is different in the UI now it's a bug in the PR.

## Motivation and Context
Offer extension points throughout our platform so that extension developers can easily hook their extensions into existing extension points.

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/extension-system/_index.md
+++ b/docs/extension-system/_index.md
@@ -134,22 +134,20 @@ You can find predefined extension point ids in the extension points section belo
 #### Extension Points
 
 There are standardized components and places where extensions are being used automatically. The following are the ones that are currently provided
-by the ownCloud Web runtime or the `files` app. If you decide to develop an extension which fulfills the type and allows the extensionPointId of the respective extension point, 
+by the ownCloud Web runtime or the `files` app. If you decide to develop an extension which fulfills the type and registers itself for the extensionPointId of the respective extension point, 
 your extension will be used automatically. 
 
-1. Left Sidebar for Navigation
-2. Any file(s) context (files app, viewer apps, editor apps)
-   1. Right sidebar. ExtensionPointId `app.files.sidebar`. Mounts extensions of type `sidebarPanel`. 
-3. Files app
-   1. Folder views for regular folders. ExtensionPointId `app.files.folder-views.folder`.
-   2. Folder views for the project spaces overview. ExtensionPointId `app.files.folder-views.project-spaces`.
-   3. Folder views for the favorites page. ExtensionPointId `app.files.folder-views.favorites`.
-   4. Right click context menu. ExtensionPointId `app.files.context-actions`.
-   5. Batch actions in the app bar above file lists. ExtensionPointId `app.files.batch-actions`.
-   6. Upload menu. ExtensionPointId `app.files.upload-menu`.
-   7. Quick actions. ExtensionPointId `app.files.quick-actions`.
-4. Search results in the search app
-5. Global progress bar for the global loading state. ExtensionPointId `app.runtime.global-progress-bar`. Mounts a single extensions of type `customComponent`.
+1. Left Sidebar for Navigation. ExtensionPointId `app.${appName}.navItems` (dynamically created for each app). Mounts extensions of type `sidebarNav`.
+2. Files app
+   1. Right sidebar. ExtensionPointId `app.files.sidebar`. Mounts extensions of type `sidebarPanel`. Used in any file(s) context (files app, file viewer apps, file editor apps)
+   2. Folder views for regular folders. ExtensionPointId `app.files.folder-views.folder`. Mounts extensions of type `folderView`.
+   3. Folder views for the project spaces overview. ExtensionPointId `app.files.folder-views.project-spaces`. Mounts extensions of type `folderView`.
+   4. Folder views for the favorites page. ExtensionPointId `app.files.folder-views.favorites`. Mounts extensions of type `folderView`.
+   5. Right click context menu. ExtensionPointId `app.files.context-actions`. Mounts extensions of type `action`.
+   6. Batch actions in the app bar above file lists. ExtensionPointId `app.files.batch-actions`. Mounts extensions of type `action`.
+   7. Upload menu. ExtensionPointId `app.files.upload-menu`. Mounts extensions of type `action`.
+   8. Quick actions. ExtensionPointId `app.files.quick-actions`. Mounts extensions of type `action`.
+3. Global progress bar for the global loading state. ExtensionPointId `app.runtime.global-progress-bar`. Mounts a single extensions of type `customComponent`. If multiple exist, the user can choose via the account page.
 
 #### User Preferences for Extensions
 
--- a/docs/extension-system/_index.md
+++ b/docs/extension-system/_index.md
@@ -148,6 +148,7 @@ your extension will be used automatically.
    7. Upload menu. ExtensionPointId `app.files.upload-menu`. Mounts extensions of type `action`.
    8. Quick actions. ExtensionPointId `app.files.quick-actions`. Mounts extensions of type `action`.
 3. Global progress bar for the global loading state. ExtensionPointId `app.runtime.global-progress-bar`. Mounts a single extensions of type `customComponent`. If multiple exist, the user can choose via the account page.
+4. Global search providers. ExtensionPointId `app.search.providers`. Utilizes extensions of type `search` as search engines for the search input in the global top bar.
 
 #### User Preferences for Extensions
 
--- a/docs/extension-system/_index.md
+++ b/docs/extension-system/_index.md
@@ -96,7 +96,7 @@ In contrast to applications, extensions usually have a rather small scope and de
 
 The globally available extension registry provided by the ownCloud Web runtime can be used to both register and query extensions. All extensions
 which are being made available via an `app` get registered in the extension registry automatically. In your custom application code you can
-then query any of the available extensions by providing a `type` string and optionally a list of `scopes`. Throughout the ownCloud Web platform
+then query any of the available extensions by providing a `type` string and optionally a list of `extensionPointIds`. Throughout the ownCloud Web platform
 and most prominently also in the `files` app we have defined some extension points which automatically use certain extensions, see the 
 `Extension Points` section below.
 
@@ -104,8 +104,8 @@ and most prominently also in the `files` app we have defined some extension poin
 
 For building an extension you can choose from the types predefined by the ownCloud Web extension system. See the full list of available extension types below.
 
-1. `ActionExtension` (type `action`) - An extension that can register `Action` items which then get shown in various places (e.g. context menus, batch actions), depending on their 
-respective scope. Most commonly used for file and folder actions (e.g. copy, rename, delete, etc.). For details, please refer to the [action docs]({{< ref "extension-types/actions.md" >}})
+1. `ActionExtension` (type `action`) - An extension that can register `Action` items which then get shown in various places (e.g. context menus, batch actions), depending on the 
+extension points defined in the extension respectively. Most commonly used for file and folder actions (e.g. copy, rename, delete, etc.). For details, please refer to the [action docs]({{< ref "extension-types/actions.md" >}})
 2. `SearchExtension` (type `search`) - An extension that can register additional search providers. For details, please refer to the [search docs]({{< ref "extension-types/search.md" >}})
 3. `SidebarNavExtension` (type `sidebarNav`) - An extension that can register additional navigation items to the left sidebar. These can be scoped to specific apps, and programmatically enabled/disabled.
 For details, please refer to the [sidebar nav docs]({{< ref "extension-types/left-sidebar-menu-item.md" >}})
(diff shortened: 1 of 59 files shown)


docs/extension-system/_index.md
```diff
@@ -134,17 +134,22 @@ You can find predefined extension point ids in the extension points section belo
...
-2. Right Sidebar in any file(s) context 
-3. Folder Views in the files app 
-4. Right click context menu in the files app 
-5. Batch actions in the files app
-6. Upload menu in the files app 
-7. Quick actions in the files list of the files app
-8. Search results in the search app
-9. Global progress bar for the global loading state. Extension point id `app.runtime.global-progress-bar`. Allows to render a single custom component.
+2. Any file(s) context (files app, viewer apps, editor apps)
+   1. Right sidebar. ExtensionPointId `app.files.sidebar`. Mounts extensions of type `sidebarPanel`.
```
JammingBen, line 142: I might be mixing things up, but shouldn't this be part of the next point "Files app"? Since the files app introduces this extension point?

JammingBen (commented):
Love to see the scopes go in favour of this 🥳

packages/web-app-files/tests/unit/views/spaces/GenericSpace.spec.ts
```diff
@@ -312,7 +317,7 @@ function getMountedWrapper({
     {
       id: 'com.github.owncloud.web.files.folder-view.resource-table',
       type: 'folderView',
-      scopes: ['resource', 'space', 'favorite'],
+      extensionPointIds: [folderViewsFolder, folderViewsProjectSpaces, folderViewsFavorites],
```
dschmidt, line 320: I'm wondering where to draw the line between a new extension point id and categories

packages/web-app-files/src/extensionPoints.ts
dschmidt, line 1: Should all of this really live in the files app? 
Then we can't import it from other apps

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
```diff
@@ -30,21 +36,27 @@ export const useExtensionRegistry = defineStore('extensionRegistry', () => {
...
-      return true
-    }) as T[]
+    return unref(extensionPoints).flatMap(
+      (e) =>
+        unref(e).filter((e) => {
+          if (Object.hasOwn(options, 'type') && e.type !== options.type) {
+            return false
+          }
+          return true
+        }) as T[]
```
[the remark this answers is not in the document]
kulmann: When I define `T[]` as return type for the function, I get `Type ExtensionPoint is not assignable to type T` as error.

packages/web-pkg/src/composables/piniaStores/extensionRegistry/types.ts
```diff
@@ -7,7 +7,11 @@ import { Component, Slot } from 'vue'
...
 export type ExtensionType = StringUnionOrAnyString<
-  'action' | 'search' | 'sidebarNav' | 'sidebarPanel' | 'folderView' | 'customComponent'
+  'action' | 'customComponent' | 'folderView' | 'search' | 'sidebarNav' | 'sidebarPanel'
+>
+
+export type ExtensionPointType = StringUnionOrAnyString<
+  'actions' | 'customComponents' | 'folderViews' | 'sidebar'
 >
 
 export type ExtensionScope = StringUnionOrAnyString<'resource' | 'user' | 'group'>
```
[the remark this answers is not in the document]
kulmann: Yes, when I killed all scopes I'll remove this type here as well.

docs/extension-system/_index.md
kulmann: I wanted to express where the extension point is available (not so much by which application it's registered)... does that make sense? Can also be in the next section, I'm indifferent.

packages/web-app-files/src/extensionPoints.ts
```diff
@@ -0,0 +1,55 @@
...
+export const quickActionsExtensionPointId = 'app.files.quick-actions'
+export const batchActionsExtensionPointId = 'app.files.batch-actions'
+export const contextActionsExtensionPointId = 'app.files.context-actions'
+export const fileSideBarExtensionPointId = 'app.files.sidebar'
+export const folderViewsFolder = 'app.files.folder-views.folder'
+export const folderViewsFavorites = 'app.files.folder-views.favorites'
+export const folderViewsProjectSpaces = 'app.files.folder-views.project-spaces'
+
+export const extensionPoints = () => {
+  return computed<ExtensionPoint[]>(() => {
```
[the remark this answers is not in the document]
kulmann: I wanted to be consistent with the way we register `extensions`

packages/web-app-files/src/components/FilesList/QuickActions.vue
```diff
@@ -39,7 +40,9 @@ export default defineComponent({
 
     const filteredActions = computed(() => {
       return unref(extensionRegistry)
-        .requestExtensions<ActionExtension>('action', { scopes: ['resource.quick-action'] })
+        .requestExtensions<ActionExtension>('action', {
+          extensionPointIds: [quickActionsExtensionPointId]
+        })
```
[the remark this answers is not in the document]
kulmann: Nice idea 💪 did that now in the latest commit. Needed to finally change the signature of the `requestExtensions` function for this, so that it only expects one `options` object (which can contain either an `extensionPoint` object or an `extensionType` string and optionally a list of `extensionPointIds`).

packages/web-app-files/src/extensionPoints.ts
kulmann: Not sure if the files app extension points are a special snow flake that should live in `web-pkg`... opinions? @JammingBen @AlexAndBear ?

kulmann:
@dschmidt @JammingBen ready for a thorough re-review. As far as I can tell the last remaining open question is whether the files app extension points should be defined and registered via the files app or defined in `web-pkg` and registered via the runtime... I added all extension point ids to the dev docs, so we might as well say that that's the source of truth for existing extension point ids. But might also be a good idea to have the extension point objects available in other apps. What do you think?

docs/extension-system/_index.md
dschmidt: then I'd say it should be a files.sidebar - not app.files.sidebar. I think app.files is misleading

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
```diff
@@ -11,40 +12,57 @@ export const useExtensionRegistry = defineStore('extensionRegistry', () => {
...
-  const requestExtensions = <T extends Extension>(
-    type: ExtensionType,
-    options?: {
-      scopes?: ExtensionScope[]
-      extensionPointIds?: string[]
+  const requestExtensions = <T extends Extension>(options: {
+    extensionType?: ExtensionType
+    extensionPointIds?: string[]
+    extensionPoint?: ExtensionPoint
+  }) => {
```
dschmidt, line 19: This feels inconsistent - why an array of ids but a single extension?

I don't think we query multiple `extensionPointId`s anywhere and conceptually it's not clear at all to me, why we would need it.. 

Moreover, I only see only one occurrence of requesting only by type (in search) - do we really need that? Wouldn't it be more consistent to use an extension point there too? If we need to request extensions by type at all - I'd argue that would justify an independent function `requestExtensionsByType` - to me it feels this function does too much as it is implemented right now.

My suggestion would be (basically as I wrote in my original comment):
```typescript
// add a type parameter to ExtensionPoint interface
export type ExtensionPoint<T> = {
…
}

// specify it when defining the extension point
export const batchActionsExtensionPoint: ExtensionPoint<ActionExtension> = {
  id: 'app.files.batch-actions',
  extensionType: 'action',
  multiple: true
}

// infer extension type from ExtensionPoint param
const requestExtensions = <T>(extensionPoint: ExtensionPoint<T>): T[] => {
…
}

// batchActions is of type ActionExtension[]
const batchActions = requestExtensions(batchActionsExtensionPoint)
```

P.S.: this would make having the extension points in web-pkg even more handy

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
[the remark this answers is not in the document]
dschmidt: Ah, you skipped the interesting part of the error message:

` 'ExtensionPoint' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'ExtensionPoint'.ts(2322)`

Makes sense, nevermind then.

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
kulmann: I don't get it. As of now it's still valid to request extensions by extension type, without any extensionPoint. E.g. we currently don't have a `search` extension point, we just request extensions of type `search`. Also, if I develop an extension and two apps that want to use the extension, where do I define the `extensionPoint`? Either results in lots of code duplication or the necessity to have a shared package. Being able to request extensions simply by an extensionType or an extensionPointId seems to be a lot more flexible.

kulmann: Could you please check again? Looks a lot cleaner now, thanks for the input ❤️

docs/extension-system/_index.md
```diff
@@ -96,16 +96,16 @@ In contrast to applications, extensions usually have a rather small scope and de
 
 The globally available extension registry provided by the ownCloud Web runtime can be used to both register and query extensions. All extensions
 which are being made available via an `app` get registered in the extension registry automatically. In your custom application code you can
-then query any of the available extensions by providing a `type` string and optionally a list of `scopes`. Throughout the ownCloud Web platform
+then query any of the available extensions by providing a `type` string and optionally a list of `extensionPointIds`. Throughout the ownCloud Web platform
 and most prominently also in the `files` app we have defined some extension points which automatically use certain extensions, see the 
 `Extension Points` section below.
```
dschmidt, line 101: I think this section is outdated

packages/web-app-files/src/extensionPoints.ts
dschmidt: For registering extensions we only need the id, we agreed that it's fine to copy that.

packages/web-app-search/src/index.ts
```diff
@@ -1,48 +1,49 @@
...
-    portal
-  }: {
-    portal: {
-      open: (toApp: string, toPortal: string, order: number, components: Component[]) => void
-    }
-  }): void {
-    portal.open('runtime', 'header', 1, [SearchBar])
   }
-}
+})
```
dschmidt, line 49: Sweeeeet 😍 

Do we need to mention this in the changelog?

packages/web-pkg/src/components/AppBar/AppBar.vue
```diff
@@ -207,6 +210,14 @@ export default defineComponent({
         ] as FileAction[]
       }
 
+      const actionExtensions = requestExtensions<ActionExtension>({
+        id: 'app.files.batch-actions',
```
dschmidt, line 214: Hmm - it's a bit weird to reference an extension point defined in an app in web-pkg ... 👀

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
```diff
@@ -11,40 +11,45 @@ export const useExtensionRegistry = defineStore('extensionRegistry', () => {
...
+          (!e.extensionPointIds || e.extensionPointIds?.includes(extensionPoint.id))
       )
     ) as T[]
   }
 
-  const extensionPoints = ref<ExtensionPoint[]>([])
-  const registerExtensionPoint = (e: ExtensionPoint) => {
+  const extensionPoints = ref<Ref<ExtensionPoint<Extension>[]>[]>([])
+  const registerExtensionPoint = <T extends Extension>(e: ExtensionPoint<T>) => {
+    extensionPoints.value.push(ref([e]))
```
dschmidt, line 31: It's somewhat fishy to create the ref on demand here - why not pass the ref as with registerExtensionPoints?

packages/web-runtime/src/extensionPoints.ts
```diff
@@ -0,0 +1,14 @@
...
+import { computed } from 'vue'
+
+export const topBarCenterExtensionPoint: ExtensionPoint<CustomComponentExtension> = {
+  id: 'app.runtime.header.center',
+  extensionType: 'customComponent',
+  multiple: true
+}
+
+export const extensionPoints = () => {
+  return computed<ExtensionPoint<any>[]>(() => {
```
dschmidt, line 11:
```suggestion
  return computed<ExtensionPoint<unknown>[]>(() => {
```

packages/web-runtime/src/pages/account.vue
```diff
@@ -461,17 +462,15 @@ export default defineComponent({
     const extensionRegistry = useExtensionRegistry()
     const extensionPointsWithUserPreferences = computed(() => {
       return extensionRegistry
-        .getExtensionPoints<ExtensionPoint>()
-        .filter((extensionPoint: ExtensionPoint) => {
+        .getExtensionPoints<ExtensionPoint<Extension>>()
```
dschmidt, line 465: feels like this type argument could be the default as well

dschmidt (approved):
LGTM 😍

docs/extension-system/_index.md
kulmann: thanks, updated it!

packages/web-runtime/src/extensionPoints.ts
kulmann: `Vue: Type unknown does not satisfy the constraint Extension`

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
kulmann: Because when I pass a `Ref<ExtensionPoint<T>>` I need to `unref` it and then `ref` it again as array.

packages/web-app-search/src/index.ts
kulmann: Why in the changelog?

dschmidt: The new search extension api

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
dschmidt: Urgh..........

packages/web-runtime/src/extensionPoints.ts
dschmidt: then Extension? 😅

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
dschmidt: Then I agree it's at least (somewhat) clear that this breaks reactivity ...

docs/extension-system/_index.md
dschmidt: Yeah, happy you chose `global.` and implemented that already

packages/web-pkg/src/composables/piniaStores/extensionRegistry/extensionRegistry.ts
kulmann: I'm going to remove the function for registering single extension points, so that we can always require a `Ref` to be passed.

packages/web-runtime/src/extensionPoints.ts
kulmann: Yep, works. Thanks 😅

dschmidt (approved):

packages/web-pkg/src/components/AppBar/AppBar.vue
kulmann: agreed. as discussed in chat: we'll tackle that in a followup.

Issue #10718, opened 2024-04-05
Introduce interface for resource status indicators 


JammingBen:
The [resource status indicators](https://github.com/owncloud/web/blob/master/packages/web-pkg/src/helpers/statusIndicators.ts) should have a proper interface that is reflected in the `Resource` type.

PR #10759, opened 2024-04-10
fix: share indicators for incoming shares

JammingBen:
## Description
Fixes an issue where share indicators would falsely show for and in incoming shares. These are only needed where outgoing shares are being displayed, meaning in project spaces the user has access to and their personal space.

Refactoring and clean-up in the course of this:

* Introduce a `ResourceStatusIndicator` type
* Remove the unnecessary property `target` from status indicators

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10718

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


kulmann (changes requested):
- If I create a user-share I see a user-share indicator in the file list (correct so far)
- if I then create a link-share, I don't see the user-share indicator anymore, but only the link-share indicator
- both share indicators at the same time are only visible after a page reload / navigation to a different folder and back to the original location

JammingBen:
> * If I create a user-share I see a user-share indicator in the file list (correct so far)
> * if I then create a link-share, I don't see the user-share indicator anymore, but only the link-share indicator
> * both share indicators at the same time are only visible after a page reload / navigation to a different folder and back to the original location

Hmm I can't reproduce this. But I remember seeing this issue a while ago. Anything I'm missing in the steps to reproduce? 🤔

kulmann (approved):
Can't reproduce my recent issue anymore. Looks good to me, but needs a rebase now 😉

Issue #10602, opened 2024-03-14
Incoming notifications don't render properly when notifications open


JammingBen:
## Describe the bug
Incoming notifications do not display properly when having the notifications opened.

## Steps to reproduce
1. Open 2 browser tabs, one being logged in as "admin", one as "einstein"
2. As "einstein", open the notifications and leave them open
3. As "admin", invite "einstein" to a space

## Expected behavior
The notification doesn't get rendered properly for "einstein":

![image](https://github.com/owncloud/web/assets/50302941/e5ec7cf4-3be3-4090-98a0-ce180d11e83a)

## Actual behavior
The notification gets rendered properly for "einstein".

kulmann:
@JammingBen is this relevant for `stable-8.0`?

JammingBen:
> @JammingBen is this relevant for `stable-8.0`?

I mean it's not critical IMO, but yes, the error also happens on `stable-8.0`.

kulmann:
Ok 👍 then for whoever works on this: please fix with PR to `stable-8.0` and then we can do a forward port to `master`.

AlexAndBear:
Was not able to fix this on stable-8.0 since the code differs too much, and I am not quite sure about changing the whole file for a patch release and an issue that is p4

PR #10760, opened 2024-04-10
Fix new notifications broken while notification center is open

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/10602

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/components/Topbar/Notifications.vue
+++ b/packages/web-runtime/src/components/Topbar/Notifications.vue
@@ -212,12 +212,12 @@ export default {
 
     const computeNotificationDataTask = useTask(function* (signal, notifications) {
       for (const notification of unref(notifications)) {
-        if (!notification.hasOwnProperty('computedMessage')) {
+        if (!notification.computedMessage) {
           notification.computedMessage = yield new Promise((resolve) =>
             resolve(getMessage(notification))
           )
         }
-        if (!notification.hasOwnProperty('computedLink')) {
+        if (!notification.computedLink) {
           notification.computedLink = yield new Promise((resolve) => resolve(getLink(notification)))
         }
       }
@@ -269,11 +269,11 @@ export default {
 
     watch(
       [notifications, dropdownOpen],
-      async () => {
+      () => {
         if (!unref(dropdownOpen)) {
           return false
         }
-        await computeNotificationDataTask.perform(notifications)
+        computeNotificationDataTask.perform(notifications)
       },
       { immediate: true }
     )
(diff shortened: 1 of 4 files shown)


packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -198,28 +193,34 @@ export default {
...
       }
-    }
+    }).restartable()
+
+    const computeNotificationDataTask = useTask(function* (signal, notifications) {
+      for (const notification of unref(notifications)) {
+        if (!notification.computedMessage) {
+          notification.computedMessage = yield new Promise((resolve) =>
+            resolve(getMessage(notification))
+          )
```
dschmidt, line 217: This looks superweird - why not just `notification.computedMessage = getMessage(notification)`?

packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -198,28 +193,34 @@ export default {
...
+        }
+        if (!notification.computedLink) {
+          notification.computedLink = yield new Promise((resolve) => resolve(getLink(notification)))
```
dschmidt, line 220: same here

packages/web-runtime/src/components/Topbar/Notifications.vue
AlexAndBear: sonarcloud is reporting a bug, if you don't use yield in a useTask generator func

packages/web-runtime/src/components/Topbar/Notifications.vue
AlexAndBear: sonarcloud is reporting a bug, if you don't use yield in a useTask generator func

packages/web-runtime/src/components/Topbar/Notifications.vue
JammingBen: SonarCloud has a point here, the question is: why use a generator function when you don't have async behaviour? I guess you just want to rely on the `isRunning` functionality of a task? I'd prefer not to use a task here, although I don't have the strongest opinion about it 🤷 A comment would be nice at least, explaining why this is being wrapped in a promise.

AlexAndBear: I really prefer to have a task here, since a loading state seems to be needed here (see previous) implementation. It's quite more convenient to have all loading states originate from the same interface. The previous implementation was hard to understand.

AlexAndBear: moved to fetchNotificationTasj / SSE listener

dschmidt (approved):
LGTM 👍🏻

PR #10764, opened 2024-04-11
Limit useKeyboardAction listener to one modifier key

### LINKED ISSUE #10477: Ctrl+S key listener in text editor / AppWrapper fires on Ctrl+Shift+S as well
## Describe the bug
The key listener we have in the AppWrapper is not an exact match on the keys 😨 

## Steps to reproduce
1. Open up your Firefox
2. Create a markdown file from within the files app (opens in text-editor app)
3. Write some content (optional)
4. Try to make a screenshot of your awesome content with Ctrl+Shift+S

## Expected behavior
Firefox keyboard shortcut for making screenshots works as expected.

## Actual behavior
Firefox keyboard shortcut for making screenshots doesn't work.

## Additional context
The key binding which is supposed to save the content via `Ctrl+S` is defined in https://github.com/owncloud/web/blob/1674c8e17da2c0385d5b0bfed2696dad864949b0/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue#L407

Unfortunately this fires for e.g. `Ctrl+Shift+S` as well. We somehow need to make this an exact match for the pressed keys....

pascalwengerter:
## Related Issue
- Fixes #10477

### DIFF
--- a/packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
+++ b/packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
@@ -78,17 +78,30 @@ export const useKeyboardActions = (options?: KeyboardActionsOptions): KeyboardAc
       return
     }
 
+    type PossibleModifier = 'altKey' | 'ctrlKey' | 'metaKey' | 'shiftKey'
+
     const { key, ctrlKey, metaKey, shiftKey } = event
     let modifier = null
+    const disallowedModifierkeys: PossibleModifier[] = []
+
     if (metaKey || ctrlKey) {
       modifier = ModifierKey.Ctrl
+      disallowedModifierkeys.push('altKey', 'shiftKey')
     } else if (shiftKey) {
       modifier = ModifierKey.Shift
+      disallowedModifierkeys.push('altKey', 'ctrlKey', 'metaKey')
     }
+
+    const hasDisallowedModifier = (event: KeyboardEvent, disallowedKeys: PossibleModifier[]) =>
+      disallowedKeys.some((key) => event[key])
 
     unref(actions)
       .filter((action) => {
-        return action.primary === key && action.modifier === modifier
+        return (
+          action.primary === key &&
+          action.modifier === modifier &&
+          !hasDisallowedModifier(event, disallowedModifierkeys)
+        )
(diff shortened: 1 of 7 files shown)


pascalwengerter:
Not sure if I like the approach, looking for feedback...

dschmidt:
urgh. That seems super odd.

Can't we simply define our api so that we want *only* the specified modifiers to be active? 
I.e. if we specify ctr+c and shift is active we don't recognize the key pressure...

pascalwengerter:
> urgh. That seems super odd.
> 
> Can't we simply define our api so that we want _only_ the specified modifiers to be active? I.e. if we specify ctr+c and shift is active we don't recognize the key pressure...

That would be desireable, but requires quite a refactoring of the current implementation

dschmidt (commented):
Please update the PR title

The naming of vars, types and function names is pretty inconsistent. Maybe you can improve that..

pascalwengerter:
> Please update the PR title
> 
> The naming of vars, types and function names is pretty inconsistent. Maybe you can improve that..

PR name has been updated, also slightly changed the naming of the `disallowedModifiers` variable - not sure if that was what you were hinting at?

packages/web-pkg/src/composables/keyboardActions/useKeyboardActions.ts
```diff
@@ -78,25 +78,38 @@ export const useKeyboardActions = (options?: KeyboardActionsOptions): KeyboardAc
...
+    const hasDisallowedModifier = (event: KeyboardEvent, disallowedModifierKeys: ModifierKey[]) =>
+      disallowedModifierKeys.some((key) => event[key])
+
     unref(actions)
       .filter((action) => {
-        return action.primary === key && action.modifier === modifier
+        return (
+          action.primary === key &&
+          action.modifier === modifier &&
+          !hasDisallowedModifier(event, disallowedModifierKeys)
```
dschmidt, line 103: lol ... we don't need to check this in the loop, if we don't pass in the `action` anyway - we can just do

```typescript
if(disallowedModifierKeys.some((key) => event[key])) {
  return
}
```

above the loop.

kulmann (approved):
nice 👍

Issue #10773, opened 2024-04-12
Properly publish web-client package


JammingBen:
The `web-client` package is supposed to be the successor of the deprecated `owncloud-sdk`. It contains abstraction layers for all currently used APIs (webdav, graph and a few ocs endpoints).

We already published the package a while ago, but it's not up to date anymore and we need to do a little clean up to get it out again.

### To-Dos

- [x] Integrate the webdav implementation in the [exported client](https://github.com/owncloud/web/blob/master/packages/web-client/src/index.ts#L21)
- [x] Clarify: the exported client takes arguments such as the computed user, access token etc. Can we expect users of this package to provide information in such format?
- [x] Publish a latest release of the package
- [x] Implement a proper build process/pipeline (?)
- [x] Add docs for the `web-client` package
- [x] Add a readme file for the `web-client` package
- [x] Decide on how we want to handle releases in the future (couple with Web releases or independent) -> outcome: we couple the release of the package to Web releases. Meaning e.g. when Web v9 gets released, we release web-client v9 as well.

PR #10785, opened 2024-04-15
refactor: prepare web-client package for publish

JammingBen:
## Description
Prepares the `web-client` package for publishing in the future. The main changes:

* The webdav implementation has been added to the main method `client()` of this package. This method initializes and returns all available clients (graph, webdav, ocs).
* The `client()` method now only takes a base url and an axios instance containing all relevant headers. All clients either use this axios instance, or at least the headers it provides (in case of the webdav implementation). 
* Other potential parameters (like e.g. username) need to be passed as options if needed for individual methods.
* Vue concepts like `ref`'s have been removed from the implementation, as we can't expect users of the package to use Vue.
* Moves the `clientInitiatorId` to the `ClientService` since this is supposed to be the single source of truth for such client related stuff.

The commit messages also explain the changes quite well.

## Related Issue

- refs https://github.com/owncloud/web/issues/10773

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-client/src/webdav/listFavoriteFiles.ts
+++ b/packages/web-client/src/webdav/listFavoriteFiles.ts
@@ -1,11 +1,14 @@
 import { urlJoin } from '../utils'
 import { WebDavOptions } from './types'
 import { DAV } from './client'
-import { DavProperties } from './constants'
+import { DavProperties, DavPropertyValue } from './constants'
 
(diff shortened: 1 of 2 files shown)


dschmidt (approved):
LGTM 😍

PR #10787, opened 2024-04-15
Add loading state for oc modal 

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


JammingBen:
> I think I will add a secondaryButton to the OcModal anyway and adjust the loading logic (button disabled) here as well, to be future proof.

Actually we has this and @dschmidt and I decided to remove it because it's just way to specific for 1-2 use cases 😅  You can always work with events, see e.g. how this is done with the `confirmDisabled` event. You could add a similar event to control the loading state from inside a custom modal component.

packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -15,7 +15,7 @@
           <h2 id="oc-modal-title" class="oc-text-truncate" v-text="title" />
         </div>
         <div class="oc-modal-body">
-          <div v-if="$slots.content" key="modal-slot-content" class="oc-modal-body-message">
+          <div v-if="false" key="modal-slot-content" class="oc-modal-body-message">
```
JammingBen, line 18: Seems like a dev left-over?

AlexAndBear: damned, yes

JammingBen:
The icon is a bit hard to see and feels cramped to me 🤔 

image

The share panel has a bit more margin and switches the button appearance to `outline` when loading, maybe we can do this here as well?

image

JammingBen (approved):
Nice 👍

PR #10806, opened 2024-04-19
docs: add proper readme content for web-client package

JammingBen:
## Related Issue

- refs https://github.com/owncloud/web/issues/10773

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [x] Documentation

### DIFF
--- a/packages/web-client/README.md
+++ b/packages/web-client/README.md
@@ -1,3 +1,69 @@
 # web-client
 
-This is experimental.
+The `web-client` is a standalone package that allows you to interact with the [ownCloud Infinite Scale (oCIS)](https://github.com/owncloud/ocis/) APIs via TypeScript. It provides an abstraction layer between the server and a (web-) application that converts API data into objects with helpful types and utilities. This abstraction ensures that users of the APIs don't need in-depth knowledge about them, such as required methods or returned status codes (note that the Graph layer is currently an exception to this, meaning it doesn't process responses in any way).
+
+The supported APIs are:
+
+- Graph (drives, sharing, user & group management)
+- OCS (capabilities & url signing)
+- WebDAV (file operations)
+
+## Installation
+
+Depending on your package manager, run one of the following commands:
+
+```
+$ npm install @ownclouders/web-client
+
+$ pnpm add @ownclouders/web-client
+
+$ yarn add @ownclouders/web-client
+```
+
+## Usage
+
+To utilize the `web-client`, you must instantiate it with a base URI corresponding to your oCIS deployment and an axios instance. The axios instance, or at least its headers, is being used for all requests, requiring the inclusion of all relevant headers such as authorization.
+
+```
+import axios from axios
+import { client } from '@ownclouders/web-client'
+
+const accessToken = 'some_access_token'
+const baseUri = 'some_base_uri'
+
+const axiosClient = axios.create({
+	headers: { Authorization: accessToken }
+})
+
+const { graph, ocs, webdav } = client({ axiosClient, baseURI })
+```
+
+### Graph
+
+The following example demonstrates how to retrieve all drives accessible to the user and subsequently convert them into a `SpaceResource`. Such a resource can then be used to e.g. fetch files and folders (see example down below).
+
+```
+import { buildSpace } from '@ownclouders/web-client'
+
+const myDrives = await graph.drives.listMyDrives()
+
+const spaceResources = myDrives.data.value.map(buildSpace)
+```
+
+### OCS
+
+The following examples demonstrate how to fetch capabilities and sign URLs.
+
+```
+const capablities = await ocs.getCapabilities()
+
+const signedUrl = await ocs.signUrl('some_url_to_sign', 'your_username')
+```
+
+### WebDAV
+
+The following example demonstrates how to list all resources of a given `SpaceResource` (see above how to fetch space resources).
+
+```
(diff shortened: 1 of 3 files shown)


JammingBen:
@kulmann Do you feel like additional dev docs are necessary for the package, or is the `README.md` file sufficient?

packages/web-client/README.md
```diff
@@ -1,3 +1,69 @@
 # web-client
 
-This is experimental.
+The `web-client` is a standalone package that allows you to interact with the [ownCloud Infinite Scale (oCIS)](https://github.com/owncloud/ocis/) APIs via TypeScript. It also provides convenient utilities and interfaces that are useful when working with these APIs.
```
kulmann, line 3: I'd like to add to this, that `web-client` also serves as an abstraction layer for the APIs. We never let the user deal with raw requests and responses. Instead we're transforming data into objects with useful types and helpers (like e.g. permission check helpers).

packages/web-client/README.md
```diff
@@ -1,3 +1,69 @@
...
+$ npm install @ownclouders/web-client
+
+$ pnpm add @ownclouders/web-client
+
+$ yarn add @ownclouders/web-client
+```
+
+## Usage
+
+To utilize the `web-client`, you must instantiate it with a base URI corresponding to your oCIS deployment and an axios instance. The axios instance, or at least its headers, is being used for all requests, necessitating the inclusion of all pertinent headers such as authorization.
```
kulmann, line 25:
```suggestion
To utilize the `web-client`, you must instantiate it with a base URI corresponding to your oCIS deployment and an axios instance. The axios instance, or at least its headers, is being used for all requests, necessitating the inclusion of all pertinent headers such as authorization.
```

Could you use simpler words for `necessitating` and `pertinent`? 😆

packages/web-client/README.md
```diff
@@ -1,3 +1,69 @@
...
+const axiosClient = axios.create({
+	headers: { Authorization: accessToken }
+})
+
+const { graph, ocs, webdav } = client({ axiosClient, baseURI })
+```
+
+### Graph
+
+The following example demonstrates how to retrieve all drives accessible to the user and subsequently convert them into a `SpaceResource`. Such a resource can then be used to e.g. fetch resources (see example down below).
```
kulmann, line 43:
```suggestion
The following example demonstrates how to retrieve all drives accessible to the user and subsequently convert them into a `SpaceResource`. Such a resource can then be used to e.g. fetch files and folders (see example down below).
```

packages/web-client/README.md
kulmann: This is mostly the case for the webdav part of the client. There you could state as an example that consumers of the client would never need to deal with the webdav domain knowledge of what a `PROPFIND` is, instead we offer functions like `listFiles` and `getFileInfo`.

kulmann (changes requested):

packages/web-client/README.md
JammingBen: > I'd like to add to this, that web-client also serves as an abstraction layer for the APIs. We never let the user deal with raw requests and responses. Instead we're transforming data into objects with useful types and helpers (like e.g. permission check helpers).

For Graph this is not true though :/ The Graph abstraction just wraps the generated SDK but adds nothing to it. (something I'd really like to change in the future).

A general note might be misleading because of that. Maybe a dedicated paragraph just for WebDAV underneath this one?

packages/web-client/README.md
JammingBen: Damn you, ChatGPT, use simpler words! 🤖

packages/web-client/README.md
JammingBen: Or write a general note and with an exception for Graph 🤔

JammingBen: ...or just a general note without any reference to one of the layers 🙈 I mean (or I hope) we will adjust the Graph layer at some point in time in the future to reflect this.

kulmann: I'd prefer the general note and an exception for Graph, because I like to imagine the web-client as an abstraction layer.

kulmann (approved):
🥳

kulmann:
> @kulmann Do you feel like additional dev docs are necessary for the package, or is the `README.md` file sufficient?

We already mention the `web-client` package here: https://owncloud.dev/clients/web/development/repo-structure/#web-client However it states, that no dedicated documentation is available, yet. Even with this readme, that's still true I think. If, at all, you could add a link to the readme to that section of the dev docs. On top of that I'd like to have autogenerated docs about the namespaces, interfaces and functions (which requires a lot of code doc blocks first). Maybe a good time now to look into existing tooling. :-)

kulmann:
> > @kulmann Do you feel like additional dev docs are necessary for the package, or is the `README.md` file sufficient?
> 
> We already mention the `web-client` package here: https://owncloud.dev/clients/web/development/repo-structure/#web-client However it states, that no dedicated documentation is available, yet. Even with this readme, that's still true I think. If, at all, you could add a link to the readme to that section of the dev docs. On top of that I'd like to have autogenerated docs about the namespaces, interfaces and functions (which requires a lot of code doc blocks first). Maybe a good time now to look into existing tooling. :-)

Also, mentioning the web-client readme in the `extension system` docs section would make sense as well, what do you think?

kulmann (approved):
🎉

PR #10823, opened 2024-04-23
Add draft for fixing (aria-)label bug on OcSelect

### LINKED ISSUE #10736: [a11y] 9.2.5.3 Visible labeling part of the accessible name
audit: https://infinite.owncloud.com/s/PnaAukHGGsLnrMM

https://bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-2-5-3-sichtbare-beschriftung-teil-des-zugaenglichen-namens

### Personal

- [ ] In the "Keywords" area, the combo box element has a visual text
"Adding or editing keywords", but this does not appear in the accessible name of the
element, as it is overwritten by the use of the "aria-label" attribute.
attribute.
image

### Spaces 

- [ ] In the "Quota for Space 8A - Change art" dialog, the visible label of the "Quota" field is changed by using the
Quota" field is overwritten by using the "aria-label" attribute "Search for option".
is overwritten.

### Account

- [ ] The visible label of the "Language" field is overwritten by the use of the "aria-label"
attribute "Search for option".

pascalwengerter:
## Description
From my understanding, the `OcSelect` component currently always gets a hardcoded "Search for option" aria-label, which overwrites the optional label that can be passed as a prop.
I've fixed that by only setting the "Search for option" aria-label if no label is passed via prop.
Another consideration is the visibility of the custom label, for which I added another prop - now, a custom label can be passed but also hidden in the GUI, while still available for screenreader users.

The next step for finalizing this PR (granted that my approach is approved) would be checking all occurances of the `OcSelect` and decide on whether they need a (hidden) custom label.

## Related Issue
- Fixes #10736

## Motivation and Context
Accessibility improvements

### DIFF
--- a/changelog/9.0.0_2024-06-17/enhancement-accessibility-improvements
+++ b/changelog/9.0.0_2024-06-17/enhancement-accessibility-improvements
@@ -6,4 +6,3 @@ https://github.com/owncloud/web/issues/5383
 https://github.com/owncloud/web/issues/5391
 https://github.com/owncloud/web/issues/10731
 https://github.com/owncloud/web/pull/10802
-https://github.com/owncloud/web/pull/10823
--- a/changelog/9.0.0_2024-06-17/enhancement-accessibility-improvements
+++ b/changelog/9.0.0_2024-06-17/enhancement-accessibility-improvements
@@ -6,3 +6,4 @@ https://github.com/owncloud/web/issues/5383
 https://github.com/owncloud/web/issues/5391
 https://github.com/owncloud/web/issues/10731
 https://github.com/owncloud/web/pull/10802
+https://github.com/owncloud/web/pull/10823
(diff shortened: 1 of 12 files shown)


pascalwengerter:
> While I usually like is and has prefixes I don't think they align with the props already present...
> 
> I'm not sure it's more elegant, but another approach might be to be able to pass in a label or an aria label... lf no aria label is passed, it uses the regular label. If only the aria label is passed, no label is shown.
> 
> No idea if that is a good or a bad thing but it allows to pass in two different values... and it avoids a boolean prop which is not super bad but always smells a bit to me for things like this

Good idea, let me get started on it!

pascalwengerter:
I've decided to switch the used `vue-select` library for `vs-vue3-select`, a fork of a fork that is maintained a little better (the bar is low, though), esp. regarding custom translations which we'll need for the next a11y audit.

AlexAndBear:
Let's check if we really need to integrate the fork.

I created an issue on vue-select repo, which was about setting the aria-label, this has been fixed in the meantime

https://github.com/sagalbot/vue-select/pull/1765

JammingBen:
I think this has been superseded by https://github.com/owncloud/web/pull/11396.

pascalwengerter:
> I think this has been superseded by #11396.

True, thanks for the housekeeping!

Issue #10689, opened 2024-03-28
[Sharing NG] Replace tokeninfo endpoint


JammingBen:
The tokeninfo endpoint is deprecated, hence we need to remove it. All necessary information is now available when requesting files for a public link via webdav. See https://github.com/owncloud/ocis/issues/8858 for more detail.

Currently blocked by https://github.com/owncloud/ocis/issues/8858.

PR #10869, opened 2024-05-02
refactor: remove tokeninfo endpoint usage

JammingBen:
## Description
Removes the tokeninfo endpoint usage and uses the webdav response instead to retrieve all necessary information to resolve a public/internal link.

The biggest change here is that we don't get the file id anymore when resolving an internal link. To solve this we need to login the user first and then redirect to the link resolving page again (only this time with an authenticated context, hence the new route `/i/...`). After that last redirect the authenticated user can now fetch all necessary information for resolving.

See https://github.com/owncloud/ocis/issues/8858 for more detail.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10689

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


kulmann (approved):
👍

Issue #10839, opened 2024-04-26
e2e tests for sse (server sent events)


ScharfViktor:
related tasks and PRs: 
- https://github.com/owncloud/web/issues/10650 
- https://github.com/owncloud/web/pull/10807
- https://github.com/owncloud/web/issues/10833
- https://github.com/owncloud/ocis/issues/8950
- https://github.com/owncloud/web/issues/10648

#### creating: 
- [x] upload file/folder
- [x] create file/folder (`.txt` `.odt`)
- [x] restore file/folder from trash
- [x] move file/folder
- [x] copy file/folder
- [x] lock/unlock file

#### updating: 
- [x] rename file/folder
- [x] change file content
- [x] restore file version

### space: 
- [ ] disable
- [ ] enable 
- [ ] delete
- [ ] change description/space image/space icon
- [x] space-member-added
- [x] space-member-removed
- [x] space-share-updated

### share:
- [x] share created
- [x] share updated
- [x] share deleted
 
event in the `eventStream`  in the GET `api/v1/notifications/sse` request: 
  ```
NOTIFICATION = 'userlog-notification',
  POSTPROCESSING_FINISHED = 'postprocessing-finished',
  FILE_LOCKED = 'file-locked',
  FILE_UNLOCKED = 'file-unlocked',
  FILE_TOUCHED = 'file-touched',
  ITEM_RENAMED = 'item-renamed',
  ITEM_TRASHED = 'item-trashed',
  ITEM_RESTORED = 'item-restored',
  ITEM_MOVED = 'item-moved',
  FOLDER_CREATED = 'folder-created',
  SPACE_MEMBER_ADDED = 'space-member-added',
  SPACE_MEMBER_REMOVED = 'space-member-removed',
  SPACE_SHARE_UPDATED = 'space-share-updated',
  SHARE_CREATED = 'share-created',
  SHARE_REMOVED = 'share-removed',
  SHARE_UPDATED = 'share-updated',
  LINK_CREATED = 'link-created',
  LINK_REMOVED = 'link-removed',
  LINK_UPDATED = 'link-updated'
```

Besides the appearance/disappearance of resources on the web page, we usually check the api responses. but checking `notifications/sse` seems tricky. We need to somehow listen `GET api/v1/notifications/sse` during the whole test, which is triggered only once after login and gets 200 with an empty response.

ScharfViktor:
The sse is not yet fully implemented. Therefore, only this test suite is available for now

[...]

Issue #10507, opened 2024-02-22
Handle the remove/update share permissions response.


2403905:
## Describe the bug
Regarding the backend changes [https://github.com/owncloud/ocis/issues/8273](#8273) we should handle the remove/update share permissions response 423 when the file is locked.

## Steps to reproduce
1. admin created the `file.txt` and shared to einstein with the editor permissions.
2. einstein locked the file
```
curl -XLOCK 'https://localhost:9200/remote.php/dav/spaces/storage-users-1$some-admin-user-id-0000-000000000000!some-admin-user-id-0000-000000000000/file.txt' -ueinstein:relativity \
-H "Content-Type: application/xml" -H "timeout: Second-300" -d "<?xml version='1.0' encoding='UTF-8'?><d:lockinfo xmlns:d='DAV:'><d:lockscope><d:shared/></d:lockscope></d:lockinfo>" -vk
```
3. admin tries to remove/update share permissions.

## Expected behavior
admin get the message that the file is locked.

## Actual behavior
On the delete share action user recursively gets the removing dialog window

## Setup
local

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

ScharfViktor:
we disable on web delete rename or move action if file is locked
<img width="473" alt="Screenshot 2024-02-23 at 09 47 17" src="https://github.com/owncloud/web/assets/84779829/d5c46a72-6264-4631-accb-78f8c83ce2cf">

If possible, I would prefer that the user not have the ability to change or delete the resource share.

my expected behavior: 
<img width="478" alt="Screenshot 2024-02-23 at 09 48 54" src="https://github.com/owncloud/web/assets/84779829/64e540df-fe9a-4472-8fea-d7be7ff732cb">

kulmann:
@AlexAndBear yes, that would be fine. The file name should have the lock information, both in the file list and in the right sidebar header section. As long as that shows the lock information (i.e. small icon with a lock) I wouldn't clutter the UI with even more lock information. Disabling `write` actions is sufficient then. 👍

ScharfViktor:
re-tested. can confirm that case with file is fixed 
<img width="910" alt="image" src="https://github.com/owncloud/web/assets/84779829/8282a270-b8bd-48cc-9dde-a4b94851cd0a">

@kulmann @micbar I'm sorry to say, but we didn't have all the cases covered.
[...]

PR #10873, opened 2024-05-03
[test-only] check lock sse event

ScharfViktor:
related #10839
check SSE events
- file-unlocked
- file-locked

check that user: 
- can see lock icon
- cannot edit locked file
- cannot edit/delete share of the locked file #10507
- can create link or share to the locked file

### DIFF
--- a/tests/e2e/cucumber/steps/ui/shares.ts
+++ b/tests/e2e/cucumber/steps/ui/shares.ts
@@ -340,8 +340,8 @@ Then(
     )
 
     if (actionType === 'should') {
-      await expect(changeRole).not.toBeDisabled
-      await expect(changeShare).not.toBeDisabled
+      await expect(changeRole).not.toBeDisabled()
+      await expect(changeShare).not.toBeDisabled()
     } else {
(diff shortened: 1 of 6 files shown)


saw-jan (approved):
There are some lint issues. but the PR looks good :+1:

tests/e2e/cucumber/steps/ui/shares.ts
```diff
@@ -321,3 +321,30 @@ Then(
...
+    const changeRole = shareObject.changeRoleLocator(
+      this.usersEnvironment.getUser({ key: recipient })
+    )
+    const changeShare = shareObject.changeShareLocator(
+      this.usersEnvironment.getUser({ key: recipient })
+    )
+
+    if (actionType === 'should') {
+      await expect(changeRole).not.toBeDisabled
+      await expect(changeShare).not.toBeDisabled
```
saw-jan, line 344: The lint problem was here
```suggestion
      await expect(changeRole).not.toBeDisabled()
      await expect(changeShare).not.toBeDisabled()
```

PR #10876, opened 2024-05-06
fix: types in web-pkg

JammingBen:
## Description
Fix TypeScript types in web-pkg.

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-pkg/src/composables/actions/files/useFileActions.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActions.ts
@@ -224,11 +224,12 @@ export const useFileActions = () => {
   }
 
   const getDefaultAction = (options: GetFileActionsOptions): Action | null => {
-    const filterCallback = (action) =>
+    const filterCallback = (action: FileAction) =>
       action.isVisible({
         ...options,
+        // FIXME: parent does not exit in FileActionOptions, what is this supposed to do?
         parent: unref(currentFolder)
-      })
+      } as any)
 
     // first priority: handlers from config
     const enabledEditorActions = unref(editorActions).filter(filterCallback)
--- a/packages/web-pkg/src/composables/localStorage/useLocalStorage.ts
+++ b/packages/web-pkg/src/composables/localStorage/useLocalStorage.ts
@@ -1,6 +1,6 @@
-import { ref, watch, unref } from 'vue'
+import { ref, watch, unref, Ref } from 'vue'
 
-export const useLocalStorage = (key: string, defaultValue: any = undefined): any => {
+export const useLocalStorage = <T>(key: string, defaultValue: T = undefined) => {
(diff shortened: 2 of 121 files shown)


packages/web-pkg/src/composables/actions/files/useFileActions.ts
```diff
@@ -224,11 +224,12 @@ export const useFileActions = () => {
   }
 
   const getDefaultAction = (options: GetFileActionsOptions): Action | null => {
-    const filterCallback = (action) =>
+    const filterCallback = (action: FileAction) =>
       action.isVisible({
         ...options,
+        // FIXME: parent does not exit in FileActionOptions, what is this supposed to do?
```
dschmidt, line 230: uargh

packages/web-pkg/src/composables/localStorage/useLocalStorage.ts
```diff
@@ -1,14 +1,14 @@
...
+export const useLocalStorage = <T>(key: string, defaultValue: T = undefined) => {
   const existingValue = window.localStorage.getItem(key)
   const variable = ref(defaultValue)
 
   if (existingValue) {
     try {
       variable.value = JSON.parse(existingValue)
     } catch {
-      variable.value = existingValue
+      ;(variable as Ref<string>).value = existingValue
```
dschmidt, line 11: This doesn't look like it should be necessary or even completely correct

packages/web-pkg/tests/unit/components/FilesList/ResourceTable.spec.ts
```diff
@@ -351,7 +351,9 @@ describe('ResourceTable', () => {
       const tr = await wrapper.find('.oc-tbody-tr-forest .oc-resource-name')
       await tr.trigger('click')
 
-      expect(wrapper.emitted().fileClick[0][0].resources[0].name).toMatch('forest.jpg')
+      expect(
+        (wrapper.emitted('fileClick')[0][0] as { resources: Resource[] }).resources[0].name
```
dschmidt, line 355: can we maybe type the emit better? Or is that only possible with script setup?

dschmidt (approved):
nice! Thanks! 🚀

packages/web-pkg/src/composables/localStorage/useLocalStorage.ts
JammingBen: Problem is that `existingValue` is a string, so in that catch-phrase here `variable` will eventually be a string as well.

packages/web-pkg/tests/unit/components/FilesList/ResourceTable.spec.ts
JammingBen: As discussed, seems only be possible with script setup.

Issue #10852, opened 2024-04-29
Make `Shortcuts` clearer


tbsbdr:
# Description

## User Stories

* > As a user who is unfamiliar with shortcuts I want clear guidance so that I understand what will happen.

## Value
- Error prevention
- Help and documentation

## Acceptance Criteria
- hide shortcut name if it is empty
- add placeholder text `example.org`
- Show shortcut name if filled
- remove `.url` from the shortcut name form
- URL and name form is in sync
- Add contextual helpers:
  -   > Enter the target URL of a webpage or the name of a file. Users will be directed to this webpage or file.
  -   > Shortcut name as it will appear in the file list.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

### hide shortcut name if it is empty
- add placeholder text `example.org`
![image](https://github.com/owncloud/web/assets/26610733/23683d31-53cd-482a-9b04-021fdc50b4d5)
![image](https://github.com/owncloud/web/assets/26610733/4cf49af4-80c3-4951-b911-d94b2e30709b)
### Show shortcut name if filled
- remove `.url` from the shortcut name form
![image](https://github.com/owncloud/web/assets/26610733/55f82fbe-d9eb-4617-982a-781fb13eec74)
### Add contextual helper to Webpage or file

> Enter the target URL of a webpage or the name of a file. Users will be directed to this webpage or file.

![image](https://github.com/owncloud/web/assets/26610733/5c3ac229-5e1d-4e84-a691-7d7ed536a316)

### Add contextual helper to shortcut name

> Shortcut name as it will appear in the file list.

![image](https://github.com/owncloud/web/assets/26610733/c4463019-5db1-4498-a79c-684d906c7598)

PR #10877, opened 2024-05-06
Make create shortcut modal ux more easy

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/10852

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/CreateShortcutModal.vue
+++ b/packages/web-pkg/src/components/CreateShortcutModal.vue
@@ -2,14 +2,33 @@
   <oc-text-input
     id="create-shortcut-modal-url-input"
     v-model="inputUrl"
-    :label="$gettext('Shortcut to a webpage or file')"
+    :placeholder="'example.org'"
+    :label="$gettext('Webpage or file')"
     @keydown.up="onKeyUpDrop"
     @keydown.down="onKeyDownDrop"
     @keydown.esc="onKeyEscDrop"
     @keydown.enter="onKeyEnterDrop"
     @input="onInputUrlInput"
     @click="onClickUrlInput"
-  />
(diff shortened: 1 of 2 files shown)


AlexAndBear:
I've decided to use a slot here for the ct helper, but we could also add some properties for the component itself like 'labelContextualHelperData' or something, but not too sure about that, any opinions @JammingBen

JammingBen:
> I've decided to use a slot here for the ct helper, but we could also add some properties for the component itself like 'labelContextualHelperData' or something, but not too sure about that, any opinions @JammingBen

I prefer the solution with the slot like you did it 👍

packages/web-pkg/src/components/CreateShortcutModal.vue
```diff
@@ -2,14 +2,29 @@
...
     @keydown.down="onKeyDownDrop"
     @keydown.esc="onKeyEscDrop"
     @keydown.enter="onKeyEnterDrop"
     @input="onInputUrlInput"
     @click="onClickUrlInput"
-  />
+  >
+    <template #label>
+      <div class="oc-flex oc-flex-middle create-shortcut-modal-label">
+        <label for="create-shortcut-modal-url-input" v-text="$gettext('Webpage for file')"></label>
```
JammingBen, line 16: I think?

```suggestion
        <label for="create-shortcut-modal-url-input" v-text="$gettext('Webpage or file')"></label>
```

JammingBen (approved):

PR #10881, opened 2024-05-07
fix: idp logout issues

JammingBen:
## Description
Fixes falsely showing the logout page after opening ownCloud Web with an expired token. The issue seemed to be related to the fact that we manually logged out the user after token expiry, whereas the correct way is to let the IDP auth flow handle this. There is no need for a custom implementation on our side.

Also removes the threshold because it doesn't really make sense IMO. Either the token is expired, or it's still valid.

## Related Issue
- refs https://github.com/owncloud/enterprise/issues/6639
- refs https://github.com/owncloud/enterprise/issues/6560
- refs https://github.com/owncloud/enterprise/issues/6627

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-idp-logout-issues
+++ b/changelog/unreleased/bugfix-idp-logout-issues
@@ -1,0 +1,5 @@
+Bugfix: IDP logout issues
+
+Falsely showing the logout page after opening ownCloud Web with an expired token has been fixed.
+
+https://github.com/owncloud/web/pull/10881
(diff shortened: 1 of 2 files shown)


JammingBen:
@kulmann @dschmidt Pinging you 2 for review since you got the most knowledge about IDP stuff. I'm still not 100% sure about the solution, but it looks good to me and seems to work.

kulmann (approved):
Nice finding 💪

dschmidt (approved):
LGTM (i guess)

AlexAndBear (approved):

Issue #9784, opened 2023-10-09
[web] Open shares without having to accept it


tbsbdr:
# Description

## User Stories

* > As a user I don't want to accept shares so that I can just klick on a share to access it.

## Acceptance Criteria
- klick on a share shows the resource
- no "accept" aka "sync" needed

## Definition of ready
[ ] everybody needs to understand the value written in the user story
[ ] acceptance criteria has to be defined
[ ] all dependencies of the user story need to be identified
[ ] feature should be seen from an end user perspective
[ ] user story has to be estimated
[ ] story points need to be less then 20

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] code review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

tbsbdr:
depend on  Sharing NG and #9709

tbsbdr:
depends on sharing NG

tbsbdr:
@JammingBen could you list what would be needed to move this issue forward?

JammingBen:
> @JammingBen could you list what would be needed to move this issue forward?

@dschmidt Do you know from the top of your head? 🤔 I need to dig into this topic again...

JammingBen:
Okay so I digged into it, looks like we have all the necessary information from the server now. So all that's necessary is to get started in Web.

PR #10887, opened 2024-05-08
feat: make unsynced shares accessible

JammingBen:
## Description
Makes unsynced shares accessible. The way to make this work is by using the remote item id instead of the share jail id for opening/resolving into incoming shares.

Notable changes in the course of this:

* Renames `shareId` to `remoteItemId` and `shareRoot` to `remoteItemPath` on resources to match with the Graph terminology.
* Removes the `remoteItemId` on space resources because it only matters for share spaces, in which case this is simply the space id. That means instead of blindly going for `space.remoteItemId`, it now checks if the space is actually a share space and then uses its id.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/9784

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/Search/List.vue
+++ b/packages/web-app-files/src/components/Search/List.vue
@@ -421,10 +421,10 @@ export default defineComponent({
       ]
     })
 
-    const resourceDomSelector = ({ id, shareId }: Resource) => {
+    const resourceDomSelector = ({ id, remoteItemId }: Resource) => {
       let selectorStr = id.toString()
-      if (shareId) {
-        selectorStr += shareId
+      if (remoteItemId) {
+        selectorStr += remoteItemId
       }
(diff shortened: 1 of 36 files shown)


dschmidt:
Todo: handle legacy links

JammingBen:
> Todo: handle legacy urls

Discussed with @tbsbdr , it's okay for the URLs to break because of the `shareId`.

AlexAndBear (approved):
LGTM, 
we should add E2E tests for this

JammingBen:
> LGTM, we should add E2E tests for this

Agreed, I created https://github.com/owncloud/web/issues/10931 to keep track.

dschmidt:
Congrats to us! 🎉

PR #10893, opened 2024-05-13
Update vue-concurrency and bump required node version (plus types) to sort out peer dependency warnings

pascalwengerter:
## Description
I contributed a little patch upstream to remove vitepress from the dependency chain, it reduces the noise in the install logs quite a bit


dschmidt (approved):

kulmann (approved):
Really no idea why `chromedriver` fails to install in CI, seems unrelated. Maybe your PR needs a rebase.
Thanks for the version bump!

Issue #6651, opened 2022-03-23
Buttons "New" and "Upload" are still available after changing the role of the space


ScharfViktor:
branch: 5.3.0-rc.3 docker

steps:
- the administrator creates the space
- the administrator gives the space to user1 with the editor role
- user1 can create a file/folder inside the space. The "Create" and "Upload" buttons are available.
- admin changes the role to viewer
- user1 refreshes the page but buttons still available although the answer is that the role of viewer

Expected result: "Create" and "Upload" buttons are not available
actual result: "Create" and "Upload" buttons are available, if user1 tries to create file/folder - get 403

https://user-images.githubusercontent.com/84779829/159697305-bce2b454-b234-46a8-bde8-42dc0c915277.mov

JammingBen:
Thx for reporting! However, this is a backend issue I believe.

At the end, you're checking how the permissions look like in the backend response. But you're checking the wrong request. The right one would be the request where the single drive is fetched (I believe the second one in the list), not the one where all drives are listed.

Ideally, both responses have the same data for this space. But because of a caching issue in the backend this is not the case yet.

ScharfViktor:
Yes, you are right. I'll create issue in ocis
request GET `https://host.docker.internal:9200/graph/v1.0/drives/3ea1ce3d-c34f-4b8e-bc6c-3ffca95e2ae0` give us wrong role after updating role:
<img width="1431" alt="Screenshot 2022-03-23 at 14 15 11" src="https://user-images.githubusercontent.com/84779829/159708242-55461781-6112-4f19-9271-7ee9c894c7fa.png">

request GET `https://host.docker.internal:9200/graph/v1.0/me/drives?%24top=0&%24skip=0&%24orderby=name+asc&%24filter=driveType+eq+project`
<img width="1431" alt="Screenshot 2022-03-23 at 14 15 55" src="https://user-images.githubusercontent.com/84779829/159708599-e9981916-0239-421d-8474-2fc572b93f61.png">

ScharfViktor:
closed in favor of https://github.com/owncloud/ocis/issues/3378

PR #10894, opened 2024-05-13
fix TUS uploading #6651

ScharfViktor:
related https://github.com/owncloud/enterprise/issues/6651#issuecomment-2107052192

@butonic I changed `OCIS_CORS_ALLOW_ORIGINS` and now TUS upload works again

CC @dschmidt @JammingBen

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,7 +87,7 @@ services:
     environment:
       <<: *ocis-environment
       OCIS_URL: ${OCIS_URL:-https://host.docker.internal:9200}
-      OCIS_CORS_ALLOW_ORIGINS: 'https://host.docker.internal:9200'
+      OCIS_CORS_ALLOW_ORIGINS: https://host.docker.internal:9200,https://host.docker.internal:9201
       OCM_WEBAPP_TEMPLATE: https://host.docker.internal:9201/o/{{.Token}}/{relative-path-to-shared-resource}
       # make the registry available to the app provider containers
       MICRO_REGISTRY: 'nats-js-kv'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,7 +87,7 @@ services:
     environment:
       <<: *ocis-environment
       OCIS_URL: ${OCIS_URL:-https://host.docker.internal:9200}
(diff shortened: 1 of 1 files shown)


butonic:
Some more explanation:
1. The first POST request works because the tus.go handlar implementation in ocdav internally makes an InitiateUpload request and then directly a PATCH request without setting an origin header. As a result the tus handler does not check the origin / no CORS check needs to be made (the request is not made by a browser)
2. The next PATCH request then fails because the origin header sent by the browser is forwardet to the datahandler where the tus handler now fails the origin check.

IMO the datagateway should not swallow that error ... may it does not, but that is only visible when enabling prefilight requests in the devtools. Maybe we can automatically add the OCIS_URL to the allowed domains? And use an OCIS_CORS_ADDITIONAL_DOMAINS?

dschmidt (changes requested):
This will most likely break uploads in Vite (our development server). imho it's an ocis bug if uploading from the same host and port requires cors settings... and even if it's considered wontfix, I'm still against breaking our dev server

ScharfViktor:
Do you mean we can implement new `OCIS_CORS_ADDITIONAL_DOMAINS`? because I didn't find this env in docs

I see that OCIS_CORS_ALLOW_ORIGINS  can be used as list: https://doc.owncloud.com/ocis/next/deployment/services/s-list/webdav.html so we can add more url to the CORS origins like: `OCIS_CORS_ALLOW_ORIGINS: https://host.docker.internal:9200,https://host.docker.internal:9201`

dschmidt:
@ScharfViktor have you checked this works? on 9200 and on 9201 with vite?

ScharfViktor:
> @ScharfViktor have you checked this works? on 9200 and on 9201 with vite?

I don't know much about how work with vite, but seems that works

- pnpm install
- pnpm buld
- docker-compose up -d
- pnpm vite

Result: https://host.docker.internal:9201/ and https://host.docker.internal:9200 work. (TUS upload works fine)

Screenshot 2024-05-13 at 14 12 43

dschmidt (approved):
LGTM, thanks for figuring this out! <3

Issue #10855, opened 2024-04-29
[Web] Research for `long running actions` (~ Web Worker)


tbsbdr:
research ticket for: https://github.com/owncloud/web/issues/7039

- expect outcome: preferred way how to technically tackle https://github.com/owncloud/web/issues/7039
- Timebox: 3 pd

JammingBen:
See the proof of concept and all related information here: https://github.com/owncloud/web/pull/10916

It feels pretty stable to me, however, I'd recommend to start using it in small incrementions. First we should think about where exactly we want to use this functionality. Looking at https://github.com/owncloud/web/issues/7039, the following 3 are probably good candidates:

* copy
* move
* delete

Then there are 2 more scenarios to think of: uploading and token refresh.

Uploading is not necessary IMO. Yes, it happens in the main thread, but it's already a non-blocking async operation and the user can navigate freely within the application. Also, I don't think we can simply introduce web workers in Uppy.

Token refresh is a bit more complicated. I'm gonna start with saying that I have absolutely no idea if and how we could introduce web workers here, especially since a big chunk of the functionality is located in the oidc lib we're using (btw this is a known issue: https://github.com/authts/oidc-client-ts/issues/954). I'd handle this topic completely separate to be honest, starting with optimizing our token refresh process. It currently re-loads hundreds of assets which results in an error when the network connection is slow. Maybe something similar happens when the browser tab is being throttled? Then we could try doing the token renewal earlier than 10 seconds before expiry, at least if the the token lifespan is > 5 minutes.

JammingBen:
Update to this statement:

> Token refresh is a bit more complicated. I'm gonna start with saying that I have absolutely no idea if and how we could introduce web workers here, especially since a big chunk of the functionality is located in the oidc lib we're using (btw this is a known issue: https://github.com/authts/oidc-client-ts/issues/954). I'd handle this topic completely separate to be honest, starting with optimizing our token refresh process. It currently re-loads hundreds of assets which results in an error when the network connection is slow. Maybe something similar happens when the browser tab is being throttled? Then we could try doing the token renewal earlier than 10 seconds before expiry, at least if the the token lifespan is > 5 minutes.

@kulmann and I did a brainstorming yesterday and found a promising approach. The idea is to just put the timer in a worker, but keep the actual token renewal in the main thread. I'm currently implementing this approach.

For the other operations (copy, move, retry) I'm going to create separate stories so we can tackle them in the next sprint(s).

[...]

Issue #9182, opened 2023-06-07
[files] Batch actions need some kind of working in progress-feedback (loading spinner)


hurradieweltgehtunter:
1. Create folder
2. Select multiple files
3. Move the files via drag&drop to folder

UI seems to do nothing, Dev Console shows sequential XHR MOVE requests.
-> Provide some sort of feedback to the user, that the moving process in progress.

Same goes for
- Batchaction pasting copied files
- Batchaction deleting files (Modal stays unresponsive)

mbuonaccorsi:
The entire implementation of copy/move seems inefficient.  Instead of individual requests from the browser to the server for each folder, which if the folders/files are HUGE are likely to take forever and time out, send a compact request to initiate the process and indicate what needs to be done, then have the UI poll the status for this process and provide a progress indicator with a percentage.   The entire process should run server side and not need to involve the browser once initiated.

This would prevent the frequent freezes in the UI and loss of the entire copy\move operation if you were to close the tab or navigate to another folder in the UI.

kulmann:
> The entire implementation of copy/move seems inefficient. Instead of individual requests from the browser to the server for each folder, which if the folders/files are HUGE are likely to take forever and time out, send a compact request to initiate the process and indicate what needs to be done, then have the UI poll the status for this process and provide a progress indicator with a percentage. The entire process should run server side and not need to involve the browser once initiated.
> 
> This would prevent the frequent freezes in the UI and loss of the entire copy\move operation if you were to close the tab or navigate to another folder in the UI.

Under the hood it's still WebDAV - the requests are server side, but we need to wait for the responses. Of course we can imagine different approaches, but then there's client compatibility - e.g. Cyberduck recently added support for ownCloud Infinite Scale, which would not have been possible if we came up with a custom API. And we're proud of writing open source software and supporting open standards. That being said, contributions are always welcome! 😉

kulmann:
> This would prevent the frequent freezes in the UI and loss of the entire copy\move operation if you were to close the tab or navigate to another folder in the UI.

Oh by the way, navigating to another folder in the UI should not be a problem. Could you try that for smaller files than you are currently trying?

mbuonaccorsi:
> > This would prevent the frequent freezes in the UI and loss of the entire copy\move operation if you were to close the tab or navigate to another folder in the UI.
> 
> Oh by the way, navigating to another folder in the UI should not be a problem. Could you try that for smaller files than you are currently trying?

[...]

Issue #7039, opened 2022-05-23
Async interactions - fix "hanging" UI during long running actions


kulmann:
Async tasks currently have the tendency of blocking the UI. The also get hibernated by the browser when the tab becomes a background tab.

Needs e.g. web workers / separate threads for:
- copy
- move
- delete (especially in batch)
- upload

kulmann:
Related: we want progress indicators for those interactions, see https://github.com/owncloud/web/issues/6183

kulmann:
For whoever works on this: have a look at https://github.com/GoogleChrome/workbox first

ScharfViktor:
When a user deletes a large number of files, the web looks like it is hanging.

- dialog "delete files" hangs
- web is disabled while deleting process
- If we click "cancel" button -> deleting process will contunue until all files are deleted

lookacat:
Related: https://github.com/owncloud/web/pull/7902

kulmann:
Related / a very good reason for Web Workers: https://github.com/owncloud/ocis/issues/8427#issuecomment-1939668580

tbsbdr:
Needs research first

JammingBen:
Closing in favour of these stories:

* https://github.com/owncloud/web/issues/10951
* https://github.com/owncloud/web/issues/10952
* https://github.com/owncloud/web/issues/10953

PR #10916, opened 2024-05-16
feat(poc): implement poc for web workers

JammingBen:
## Description
This is a proof of concept web worker implementation. It implements the basic architecture (or how it could look like) and demonstrates the functionality by handling file deletion in a web worker.

I was facing mainly 2 challenges here. Firstly, a web worker lives completely isolated from the main thread, hence it has no knowledge about things like the DOM, browser storage etc. All things a worker might need must be serialized and passed into the worker. This also means that only serializable data can be passed, no functions e.g. ~This hurts especially since it means we can't use the factories from `web-client` to build requests.~ We can still use the `web-client` here, just not the `ClientService`.

The second challenge I faced was that workers need to be informed about potential token updates during an ongoing worker process. This means, if a token is being updated, all potential existing web workers need to be updated with the new token.

## Architecture

The implementation consists of 3 main parts. First we have the `WebWorkersStore` that acts like a registry. It can create and terminate web workers and update access tokens for all active workers. The idea behind this store is to have some instance that always knows which and how many workers are currently active.

Then there are the actual workers. These are isolated `ts` files that always have a `self.onmessage` method to receive data from the main thread.

At last there are worker composables that are responsible for creating instances of web workers and registering them in the `WebWorkersStore`. They also handle serialization and termination after a process has finished. These composables act like an interface between the store, the actual worker and the "outside world", where they are supposed to be used. They need to be given all relevant data and a potential callback.

## Related Issue
- refs https://github.com/owncloud/web/issues/10855
- refs https://github.com/owncloud/web/issues/7039
- refs https://github.com/owncloud/web/issues/9182

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


JammingBen:
This is ready for a general review. I added a 2 second timer to the delete action to better visualize the process (hence this is still a draft). Once we decide if/how to move forward here I'll remove it of course.

packages/web-pkg/src/composables/piniaStores/webWorkers.ts
```diff
@@ -0,0 +1,52 @@
+import { defineStore } from 'pinia'
+import { Ref, ref, unref } from 'vue'
+import { v4 as uuidV4 } from 'uuid'
+import { UseWebWorkerReturn, useWebWorker } from '@vueuse/core'
+
+export interface WebWorker<T = any> extends UseWebWorkerReturn<T> {
+  id: string
+}
+
+export type WorkerTopics = 'startProcess' | 'tokenUpdate'
```
kulmann, line 10:
```suggestion
export type WorkerTopic = 'startProcess' | 'tokenUpdate'
```

You're not defining an enum but a type, so the var name should be singular (I know, ultra nitpick, sorry).

packages/web-pkg/src/composables/piniaStores/webWorkers.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const useWebWorkersStore = defineStore('webWorkers', () => {
+  const workers = ref([]) as Ref<WebWorker[]>
+
+  const createWorker = <T = any>(workerUrl: string): WebWorker<T> => {
+    const result = useWebWorker(workerUrl, { type: 'module', name: uuidV4() })
+    const worker = { id: uuidV4(), ...result }
```
kulmann, line 17: Why don't you re-use the `name` as the id? Any benefit in having two different uuids for the `id` and `name` props? 🤔

packages/web-pkg/src/composables/webWorkers/deleteWorker/worker.ts
```diff
@@ -0,0 +1,76 @@
...
+  const { baseUrl, headers, resources } = data
+
+  const successful: Resource[] = []
+  const failed: { resource: Resource; status: number }[] = []
+  const queue = new PQueue({ concurrency: 2 })
+
+  const doDelete = async (path: string) => {
+    // fake timer so requests take more time
+    await new Promise((res) => setTimeout(res, 2000))
+    return fetch(`${baseUrl}/${encodePath(path)}`, {
```
kulmann, line 45: Using our `web-client` here is a problem, right? Could you remind me why? I don't remember... but using raw `fetch` seems like a huge step back in DX. :-(

packages/web-pkg/src/composables/piniaStores/webWorkers.ts
JammingBen: I just didn't see it because I added the name later I guess, good catch!

packages/web-pkg/src/composables/webWorkers/deleteWorker/worker.ts
JammingBen: Using the `ClientService` doesn't work because it needs to be initiated with so much stuff. Using the `web-client` could work though... I didn't think of that. It probably comes down to the question if we can use axios here. I'll give it a try 👍

kulmann: That would be nice. Thank you for checking it! 🤩

JammingBen: It actually works 🙌  Feels a lot better this way!

kulmann: 😍

kulmann (approved):
I like this big time! 😍

Issue #10917, opened 2024-05-16
Trash bin is empty after deleting folder which name is int


ScharfViktor:
`Version | 5.1.0-prealpha+0ac03d8a8`
`web 9.0.0-alpha.5`

Steps: 
- create folders `123` - use only digits in the name - this is important to reproduce the bug
- delete folder `123`
- go to the trash bin

Expected: folder `123` is in the trash bin
Actual: trash bin is empty.` PROPFIND https://host.docker.internal:9200/remote.php/dav/spaces/trash-bin/personalSpaceUUID` contains the folder `123`

```
web-app-files-XHrTtWcM.mjs:1 TypeError: Path must be a string. Received 123
    at No (vendor-D6wJIj-w.mjs:26:177639)
    at Object.basename (vendor-D6wJIj-w.mjs:26:180163)
    at $p (TextEditor.vue_vue_type_style_index_0_lang-DJRdX2Ka.mjs:1:7478)
    at Array.map (<anonymous>)
    at Object.listFiles (index.html-BQAoY_iH.mjs:1:17053)
```

https://github.com/owncloud/web/assets/84779829/6eb7ebde-0594-491b-a734-5be48bbd9000

PR #10920, opened 2024-05-16
Fix bug where resource with name consists of only numbers won't showup in the trash bin

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/10917

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


dschmidt (commented):
Why is the value parsed to a number in the first place?! 

This seems more like fighting the symptom rather than the root cause

AlexAndBear:
> Why is the value parsed to a number in the first place?!
> 
> This seems more like fighting the symptom rather than the root cause

We do this string parsing in L79 as well, the root issue might go down to the XML builder (webdav) that seems to automatically parse the data to the most matching data type as we don't hand it a scheme. 
We might look into that later but I think that would go too far to fix a p2 bug, while we already to the same kind of parsing in the code already

dschmidt (approved):
Clarified: I still think it's suspicious, that we get a number here - but web-client seems to be the right level to handle this and we handle it analogously for other props.

LGTM

JammingBen (approved):

Issue #10933, opened 2024-05-21
Deleting a new folder inside a shared folder doesn't work


JammingBen:
## Describe the bug
Deleting a newly created folder inside a shared folder doesn't work.

## Steps to reproduce
1. Navigate into an incoming share with write permissions
2. Create a folder
3. Without any reload, delete the folder

## Expected behavior
The folder can be deleted.

## Actual behavior
The folder can't be deleted, an error ocourrs.

PR #10934, opened 2024-05-21
fix: deleting newly created folders/pasted resources inside shares

JammingBen:
## Description
Fixes deleting newly created folders and pasted resources inside shares.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10933

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFolder.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFolder.ts
@@ -1,4 +1,4 @@
-import { SpaceResource } from '@ownclouders/web-client'
+import { SpaceResource, isShareSpaceResource } from '@ownclouders/web-client'
 import { computed, nextTick, unref } from 'vue'
 import { useClientService } from '../../clientService'
 import { useRouter } from '../../router'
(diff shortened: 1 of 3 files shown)


packages/web-pkg/src/composables/actions/files/useFileActionsCreateNewFolder.ts
```diff
@@ -76,6 +76,10 @@ export const useFileActionsCreateNewFolder = ({ space }: { space?: SpaceResource
         })
       }
 
+      if (isShareSpaceResource(space)) {
+        resource.remoteItemId = space.id
```
AlexAndBear, line 80: we can't do this, when we build the resource ?

JammingBen: Because we don't have the space in that context. It would be great to add it in the future, but that needs more refactoring.

AlexAndBear: Fine for now, could you add a fixme or todo comment anyways ?

AlexAndBear (approved):

Issue #10474, opened 2024-02-10
Access token expiry causes failed uploads on OCIS test server


prohtex:
I've been scratching my head over token expiry errors that cause failed uploads on OCIS documented on owncloud/ocis#8368 and owncloud/ocis#8372. Today I reproduced these errors on OCIS continuous deployment test server.

I obviously don't have access to those logs, but screenshots are below.

Steps to reproduce:

1. Log into OCIS test server: https://ocis.ocis-wopi.latest.owncloud.works admin/admin
2. Create a new space
3. Upload a directory containing large flies to the new space
4. Upload fails with unknown error. Observe console errors caused by token expiry

![Screenshot 2024-02-10 at 2 42 39 PM](https://github.com/owncloud/ocis/assets/12946938/af5907b9-a581-45bd-91cd-d9e4a9153f02)

tbsbdr:
All I can say is that uploads must continue in "background" tabs; assessing as p2 as uploading is essential for us. as mentioned by you, @kulmann solving this issue via web worker might be heavy lifting, so a effort estimation / reasearch beforehand should happen before we go in.

prohtex:
> **Could you please try the following?**
> 
> 1. Start a long running upload (e.g. like in your reproducer)
> 2. Open the network tab in the dev tools, so that network requests get recorded
> 3. Grab a coffee or click around in the ownCloud browser tab. Make sure to not put the browser or the tab into hibernation.

Hi @kulmann, I am glad the issue was found! I can confirm that taking further actions keeps the tab active and the upload going. This also explains why the point at which the upload failed was rather inconsistent (as I was taking various actions with the browser, and not always keeping it in the background). I can also confirm that setting IDP_ID_TOKEN_EXPIRATION mitigates the issue; I had mistakenly tried to set the value in the .env file previously and it was not taking effect.

Our use case is for photo retouching, and so we are frequently sending massive files over the web client. A big part of why we wanted to migrate to OCIS was to improve reliability sending large files this way; we often encounter timeouts and broken uploads with OC10. I am assuming sites like DropBox and Wetransfer use Web Workers for their uploads.

For the meantime, we will likely deploy more than one instance of OCIS so we can continue to test, and to keep one with a very long token timeout for these purposes. It doesn't appear possible to get the WOPI deployment working without Docker so we will continue to use the Ubuntu VM. I can always test anything in a fresh Ubuntu VM or on macOS.

[...]

PR #10955, opened 2024-05-23
feat: move access token timer to web worker

JammingBen:
## Description
Introduces a web worker for the access token timer to run in. As soon as a token is about to expire, the web worker will inform the main thread that it needs to retrieve a new token from the server.

This ensures that the timer runs es expected even if the browser tab is being throttled or suspended. As a result, long running operation such as uploads or move actions can continue in the background.

## Related Issue

- refs https://github.com/owncloud/web/issues/10855#issuecomment-2126553354
- fixes https://github.com/owncloud/web/issues/10474 (hopefully)

## How has this been tested

- Set `IDP_ACCESS_TOKEN_EXPIRATION` to `120`
- Add a time log to the user loaded event (takes place immediately after each token renewal)
- Leave the browser tab open in background, do other things
- Check the tab after some time period (~ 15-20 min)
- The time difference between the time logs should be a consistent 110 + 2-3 seconds: 120 seconds minus 10 seconds (because we're triggering the renewal 10 seconds before expiry) plus 2-3 seconds for the token renewal to take place (this varies a bit because of server & network)

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


kulmann (approved):
Code looks good an my localhost tests seemed to work. Let's try this 👍

prohtex:
Amazing. Can't wait to try this out and set my tokens back to something reasonable...! Thank you!

PR #10956, opened 2024-05-23
Exif and audio panels

kulmann:
## Description
This PR introduces an `EXIF` panel, which displays image, photo and location metadata if that's available on the given resource as well as an `Audio Info` panel, which displays audio metadata if that's available on the given resource.

## Related Issue
- Relates to https://github.com/owncloud/web/issues/7930

## Motivation and Context
Provide useful additional information.

## Screenshots
Screenshot 2024-05-23 at 16 16 12

Screenshot 2024-05-24 at 14 57 39

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:
- [x] Changelog item
- [x] Unit tests
- [x] Blocked by https://github.com/cs3org/reva/pull/4695

### DIFF
--- a/changelog/unreleased/enhancement-exif-panel
+++ b/changelog/unreleased/enhancement-exif-panel
@@ -1,6 +1,6 @@
 Enhancement: EXIF metadata panel
 
-We've added a right sidebar panel with the name `Image Info`, which display `image`, `photo` and `location`
+We've added a right sidebar panel with the name `Image Info`, which displays `image`, `photo` and `location`
 metadata if present on the given resource. It is registered by the files app and becomes visible in the global right sidebar
 in all file contexts, e.g. file listing with single select on an image or the preview app displaying an image.
 
--- a/changelog/unreleased/enhancement-exif-panel
+++ b/changelog/unreleased/enhancement-exif-panel
@@ -1,0 +1,7 @@
+Enhancement: EXIF metadata panel
+
+We've added a right sidebar panel with the name `Image Info`, which display `image`, `photo` and `location`
+metadata if present on the given resource. It is registered by the files app and becomes visible in the global right sidebar
+in all file contexts, e.g. file listing with single select on an image or the preview app displaying an image.
+
+https://github.com/owncloud/web/pull/10956
--- a/packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
+++ b/packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
@@ -1,0 +1,133 @@
+<template>
+  <div id="files-sidebar-panel-audio">
+    <dl class="audio-data-list">
+      <template v-if="title">
+        <dt v-text="$gettext('Title')" />
+        <dd v-text="title" />
+      </template>
+      <template v-if="duration">
+        <dt v-text="$gettext('Duration')" />
+        <dd v-text="duration" />
+      </template>
+      <template v-if="artist">
+        <dt v-text="$gettext('Authors')" />
+        <dd v-text="artist" />
+      </template>
+      <template v-if="album">
+        <dt v-text="$gettext('Album')" />
+        <dd v-text="album" />
+      </template>
+      <template v-if="genre">
+        <dt v-text="$gettext('Genre')" />
+        <dd v-text="genre" />
+      </template>
+      <template v-if="year">
+        <dt v-text="$gettext('Year recorded')" />
+        <dd v-text="year" />
+      </template>
+      <template v-if="track">
+        <dt v-text="$gettext('Track')" />
+        <dd v-text="track" />
+      </template>
+      <template v-if="disc">
+        <dt v-text="$gettext('Disc')" />
+        <dd v-text="disc" />
+      </template>
+    </dl>
+  </div>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent, inject, Ref, unref } from 'vue'
+import { Resource } from '@ownclouders/web-client'
+import { Duration } from 'luxon'
+
+export default defineComponent({
+  name: 'AudioMetaPanel',
+  setup() {
+    const resource = inject<Ref<Resource>>('resource')
+
+    const album = computed(() => {
+      return unref(resource).audio.album
+    })
+
+    const artist = computed(() => {
+      return unref(resource).audio.artist
+    })
+
+    const albumArtist = computed(() => {
+      return unref(resource).audio.albumArtist
+    })
+
+    const genre = computed(() => {
+      return unref(resource).audio.genre
+    })
+
+    const title = computed(() => {
+      return unref(resource).audio.title
+    })
+
+    const duration = computed(() => {
+      const milliseconds = unref(resource).audio.duration
+      if (milliseconds) {
+        const d = Duration.fromMillis(milliseconds)
+        if (d.hours > 0) {
+          return d.toFormat('hh:mm:ss')
+        }
+        return d.toFormat('mm:ss')
+      }
+      return ''
+    })
+
+    const track = computed(() => {
+      const audio = unref(resource).audio
+      if (audio.track && audio.trackCount) {
+        return `${audio.track} / ${audio.trackCount}`
+      }
+      return audio.track
+    })
+
+    const disc = computed(() => {
+      const audio = unref(resource).audio
+      if (audio.disc && audio.discCount) {
+        return `${audio.disc} / ${audio.discCount}`
+      }
+      return audio.disc
+    })
+
+    const year = computed(() => {
+      return unref(resource).audio?.year
+    })
+
+    return {
+      album,
+      artist,
+      albumArtist,
+      genre,
+      title,
+      duration,
+      track,
+      disc,
+      year
+    }
+  }
+})
+</script>
+
+<style lang="scss">
+.audio-data-list {
+  display: grid;
+  grid-template-columns: auto minmax(0, 1fr);
+  dt,
+  dd {
+    margin-bottom: var(--oc-space-small);
+  }
+  dt {
+    font-weight: bold;
+    white-space: nowrap;
+  }
+  dd {
+    margin-inline-start: var(--oc-space-medium);
+  }
+}
+</style>
--- a/packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
+++ b/packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
@@ -69,14 +69,14 @@ export default defineComponent({
 
     const duration = computed(() => {
       const milliseconds = unref(resource).audio.duration
-      if (milliseconds) {
-        const d = Duration.fromMillis(milliseconds)
-        if (d.hours > 0) {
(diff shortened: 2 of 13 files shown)


AlexAndBear:
Would love to change "Musical genre" to "Genre"

kulmann:
> Would love to change "Musical genre" to "Genre". Sounds a little too lofty.

I was getting inspiration from the field labels in mac finder file info. :-) But also fine with `Genre`, changed it 👍

packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <div id="files-sidebar-panel-audio">
+    <dl class="audio-data-list">
+      <template v-if="title">
+        <dt v-text="$gettext('Title')" />
+        <dd v-text="title" />
+      </template>
+      <template v-if="duration">
```
AlexAndBear, line 8: I like the idea to only show the available data. But I think while showing the exif/meta data, it makes more sense to actually show the existing and non existing data e.G with a dash(-). 

Could you check how the mac is handling missing data ?

changelog/unreleased/enhancement-exif-panel
```diff
@@ -0,0 +1,7 @@
+Enhancement: EXIF metadata panel
+
+We've added a right sidebar panel with the name `Image Info`, which display `image`, `photo` and `location`
```
AlexAndBear, line 3:
```suggestion
We've added a right sidebar panel with the name `Image Info`, which displays `image`, `photo` and `location`
```

AlexAndBear (changes requested):

packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
```diff
@@ -0,0 +1,133 @@
...
+      return unref(resource).audio.genre
+    })
+
+    const title = computed(() => {
+      return unref(resource).audio.title
+    })
+
+    const duration = computed(() => {
+      const milliseconds = unref(resource).audio.duration
+      if (milliseconds) {
```
AlexAndBear, line 72: I am more a fan of early returns, maybe we can swap the condition and return '' early

packages/web-app-files/src/components/SideBar/Exif/ExifPanel.vue
```diff
@@ -0,0 +1,189 @@
...
+
+    const fNumber = computed(() => {
+      const photo = unref(resource).photo
+      return photo?.fNumber ? `f/${photo.fNumber}` : ''
+    })
+
+    const exposureTime = computed(() => {
+      const photo = unref(resource).photo
+      return photo?.exposureDenominator
+        ? `${photo.exposureNumerator}/${photo.exposureDenominator}`
```
AlexAndBear, line 115: are we sure exposeNumerator is always set otherwise we might get an error

AlexAndBear (changes requested):

packages/web-pkg/src/composables/piniaStores/config/config.ts
```diff
@@ -43,6 +43,9 @@ const defaultOptions = {
   sidebar: {
     shares: {
       showAllOnLoad: false
+    },
+    exif: {
+      showLocation: true
```
dschmidt, line 48: hum - maybe now I understand why you were reluctant to add the config setting. I had thought of a setting to a specific app. Now it's weird that the setting is in web-pkg and the panel in web-app-files. But having settings for files app would be 💩  as well... yeah well 🙈

kulmann: Yeah, now I built it so let's keep it. ;-)

packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
kulmann: Mac Finder is only showing the fields that were found. That's why I built it as well. But no hard feelings, I'll make a screenshot with `-` for missing fields and then we can discuss.

kulmann: Screenshot 2024-05-29 at 10 30 40
Screenshot 2024-05-29 at 10 32 21

AlexAndBear: I kind a like it that way

dschmidt: I think it looks a bit weird when you have hardly any data in there - but no strong opinion either. I'm happy either way that this is finally happening after I started metadata extraction it last year :)

JammingBen: No strong opinion either from my side. I'd probably hide the fields without data, but I can still see the point in having them 🤷

packages/web-app-files/src/components/SideBar/Exif/ExifPanel.vue
```diff
@@ -67,6 +69,7 @@ export default defineComponent({
         return '-'
       }
       if ([5, 6, 7, 8].includes(unref(resource).photo?.orientation)) {
+        // these orientations indicate portrait mode. tika normalizes width and height according to orientation.
```
dschmidt, line 72: What a weird thing to do, should we maybe handle this on server side instead?

packages/web-app-files/tests/unit/components/SideBar/Exif/ExifPanel.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+          fNumber: 5.6,
+          exposureNumerator: 1,
+          exposureDenominator: 20,
+          iso: 320,
+          orientation: 1,
+          takenDateTime: '2017-02-11T14:54:50Z'
+        })
+      })
+      const { wrapper } = createWrapper({ resource })
+      expect(wrapper.find(`[data-testid="exif-panel-${key}"]`).text()).toMatchSnapshot()
```
dschmidt, line 34: Isn't it a bit weird to have a snapshot for a single value?

kulmann: Hm... actually, what's your take on that? I made it a snapshot test because most of the values get some string modification for being displayed. E.g. if the `fNumber` value is `16.6` it will be rendered as `16.6 mm`. I didn't want to rebuild the string modification logic in the tests again, so that's why I decided for snapshot tests. And because I hate snapshot tests I made the scope as small as possible (only the value, no surrounding html).

dschmidt: Ah, fair enough. 👍🏻

packages/web-app-files/src/components/SideBar/Exif/ExifPanel.vue
kulmann: please not for this PR 🙈 maybe as a followup 😅

dschmidt: sure sure...

packages/web-app-files/src/components/SideBar/Audio/AudioMetaPanel.vue
```diff
@@ -0,0 +1,117 @@
...
+    const disc = computed(() => {
+      const audio = unref(resource).audio
+      if (audio.disc && audio.discCount) {
+        return `${audio.disc} / ${audio.discCount}`
+      }
+      return audio.disc || '-'
+    })
+
+    const year = computed(() => {
+      return unref(resource).audio?.year || '-'
```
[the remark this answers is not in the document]
kulmann: Good spot, thank you, that's a copy-paste-leftover. The exif panel shows data from `image`, `photo` and `location`, but we don't know which object exists. Could be only the `image`, only the `photo`, all three, whatever. That's why I needed optional chaining there. For audio we only show the panel at all if `audio` is present on the resource, so optional chaining is not needed.

packages/web-client/src/helpers/resource/functions.ts
```diff
@@ -66,6 +67,17 @@ export const isShareRoot = (resource: Resource) => {
   return typeof resource.isShareRoot === 'function' && resource.isShareRoot()
 }
 
+const convertObjectToCamelCaseKeys = (data: Record<string, any>) => {
+  if (!data) {
+    return data
+  }
+  const converted = {}
```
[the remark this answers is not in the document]
kulmann: I don't have a better type than `Record<string, any>` here 🙈

packages/web-app-files/src/components/SideBar/Exif/ExifPanel.vue
```diff
@@ -0,0 +1,181 @@
+<template>
+  <div id="files-sidebar-panel-exif">
+    <dl class="exif-data-list">
+      <dt v-text="$gettext('Dimensions')" />
+      <dd data-testid="exif-panel-dimensions" v-text="dimensions" />
+      <dt v-text="$gettext('Device make')" />
+      <dd data-testid="exif-panel-cameraMake" v-text="cameraMake" />
```
AlexAndBear, line 7: Why do we have the mix aus Kebap and camelCase here? 😅

kulmann: Because it's only data-testids? 🙈 Normally we use kebab case in html. I'm using them in a generic way and just inline the object keys. The object keys are camelCase. If you're super unhappy with that I can transform the object keys to kebab case first. 🤷

packages/web-client/src/helpers/resource/functions.ts
```diff
@@ -71,7 +71,7 @@ const convertObjectToCamelCaseKeys = (data: Record<string, any>) => {
   if (!data) {
     return data
   }
-  const converted = {}
+  const converted: Record<string, any> = {}
```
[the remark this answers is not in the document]
dschmidt: then again... it's not exported, so maybe not worth the effort... 🤷🏻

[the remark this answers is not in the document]
kulmann: I also thought about a `T` that's either `Audio`, `Image`, Photo` or `GeoLocation`, but you can't initialize that with an empty object `{}` in this line. So I thought exactly `not worth the effort` and made it `Record<string, any>`. Also not a fan of the `any`....

packages/web-app-files/src/components/SideBar/Exif/ExifPanel.vue
AlexAndBear: fine for me, is wa just confused

AlexAndBear (approved):
🚀 Wonderful, hope to do some of the calculations in the backend later so its easier for other clients to implement

dschmidt:
🎉 🎉 🎉 

I love it so so much! Thanks!

Now ... multiple root panels? ;-)

Issue #10458, opened 2024-02-09
[QA] run e2e tests with external idp (keycloak) in CI


saw-jan:
### Description
e2e tests can run with keycloak as an idp (Done in https://github.com/owncloud/web/issues/9904). So now, we want to have e2e tests running with keycloak setup in CI.

### Motivation
Previously, we found bugs related to ocis with external idp like this one https://github.com/owncloud/ocis/pull/8074. Issues related to ocis with external idp are not possible to find out as early as possible. And only can be found when playing around manually.

Since we have e2e tests that can run with keycloak idp, bugs like that can be found early if we have automated tests running in CI.

### Open Tasks
- [ ] Filter out e2e test suites to run with keycloak (We may want to run only the few selected scenarios)
- [ ] Bring e2e-with-keycloak to Drone CI

CC @ScharfViktor

amrita-shrestha:
@PrajwolAmatya and @saw-jan i need some basic info how do we run test before

PrajwolAmatya:
I used to run as follow:
```md
### on oCIS repo
goto: deployments/examples/ocis_keycloak
run: docker compose up

### first login to
https://keycloak.owncloud.test

### then login to
https://ocis.owncloud.test
run e2e test:
BASE_URL_OCIS=ocis.owncloud.test KEYCLOAK=ture pnpm run test:e2e:cucumber tests/e2e/cucumber/features/smoke/copyMove.feature:6
```

saw-jan:
> @PrajwolAmatya and @saw-jan i need some basic info how do we run test and keycloak before

currently, we do not have any docs related to that. but here are the available envs:

https://github.com/owncloud/web/blob/71cbae47a518c0d0b2bf446fce7fb6bc3ee4813c/tests/e2e/config.js#L9-L14

amrita-shrestha:
> > @PrajwolAmatya and @saw-jan i need some basic info how do we run test and keycloak before
> 
> currently, we do not have any docs related to that. but here are the available envs:
> 
> https://github.com/owncloud/web/blob/71cbae47a518c0d0b2bf446fce7fb6bc3ee4813c/tests/e2e/config.js#L9-L14

i already seen had i just want to know if we already had setup docs for testing it against local setup but nevermind i will look over it

amrita-shrestha:
Problems to look 
- if scenario fails then browser doesn't close at after scenario (running tests on keycloak )

amrita-shrestha:
Todo
[...]

PR #10961, opened 2024-05-24
[tests-only][full-ci]Run more e2e tests on keycloak

amrita-shrestha:
## Description
- Keycloak supports assigning multiple realm roles to a user but oCIS supports only one role for a user. 
If multiple roles are assigned to a user in Keycloak, oCIS map the highest priority role among Keycloak assigned roles.
So, before assigning role we have to unassign previous role using keycloak api. This part has been implemented in this PR.
 - added more tests to run on CI admin-settings/users.feature:20 related to quota change

## Related Issue

- https://github.com/owncloud/web/issues/10458

## How Has This Been Tested?

- LOCALLY

## Types of changes

- [ ] Technical debt
- [x] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/e2e/cucumber/steps/api.ts
+++ b/tests/e2e/cucumber/steps/api.ts
@@ -22,6 +22,15 @@ Given(
     const admin = this.usersEnvironment.getUser({ key: stepUser })
     for await (const info of stepTable.hashes()) {
       const user = this.usersEnvironment.getUser({ key: info.id })
+      /**
+         The oCIS API request for assigning roles allows only one role per user,
+         whereas the Keycloak API request can assign multiple roles to a user.
+         If multiple roles are assigned to a user in Keycloak,
+         oCIS map the highest priority role among Keycloak assigned roles.
+         Therefore, we need to unassign the previous role before
+         assigning a new one when using the Keycloak API.
+      */
+      await api.provision.unAssignRole({ admin, user })
       await api.provision.assignRole({ admin, user, role: info.role })
     }
   }
--- a/tests/e2e/support/api/keycloak/user.ts
+++ b/tests/e2e/support/api/keycloak/user.ts
@@ -46,11 +46,12 @@ export const createUser = async ({ user, admin }: { user: User; admin: User }): 
   const uuid = getUserIdFromResponse(creationRes)
 
   // assign realmRoles to user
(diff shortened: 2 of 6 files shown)


tests/e2e/cucumber/steps/api.ts
```diff
@@ -22,6 +22,7 @@ Given(
     const admin = this.usersEnvironment.getUser({ key: stepUser })
     for await (const info of stepTable.hashes()) {
       const user = this.usersEnvironment.getUser({ key: info.id })
+      await api.provision.unAssignRole({ admin, user })
```
saw-jan, line 25: Please, add a comment here explaining why we need to unassign role and for which setup

tests/e2e/support/api/keycloak/user.ts
```diff
@@ -67,7 +68,8 @@ export const assignRole = async ({
   uuid: string
   role: string
 }) => {
-  return request({
+  // can assign multiple realm role at once
+  return await request({
```
saw-jan, line 72: I don't see reason for using `await` but not a big deal
```suggestion
  return request({
```

tests/e2e/support/api/keycloak/user.ts
```diff
@@ -78,6 +80,26 @@ export const assignRole = async ({
     header: { 'Content-Type': 'application/json' }
   })
 }
+export const unAssignRole = async ({
```
saw-jan, line 83:
```suggestion
}

export const unAssignRole = async ({
```

tests/e2e/support/api/provision/user.ts
```diff
@@ -45,3 +46,19 @@ export const assignRole = async ({
...
+
+/**
+   The oCIS API request for assigning roles allows only one role per user,
+   whereas the Keycloak API request can assign multiple roles to a user.
+   If multiple roles are assigned to a user in Keycloak,
+   oCIS map the highest priority role among Keycloak assigned roles.
+   Therefore, we need to unassign the previous role before
+   assigning a new one when using the Keycloak API.
+*/
+export const unAssignRole = async ({ admin, user }: { admin: User; user: User }): Promise<void> => {
```
saw-jan, line 58: I would recommend to put the comment where it's used not where it is defined

saw-jan (approved):
LGTM

PR #10974, opened 2024-05-28
fix: remove user locally if no logout url in IdP

kulmann:
## Description
There are IdPs without an `endSessionEndpoint` (e.g. Authelia). In those cases we just need to unload the currently authenticated user.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/6631

## Motivation and Context
Hardening.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-local-logout
+++ b/changelog/unreleased/bugfix-local-logout
@@ -1,0 +1,8 @@
+Bugfix: Local logout if IdP has no logout support
+
+Some IdPs don't support a logout endpoint. In those cases the web UI ran into a fatal error an showed an empty screen without
+further redirects. Fixed by forgetting the currently authenticated user when the OpenID Connect configuration doesn't contain
+an `endSessionEndpoint` url.
+
+https://github.com/owncloud/web/pull/10974
+https://github.com/owncloud/enterprise/issues/6631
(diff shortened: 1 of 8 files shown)


kulmann:
Currently thinking about how to test this properly. I'm looking into setting up authelia (because that doesn't have an `endSessionEndpoint`) and testing it with that... anyone different ideas?

packages/web-runtime/src/pages/accessDenied.vue
```diff
@@ -82,7 +82,6 @@ export default defineComponent({
         }
       }
       return {
-        name: 'login',
```
kulmann, line 85: No idea why we had the `name` key here... maybe just by mistake and it was really meant for the `to` route?!

changelog/unreleased/bugfix-local-logout
```diff
@@ -0,0 +1,8 @@
+Bugfix: Local logout if IdP has no logout support
+
+Some IdPs don't support a logout endpoint. In those cases the web UI ran into a fatal error an showed an empty screen without
```
JammingBen, line 3:
```suggestion
Some IdPs don't support a logout endpoint. In those cases the web UI ran into a fatal error and showed an empty screen without
```

JammingBen (approved):
Can't test it obviously, but code LGTM (aside from one spelling mistake :trollface: )

changelog/unreleased/bugfix-local-logout
kulmann: Really need to clean my keyboard 😆 this has been a frequent issue lately :trollface:

Issue #10954, opened 2024-05-23
[web] Warn user about pending requests when attempting to close the browser(-tab)


kulmann:
# Description
When a user attempts to close the browser(-tab) we should warn them about pending requests (e.g. from batch operations) that would be aborted by closing.

## User Stories

* > As a user who started a batch operation for deleting a large number of files, I'd be annoyed if closing the browser aborted my batch operation, because I expect the process to finish. 

## Value
Educate users about batch operations being orchestrated by the web ui. -> less bug reports.

## Acceptance Criteria
- if the `loadingService` has any unfinished process and the user attempts to close the browser(-tab), show a warning that closing the browser(-tab) would abort pending requests.
- Closing the browser while an upload is running already has such a message. There should not be two after this story is done (i.e. a: have a new concept and also use it for uploads or b: make an exception for uploads so that we continue to use the already existing mechanism there, slightly preferring a, but might not be feasible. needs to be checked.)
- Closing the browser while a document editing session is in a dirty state (i.e. file not saved) also already has such a message. see above.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

## Additional context
This was pointed out by @mbuonaccorsi in https://github.com/owncloud/web/issues/9182#issuecomment-1832807018 - thanks for the input.

AlexAndBear:
Maybe some kind of axios http interceptor possible

JammingBen:
I guess this is what we're looking for: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event. Then just check for `loadingService.isLoading()` and show the user some prompt.

kulmann:
> I guess this is what we're looking for: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event. Then just check for `loadingService.isLoading()` and show the user some prompt.

We're already using exactly that in the AppWrapper and for active upload queues :-)

PR #10986, opened 2024-06-03
Show leave site confirmation dialog when loadingService has active tasks

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/10954

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/services/loadingService.ts
+++ b/packages/web-pkg/src/services/loadingService.ts
@@ -67,6 +67,12 @@ export class LoadingService {
       active: false,
       ...(!indeterminate && { state: { total: 0, current: 0 } })
     }
+
+    // If no tasks are in progress, attach an event listener for 'beforeunload'.
+    if (!this.tasks.length) {
+      window.addEventListener('beforeunload', this.onBeforeUnload)
+    }
+
     this.tasks.push(task)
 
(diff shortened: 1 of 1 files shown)


dschmidt (approved):
little weird indention in the comment, but otherwise LGTM. Good stuff 👍🏻 

It's a bit unfortunate we can't show the user what exactly is going on and why the popup is showing...

packages/web-pkg/src/services/loadingService.ts
```diff
@@ -67,6 +67,16 @@ export class LoadingService {
       active: false,
       ...(!indeterminate && { state: { total: 0, current: 0 } })
     }
+
+    /**
+     * Add a task only if no other tasks are in progress.
```
kulmann, line 72: What does this line of the comment-block do? can't you remove it? The other two lines read like that's already everything I need to know...

AlexAndBear: I asked chad gpt to make it a easy to read as possibly :D

PR #10987, opened 2024-06-03
feat: web worker for copy/move operations

JammingBen:
## Description
Adds a web worker for pasting resources that were copied/cut before.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10951

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsPaste.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsPaste.ts
@@ -75,6 +75,10 @@ export const useFileActionsPaste = () => {
     startWorker(transferData, async ({ successful, failed }) => {
       resourceTransfer.showResultMessage(failed, successful, unref(transferType))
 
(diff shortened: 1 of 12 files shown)


packages/web-pkg/src/composables/actions/files/useFileActionsPaste.ts
```diff
@@ -46,28 +54,37 @@ export const useFileActionsPaste = () => {
...
+      return
     }
-    return movedResourcesPromise.then((movedResources) => {
-      const loadingResources = []
+
+    const originalCurrentFolderId = unref(currentFolder)?.id
+
+    startWorker(transferData, async ({ successful, failed }) => {
+      resourceTransfer.showResultMessage(failed, successful, unref(transferType))
+
```
kulmann, line 77: Wouldn't it make sense to have an early return here after the result message if the length of `successful` is `0`?

Issue #10808, opened 2024-04-19
[web-client] Make graph abstraction layer useful


JammingBen:
The `web-client` package follows the paradigm of providing an abstraction layer between the server and a (web-) application that converts API data into objects with helpful types and utilities. This ensures that users of the server APIs don't need in-depth knowledge about them.

However, the Graph layer is currently an exception to this as it doesn't process any data returned by the server. We want to change that so it aligns with the other layers.

### Acceptance criteria

- [ ] Split up the `index.ts` file into separate factories (similar to the WebDAV implementation)
- [ ] Drives are being returned as `SpaceResource`
- [ ] Drive items and their permissions are being returned as `OutgoingShareResource` or `IncomingShareResource`
- [ ] Permissions are being returned as `Share`
- [ ] Think about a convenient way to provide options such as filters

PR #10997, opened 2024-06-04
refactor: useful graph abstraction for users

JammingBen:
## Description
Refactors the graph abstraction layer for users to make it more practical and convenient to use.

The new implementation has improved method parameters, types and destructures API responses so users of the abstraction don't need to care about such things.

This is the first PR of a whole series, groups will be next. I currently work on it when I have some spare time, although it would be nice to have this completed for the `9.0.0` release (especially now that we decided to couple web-client releases to Web releases).

## Related Issue

- refs https://github.com/owncloud/web/issues/10808

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-admin-settings/src/views/Users.vue
+++ b/packages/web-app-admin-settings/src/views/Users.vue
@@ -270,12 +270,12 @@ export default defineComponent({
         .filter(Boolean)
         .join(' and ')
 
-      const usersResponse = yield clientService.graphAuthenticated.users.listUsers(
-        'displayName',
+      const usersResponse = yield clientService.graphAuthenticated.users.listUsers({
+        orderBy: ['displayName'],
         filter,
-        ['appRoleAssignments']
-      )
-      userSettingsStore.setUsers(usersResponse.data.value || [])
+        expand: ['appRoleAssignments']
+      })
+      userSettingsStore.setUsers(usersResponse || [])
     })
 
     const isLoading = computed(() => {
@@ -306,7 +306,7 @@ export default defineComponent({
         return
       }
 
-      const { data } = yield clientService.graphAuthenticated.users.getUser(user.id)
+      const data = yield clientService.graphAuthenticated.users.getUser(user.id)
       unref(additionalUserDataLoadedForUserIds).push(user.id)
 
       Object.assign(user, data)
--- a/packages/web-runtime/src/pages/account.vue
+++ b/packages/web-runtime/src/pages/account.vue
@@ -314,8 +314,7 @@ export default defineComponent({
       }
 
       try {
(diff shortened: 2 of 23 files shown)


packages/web-runtime/src/services/auth/userManager.ts
```diff
@@ -189,22 +189,22 @@ export class UserManager extends OidcUserManager {
 
     const graphClient = this.clientService.graphAuthenticated
     const [graphUser, roles] = await Promise.all([graphClient.users.getMe(), this.fetchRoles()])
-    const role = await this.fetchRole({ graphUser: graphUser.data, roles })
+    const role = await this.fetchRole({ graphUser: graphUser, roles })
```
kulmann, line 192:
```suggestion
    const role = await this.fetchRole({ graphUser, roles })
```

packages/web-runtime/src/pages/account.vue
```diff
@@ -314,7 +314,7 @@ export default defineComponent({
       }
 
       try {
-        const { data } = yield* call(clientService.graphAuthenticated.users.getMe())
+        const data = yield* call(clientService.graphAuthenticated.users.getMe())
```
kulmann, line 317: Could be one line now:

`graphUser.value = yield* call(clientService.graphAuthenticated.users.getMe())`

packages/web-app-admin-settings/src/views/Users.vue
```diff
@@ -270,12 +270,12 @@ export default defineComponent({
         .filter(Boolean)
         .join(' and ')
 
-      const usersResponse = yield clientService.graphAuthenticated.users.listUsers(
-        'displayName',
+      const usersResponse = yield clientService.graphAuthenticated.users.listUsers({
+        orderBy: new Set(['displayName']),
```
kulmann, line 274: all the `new Set(...)` are irritating me. What's the benefit over a simple array? Do we really need to enforce uniqueness?

kulmann (changes requested):

packages/web-app-admin-settings/src/views/Users.vue
JammingBen: This is how the generated code expects it, so it's basically a question where to handle: by our abstraction layer or by the consuming party. So I guess you'd rather expect it in the abstraction layer, right?

kulmann: Yep, I'd be in favour of building the `Set`s in the abstraction layer and only expecting simple arrays as input.

kulmann (approved):
🥳

PR #10998, opened 2024-06-04
fix: viewer light role permission loading

fschade:
## Description
loading role permissions is shadowed by a early return which get triggered if the user does not have a personal space.
In some cases, for example if the logged in user is of type (role), user-light, this is a valid use-case.

This is fixed by the pr, but we have to discuss (and test) possible side effects of the change.

## How Has This Been Tested?
- ci
- manuell testing

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/index.ts
+++ b/packages/web-runtime/src/index.ts
@@ -216,15 +216,13 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> => 
       await spacesStore.loadSpaces({ graphClient: clientService.graphAuthenticated })
       const personalSpace = spacesStore.spaces.find(isPersonalSpaceResource)
 
-      if (!personalSpace) {
-        return
+      if (personalSpace) {
+        spacesStore.updateSpaceField({
+          id: personalSpace.id,
+          field: 'name',
+          value: app.config.globalProperties.$gettext('Personal')
+        })
       }
-
-      spacesStore.updateSpaceField({
-        id: personalSpace.id,
-        field: 'name',
-        value: app.config.globalProperties.$gettext('Personal')
-      })
 
       // load sharing roles from graph API
       const { data } =


AlexAndBear (dismissed):

packages/web-runtime/src/index.ts
```diff
@@ -212,6 +212,13 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> =>
         })
       }
 
+      // load sharing roles from graph API
```
kulmann, line 215: I'd also be fine keeping the previous order and removing the early return. 
```js
if (personalSpace) {
  spacesStore.updateSpaceField... for renaming the space locally to `Personal`
}
```
Is legit, I don't see the benefit of that being an early return.

JammingBen: Agreed! While early returns are nice in certain situations, it can quite easily lead to such issues.

changelog/unreleased/bugfix-viewer-light-permissions
JammingBen, line 1: In general I'd remove the changelog, since this issue has never been part of any release. We're handling things a bit different than oCIS in that regard.

packages/web-runtime/src/index.ts
```diff
@@ -216,10 +216,6 @@ export const bootstrapApp = async (configurationPath: string): Promise<void> =>
       await spacesStore.loadSpaces({ graphClient: clientService.graphAuthenticated })
       const personalSpace = spacesStore.spaces.find(isPersonalSpaceResource)
 
-      if (!personalSpace) {
-        return
-      }
-
```
JammingBen, line 222: The idea was to have the next statement inside the block:

```
if (personalSpace) {
  spacesStore.updateSpaceField(...)
}
```

fschade: got it, updated

JammingBen (approved):
👍

Issue #10990, opened 2024-06-04
Search field in emoji picker can


kulmann:
## Steps to reproduce
1.
2.
3.

## Actual behavior
A clear and concise description of what happened.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

PR #11009, opened 2024-06-07
fix emoji picker search field not focused nor usable

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/10990

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcModal/OcModal.vue
+++ b/packages/design-system/src/components/OcModal/OcModal.vue
@@ -277,7 +277,10 @@ export default defineComponent({
     focusTrapInitial: {
       type: [String, Boolean],
       required: false,
-      default: null
+      default: null,
+      validator: (focusTrapInitial: string | boolean) => {
+        return !(typeof focusTrapInitial === 'boolean' && focusTrapInitial === true)
+      }
     },
     /**
      * Hide the actions at the bottom of the modal
@@ -349,7 +352,7 @@ export default defineComponent({
   computed: {
     initialFocusRef(): FocusTargetOrFalse {
       if (this.focusTrapInitial || this.focusTrapInitial === false) {
-        return this.focusTrapInitial
+        return this.focusTrapInitial as FocusTargetOrFalse
       }
       return () => {
         // FIXME: according to the types it's incorrect to pass this.$refs.ocModalInput
--- a/packages/design-system/src/components/OcModal/OcModal.vue
+++ b/packages/design-system/src/components/OcModal/OcModal.vue
@@ -303,7 +303,7 @@ export default defineComponent({
     const buttonConfirmAppearance = ref('filled')
 
     const tabbableOptions = computed((): FocusTrapTabbableOptions => {
-      // Enable shadowDom support for e.G emoji-picker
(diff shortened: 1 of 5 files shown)


packages/web-pkg/src/composables/actions/spaces/useSpaceActionsSetIcon.ts
```diff
@@ -31,6 +31,7 @@ export const useSpaceActionsSetIcon = () => {
       title: $gettext('Set icon for %{space}', { space: resources[0].name }),
       hideConfirmButton: true,
       customComponent: EmojiPickerModal,
+      focusTrapActive: false,
```
kulmann, line 34: Ooooh wait. The focus trap is required for all modals, otherwise it's an a11y violation. Why is this line required?

AlexAndBear: This is JUST for the emoji picker modal, the emoji picker is not compatible with the shadowRoot implementation of the emoji picker and so the emoji picker handles the focus by itself. What does the focus trap in addition ?

AlexAndBear: Ok understood what the trap should do, somehow the focus trap breaks the emoji picker ...

AlexAndBear: Got it, I don't have a solution to this on hand rn

AlexAndBear: Fixed

packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -302,6 +302,13 @@ export default defineComponent({
     const showSpinner = ref(false)
     const buttonConfirmAppearance = ref('filled')
 
+    const tabbableOptions = computed((): FocusTrapTabbableOptions => {
+      // Enable shadowDom support for e.G emoji-picker
```
dschmidt, line 306:
```suggestion
      // Enable shadow DOM support for e.g. emoji-picker
```

dschmidt (approved):

kulmann (approved):
Thank you 😍

packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -340,8 +351,8 @@ export default defineComponent({
   },
   computed: {
     initialFocusRef(): FocusTargetOrFalse {
-      if (this.focusTrapInitial) {
-        return this.focusTrapInitial
+      if (this.focusTrapInitial || this.focusTrapInitial === false) {
+        return this.focusTrapInitial as FocusTargetOrFalse
```
dschmidt, line 355: Why not use `FocusTargetOrFalse` as type for `focusTrapInitial` directly?

Issue #11011, opened 2024-06-07
Uncaught console error in web


ScharfViktor:
how to find error see here: https://github.com/owncloud/web/pull/10993#issue-2333240779
see all fails here: https://drone.owncloud.com/owncloud/web/44914

<details>
<summary>  ##  TypeError: Cannot read properties of undefined (reading 'storageId') </summary>
```

- `user1` creates `f1/file.txt`
- `user1` shares `f1` with edit role
- `user2` renames `file.txt` 

Result: `user1` gets error `Cannot read properties of undefined (reading 'storageId')`
</details>
<details>
<summary>  ##  "TypeError: Cannot read properties of undefined (reading 'split')" </summary>

run test `tests/e2e/cucumber/features/spaces/participantManagement.feature:3`

</details>
<details>
<summary>  ##  "TypeError: cancel" </summary>

run test `tests/e2e/cucumber/features/shares/link.feature:312`
- `user1` creates `f1/file.txt`
- `user1` create link to `f1` with edit role
- `public` opens link
-  `user1` changes password of link
- `public` refresh page

Result: `public` gets error `Uncaught (in promise) cancel`
</details>
<details>
<summary>  ##  "TypeError: Cannot read properties of null (reading 'getTabs')" </summary>

run test `tests/e2e/cucumber/features/app-provider/officeSuites.feature:17`

AlexAndBear:
I decided to give it a P2 since this will enhance our automated testing and should be quite relevant for rolling releases

JammingBen:
`Uncaught (in promise) cancel` looks like an error you get when canceling a task. There are situations where we do this intentionally, so it might be sufficient to wrap this in a `try { ... } catch (e) { }` block.

AlexAndBear:
> `Uncaught (in promise) cancel` looks like an error you get when canceling a task. There are situations where we do this intentionally, so it might be sufficient to wrap this in a `try { ... } catch (e) { }` block.

Hm I can't find anything that cancel or cancelAll should throw an error in the docs 🤔

AlexAndBear:
@ScharfViktor 
"TypeError: Cannot read properties of null (reading 'getTabs')"
run test tests/e2e/cucumber/features/app-provider/officeSuites.feature:17"

that is an issue in collabora, unfortantly we can't fix this. I would propose to not fail on external app errors maybe ?

ScharfViktor:
[...]

PR #11024, opened 2024-06-11
[check-console-errors] Console error e2e fixes

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/11011

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/container/sse/helpers.ts
+++ b/packages/web-runtime/src/container/sse/helpers.ts
@@ -21,7 +21,7 @@ export const isItemInCurrentFolder = ({
   parentFolderId: string
 }) => {
   const currentFolder = resourcesStore.currentFolder
-  if (!currentFolder) {
+  if (!currentFolder || !currentFolder.id) {
     return false
   }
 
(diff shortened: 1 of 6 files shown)


ScharfViktor:
@AlexAndBear I changed title. If we want to check CI we need -> `[check-console-errors]` in the title

ScharfViktor:
> But looks like this test is successful in the re-run and is not web related, worth looking into @ScharfViktor

done, all tests are green)). we didn't close the page to the public. we closed it at the end if the test was failed (so after a retry, the test would turn green)

packages/web-runtime/src/container/sse/helpers.ts
```diff
@@ -21,7 +21,7 @@ export const isItemInCurrentFolder = ({
   parentFolderId: string
 }) => {
   const currentFolder = resourcesStore.currentFolder
-  if (!currentFolder) {
+  if (!currentFolder || !currentFolder.id) {
```
dschmidt, line 24:
```suggestion
  if (!currentFolder?.id) {
```

dschmidt (approved):

Issue #10502, opened 2024-02-21
Modals can overlay the logged out page


dschmidt:
## Steps to reproduce
open a modal, be logged out because of session timeout

## Expected behavior
Modal is closed and the logged out page is visible and accessible

## Actual behavior
Modal is not closed and prevents access to the underlying page

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

AlexAndBear:
@dschmidt what would be your technically advise to solve this ? Iterating trough modal elements and call a close method during logout or whatever page will be rendered

pascalwengerter:
> @dschmidt what would be your technically advise to solve this ? Iterating trough modal elements and call a close method during logout or whatever page will be rendered

I've faced a similar issue with wanting to remove the "UploadInfo" popup on the FilesDrop page, my way was adding a property to the route `meta` and then dynamically modifying the global `App.vue`

PR #11029, opened 2024-06-12
Remove all modals on logout

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/10502

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


dschmidt (approved):
LGTM - it feels a bit like it's handled on the wrong end but it's a non invasive approach that won't break anything, so I'm fine with it.

Another idea would be to handle it in router.beforeEach - but that could break way more, even if we removed modals only on authContext change

Issue #11035, opened 2024-06-12
Prevent loading all languages for all apps on initial page load


JammingBen:
Since we've introduced so many languages, the initial page loading times have gone down quite a lot. This is happens due to the fact that we're loading all available languages for all apps during the initial load.

The goal is this issue is to prevent that and to just load the current language initially. Other languages are then be loaded on language change.

AlexAndBear:
We need to split the translations.json into separate json files (e.g. de.json, ru.json etc.) 
https://jshmrtn.github.io/vue3-gettext/extraction.html
This can happen via
```
    splitJson: false, // create separate json files for each locale. If used, jsonPath must end with a directory, not a file
```
set to true

dschmidt:
Could possibly be implemented in multiple iterations as well... 
1) do not merge and register everything all at once
2) really load on demand (async, so probably more complicated)

JammingBen:
Okay so https://github.com/owncloud/web/pull/11040 prevents merging and registering all app languages on page load.

Splitting the languages into multiple files and loading them dynamically is not easy though. The splitting itself can be done without too much effort, but then we would need to implement some architecture so that translations can be dynamically loaded per app. So each app would need to provide a way to load translations for a given language on demand.

We don't have anything in that regard yet, and to be honest, I'm not sure if it's worth the effort currently.

DeepDiver1975:
Closing as fixed. The titled ask — preventing all app languages from being loaded/registered on initial page load — shipped in #11040 (merged 2024-06-13).

That PR removed `announceTranslations` in `web-runtime/src/container/api.ts` (which merged every language for every app into gettext at startup) and now only registers translations for the currently selected language. Still gone on current `master`. Changelog `enhancement-page-loading-times` lists "Only register translations for the currently selected language" and references this issue.

The broader idea — splitting `translations.json` into per-locale files + true async per-app on-demand loading — was deliberately deemed not worth the effort here (per @JammingBen above). If that's still wanted, please open a fresh issue scoped to that.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11040, opened 2024-06-13
perf: prevent setting all languages for all apps on initial load

JammingBen:
## Description
Changes the way how app languages get set in `gettext` so only the current language is being considered initially. Other languages will be set dynamically when changing the current language.

Also refactors the language setup process a bit to make it simpler.

`announceGettext` sets up gettext in our Vue app, without any translations yet though. `announceTranslations` gets called later during the bootstrap process and provides all translations to `gettext`: the ones for core, additional ones and the ones for the apps for the current language. That way we have all that stuff in one place.

## Related Issue

- refs https://github.com/owncloud/web/issues/11035

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/enhancement-page-loading-times
+++ b/changelog/unreleased/enhancement-page-loading-times
@@ -2,7 +2,7 @@ Enhancement: Faster page loading times
 
 We've massively decreased the initial page loading times. This is especially noticeable with slower network connections.
 
-According to the lighthouse performance metrics, we improved from a score of 60 to 94.
+According to the lighthouse performance metrics, we improved from a score of 60 to 94 out of 100.
 
 This has been achieved by taking several different actions:
 
--- a/changelog/unreleased/enhancement-page-loading-times
+++ b/changelog/unreleased/enhancement-page-loading-times
@@ -1,0 +1,19 @@
+Enhancement: Faster page loading times
+
+We've massively decreased the initial page loading times. This is especially noticeable with slower network connections.
+
+According to the lighthouse performance metrics, we improved from a score of 60 to 94.
+
+This has been achieved by taking several different actions:
+
+- Lazy load large dependencies
+- Improve chunking
+- Enable server side compression
+- Reduce the size of the main font we're using
+- Only register translations for the currently selected language
+
+https://github.com/owncloud/web/pull/10976
+https://github.com/owncloud/web/pull/10979
+https://github.com/owncloud/web/pull/11040
+https://github.com/owncloud/web/issues/7964
+https://github.com/owncloud/web/issues/11035
--- a/packages/web-runtime/src/container/bootstrap.ts
+++ b/packages/web-runtime/src/container/bootstrap.ts
@@ -3,7 +3,7 @@ import { buildApplication, NextApplication } from './application'
 import { RouteLocationRaw, Router, RouteRecordNormalized } from 'vue-router'
 import { App, watch } from 'vue'
 import { loadTheme } from '../helpers/theme'
-import { createGettext, GetTextOptions, Language } from 'vue3-gettext'
+import { createGettext, GetTextOptions, Language, Translations } from 'vue3-gettext'
 import { getBackendVersion, getWebVersion } from './versions'
 import {
   useModals,
@@ -74,6 +74,7 @@ import {
   onSSELinkUpdatedEvent
 } from './sse'
 import { useWebWorkersStore, WebWorkersStore } from '@ownclouders/web-pkg'
+import { loadAppTranslations } from '../helpers/language'
 
 const getEmbedConfigFromQuery = (
   doesEmbedEnabledOptionExists: boolean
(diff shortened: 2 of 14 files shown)


changelog/unreleased/enhancement-page-loading-times
```diff
@@ -0,0 +1,19 @@
+Enhancement: Faster page loading times
+
+We've massively increased the initial page loading times. This is especially noticiable with slower network connections.
```
phil-davis, line 3:
```suggestion
We've massively decreased the initial page loading times. This is especially noticeable with slower network connections.
```

packages/web-runtime/src/helpers/language.ts
```diff
@@ -16,3 +18,31 @@ export const setCurrentLanguage = ({
     document.documentElement.lang = currentLanguage
   }
 }
+
+/**
+ * Loads all app translations for one given language.
+ * This is adviced to call each time the language is being changed.
```
phil-davis, line 24:
```suggestion
 * This should be called each time the language is being changed.
```

changelog/unreleased/enhancement-page-loading-times
JammingBen: Haha 😂 Thanks for the hint!

changelog/unreleased/enhancement-page-loading-times
```diff
@@ -0,0 +1,19 @@
...
+
+According to the lighthouse performance metrics, we improved from a score of 60 to 94.
+
+This has been achieved by taking several different actions:
+
+- Lazy load large dependencies
+- Improve chunking
+- Enable server side compression
+- Reduce the size of the main font we're using
+- Stop setting all app languages on the initial page load
```
dschmidt, line 13: Can we improve this wording? As a user I have no idea what this means

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -382,13 +364,26 @@ export const announceTranslations = ({
...
+  customTranslations?: Translations
+  appsStore?: AppsStore
+}) => {
+  gettext.translations = merge(coreTranslations, customTranslations || {})
+
+  if (appsStore) {
+    loadAppTranslations({
+      apps: appsStore.apps,
+      gettext,
+      lang: gettext.current
```
dschmidt, line 384: Is this correctly set on application boot? When is the language setting loaded?

packages/web-runtime/src/helpers/language.ts
```diff
@@ -16,3 +18,31 @@ export const setCurrentLanguage = ({
...
+}) => {
+  const appTranslations = Object.values(apps).reduce<Translations>((acc, app) => {
+    const { translations } = app
+
+    if (gettext.translations[lang] && translations[lang]) {
+      acc = merge(acc, translations[lang])
+    }
+
+    return acc
+  }, {})
```
dschmidt, line 43: It's my personal taste, but I always feel it's weird to use `reduce` in this way. This is basically the same as:

```
let acc: Translations = {}
Object.values(apps).forEach(app => {
    const { translations } = app
    if (gettext.translations[lang] && translations[lang]) {
      acc = merge(acc, translations[lang])
    }
})
```
isn't it?

dschmidt (commented):
LGTM 😍 

Haven't tested it, but code wise looks so much better than before

packages/web-runtime/src/helpers/language.ts
dschmidt: possibly could also be `const acc` and then just `Object.assign` instead of merge

JammingBen: Yup that works as well. No hard feelings from my side 🤷

packages/web-runtime/src/container/bootstrap.ts
JammingBen: We need to set it here initially to make sure the correct language has been loaded in an unauthenticated context. In an authenticated context, we set it a second time after the user has been fetched via the Graph API.

packages/web-runtime/src/helpers/language.ts
```diff
@@ -16,3 +18,29 @@ export const setCurrentLanguage = ({
...
+  lang
+}: {
+  apps: Record<string, ApplicationInformation>
+  gettext: Language
+  lang: string
+}) => {
+  const appTranslations: Translations = {}
+  Object.values(apps).forEach((app) => {
+    const { translations } = app
+    if (gettext.translations[lang] && translations[lang]) {
```
kulmann, line 38: What is the `gettext.translations[lang]` check for here? 🙈

JammingBen: As discussed: It checks if we support the requested language and that there are translations available for it. This is basically determined by the core translations.

dschmidt (approved):

PR #11049, opened 2024-06-17
Rename enabled only filter to include disabled, turn off by default

AlexAndBear:
## Description
image

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/ItemFilterToggle.vue
+++ b/packages/web-pkg/src/components/ItemFilterToggle.vue
@@ -39,7 +39,7 @@ export default defineComponent({
       return router.push({
         query: {
           ...omit(unref(currentRoute).query, [queryParam]),
-          [queryParam]: unref(filterActive).toString()
+          ...(unref(filterActive) && { [queryParam]: 'true' })
         }
       })
     }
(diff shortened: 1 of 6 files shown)


packages/web-pkg/src/components/ItemFilterToggle.vue
```diff
@@ -39,7 +39,7 @@ export default defineComponent({
       return router.push({
         query: {
           ...omit(unref(currentRoute).query, [queryParam]),
-          [queryParam]: unref(filterActive).toString()
+          ...(unref(filterActive) && { [queryParam]: 'true' })
```
AlexAndBear, line 42: reverted to original state

kulmann (changes requested):
Could you please adjust the existing changelog item and add the link of this PR to the link list?

AlexAndBear:
> Could you please adjust the existing changelog item and add the link of this PR to the link list?

I did that, but then reverted, since this is in a release folder now, should I still change it, or add a new one ?

kulmann:
> > Could you please adjust the existing changelog item and add the link of this PR to the link list?
> 
> I did that, but then reverted, since this is in a release folder now, should I still change it, or add a new one ?
> 
> See [7de92ee](https://github.com/owncloud/web/commit/7de92ee4bf41c3d3db2d87c2c7049f95c1ab15bb)

Just do that in the release folder. It's from an alpha, doesn't matter. Only would be relevant if it would have been a final release.

kulmann (approved):
👍

PR #11051, opened 2024-06-17
[test-only] fix test expiration date

ScharfViktor:
I'm not very happy with my current solution, but it can unlock you @kulmann quickly. 

this is what happens: 

https://github.com/owncloud/web/assets/84779829/2cb5f01d-e3f1-4b94-8c3c-2a7098f3f978

after setting timeout this behavior doesn't exist

### DIFF
--- a/tests/e2e/support/objects/app-files/share/collaborator.ts
+++ b/tests/e2e/support/objects/app-files/share/collaborator.ts
@@ -257,15 +257,15 @@ export default class Collaborator {
       },
       { newExpiryDate }
     )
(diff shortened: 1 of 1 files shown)


kulmann:
Can you reproduce this glitch in the ui or is it only happening in playwright? 😨

ScharfViktor:
> Can you reproduce this glitch in the ui or is it only happening in playwright? 😨

no, I couldn't reproduce it in ui

kulmann (approved):

Issue #10982, opened 2024-05-31
Prevent toast-ui editor from always loading all dependencies


JammingBen:
The `TextEditor` component loads the `toast-ui` editor package + its translations + a package for syntax highlighting. However, the latter 2 are only needed for editing markdown files, so there is no need to load these packages if the editor is a) in read-only mode or b) not showing a markdown file.

PR #11060, opened 2024-06-18
perf: lazy load toast-ui editor dependencies

JammingBen:
## Description
Dynamically load toast-ui editor languages and the syntax highlight plugin for markdown files. This improves editor loading times in other scenarios where of one these 2 things are not needed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10982

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-pkg/src/components/TextEditor.vue
```diff
@@ -71,7 +50,59 @@ export default defineComponent({
...
+
+      return plugin.default
+    }
+
+    onMounted(async () => {
+      let codeSyntaxHighlight: unknown
+      if (!props.isReadOnly && unref(isMarkdown)) {
+        // only needed for md documents with write permissions
+        await loadTranslations()
+        codeSyntaxHighlight = await loadSyntaxHighlighting()
```
AlexAndBear, line 100: Syntax highlighting should be possible even if read only ? 🤷
Check space readme for example

AlexAndBear (changes requested):

packages/web-pkg/src/components/TextEditor.vue
JammingBen: Ah okay, I didn't know that 👍

AlexAndBear: I think that also applies to shared MD files with viewer permissions 👍

AlexAndBear (approved):

AlexAndBear:
Looks a little clumsy but I fear that's the downside of making it performant 😁

JammingBen:
> Looks a little clumsy but I fear that's the downside of making it performant 😁

Totally agree :( Also the weird dist-imports of the lib contribute to that unfortunately.

kulmann:
> > Looks a little clumsy but I fear that's the downside of making it performant 😁
> 
> Totally agree :( Also the weird dist-imports of the lib contribute to that unfortunately.

Do you think this is something to solve or propose upstream? ;-)

AlexAndBear:
Could you think of adding a new ts file for the translations where you import the lang files and then just importing that file Async, for now ?

JammingBen:
> Do you think this is something to solve or propose upstream? ;-)

I didn't dig too deep, but it seems that you can use the non-dist imports, but then you have to explicitly name the supported languages for the syntax highlighting.

> > Could you think of adding a new ts file for the translations where you import the lang files and then just importing that file Async, for now ?
> 
> good idea 👍

Hmm I don't see much benefit tbh. Also it would raise the question where this file would live. We're not in an app context here, this is just a single `web-pkg` component. But if you 2 want to have it that way, sure that's possible 🤷

AlexAndBear:
Just decluttering I guess. Could make a new Folder 
/Components/TextEditor/editor.vue
/Components/TextEditor/i18n.ts

PR #11067, opened 2024-06-19
[full-ci][tests-only] regenerate the access token for web-ocis while running with keycloak

nabim777:
## Description

It was found that the access token used was invalid so deletion of any users from Web-Ocis is not happened afterall clean up. The access token becomes invalid due to Keyclock's backchannel logout feature, which makes the token invalid when the user logs out. So, above problem is cover in this PR.
More describe in here: [link](https://github.com/owncloud/docs-ocis/pull/872/files#diff-634ab767980da7e3f92f7e33d92a0f46480f230cc1d6203e74056f1bdd6069cdR25-R27)

Also in this PR, there is addition of the scenario from suites `admin-settings` that will pass in drone.
The scenarios of suites `admin-settings` which are fails in drone and are not included are as follows:

```console
- Scenario: user login can be managed in the admin settings: 3
- Scenario: edit user:84
- Scenario: admin creates user: 168
```

## Related Issue

- https://github.com/owncloud/web/issues/10458

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1796,7 +1796,13 @@ def e2eTestsOnKeycloak(ctx):
     e2e_Keycloak_tests = [
         "journeys",
         "admin-settings/users.feature:20",
+        "admin-settings/users.feature:43",
+        "admin-settings/users.feature:106",
+        "admin-settings/users.feature:131",
+        "admin-settings/users.feature:185",
         "admin-settings/spaces.feature",
+        "admin-settings/groups.feature",
+        "admin-settings/general.feature",
     ]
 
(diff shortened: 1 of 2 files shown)


saw-jan (approved):
:+1:

ScharfViktor (approved):

amrita-shrestha (approved):

Issue #11063, opened 2024-06-19
Shortcuts with no protocol prefix


jesmrec:
## Describe the bug

Web client creates shortcuts without protocol prefix (if not typed by the user), that leads to unexpected behaviours in other clients that expects the URL inside the shortcuts with http/https. Mobile clients can not open URLs in a browser if the URL has no http/https protocol 

## Steps to reproduce

1.  New > Shortcut
2. Add URL without protocol prefix, like `owncloud.com`
3. Open shortcut in iOS client

## Expected behavior

iOS client opens correctly the shortcut

## Actual behavior

iOS can't open because the URL lacks of protocol

Actually, every client behaves in a different way:

- Web: doesn't add protocol in creation phase
- iOS: add protocol in creation phase
- Android: initially, add protocol in creation phase (this is still under QA)

As this unofficial format guide: https://www.lyberty.com/encyc/articles/tech/dot_url_format_-_an_unofficial_guide.html

```
URL

The URL field is self-explanatory. It’s the address location of the page to load. It should be a fully qualifying URL with the format protocol://server/page. A URL file is not restricted to the HTTP protocol. In general, at least, whatever that can be saved as a favorite is a valid URL.
```

So, my suggestion is:

- If user types the protocol in the URL, everything OK ✅ 
- If not, adding `https://` as default in creation phase

Open to discussion with @felix-schwarz @Aitorbp @JuancaG05 @TheOneRing

kulmann:
Fix for this should land in `stable-8.0` and `master`. A patch release of v8.0 makes sense.

kulmann:
I agree with the default behaviour of adding `https://` if the user didn't provide a protocol. 👍

JammingBen:
Has been fixed via https://github.com/owncloud/web/pull/11076.

PR #11076, opened 2024-06-20
fix: enforce protocol in shortcut urls

kulmann:
## Description
If the user types a URL without protocol and doesn't click on the proposed URL in the dropdown (which would include `https://` as protocol), the resulting .url file wouldn't have a protocol in the URL. This PR fixes that by enforcing `https://` as prefix for the user input if none was given. Technically this would've been possible by using `unref(dropItemUrl)` instead of `unref(inputUrl)`, but I've decided to extract the body of `dropItemUrl` into a function and use that, because it would be confusing to use `dropItemUrl` as the final input on confirm.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/11063

## Motivation and Context
Enforce url protocol so that mobile clients won't have trouble opening the URLs.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-enforce-shortcut-url-protocol
+++ b/changelog/unreleased/bugfix-enforce-shortcut-url-protocol
@@ -1,0 +1,6 @@
+Bugfix: Enforce shortcut URL protocol
+
+We've fixed a bug where the protocol in shortcut URLs was not enforced. Now, the shortcut URL always starts `https://` if no protocol is provided by the user.
+
+https://github.com/owncloud/web/issues/11063
+https://github.com/owncloud/web/pull/11076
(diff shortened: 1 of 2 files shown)


AlexAndBear (changes requested):
Since we add the protocol to the file, we can omit the protocol rewrite in packages/web-pkg/src/composables/actions/files/useFileActionsOpenShortcut.ts

kulmann:
> Since we add the protocol to the file, we can omit the protocol rewrite in packages/web-pkg/src/composables/actions/files/useFileActionsOpenShortcut.ts

I think it still makes sense there. url files which were create with a different application might have been uploaded and not have a protocol in the url, we still want to be able to use them.

AlexAndBear:
> > Since we add the protocol to the file, we can omit the protocol rewrite in packages/web-pkg/src/composables/actions/files/useFileActionsOpenShortcut.ts
> 
> I think it still makes sense there. url files which were create with a different application might have been uploaded and not have a protocol in the url, we still want to be able to use them.

Got you, but then the initial problem, that has been reported is still an issue on the clients side, when a user does this ;)

AlexAndBear (approved):

JammingBen (approved):

Issue #11069, opened 2024-06-19
Prevent loading and initializing all apps on token renewal


JammingBen:
Token renewal currently re-loads and re-initializes all apps in an iFrame, although not being necessary. We can reduce the network load by quite a bit by preventing this and only loading necessary parts.

PR #11077, opened 2024-06-20
perf: reduce network load on token renewal

JammingBen:
## Description
Reduces the network load on token renewal by not bootstrapping the whole application. Parts that are skipped loading during token renewal are:

* loading applications
* loading the inter font
* loading translations
* loading the theme

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11069

## Benchmarks

Network load during token renewal - **before**:

Pasted Graphic 4

Network load during token renewal - **after**:

Pasted Graphic 3

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/container/bootstrap.ts
+++ b/packages/web-runtime/src/container/bootstrap.ts
@@ -1,7 +1,7 @@
 import { registerClient } from '../services/clientRegistration'
 import { buildApplication, NextApplication } from './application'
 import { RouteLocationRaw, Router, RouteRecordNormalized } from 'vue-router'
-import { App, watch } from 'vue'
+import { App, computed, watch } from 'vue'
 import { loadTheme } from '../helpers/theme'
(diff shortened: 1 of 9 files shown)


packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -410,6 +411,37 @@ export const announceClientService = ({
...
+    userStore,
+    configStore.serverUrl,
+    computed(
+      () =>
+        capabilityStore.filesArchivers || [
+          {
+            enabled: true,
+            version: '1.0.0',
+            formats: ['tar', 'zip'],
+            archiver_url: `${configStore.serverUrl}index.php/apps/files/ajax/download.php`
```
kulmann, line 436: I know that this is not your code and you just moved it around, but could you use `urlJoin` here for joining the serverUrl and the path string that starts with `index.php`?

kulmann (approved):
Awesome 🥳

Issue #11083, opened 2024-06-21
Initial page loading spinner never hides in some scenarios


JammingBen:
## Describe the bug
The initial page loading spinner never hides in scenarios where we don't load all apps. This is due to the fact that the spinner only gets removed via a callback after all apps have been loaded.

## Steps to reproduce
1. Go to https://host.docker.internal:9200/access-denied
2. Enjoy the endless loading state

## Expected behavior
The loading spinner eventually disappears.

## Actual behavior
It stay forever on such pages.

This is a regression from https://github.com/owncloud/web/pull/11054.

kulmann:
I can't reproduce it with current master 🙈  neither mit `vite` nor with a production `build`.

JammingBen:
Now I can't either... but it's a timing issue. The loading spinner appears after 0.5 seconds. Meaning if the page loads faster than that, everything is fine. But if it's slower, then the loading spinner will appear and never go away.

PR #11084, opened 2024-06-21
fix: endless initial loading spinner

JammingBen:
## Description
Fixes an issue where the initial page loading spinner would not disappear in scenarios where we don't load all apps. This affects all plain layout pages such as the access-denied page, but also the password form for public links.

I'm not suuuuper happy with the solution, but I also can't find a better one currently.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11083

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


AlexAndBear (changes requested):
Should this loading spinner be always in the DOM to enable it if needed on the fly?

If yes, it makes sense to adjust the comment wording since we actually don't remove it but hide/disable it. 

If no, it makes sense to remove it from the dom

kulmann (approved):
🚀

AlexAndBear (approved):

PR #11086, opened 2024-06-24
fix: visiting public links with expired access token

JammingBen:
## Description
Fixes bug where opening public links with an expired token would falsely show the access denied page. Instead, Web now tries to login the user. If that doesn't work, the user gets logged out.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/6734

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-opening-public-links-with-expired-token
+++ b/changelog/unreleased/bugfix-opening-public-links-with-expired-token
@@ -1,0 +1,5 @@
+Bugfix: Opening public links with an expired token
+
+We've fixed a bug where opening public links with an expired token would falsely show the access denied page. Instead, Web now tries to login the user. If that doesn't work, the expired user session gets removed.
+
+https://github.com/owncloud/web/pull/11086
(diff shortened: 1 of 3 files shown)


changelog/unreleased/bugfix-opening-public-links-with-expired-token
```diff
@@ -0,0 +1,5 @@
+Bugfix: Opening public links with an expired token
+
+We've fixed a bug where opening public links with an expired token would falsely show the access denied page. Instead, Web now tries to login the user. If that doesn't work, the user gets logged out.
```
kulmann, line 3:
```suggestion
We've fixed a bug where opening public links with an expired token would falsely show the access denied page. Instead, Web now tries to login the user. If that doesn't work, the expired user session gets removed.
```

kulmann (approved):

AlexAndBear (approved):

Issue #10980, opened 2024-05-31
Make routes lazy


JammingBen:
To improve overall performance of Web we want to lazy load all of our routes, since it doesn't make sense to load all of them on the initial page load.

The implementation will also need some wrapper component that displays a loading state (with some debounce) during which a route is loading its data.

kobergj:
not planned for now

PR #11089, opened 2024-06-24
perf: make routes lazy

JammingBen:
## Description
Makes all routes lazy so their components don't get loaded with the initial load, but when navigating to the routes.

The `ComponentLoader` is basically a component wrapping the route components to display a loading spinner during the loading times.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10980

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-admin-settings/src/index.ts
+++ b/packages/web-app-admin-settings/src/index.ts
@@ -1,11 +1,8 @@
 import translations from '../l10n/translations.json'
-import General from './views/General.vue'
-import Users from './views/Users.vue'
-import Groups from './views/Groups.vue'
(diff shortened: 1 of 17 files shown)


JammingBen:
I'm a bit torn here. It adds a bit of complexity while (currently) having near to no impact (Web now loads smaller but more chunks, even resulting in ~10-20ms more initial loading time - however this is not noticable). There might be situations in the future where this would be useful though, e.g. when some route component loads a lot of dependencies.

@kulmann @dschmidt any opinions from your side?

JammingBen:
Closing here since we're currently not planing this.

PR #11096, opened 2024-06-25
Remove aria-describedby for OcDrop

### LINKED ISSUE #5398: [a11y] 9.4.1.3 - Status messages programatically available
- Context: The Web UI should be accessible according to the WCAG2.1 /BITV 2.0 guidelines. The number in the title relates to the test-steps: https://www.bitvtest.de/bitv_test/das_testverfahren_im_detail/pruefschritte.html
- "Status" relates to the evaluation scheme https://www.bitvtest.de/bitv_test/das_testverfahren_im_detail/verfahren.html#c1262 
- "a11y" is short for accessibility
- ----------------
## General notes
All pages / Popups "Apps-Menü" and "Benutzereinstellungen" (probably also other similar Popups affected): After activating the button the whole container of the displayed popup will be read out by screenreaders.
(On user settings it is the 3 links). Better solution would be to let the screenreader read out only „erweitert“ after activating the app-menue. The aria-describedby-Attribute which triggers the output should be removed.
**Note:** the issue also appears with Firefox. Edge doesn’t read out the content.

## Status: partly fulfilled

pascalwengerter:
## Related Issue
- Fixes #5398

### DIFF
--- a/changelog/unreleased/enhancement-accessibility-improvements
+++ b/changelog/unreleased/enhancement-accessibility-improvements
@@ -3,4 +3,6 @@ Enhancement: Accessibility improvements
 Across the board, we have implemented improvements in regards of accessibility for the web UI.
 
 https://github.com/owncloud/web/issues/5387
+https://github.com/owncloud/web/issues/5398
 https://github.com/owncloud/web/pull/11087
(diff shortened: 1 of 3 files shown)


kulmann (approved):
👍

PR #11098, opened 2024-06-25
Add missing aria-labels for OcDrops, hide WhiteSpaceContextMenu from screenreader

pascalwengerter:
## Description
As discussed with @kulmann


kulmann (approved):
👍

PR #11103, opened 2024-06-26
Dynamic app provider apps

kulmann:
## Description
Dynamically register one instance of the `external` app per app provider app.

## Motivation and Context
No special handling for `open file in editor/viewer` actions anymore.

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:
- [x] fix broken page reload from within editor (routes are registered too late in the boot process... maybe we can have a catch all route and remove it once external apps are registered)
- [x] remove file action composable special handling
- [x] segmentation of viewers/editors into `app provider apps` and `native apps` was very useful in the `+ New`-menu. Can we skip it in the future or do we want to have some sort of categorization for it?

### DIFF
--- a/packages/web-app-external/package.json
+++ b/packages/web-app-external/package.json
@@ -11,6 +11,7 @@
   "peerDependencies": {
     "@ownclouders/web-client": "workspace:*",
     "@ownclouders/web-pkg": "workspace:*",
+    "pinia": "2.1.7",
     "uuid": "9.0.1",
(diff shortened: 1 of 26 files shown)


AlexAndBear:
packages/web-pkg/src/composables/actions/files/useFileActions.ts:173 can be adjusted as well I guess, we don't need to provide the app name as query

AlexAndBear (approved):

Issue #11056, opened 2024-06-18
File/Folder count in Details panel for current folder


kulmann:
# Description

## User Stories

* > As a user with high file count in one folder, I want to see at a glance how many files/folders the current folder contains so that I don't have to scroll the file list to the bottom, where the footer displays that exact information.

## Value
Easy UX

## Acceptance Criteria
- in Details panel, show file/folder count above the `Size` if available
  - show a total count and additionally a detailed count (files / folders / hidden) like in the footer
  - only show for folders and space roots
  - only show for current folder/space root, not for items in the file list, because:
  - afaik this information is only available in a file listing (propfind with depth=1), not in a search response or on the children of a propfind

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

tbsbdr:
like that?
![screenshot_002778](https://github.com/owncloud/web/assets/26610733/6f91bf17-7ffc-4fd1-bd02-0c55dadd1ce8)

kulmann:
Yes, that would be cool. Having it in the `Size` field as additional information if available is a good idea 👍 

For completeness, when you also have hidden items it would read as follows:

`1.5 MB, 7 items (4 files including 1 hidden, 3 folders including 2 hidden)`

PR #11117, opened 2024-07-01
Show folder contents in details panel

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/11056

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -181,7 +181,7 @@ export default defineComponent({
     const userStore = useUserStore()
     const capabilityStore = useCapabilityStore()
     const { getMatchingSpace } = useGetMatchingSpace()
-    const { resourceContentsText } = useResourceContents()
+    const { resourceContentsText } = useResourceContents({ showSizeInformation: false })
 
     const language = useGettext()
 
@@ -349,7 +349,9 @@ export default defineComponent({
     },
     resourceSize() {
       if (isPersonalSpaceResource(this.space) && this.resource.path === '/') {
-        return this.resourceContentsText
+        return `${formatFileSize(this.resource.size, this.$language.current)}, ${
+          this.resourceContentsText
+        }`
       }
 
       return formatFileSize(this.resource.size, this.$language.current)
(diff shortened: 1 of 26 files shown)


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -345,10 +349,16 @@ export default defineComponent({
       return this.resource.owner?.displayName
     },
     resourceSize() {
+      if (this.resource.id === this.currentFolder.id) {
```
JammingBen, line 352: For places where we don't have a current folder, e.g. search results.

```suggestion
      if (this.resource.id === this.currentFolder?.id) {
```

packages/web-app-files/src/views/Favorites.vue
```diff
@@ -42,13 +42,7 @@
           </template>
           <template #footer>
             <pagination :pages="paginationPages" :current-page="paginationPage" />
-            <list-info
-              v-if="paginatedResources.length > 0"
-              class="oc-width-1-1 oc-my-s"
-              :files="totalResourcesCount.files"
-              :folders="totalResourcesCount.folders"
-              :size="totalResourcesSize"
```
JammingBen, line 50: I guess `totalResourcesCount` and `totalResourcesSize` can now be removed as well here? Same goes for the other views.

JammingBen (commented):
I think this change is worth a changelog item, no?

AlexAndBear:
> I think this change is worth a changelog item, no?

Hm idk, somehow this is: "You will see now additionally in the details panel how many files and folders are there, but only if you open the details panel for the current folder or space root, you can also see it if you just look down to the files table footer."
It's a gimmick but not really adds a lot of user value, so idk

packages/web-app-files/src/views/Favorites.vue
JammingBen: You've adjusted the `Favorites` view, but the other views are still missing. Don't know if it can be removed in all of them, but in some at least.

AlexAndBear: i hope that all now

JammingBen (approved):
👍

Issue #10371, opened 2024-01-19
Clicking the oC logo on public links causes weird issues


JammingBen:
## Steps to reproduce
1. Make sure you are running a stack with external apps such as OnlyOffice or Collabora
2. Create a public link for a folder that includes a `docx` file
3. Open the link in an incognito window
4. Click on the oC logo a few times
5. Right click the docx file to open the context menu

## Expected behavior
Clicking the oC logo always reloads the current link. Right clicking the file shows the context menu and its entries as expected.

## Actual behavior
Sometimes the page dies when trying to reload, giving an error `Cannot read properties of null (reading 'canCreate')`. Reloads that are successful on the other hand cause duplicated entries for external apps in the context menu:

<img width="290" alt="image" src="https://github.com/owncloud/web/assets/50302941/d6628003-1aea-4878-82d9-f7edf8e077d9">

The error only exists on master, not on `stable-8.0`.

JanAckermann:
Introduced via https://github.com/owncloud/web/pull/10207

PR #11137, opened 2024-07-04
fix: duplicated elements on public link page

JammingBen:
## Description
Fixes a bug where clicking the ownCloud logo on a public link page would lead to certain UI elements being duplicated.

This happens because some extensions and extension points are being registered more than once because the application layout gets unmounted on mounted again when clicking the logo. The solution is to properly unregister those extensions on unmount.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10371

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-duplicated-ui-elements-public-link
+++ b/changelog/unreleased/bugfix-duplicated-ui-elements-public-link
@@ -1,0 +1,6 @@
+Bugfix: Duplicated elements on public link page
+
+We've fixed a bug where clicking the ownCloud logo on a public link page would lead to certain UI elements being duplicated.
+
+https://github.com/owncloud/web/pull/11137
+https://github.com/owncloud/web/issues/10371
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsCopy.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsCopy.ts
@@ -66,7 +66,7 @@ export const useFileActionsCopy = () => {
           }
 
           if (isLocationPublicActive(router, 'files-public-link')) {
-            return unref(currentFolder).canCreate()
+            return unref(currentFolder)?.canCreate()
           }
 
           if (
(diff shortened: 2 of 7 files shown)


kulmann:
An alternative approach would be to not register the extensionPoint and extension in the layout but somewhere earlier in the process... I'd prefer that actually, but back then when I wrote the extensionPoint and extension for the progress bar I couldn't find a good place to do that. Do you have an idea?

JammingBen:
> An alternative approach would be to not register the extensionPoint and extension in the layout but somewhere earlier in the process... I'd prefer that actually, but back then when I wrote the extensionPoint and extension for the progress bar I couldn't find a good place to do that. Do you have an idea?

Maybe the main component wrapping the layouts, so in `App.vue`?

kulmann:
> > An alternative approach would be to not register the extensionPoint and extension in the layout but somewhere earlier in the process... I'd prefer that actually, but back then when I wrote the extensionPoint and extension for the progress bar I couldn't find a good place to do that. Do you have an idea?
> 
> Maybe the main component wrapping the layouts, so in `App.vue`?
> 
> Edit: Also not super optimal when I think about it since other layouts don't need the progress bar. But it's the only option I can think of that doesn't get re-mounted on logo click.

Doesn't even need to be in a component. All the apps provide their extensions and extensionPoints to the web runtime for being registered during app initialization. So why not register extensions and extensionPoints from within the runtime directly before or after app initialization during bootstrap? 🤔

kulmann (approved):

kulmann:
> > Doesn't even need to be in a component. All the apps provide their extensions and extensionPoints to the web runtime for being registered during app initialization. So why not register extensions and extensionPoints from within the runtime directly before or after app initialization during bootstrap? 🤔
> 
> Sounds like a good idea. However, it makes the solution for this issue more complicated 😬 That's because the watcher that initializes all apps gets also called each time the logo is being clicked. So we would again re-register existing extensions. I guess it would be best to fix that... although I don't know how (yet). The issue is that clicking the logo toggles `publicLinkContextReady` off and on, leading to the watcher re-firing.

Oooof. I've just recently added the initialization of the dynamic app provider based apps into this exact watcher. That needs fixing. 😢  I've approved this PR for the sake of having an early bug fix, but could you look into the other issue please? 🙈

kulmann (approved):

Issue #11120, opened 2024-07-01
[Web] Provide link for users with existing access


tbsbdr:
# Description

## User Stories

* > As a user I want to copy a link to a file so that I can send a direct pointer for my colleagues.

## Value

- Reduced Training Time
- Increased Efficiency & Enhanced Collaboration
- Streamlined Processes
- Error Reduction

## Acceptance Criteria
- add a "Permanent link" button next to "shared with" in the shares panel
- click on the button copies the private link (example https://infinite.owncloud.com/f/31e6d44f-f373-557c-9ab3-1748fc0c650d%24cb86f47b-4f06-4bff-92ca-186e6b8eebdf%211d683591-3529-4106-b1f3-871642e5a289 )
     - pro: robust/permanent
    - con: non-semantic (surprise package)
- Add context helper. text : "`Copy the link to point your team to this share. Works only for people with existing access.`"  
- Rename "Share via link" --> "Share public"
- Link appears next to "Shared with" or "Members" (depending on whats on top)

----------------------
Dismissed for now (we're NOT doing this!): The first part or the URL (https://infinite.owncloud.com/pdf-viewer/project/marketing/Presentations/2024-29-06_openSuSE_Conference/2024-29-06_SuSE-Conference.pdf )
    - pro: semantic (preview in file name)
    - con: long
    - con: unstable (fails if path gets renamed)

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

## releasenote

**Jumplinks for Easy Navigation**
You can now copy a direct link to any file or folder for users with existing access. This Jumplink allows your colleagues to navigate directly to the specified file without needing to click through the file tree. Simply copy the Jumplink to point your team to the location they need.

## Visual
favored and dismissed variants
![image](https://github.com/owncloud/web/assets/26610733/18467baa-de36-42cd-8f4f-70d7f6e5f30f)

PR #11161, opened 2024-07-08
Add Permanent Link copy button

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/11120

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
```diff
@@ -19,10 +19,11 @@
       data-testid="files-collaborators-no-share-permissions-message"
       v-text="noSharePermsMessage"
     />
-    <div v-if="hasSharees" class="avatars-wrapper oc-flex oc-flex-middle oc-flex-between">
-      <h4 class="oc-text-bold oc-my-rm" v-text="sharedWithLabel" />
-    </div>
     <template v-if="hasSharees">
+      <div id="files-collaborators-headline" class="oc-flex oc-flex-middle oc-flex-between">
+        <h4 class="oc-text-bold oc-my-rm" v-text="sharedWithLabel" />
+        <CopyPrivateLink :resource="resource" />
```
JammingBen, line 25: There are a 2 more occurrences in this file and in `SpaceMembers.vue`.

```suggestion
        <copy-private-link :resource="resource" />
```

JammingBen (commented):
There is an existing component `PrivateLinkItem.vue`, maybe this can be removed now?

kulmann:
> There is an existing component `PrivateLinkItem.vue`, maybe this can be removed now?

Yes, please. And the private link capability in ocis needs to be enabled and respected in web (for the new permanent link copy button).

packages/web-client/src/helpers/space/functions.ts
```diff
@@ -180,6 +180,7 @@ export function buildSpace(
     description: data.description,
     extension: '',
     path: '/',
+    webUrl: data.webUrl,
```
kulmann, line 183: can't we name it `privateLink` here as well to avoid the case differentiation in the CopyPrivateLink component? 🤔

packages/web-app-files/src/components/Shares/CopyPrivateLink.vue
```diff
@@ -0,0 +1,51 @@
...
+      copy(link)
+      showMessage({
+        title: $gettext('Permanent link copied'),
+        desc: $gettext('The permanent link has been copied to your clipboard.')
+      })
+    }
+
+    const helperText = computed(() => {
+      return $gettext(
+        'Copy the link to point your team to this share. Works only for people with existing access.'
```
kulmann, line 44: `share` might be misleading because it also works for spaces and files and folders within spaces. Maybe just `item`? Or `file` (as a collective term for files and folders)?

kulmann (approved):

packages/web-client/src/helpers/space/functions.ts
AlexAndBear: we could, but thats the raw data from graph, do we really want to coerce it to another name ?

packages/web-app-files/src/components/Shares/CopyPrivateLink.vue
AlexAndBear: Text was provided in the issue ticket, I will adjust it to item

packages/web-client/src/helpers/space/functions.ts
AlexAndBear: oh yeah, quite true we already have private link on that resourceType

AlexAndBear: done :-)

Issue #10960, opened 2024-05-24
[QA] Configure ocis collaboration service for e2e tests


saw-jan:
oCIS (master) now has a collaboration service (which is the wopi server). We want to use the wopi server provided by the ocis.

Collaboration service:
- separate app-provider services are not required
- ~can be setup with ONLY ONE office (collabora or onlyofice)~ Possible to setup with multiple online office suites

useful envs (UPDATED):
```yml
COLLABORATION_HTTP_ADDR: 0.0.0.0:9300
COLLABORATION_WOPI_SRC: https://wopiserver.owncloud.test
COLLABORATION_APP_NAME: "Collabora"
COLLABORATION_APP_ADDR: https://collabora.owncloud.test
COLLABORATION_APP_ICON: https://collabora.owncloud.test/favicon.ico
COLLABORATION_APP_INSECURE: true
COLLABORATION_CS3API_DATAGATEWAY_INSECURE: true
```

Open tasks:
- [x] configure service: [docker-compose](https://github.com/owncloud/web/blob/master/docker-compose.yml) https://github.com/owncloud/web/pull/11162
- [x] configure service: [.drone.star](https://github.com/owncloud/web/blob/master/.drone.star) #11178
- ~[ ] separate Collabora and OnlyOffice e2e tests~
- [x] remove cs3org wopiserver configs #11178
- [x] unskip e2e https://github.com/owncloud/web/pull/10946/commits/67c948d87d25be095789b661e164a357735dd878 @ScharfViktor  https://github.com/owncloud/web/pull/11185

~:information_source: **_NOTE: wait until the service is stable and there's deployment example available_**~

CC @ScharfViktor

saw-jan:
https://github.com/owncloud/ocis/pull/9113#discussion_r1620846713
>some of the service env vars will change with
>- https://github.com/owncloud/ocis/pull/9165
>- https://github.com/owncloud/ocis/pull/9277
>- https://github.com/owncloud/ocis/pull/9280

saw-jan:
>ℹ️ NOTE: wait until the service is stable and there's deployment example available

done in PR https://github.com/owncloud/ocis/pull/9442

PR #11162, opened 2024-07-09
Configure ocis collaboration service in docker compose

Salipa-Gurung:
## Description
Set up collaboration service (wopi server provided by oCIS) in docker compose.

## Related Issue
- Part of https://github.com/owncloud/web/issues/10960

## How Has This Been Tested?
- locally

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -163,76 +163,55 @@ services:
       traefik.http.routers.companion.entrypoints: ocis
       traefik.http.services.companion.loadbalancer.server.port: 3020
 
-  ocis-appprovider-collabora:
+  collaboration:
     image: ${OCIS_IMAGE:-owncloud/ocis-rolling:master}
-    command: app-provider server
-    environment:
-      APP_PROVIDER_LOG_LEVEL: error
-      # use the internal service name of the gateway
-      REVA_GATEWAY: ${REVA_GATEWAY:-com.owncloud.api.gateway}
-      # STORAGE_GATEWAY_ENDPOINT: host.docker.internal:9142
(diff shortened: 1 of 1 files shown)


dschmidt (commented):
If I understood Michael correctly this should replace the old wopiserver, correct? Then that should be removed.
Also we have Collabora and OnlyOffice in the docker compose setup a collaboration service should be used for each probably?

docker-compose.yml
```diff
@@ -224,6 +224,37 @@ services:
       - host.docker.internal:${DOCKER_HOST:-host-gateway}
     restart: always
 
+  collaboration:
```
saw-jan, line 227: we also need separate service configs for onlyoffice and collabora. And we have to think how to do it. via `.env` or `profiles`

saw-jan:
> If I understood Michael correctly this should replace the old wopiserver, correct? Then that should be removed. Also we have Collabora and OnlyOffice in the docker compose setup a collaboration service should be used for each probably?

yeah, that is true.

saw-jan:
>Also we have Collabora and OnlyOffice in the docker compose setup a collaboration service should be used for each probably?

seems like we can use both office suites at the same time by starting two different collaboration services on different ports (`COLLABORATION_GRPC_ADDR` and `COLLABORATION_HTTP_ADDR`)

saw-jan (approved):
LGTM

ScharfViktor (approved):
awesome. Thank you 👍

Issue #11167, opened 2024-07-09
Listing shares always shows the current user as creator of the share


JammingBen:
## Describe the bug
The sharing information in the right sidebar always displays the current user as creator of the share, even if the share has been created by some other user.

## Steps to reproduce
1. Create a space as "admin", invite "einstein" as space manager
2. Create a resource inside the space
3. Share this resource with "marie"
4. Log in as "einstein", head to the file and open the right sidebar
5. Check the information about the share by clicking "Access details" in the context menu of the share

## Expected behavior
"admin" is being listed as the creator of the share.

## Actual behavior
The current user (= "einstein") is being listed as the creator.

Note that we currently can't display the correct creator because of https://github.com/owncloud/ocis/issues/9571.

PR #11171, opened 2024-07-10
refactor: make graph abstraction layer useful for permissions

JammingBen:
## Description
Refactors the graph abstraction layer for permissions (= shares) to make them more practical and convenient to use.

The new implementation has improved method parameters, types and destructures API responses so users of the abstraction don't need to care about such things.

## Related Issue

- fixes https://github.com/owncloud/web/issues/10808
- refs https://github.com/owncloud/web/issues/11167

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-client/src/graph/permissions/permissions.ts
+++ b/packages/web-client/src/graph/permissions/permissions.ts
@@ -1,0 +1,232 @@
+import { buildCollaboratorShare, buildLinkShare, CollaboratorShare, LinkShare } from '../../helpers'
+import {
+  CollectionOfPermissionsWithAllowedValues,
+  DrivesPermissionsApiFactory,
+  DrivesRootApiFactory,
+  Permission,
+  RoleManagementApiFactory,
+  UnifiedRoleDefinition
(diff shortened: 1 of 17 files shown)


packages/web-client/src/graph/permissions/permissions.ts
```diff
@@ -0,0 +1,232 @@
...
+
+      return { shares, allowedActions, allowedRoles }
+    },
+
+    async updatePermission<T extends CollaboratorShare | LinkShare>(
+      driveId: string,
+      itemId: string,
+      permId: string,
+      data: Permission,
+      isSpaceRoot: boolean,
```
kulmann, line 93: do we need the `isSpaceRoot` prop? Can't we derive that from `driveId === itemId`?

JammingBen: Seems to work 👍

kulmann (approved):
Awesome 😍

Issue #11072, opened 2024-06-19
download in editor context menu downloads last saved state (what about dirty state?)


kulmann:
# Description
For downloads from within an editor it could be perceived as a bug, that the last server side known state is downloaded while changes to the open document don't appear in the download. We should save the document before downloading it.

## User Stories

* > As a user who currently edits a markdown file, I want changes to my document being saved before downloading it, so that I don't accidentally have an "old" state of the document on my local disk.

## Value
UX

## Acceptance Criteria
- if a document is in a dirty state (we already detect that in the AppWrapper... is the solution sufficient?!):
  - auto-save the file
  - wait until auto-save is finished
  - download the file
- technical detail: since we only track the dirty state within the AppWrapper, we should wrap the handler of the download action with the check if the document is in a dirty state and trigger the auto-save from there.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

tbsbdr:
I'd keep it simple and save implicitly before download. if the save is unwanted, the user can still go back in versions.
lets keep it simple, no risk no fun, quod licet Iovi, non licet bovi.

kulmann:
> if the save is unwanted, the user can still go back in versions.

... well, that's not true for share receivers, because versions are not exposed in shares. ;-) But anyway, fine by me. I'll change the acceptance criteria accordingly.

tbsbdr:
ok, lets try. if accidents happen, we can adjust it and add a safety net dialog.

PR #11174, opened 2024-07-10
Save file in app when dowload gets triggered, so the file eq the current state

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/11072

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
+++ b/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
@@ -466,6 +466,22 @@ export default defineComponent({
       }
     })
 
+    const originalDownloadFileActionHandler = unref(downloadFileActions)[0].handler
+
+    /**
+     * The interceptor is used to save the file automatically when in dirty state,
+     * so the downloaded file equals the current state
+     */
+    const downloadFileActionInterceptor = async (args: FileActionOptions) => {
+      if (unref(isDirty)) {
+        await save()
+        autosavePopup()
+      }
+      originalDownloadFileActionHandler(args)
+    }
+
+    unref(downloadFileActions)[0].handler = downloadFileActionInterceptor
+
     const menuItemsContext = computed(() => {
       return [...unref(openWithAppActions), ...unref(fileActionsSave)].filter((item) =>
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -466,6 +466,22 @@ export default defineComponent({
...
+     */
+    const downloadFileActionInterceptor = async (args: FileActionOptions) => {
+      if (unref(isDirty)) {
+        await save()
+        autosavePopup()
+      }
+      originalDownloadFileActionHandler(args)
+    }
+
+    unref(downloadFileActions)[0].handler = downloadFileActionInterceptor
```
JammingBen, line 483: Overwriting the handler of the action feels a bit hacky to me 🤔 Can't you just add a method that checks for the dirty state and then calls `unref(downloadFileActions)[0].handler`?

AlexAndBear: I don't get it, we need to save on download, how to "just add a method" ?

JammingBen: I thought the component calls `unref(downloadFileActions)[0].handler` directly, but that doesn't seem to be the case. Hmm. Then I don't have a better idea...

JammingBen (approved):

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
dschmidt: It's not suuuper nice, but I don't think it's too bad. Fine for me.

The only issue I have with this, is that you `unref` in the setup function - the current code will break when the computed in the composable is reevaluated and there's a new action object.
This needs to live inside a watcher to make it more resilient - I'm sorry 😅

dschmidt (approved):
LGTM :)

Issue #9786, opened 2023-10-10
Id-based WebDAV requests


JammingBen:
Currently, all of our WebDAV requests are path based. In the long term we want to have this id-based, which improves performance on the server side and lets us get rid of the path handling, which is quite laborious most of the time.

The WebDAV implementation needs to be refactored to the `web-client` package before this can happen (see https://github.com/owncloud/web/pull/9764).

JammingBen:
Blocked because of https://github.com/owncloud/ocis/issues/9618

dj4oC:
Before this can be acted on, the following needs to be resolved:

- This issue is explicitly blocked by `owncloud/ocis#9623` and `owncloud/ocis#9618`, with no update since 2024 on whether those have been resolved.

Could someone confirm the current status of those two ocis issues? If still open, this remains blocked and not actionable here.

🤖 Automated by Claude Code

---
_Generated by [Claude Code](https://claude.ai/code)_

DeepDiver1975:
This repository is being archived and is no longer maintained. Closing this as part of archival. Thank you for your contribution.

Issue #7799, opened 2022-10-14
Breadcrumbs don't resolve conflicting spaces / are only path based


kulmann:
Followup of https://github.com/owncloud/web/pull/7725

Breadcrumbs are currently built from parent paths segments (i.e. only path based, no fileIds present). As a result, the breadcrumbs might resolve to the wrong space if you have two or more spaces with identical drive aliases (drive alias clashes are allowed, we need fileIds to resolve the correct space).

To solve this we need a sequence of propfinds, with each propfind being on the fileId of a folder and requesting the respective parent folder id, until we reached the root of the space.

Side note: we can collect the `shareTypes` of all parent folders as well. This would resolve the issue that we currently request all shares via OCS. Doesn't need to be implemented in the same PR, but should be prepared.

tbsbdr:
imo cornercase but consequences could be really ugly. I need to talk with you about this to give it a proper prio.

kulmann:
Turns out: the breadcrumbs work correct without the `fileId` because we already have a resolved space (from `DriveResolver.vue` component / `useDriveResolver` composable) and when the drive alias doesn't change we don't try to re-resolve the space. We only extract the updated relative item path. The `fileId` gets added back to the URL (because we have a mechanism in place that does that). Definitely not a GA blocker, everything works. Just could be "more precise" with the fileId already being in the route to the parent folder. No other added value.

kulmann:
We have the `ancestorMetaData` available in the resource pinia store by now. That is all the information we need (a fileId for each and every parent folder up to the space root). The function which builds breadcrumb items needs to be refactored to take the fileIds into account as well. But as already stated, prio is really really low because it fixes no bug at all.

PR #11180, opened 2024-07-11
[wip] id-based webdav requests

JammingBen:
## Description
Introduce id-based webdav requests. This is the recommended way to work with the webdav API according to oCIS because it delivers better performance, especially with slower storages.

This is quite a big PR that tries to gather all relevant changes. The plan is to divide it into smaller parts and create separate PRs. I'm not 100% sure of we can (or should) fully switch to id-based webdav requests yet. But there are definitely parts of it that make sense to do now.

## Related Issue

- refs https://github.com/owncloud/web/issues/9786
- refs https://github.com/owncloud/web/issues/7799

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


JammingBen:
Closing since we'll never merge this PR. It's only a reference for me what still needs to be done on the road to id-based webdav requests.

PR #11185, opened 2024-07-12
[full-ci] enable secure view test

ScharfViktor:
- [x] enabled secure view tests
- [x] user opens different type of files in the collabora secure mod 
- [x] enabled collabora the secure view in the docker-compose and in the drone.star
- [ ] update collabora to version collabora/code:24.04.5.1.1 - doesn't work. We cannot open file in the collabora because `SSL verification warning (handshake): self-signed certificate| net/SslSocket.hpp:448`


AlexAndBear (approved):
🤩

PR #11190, opened 2024-07-15
feat: add file id support to webdav client

JammingBen:
## Description
Adds support for file ids to the webdav client. This currently excludes delete and restore operations because the server lacks their support at the moment.

Id-based webdav operations perform much better in general since the server doesn't need to traverse file paths.

Note that all webdav operations using a file id won't return any path information. Meaning if you need to work with a resource path, it's recommended to keep providing a path to the client.

The implementation prioritizes a path over a file id if both parameters are given.

## Related Issue

- Works towards https://github.com/owncloud/web/issues/9786

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


JammingBen:
@kulmann @dschmidt What do you think about the implicit behavior that `path` takes priority over a `fileId` if both params are given? An alternative would be to have separate factories for id-based operations, like the one we had for listing files. But that would be a lot of code duplication in the end.

kulmann:
> @kulmann @dschmidt What do you think about the implicit behavior that `path` takes priority over a `fileId` if both params are given?

The implicit behaviour sounds good 👍

kulmann (approved):

PR #11199, opened 2024-07-16
refactor: use ids for copy and move operations

JammingBen:
## Description
As per title.

Also fixes as issue where `listFiles` with only a space provided would fail. The webdav path now falls back to the space webdav path if no `path` or `fileId` are being provided.

## Related Issue

- works towards https://github.com/owncloud/web/issues/9786

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


JammingBen:
This solution currently doesn't work on public pages because we still need to work with paths on those. We could introduce checks and keep supporting paths *and* ids, but I don' really like this.

Maybe https://github.com/owncloud/ocis/issues/9623 gets resolved in the near future, then we can omit paths completely for copy- and move-operations.

JammingBen:
Closing because of https://github.com/owncloud/web/pull/11199#issuecomment-2230878625, this doesn't seem to get attention in the backend.

PR #11201, opened 2024-07-17
refactor: use fileId for breadcrumb navigation

JammingBen:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/7799

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-pkg/src/composables/breadcrumbs/useBreadcrumbsFromPath.ts
+++ b/packages/web-pkg/src/composables/breadcrumbs/useBreadcrumbsFromPath.ts
@@ -6,7 +6,8 @@ import { v4 as uuidv4 } from 'uuid'
 import { SpaceResource } from '@ownclouders/web-client'
 import { urlJoin } from '@ownclouders/web-client'
 import { useGetMatchingSpace } from '../spaces'
-import { Ref, unref } from 'vue'
+import { Ref, ref, unref } from 'vue'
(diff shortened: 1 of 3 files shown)


packages/web-pkg/src/composables/breadcrumbs/useBreadcrumbsFromPath.ts
```diff
@@ -37,7 +44,10 @@ export const useBreadcrumbsFromPath = () => {
         ...(isAccessible && {
           to: {
             path: '/' + [...current].splice(0, current.length - resource.length + i + 1).join('/'),
-            query: omit(route.query, 'fileId', 'page') // TODO: we need the correct fileId in the query. until we have that we must omit it because otherwise we would correct the path to the one of the (wrong) fileId.
```
kulmann, line 40: uhm... don't we still need to omit the `fileId` from the original query? If there is an `ancestor` in the next line you'll add the correct ancestor file id. If there is no ancestor meta data we **must** not provide a `fileId`.

kulmann (changes requested):

packages/web-pkg/src/composables/breadcrumbs/useBreadcrumbsFromPath.ts
JammingBen: Yep, good catch 👍

kulmann (approved):
nice 👍

Issue #11207, opened 2024-07-17
Thumbnails are displayed in some places and in some places not


ScharfViktor:
web 6.1.0+0b8559f91
ocis 9.1.0

Thumbnails are not displayed on the:
 -  /files/spaces 
 - /files/search pages

Thumbnails are displayed on the:
- /files/shares

Steps: 
1. `marie` uploads png files to the personal or project space
2. `marie` opens personal space `https://host.docker.internal:9200/files/spaces/personal/marie?fileId=uuid`

Actual: no thumbnails -> no GET preview request

3. `marie` clicks by `Details` 

Actual: can see file preview

4. `marie` shares png file to admin
5. `admin` goes to shares with me page

Expected: user can see thumbnails in the /files/spaces and files/search pages too. 
not sure about search result 
<img width="503" alt="Screenshot 2024-07-17 at 15 59 01" src="https://github.com/user-attachments/assets/b580995a-5deb-48b4-ad8a-366bd2a2d5ea">

Actual: admin can see thumbnails in the /files/shares page

https://github.com/user-attachments/assets/fe37c994-d173-4b1f-a130-95101c0c23a9

AlexAndBear:
@ScharfViktor it behaves complete different for me, could you check the config.json in the network tab and look up the value for disablePreviews ?

ScharfViktor:
> @ScharfViktor it behaves complete different for me, could you check the config.json in the network tab and look up the value for disablePreviews ?

I found it https://github.com/owncloud/web/blob/master/dev/docker/ocis.web.config.json#L13

Then why is it enabled on the "Share with Me" page?

I guess it behaves complete different for you because you use vite and here we enable preview: https://github.com/owncloud/web/blob/master/vite.config.ts#L81

AlexAndBear:
Yup it should not behave different between shared with me and personal space page (some vars are missing).
I would leave the ticket open, and try to make the behavior equal for all views. (assigning myself) 

I wonder why we have the option to disable thumbnails in general @kulmann any idea?

AlexAndBear:
@ScharfViktor can you check if everything is working now as expected: 
https://github.com/owncloud/web/pull/11209

I am not on a computer rn

AlexAndBear:
Discussed with @kulmann we will remove all of this restriction and configuration stuff

PR #11209, opened 2024-07-17
Remove config options "disablePreviews", "displayThumbnails" and "previewFileMimeTypes"

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/11207

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


AlexAndBear:
@ScharfViktor I have the gut feeling, that this makes the e2e tests flaky. 
Maybe these are API call checks or something,
any idea ?

ScharfViktor:
> @ScharfViktor I have the gut feeling, that this makes the e2e tests flaky.

I see this step `await sidebar.openPanel({ page, name: '' })` started failing often. if we fix it in one place - we fix all tests because a lot of the tests use it 

I'll try to reproduce it localy and fix

kulmann:
We found one more: `options.previewFileMimeTypes` can be hardcoded, because we now only have `ocis` as backend and know which mime types are supported.

ScharfViktor:
e2e tests have 2 problem
- 1. new Uncaught error
image
- 2. existing flaky (not releated this PR) 
image

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -864,7 +864,7 @@ export default defineComponent({
       return this.clipboardResources.some((r) => r.id === resource.id)
     },
     shouldDisplayThumbnails(item: Resource) {
-      return this.areThumbnailsDisplayed && !isResourceTxtFileAlmostEmpty(item)
+      return !isResourceTxtFileAlmostEmpty(item)
```
kulmann, line 867: What do you think about dropping this whole function as well? While I get the intention, I still always think that it might be confusing at the same time that sometimes .txt files have a preview/thumbnail and sometimes they don't. Yes, the thumbnail is a white box if the text file is empty. But well, so be it... I'm in favour of reducing complexity / confusion and remove the helper and its usages entirely. Would fit in this PR...

AlexAndBear: Currently idk, it seems that the server is throwing an error, when you request a preview on an empty thumbnail which triggers, the CI to fail.
Should I proceed ?

kulmann (approved):
🏅

Issue #11215, opened 2024-07-18
Check if config options `topCenterNotifications` and `displayResourcesLazy` can be dropped


kulmann:
## Description

We currently set
```json
"topCenterNotifications": true,
"displayResourcesLazy": false,
```
in the CI config. They were introduced back in the nightwatch based acceptance test days...

Please check if they are still needed by making a PR which flips both booleans. If not needed anymore or effort low to get rid of them, let's get rid of them.

PR #11220, opened 2024-07-18
[full-ci]check e2e test without redundant config

ScharfViktor:
related https://github.com/owncloud/web/issues/11215

run e2e test in the CI without: 
```
"topCenterNotifications": true,
"displayResourcesLazy": false,
```

### DIFF
--- a/dev/docker/ocis.web.config.json
+++ b/dev/docker/ocis.web.config.json
@@ -9,9 +9,7 @@
     "scope": "openid profile email"
   },
   "options": {
-    "topCenterNotifications": true,
     "disablePreviews": true,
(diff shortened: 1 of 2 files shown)


ScharfViktor:
@kulmann e2e tests work fine without these envs
we can delete it

kulmann:
> @kulmann e2e tests work fine without these envs
> 
> we can delete it

Awesome, thank you. What about duration of the test runs, are they similar with and without?

ScharfViktor:
> Awesome, thank you. What about duration of the test runs, are they similar with and without?

similar time before and after

PR #11221, opened 2024-07-18
refactor: setting space image and readme data id-based

JammingBen:
## Description
Refactors setting/updating space image and readme data to work id-based. The big difference is that we need to retrieve the id of the `.space` meta folder so it can be used for the requests.

Also fixes an issue in scenarios where we want to list files for a space root where no `fileId` and no `path` are being provided. We need to fall back to the space id in such case.

## Related Issue

- works towards https://github.com/owncloud/web/issues/9786

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-pkg/src/composables/actions/spaces/useSpaceActionsEditReadmeContent.ts
+++ b/packages/web-pkg/src/composables/actions/spaces/useSpaceActionsEditReadmeContent.ts
@@ -18,6 +18,7 @@ export const useSpaceActionsEditReadmeContent = () => {
   const { getDefaultMetaFolder } = useSpaceHelpers()
 
   const createReadme = async (space: SpaceResource, metaFolder: Resource) => {
+    // FIXME: remove path as soon as we make the full switch to id-based dav requests
     const markdownResource = await clientService.webdav.putFileContents(space, {
       path: '.space/readme.md',
       parentFolderId: metaFolder.id,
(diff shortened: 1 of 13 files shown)


packages/web-pkg/src/composables/actions/spaces/useSpaceActionsEditReadmeContent.ts
```diff
@@ -15,36 +15,45 @@ export const useSpaceActionsEditReadmeContent = () => {
...
-    try {
-      await clientService.webdav.getFileInfo(resources[0], { path: '.space' })
-    } catch (_) {
-      await createDefaultMetaFolder(resources[0])
-      markdownResource = await clientService.webdav.putFileContents(resources[0], {
-        path: '.space/readme.md'
-      })
+  const createReadme = async (space: SpaceResource, metaFolder: Resource) => {
+    const markdownResource = await clientService.webdav.putFileContents(space, {
+      path: '.space/readme.md',
```
JammingBen, line 22: We need to keep providing the path here because the resource is being used afterwards and we might need its path information. Hence this needs to wait until we make the full switch to id-based requests.

kulmann:
> [CLA assistant check](https://cla-assistant.io/owncloud/web?pullRequest=11221) All committers have signed the CLA.

@JammingBen where's this coming from? F**ked up rebase? 😅

JammingBen:
> > [CLA assistant check](https://cla-assistant.io/owncloud/web?pullRequest=11221) All committers have signed the CLA.
> 
> @JammingBen where's this coming from? F**ked up rebase? 😅

Originally, yes 😄

packages/web-pkg/src/composables/actions/spaces/useSpaceActionsEditReadmeContent.ts
kulmann: Can you add a `FIXME`? Otherwise it might be confusing that both `path` and the tuple of `parentFolderId` and `fileName` is provided.

kulmann (approved):

PR #11224, opened 2024-07-18
Purge config options topCenterNotifications and displayResourcesLazy

AlexAndBear:
## Related Issue

- Fixes [issues/11215](https://github.com/owncloud/web/issues/11215)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTable.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTable.vue
@@ -25,7 +25,7 @@
     :selection="selectedResources"
     :sort-by="sortBy"
     :sort-dir="sortDir"
-    :lazy="lazyLoading"
+    :lazy="lazy"
     :grouping-settings="groupingSettings"
     padding-x="medium"
     @highlight="fileClicked"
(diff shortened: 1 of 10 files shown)


packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -484,6 +484,12 @@ export default defineComponent({
     resourceType: {
       type: String as PropType<'file' | 'space'>,
       default: 'file'
+    } /**
+     * Determines if the table content should be loaded lazily.
+     */,
+    lazy: {
```
AlexAndBear, line 490: still needed for unit tests, that wont work with lazy loading

packages/web-runtime/src/components/MessageBar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <oc-notifications :position="notificationPosition">
+  <oc-notifications position="default">
```
kulmann, line 2: Instead of having the `position` prop here at all you could change the default value of the prop to `default` in the `OcNotifications.vue` component. Kind of ridiculous that `OcNotifications` has `top-center` as default while `default` is a valid value. 😆

AlexAndBear: hahah, yess I was baffled :D

kulmann (approved):
💪

PR #11240, opened 2024-07-19
Dont load config routing from config.json, instead set based on cernFeatures option

### LINKED ISSUE #11216: Remove config option `option.routing`
## Describe the bug
Nobody wants to set `options.routing.idBased` to false, because that would only lead to shooting yourself in the foot. We can't drop the feature flag at the moment. But let's drop the config option and instead disable id based routing if `cernFeatures` is set to `true`.

Remember to also remove the config option `options.routing` in the ocis config struct.

AlexAndBear:
## Related Issue

- Fixes #11216 
- ## Motivation and Context

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/composables/piniaStores/config/config.ts
+++ b/packages/web-pkg/src/composables/piniaStores/config/config.ts
@@ -32,7 +32,8 @@ const defaultOptions = {
   },
   openAppsInTab: false,
   routing: {
-    idBased: true
+    idBased: true,
+    fullShareOwnerPaths: false
   },
   runningOnEos: false,
   sidebar: {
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/composables/piniaStores/config/config.ts
```diff
@@ -79,7 +81,10 @@ export const useConfigStore = defineStore('config', () => {
     theme.value = data.theme
 
     if (data.options) {
-      options.value = merge({ ...defaultOptions }, data.options)
+      // routing will not be loaded from config, but set based on cernFeatures option
+      options.value = merge({ ...defaultOptions }, omit(data.options, ['routing']))
```
AlexAndBear, line 85: maybe the omit is not needed, just a security layer, so that routing will not be loaded at all via config

packages/web-pkg/src/composables/piniaStores/config/config.ts
```diff
@@ -29,7 +30,8 @@ const defaultOptions = {
   },
   openAppsInTab: false,
   routing: {
-    idBased: true
+    idBased: true,
+    fullShareOwnerPaths: false
```
AlexAndBear, line 34: Is the default correct, there is no documentation about it anywhere

kulmann: It's optional and ocis doesn't set it, so it defaults to `false`. Your chosen default is correct :-)

kulmann (approved):

Issue #11259, opened 2024-07-23
Refactor apps to use the appMenuExtensionPoint


JammingBen:
https://github.com/owncloud/web/pull/11258 deprecated the `applicationMenu` property of the app info object. Instead, the `appMenuExtensionPoint` should be used to add items to the app menu.

Apps that had `openAsEditor` set to `true` need to implement a `handler` on their own, the runtime won't provide any logic for opening editors that way in the future.

JammingBen:
Fixed via the linked PRs.

PR #11273, opened 2024-07-25
refactor: make use of menu items extension point

JammingBen:
## Description
Refactors apps to use the new `appMenuItem` extension point instead of the deprecated `applicationMenu` property.

Also deprecates 2 unused app info properties: `type` and `isFileEditor`.

## Related Issue

- refs https://github.com/owncloud/web/issues/11259

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-ocm/src/extensions.ts
+++ b/packages/web-app-ocm/src/extensions.ts
@@ -1,19 +1,22 @@
 import {
+  ApplicationInformation,
   FileActionOptions,
   useClientService,
   useConfigStore,
   useMessages,
+  useUserStore,
   useWindowOpen
 } from '@ownclouders/web-pkg'
 import { useGettext } from 'vue3-gettext'
 import { computed } from 'vue'
 import { Extension } from '@ownclouders/web-pkg'
-import { OCM_PROVIDER_ID } from '@ownclouders/web-client'
+import { OCM_PROVIDER_ID, urlJoin } from '@ownclouders/web-client'
 
-export const extensions = () => {
+export const extensions = (appInfo: ApplicationInformation) => {
   const { showErrorMessage } = useMessages()
   const clientService = useClientService()
   const configStore = useConfigStore()
+  const userStore = useUserStore()
   const { $gettext } = useGettext()
   const { openUrl } = useWindowOpen()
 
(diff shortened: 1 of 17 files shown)


AlexAndBear:
Looks like science mesh is now enabled by default, and as well is wrongly the first item:
image

packages/web-app-ocm/src/extensions.ts
```diff
@@ -69,6 +72,18 @@ export const extensions = () => {
...
+    },
+    ...((userStore.user && [
+      {
+        id: appInfo.id,
+        type: 'appMenuItem',
+        label: () => appInfo.name,
+        color: appInfo.color,
+        icon: appInfo.icon,
+        priority: 10,
+        path: urlJoin(appInfo.id)
```
dschmidt, line 84: Okay for now for me, but I somewhat dislike having to define the same name, color, icon multiple times - I assume most of the time falling back to the global definition for the app would work as well

What do you think?

I think I've built this kind of fallback mechanism for file icons that are falling back to the app icon...

AlexAndBear:
One additional question, it was possible to open drawio documents via the appMenu as well, how is this handled now, since not anymore living in this repo ?

packages/web-app-ocm/src/extensions.ts
```diff
@@ -69,6 +72,18 @@ export const extensions = () => {
         componentType: 'button',
         class: 'oc-files-actions-open-file-remote'
       }
-    }
+    },
+    ...((userStore.user && [
+      {
+        id: appInfo.id,
```
dschmidt, line 78: Shouldn't the id be a more "classic" extension id?

dschmidt (approved):
LGTM 🚀

AlexAndBear (changes requested):
Make science mesh not the first item

JammingBen:
> One additional question, it was possible to open drawio documents via the appMenu as well, is this still workling, since not anymore living in this repo ?

Yes, this still works (and it will get the same treatment: https://github.com/owncloud/web-extensions/pull/18).

packages/web-app-ocm/src/extensions.ts
JammingBen: As discussed: let's leave it as it is for now. It's not that trivial to build a fallback mechanism here because we don't know which app registers which menu item.

kulmann (approved):
Nice 👍 

Since you're removing the applicationMenu appInfo prop handling, can we have a changelog item for a breaking change for devs?

I'm still wondering if we do ourselves a favour allowing to define a priority. Why don't we sort the app switcher items lexicographically? 🤔

PR #11284, opened 2024-07-26
refactor: maintain loading and error state per file

kulmann:
## Description
I noticed that if preloaded images failed loading then we wouldn't show an error state for the image, but we'd have a broken `` instead. So I refactored the loading- and error-states to be individual for each file. And then I refactored more. And more... there's more to refactor (options-api to composition api, putting data loading into a composable, etc), but I don't want to spend more time on it now. Hope the current state is an improvement over current master.

This can be tested easily by starting ocis with `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE=500KB` and then upload some images with 100kb, some with 2mb and then cycle through the folder within the preview app. :-)

This also now makes sure that the file navigation controls are visible on files with loading- or error-state. Previously the controls were only visible for rendered files.

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-preview/src/App.vue
+++ b/packages/web-app-preview/src/App.vue
@@ -1,29 +1,35 @@
 <template>
-  <div v-if="loading" class="oc-width-1-1">
+  <div v-if="isFolderLoading" class="oc-width-1-1">
     <div class="oc-position-center">
       <oc-spinner :aria-label="$gettext('Loading media file')" size="xlarge" />
     </div>
   </div>
-  <oc-icon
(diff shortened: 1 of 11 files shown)


JammingBen (approved):
Awesome! 😍 

One tiny quirk I found (not sure if worth fixing though): the navigation buttons are obviously located differently for broken images. So you can't comfortably click through all images while resting at the same location with your mouse.

Also, CI complains about some failed type checks.

PR #11292, opened 2024-07-29
[full-ci] chore: prepare release v10.0.0

kulmann:


changelog/10.0.0_2024-07-29/change-remove-application-menu
```diff
@@ -0,0 +1,8 @@
+Change: Remove applicationMenu app property
+
+BREAKING CHANGE for developers: We've removed the `applicationMenu` app property in favour of the new `appMenuItem` extension type.
```
JammingBen, line 3: We just deprecated the prop, does that mean we can actually remove it now? 🤔

JammingBen (approved):
👍 

We should probably release a `10.0.0` of `web-client` and `web-pkg` as well now, no?

changelog/10.0.0_2024-07-29/change-remove-application-menu
kulmann: You removed all the handling in the runtime 🤔

JammingBen: No I didn't 😅   I just ported our apps, but the fallback is still active: https://github.com/owncloud/web/blob/renovate/npm-run-all-replacement/packages/web-runtime/src/components/Topbar/TopBar.vue#L186

kulmann: ah... hm. sorry 😅 removed the changelog item again.

PR #11301, opened 2024-07-30
[full-ci]collabora/code bump version to 24 in dev setup, add needed extra param

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1550,7 +1550,7 @@ def collaboraService():
             "detach": True,
             "environment": {
                 "DONT_GEN_SSL_CERT": "set",
-                "extra_params": "--o:ssl.enable=false --o:ssl.ssl_verification=false --o:ssl.termination=true --o:welcome.enable=false --o:net.frame_ancestors=https://ocis:9200",
+                "extra_params": "--o:ssl.enable=true --o:ssl.termination=true --o:welcome.enable=false --o:net.frame_ancestors=https://ocis:9200",
             },
             "commands": [
                 "coolconfig generate-proof-key",
--- a/.drone.star
+++ b/.drone.star
@@ -1,6 +1,6 @@
 ALPINE_GIT = "alpine/git:latest"
 APACHE_TIKA = "apache/tika:2.8.0.0"
-COLLABORA_CODE = "collabora/code:24.04.5.1.1"
+COLLABORA_CODE = "collabora/code:23.05.6.5.1"
 CS3ORG_WOPI_SERVER = "cs3org/wopiserver:v10.3.0"
 KEYCLOAK = "quay.io/keycloak/keycloak:24.0.1"
 MINIO_MC = "minio/mc:RELEASE.2021-10-07T04-19-58Z"
--- a/.drone.star
+++ b/.drone.star
@@ -1,6 +1,6 @@
 ALPINE_GIT = "alpine/git:latest"
 APACHE_TIKA = "apache/tika:2.8.0.0"
-COLLABORA_CODE = "collabora/code:23.05.6.5.1"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -202,11 +202,11 @@ services:
     restart: always
 
   collabora:
-    image: collabora/code:23.05.10.1.1
+    image: collabora/code:24.04.5.1.1
     command: ['bash', '-c', 'coolconfig generate-proof-key ; /start-collabora-online.sh']
     environment:
       DONT_GEN_SSL_CERT: YES
-      extra_params: --o:ssl.enable=false --o:ssl.termination=true --o:welcome.enable=false --o:net.frame_ancestors=${OCIS_URL:-https://host.docker.internal:9200}
+      extra_params: --o:ssl.enable=false --o:ssl.ssl_verification=false --o:ssl.termination=true --o:welcome.enable=false --o:net.frame_ancestors=${OCIS_URL:-https://host.docker.internal:9200}
```
micbar, line 209: Could we use the value from OCIS_INSECURE?

AlexAndBear: I am not sure about this, we would need to negotiate it, and the other extra params, doesn't it add to much confusion ?

ScharfViktor:
I tested it on deployments->ocis_full. Now it works fine 👍 

if everything is okay I can create PR to solve this issue in ocis https://github.com/owncloud/ocis/compare/addNeededCollaboraExtraParam?expand=1

docker-compose.yml
AlexAndBear: Checked: it's not easy to negotiate it, we would need to write bash script to the docker-compose, which would lead to spaghetti-code, so sorry but no :( :( :(

ScharfViktor:
I updated collabora in CI. let see what says our tests

.drone.star
```diff
@@ -1550,7 +1550,7 @@ def collaboraService():
             "detach": True,
             "environment": {
                 "DONT_GEN_SSL_CERT": "set",
```
AlexAndBear, line 1552: is it ok, that the value is different, that the value in the docker compose ?

ScharfViktor: I don't think it's related, but we could try changing it to “yes”.

micbar (approved):

.drone.star
ScharfViktor: if we use "extra_params":`--o:ssl.enable=false` -> we need to switch using `http` so https://github.com/owncloud/web/pull/11301/files#diff-6a968d4b7901c8526c8d9a86d1f434737dbfe2b00da21a7d3041ed4e3d58d62eL1598-L1599
should be: 
```
environment["COLLABORATION_APP_ADDR"] = "http://collabora:9980"
        environment["COLLABORATION_APP_ICON"] = "http://collabora:9980/favicon.ico"
```
right?

AlexAndBear: I don't think we need all those changes, just bump code version in drone and hope for the best 😮

ScharfViktor: I'll try. if it doesn't work, we'll do it in a separate PR

ScharfViktor: Screenshot 2024-07-30 at 15 53 14

tests are green due the new popup. I'll look how disable popup in the separate PR

ScharfViktor (approved):

.drone.star
AlexAndBear: Whooopsie. Thanks for the engagement 👍

PR #11302, opened 2024-07-30
feat: app store app

kulmann:
## Description
New app store app

## Related Issue

- submarine

## Motivation and Context
Make Web great again

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open Tasks
- [x] tiles view
  - [x] show tags
- [x] app details page
  - [x] screenshots
  - [x] long description
  - [x] full version history
- [x] search


packages/web-app-app-store/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    })
+
+    return {
+      appInfo,
+      routes,
+      extensions,
+      ready: () => {
+        const appsStore = useAppsStore()
+        return appsStore.loadApps()
+      }
```
dschmidt, line 97: Why the ready hook? Why not ln AppsList?

packages/web-app-app-store/src/piniaStores/apps.ts
```diff
@@ -0,0 +1,37 @@
+import { defineStore } from 'pinia'
+import { App, RawAppListSchema } from '../types'
+import { ref } from 'vue'
+import { APPID } from '../index'
+import { useRepositoriesStore } from './repositories'
+
+export const useAppsStore = defineStore(`${APPID}-apps`, () => {
```
dschmidt, line 7: Be careful not to introduce circular import issues here

packages/web-app-app-store/src/piniaStores/apps.ts
```diff
@@ -0,0 +1,37 @@
...
+  const apps = ref<App[]>([])
+
+  const repositoriesStore = useRepositoriesStore()
+
+  const loadApps = async () => {
+    for (const repo of repositoriesStore.repositories) {
+      try {
+        const data = await fetch(repo.url)
+        const appsListData = await data.json()
```
dschmidt, line 16: Could be parallelized

dschmidt (commented):
🚀

packages/web-app-app-store/src/index.ts
kulmann: Because we need the apps loaded in every view/route. A wrapper for all views with `loadApps` in the `onMounted()` hook would be better probably...

kulmann: introduced a `LayoutContainer` component which wraps both the `list` and the `details` views and takes care of loading the apps before rendering the views.

packages/web-app-app-store/src/index.ts
```diff
@@ -0,0 +1,110 @@
...
+import { AppStoreConfigSchema } from './types'
+import { APPID } from './appid'
+export default defineWebApplication({
+  setup({ applicationConfig }) {
+    const { $gettext } = useGettext()
+    const { can } = useAbility()
+    const userStore = useUserStore()
+    const repositoryStore = useRepositoriesStore()
+
+    const defaultRepositories = [
```
JammingBen, line 22:
```suggestion
    const defaultRepositories: AppStoreRepository[] = [
```

packages/web-app-app-store/src/views/AppList.vue
```diff
@@ -0,0 +1,95 @@
...
+      return searchEngine.search(filterTerm).map((r) => r.item)
+    }
+    const filteredApps = computed(() => {
+      // TODO: debounce the filtering by 100-300ms
+      return filter(unref(apps), unref(filterTerm)).sort((a, b) =>
+        a.name.toLowerCase().localeCompare(b.name.toLowerCase())
+      )
+    })
+
+    const markInstance = ref(null)
```
JammingBen, line 63:
```suggestion
    const markInstance = ref<Mark>(null)
```

JammingBen (commented):
Love it! 😍 

What do you think about debouncing the initial loading state a bit? I always see this blue box showing up for a split-second.

packages/web-app-app-store/src/views/AppList.vue
```diff
@@ -0,0 +1,95 @@
...
+      markInstance.value = new Mark('.mark-element')
+    })
+    watch(filterTerm, () => {
+      unref(markInstance)?.unmark()
+      if (unref(filterTerm)) {
+        unref(markInstance)?.mark(unref(filterTerm), {
+          element: 'span',
+          className: 'mark-highlight'
+        })
+      }
```
AlexAndBear, line 75: we have a little problem here, mark-highlight has a  font-weight of 600, which means that the app title will be smaller when 'marked'
image

AlexAndBear (changes requested):

AlexAndBear:
The app details view looks a little empty, I would suggest smth like that:

image

image

packages/web-app-app-store/src/views/AppList.vue
```diff
@@ -0,0 +1,95 @@
...
+    <h2 class="oc-mt-rm" v-text="$gettext('App Store')" />
+    <div class="oc-flex oc-flex-middle">
+      <oc-text-input
+        id="apps-filter"
+        v-model.trim="filterTerm"
+        :label="$gettext('Search')"
+        :clear-button-enabled="true"
+        autocomplete="off"
+      />
+    </div>
```
AlexAndBear, line 12: We have moved most of our search terms to the right, I'd keep up with that here as well too

AlexAndBear (changes requested):

packages/web-app-app-store/src/views/AppDetails.vue
```diff
@@ -0,0 +1,89 @@
...
+      </div>
+      <p class="oc-my-rm">{{ app.subtitle }}</p>
+    </div>
+    <div v-if="app.description">
+      <h3>{{ $gettext('Details') }}</h3>
+      <p class="oc-my-s">{{ app.description }}</p>
+    </div>
+    <div v-if="app.tags">
+      <h3>{{ $gettext('Tags') }}</h3>
+      <app-tags :app="app" />
```
AlexAndBear, line 25: somehow the app tags doesn't emit search here, but are buttons

AlexAndBear (changes requested):

packages/web-app-app-store/src/views/AppList.vue
AlexAndBear: Doesnt look good, don't do it

packages/web-app-app-store/src/components/AppTile.vue
```diff
@@ -72,5 +72,12 @@ export default defineComponent({
     justify-content: space-between;
     height: 100%;
   }
+
+  .app-tile-title {
+    .mark-highlight {
+      font-weight: var(--oc-font-weight-extrabold) !important;
```
AlexAndBear, line 78: maybe set to initial ?

AlexAndBear (changes requested):

packages/web-app-app-store/src/components/AppTags.vue
```diff
@@ -0,0 +1,45 @@
...
+
+export default defineComponent({
+  name: 'AppTags',
+  props: {
+    app: {
+      type: Object as PropType<App>,
+      required: true
+    }
+  },
+  emits: ['search'],
```
AlexAndBear, line 27: Why does this component makes assumptions and emit a search event, shouldn't it just emit a click event ? 🤔

AlexAndBear (changes requested):

packages/web-app-app-store/src/components/AppTags.vue
kulmann: fine by me... care to change it? ;-)

dev/docker/ocis/csp.yaml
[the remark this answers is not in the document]
kulmann: went back to using the good old app-loading-spinner component. Showing text content feels just like too complex for a simple loading state.

kulmann:
> What do you think about debouncing the initial loading state a bit? I always see this blue box showing up for a split-second.

Went back to using the good old app-loading-spinner component. Showing text content feels just like too complex for a simple loading state.

packages/web-app-app-store/src/views/AppList.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <div class="app-list oc-mb-m">
+    <h2 class="oc-mt-rm" v-text="$gettext('App Store')" />
```
pascalwengerter, line 3: I'd suggest adding a link to the install guides for apps here? I naively thought I'd install them directly into my instance and was a bit surprised when it "just" downloaded a zip file 😅

kulmann: Phase 1 is `discovery`, phase 2 is `easy install`, we need backend features for phase 2.

Do you have a good idea where to put such a link? It would be helpful indeed...

pascalwengerter: I guess an ℹ️  icon next to the heading with a tooltip that provides a link to https://owncloud.dev/services/web/#loading-applications would suffice for now?

AlexAndBear (approved):
At this point it looks pretty mature to me 👍

I suggested to add a link to our docs, so the user knows how to install a an app. 

Maybe we can add a context helper next to the download buttons or a link with text (How to install) under them. 

Approving anyways

JammingBen (approved):
Super awesome! 🚀

Issue #11264, opened 2024-07-24
[web] Unzip in Web


tbsbdr:
# Description
- Unzip is performed on the user side (not server side), because: 
  - filefirewall should take effect
  - simpler handling of quota-check
  - zip-bombing will not affect the server (only the client)

## User Stories

* > As a user I want to unzip files so that I dont rely on a local unzip app.

## Acceptance Criteria
- new web extension "Unzip" in web-extensions repo
- new context menu item for a compressed file: "Extract here" 
- supported formats: `tar` and `zip`
- standard maximum filesize is 64MB 
- click on "Extract here" extracts all files from the archive to the current location into a new folder with the foldername of the archive ( `archive.zip` ->`archive` )
  - filename conflict: If the new extracted folder already exists, the new foldername gets implicitly an appended `archive.zip` ->`archive (2)`
- If extraction for a file is not available, disable the action and show a tooltip "Exceeds maximum size of 64MB"
- while the unzip is in progress, show the top loading bar
- ❌ Not available: batch actions for unzip
- ❌ Not available: extract password protected archives (follow up story)
  - Show error message "File is password protected. Password protected files can not be extracted."
 

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

## Visual
![screenshot_002939](https://github.com/user-attachments/assets/a695f233-7e82-4517-a6e7-9f0b5c6f344b)

AlexAndBear:
Nice 🙂 

But I wonder why we need a new config variable. Will there be systems or user clients where the admin knows that they can safely set a higher value ?

JammingBen:
This is predestined to run in a web worker since this is most likely a synchronous (=UI blocking) task.

> But I wonder why we need a new config variable. Will there be systems or user clients where the admin knows that they can safely set a higher value ?

[...]

PR #11304, opened 2024-07-31
chore: add new csp rules to allow unzipping files

JammingBen:
## Description
Adds these 2 CSP rules to our dev setup to make unzipping via wasm work:

- `child-src: blob`
- `script-src: wasm-unsafe-eval`

Tbh I'm not sure how "unsafe" the second rule is... but AFAIK it's needed for wasm to work. If that isn't an option, we need to look for a different library for unzipping.

## Related Issue

- Works towards https://github.com/owncloud/web/issues/11264

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/deployments/examples/ocis_web/config/ocis/csp.yaml
+++ b/deployments/examples/ocis_web/config/ocis/csp.yaml
@@ -1,6 +1,7 @@
 directives:
   child-src:
     - '''self'''
+    - 'blob:'
   connect-src:
     - '''self'''
     - 'https://${COMPANION_DOMAIN|companion.owncloud.test}/'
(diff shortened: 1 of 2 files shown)


JammingBen:
@DeepDiver1975 Do you have an opinion about the second CSP rule `script-src: wasm-unsafe-eval`? Is it generally "safe" to use, or should it be avoided?

DeepDiver1975:
from my pov it is as unsafe as `unsafe-inline` and from my understanding both shall be avoided.

I have little understanding if there is a proper way to make things work without `wasm-unsafe-eval`.

Usually there is a way to make things work without specing unsafe CSPs - no idea about wasm.

It might be worth to spend some time to research into this and potentially reach out to the libarchive project.

JammingBen:
Thanks for the feedback! I'll do a little more research then and check potential other libs.

JammingBen:
I've found another lib that seems to be suitable, hence closing here. Let's be on the safe side.

PR #11308, opened 2024-07-31
[full-ci] update collabora in CI

ScharfViktor:
I tried to update collabora in CI here https://github.com/owncloud/web/pull/11301/commits/05049857107d6511666adc83be9552f36d9c24e0

but due new copy-paste warning popup tests were red. The popup window appears after pressing `ctr+C`.
I didn't find how disable this popup so I close it using UI when if it exists


saw-jan (approved):
LGTM

Issue #11204, opened 2024-07-17
Simplify Link sharing dialog


tbsbdr:
# Description

## User Stories

* > As a user I want a simple link sharing dialog so that I can quickly share links.

## Acceptance Criteria
- Dialog Title: "Copy link for "resourcename" "
- Show default persmission of the link eg "Anyone with the link can view" (same wording as in right sidebar -> Link permission)
- Show autogenerated PW
- show ⚙️Options Option, click on option shows:
  - show drop menu for link permission
  - show password field, prefilled with generated password
  - show 3-dots next to `cancel`
 - If a password is set, label of copy link button is "Copy link and password"
   - Link: https://ocis.ocis-web.master.owncloud.works/s/zwAPFgepKGNFfhk 
   - Password: xuK7[R>1q[9B
## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

## Visual
![image](https://github.com/user-attachments/assets/f27d8224-e1f6-4255-867a-37f8aaea85b0)

tbsbdr:
cc @AlexAndBear

PR #11310, opened 2024-07-31
Redesign copy link modal

AlexAndBear:
## Description
image

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11204

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcModal/OcModal.vue
+++ b/packages/design-system/src/components/OcModal/OcModal.vue
@@ -499,10 +499,6 @@ export default defineComponent({
     line-height: 1.625;
     padding: var(--oc-space-medium) var(--oc-space-medium) 0;
 
-    span {
-      color: var(--oc-color-text-default);
-    }
-
     &-message {
       margin-bottom: var(--oc-space-medium);
       margin-top: var(--oc-space-small);
--- a/packages/web-app-files/src/components/Modals/SetLinkPasswordModal.vue
+++ b/packages/web-app-files/src/components/Modals/SetLinkPasswordModal.vue
@@ -92,7 +92,7 @@ export default defineComponent({
       onInput,
       errorMessage,
       passwordPolicyService,
-      inputPasswordPolicy: passwordPolicyService.getPolicy(),
+      inputPasswordPolicy: passwordPolicyService.getPolicy({ enforcePassword: true }),
       inputGeneratePasswordMethod: () => passwordPolicyService.generatePassword(),
 
       // unit tests
--- a/packages/web-app-files/src/components/Modals/SetLinkPasswordModal.vue
+++ b/packages/web-app-files/src/components/Modals/SetLinkPasswordModal.vue
@@ -5,7 +5,6 @@
     type="password"
     :password-policy="inputPasswordPolicy"
     :generate-password-method="inputGeneratePasswordMethod"
-    :fix-message-line="true"
     :placeholder="link.hasPassword ? '●●●●●●●●' : null"
     :error-message="errorMessage"
     class="oc-modal-body-input"
--- a/packages/web-pkg/src/components/CreateLinkModal.vue
+++ b/packages/web-pkg/src/components/CreateLinkModal.vue
@@ -67,8 +67,8 @@
       </oc-button>
       <oc-drop
         v-if="!onlyInternalLinksAllowed"
-        X
         ref="contextMenuDrop"
+        X
         drop-id="link-modal-context-menu-drop"
         toggle="#link-modal-context-menu-toggle"
         padding-size="small"
--- a/packages/web-pkg/src/components/CreateLinkModal.vue
+++ b/packages/web-pkg/src/components/CreateLinkModal.vue
@@ -166,6 +166,11 @@ import LinkRoleDropdown from './LinkRoleDropdown.vue'
 
 type RoleRef = ComponentPublicInstance<typeof OcButton>
 
+interface CallbackArgs {
+  result: PromiseSettledResult<LinkShare>[]
+  password: string
+}
+
 export default defineComponent({
   name: 'CreateLinkModal',
   components: { LinkRoleDropdown },
@@ -175,9 +180,7 @@ export default defineComponent({
     space: { type: Object as PropType<SpaceResource>, default: undefined },
     isQuickLink: { type: Boolean, default: false },
     callbackFn: {
-      type: Function as PropType<
-        (result: PromiseSettledResult<LinkShare>[], password: string) => Promise<void> | void
-      >,
+      type: Function as PropType<(args: CallbackArgs) => Promise<void> | void>,
       default: undefined
     }
   },
@@ -323,7 +326,7 @@ export default defineComponent({
       }
 
       if (props.callbackFn) {
-        props.callbackFn(result, password.value)
+        props.callbackFn({ result, password: password.value })
       }
     }
 
--- a/packages/web-pkg/src/components/CreateLinkModal.vue
+++ b/packages/web-pkg/src/components/CreateLinkModal.vue
@@ -285,11 +285,6 @@ export default defineComponent({
 
     const onConfirm = async () => {
       if (!unref(selectedLinkTypeIsInternal)) {
-        if (unref(passwordEnforced) && !unref(password).value) {
-          password.error = $gettext('Password must not be empty')
-          return Promise.reject()
-        }
-
         if (!passwordPolicy.check(unref(password).value)) {
           password.error = $gettext('Password must fulfill the password policy')
           return Promise.reject()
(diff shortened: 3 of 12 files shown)


packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -499,10 +499,6 @@ export default defineComponent({
     line-height: 1.625;
     padding: var(--oc-space-medium) var(--oc-space-medium) 0;
 
-    span {
-      color: var(--oc-color-text-default);
-    }
-
```
AlexAndBear, line 505: overwrites all other classes, that we may set: e.G  oc-text-muted

packages/web-app-files/src/components/Modals/SetLinkPasswordModal.vue
```diff
@@ -5,7 +5,6 @@
     type="password"
     :password-policy="inputPasswordPolicy"
     :generate-password-method="inputGeneratePasswordMethod"
-    :fix-message-line="true"
```
AlexAndBear, line 8: we dont need an extra spacer

packages/web-pkg/src/services/passwordPolicy/passwordPolicy.ts
```diff
@@ -91,7 +90,7 @@ export class PasswordPolicyService {
     }
     const rules = {} as Record<string, unknown>
 
-    if (this.useDefaultRules()) {
+    if (enforcePassword && !this.hasRules()) {
       rules.mustNotBeEmpty = {}
     }
```
AlexAndBear, line 95: We might have password fields, that don't require a password, that is in conflict with the always added not empty rule

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -53,75 +38,82 @@
           : undefined
       "
       :label="passwordEnforced ? `${$gettext('Password')}*` : $gettext('Password')"
-      class="link-modal-password-input oc-mt-l"
+      class="link-modal-password-input"
       :disabled="selectedLinkTypeIsInternal"
       @update:model-value="updatePassword"
     />
+    <div v-else-if="password.value" class="link-modal-password-text oc-text-small oc-text-muted">
+      <span v-text="$gettext('Password: ')" />
```
JammingBen, line 46: The blank shouldn't be needed because the password is in a new line here in the code.

```suggestion
      <span v-text="$gettext('Password:')" />
```

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -169,20 +161,26 @@ import { Resource } from '@ownclouders/web-client'
 import { formatRelativeDateFromDateTime } from '../helpers'
 import { OcButton, OcDrop } from 'design-system/src/components'
 import { SharingLinkType } from '@ownclouders/web-client/graph/generated'
+import { $gettext } from '../router/utils'
```
JammingBen, line 164: 👀

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -370,11 +387,14 @@ export default defineComponent({
...
       confirmButtonText,
+      advancedMode,
+      setAdvancedMode,
 
       // unit tests
       onConfirm
     }
-  }
+  },
+  methods: { $gettext }
```
JammingBen, line 397:
```suggestion
  }
```

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -384,6 +404,10 @@ export default defineComponent({
   justify-content: flex-start;
 }
 
+.link-modal-password {
+  margin-left: calc(var(--oc-space-small) + 22px);
```
JammingBen, line 408: I guess this is the alignment of `Password` with `Anyone can...`? Maybe better go for a "grid-based" solution if possible? So the icon in the first "column", the rest in the second so it automatically aligns. Working with pixels always feels so fragile.

packages/web-pkg/src/composables/actions/files/useFileActionsCreateLink.ts
```diff
@@ -38,7 +34,14 @@ export const useFileActionsCreateLink = ({
...
 
-  const proceedResult = async (result: PromiseSettledResult<LinkShare>[]) => {
+  const proceedResult = async ({
+    result,
+    password
+  }: {
+    result: PromiseSettledResult<LinkShare>[]
+    password?: string
+  }) => {
+    console.log(password)
```
JammingBen, line 44:
```suggestion
```

packages/web-pkg/src/composables/actions/files/useFileActionsCreateLink.ts
```diff
@@ -49,22 +52,21 @@ export const useFileActionsCreateLink = ({
       if (result.length === 1) {
         // Only copy to clipboard if the user tries to create one single link
         try {
-          await copyToClipboard(succeeded[0].value.webUrl)
+          const copyToClipboardText = password
+            ? `Link:${succeeded[0].value.webUrl} Password:${password}`
+            : succeeded[0].value.webUrl
+
+          console.log(copyToClipboardText)
```
JammingBen, line 59:
```suggestion
```

packages/web-pkg/src/composables/actions/files/useFileActionsCreateLink.ts
```diff
@@ -49,22 +52,21 @@ export const useFileActionsCreateLink = ({
       if (result.length === 1) {
         // Only copy to clipboard if the user tries to create one single link
         try {
-          await copyToClipboard(succeeded[0].value.webUrl)
+          const copyToClipboardText = password
+            ? `Link:${succeeded[0].value.webUrl} Password:${password}`
```
JammingBen, line 56: Should we translate this?

JammingBen (commented):
Uh I like it a lot! Feels to much cleaner.

Just a few nitpicks here and there. And creating links with an expiration date gives me a weird error underneath the password form: `Invalid body schema definition` 🤔

packages/web-pkg/src/components/CreateLinkModal.vue
AlexAndBear: I am not sure about this, we would need to fake an empty element in order to archive this, it also makes the HTML markup less easy to read. 
Any opinions on this @kulmann 

The current implementation calculated the margin based on the icon width, which is 22px

JammingBen: Why do we need to fake an empty element? What I mean is something like this:

```
<div class="oc-flex">
  <div><oc-icon ... /></div>
  <div>
    <div>role select</div>
    <div>password</div>
  </div>
</div>
```

AlexAndBear: Because we need something like that:


icon  |   roleselect   |   options|
         |   password 

That's not gonna work with your proposal, a margin or adding an empty element to a grid in order to archive this visual appearance is inevitable

JammingBen: I'm pretty sure that works with my proposal, unless I'm missing something 😅 It just needs to be extended via the third column:

```
<div class="oc-flex">
  <div><oc-icon ... /></div>
  <div>
    <div>role select</div>
    <div>password</div>
  </div>
  <div>options</div>
</div>
```

AlexAndBear: Feel free to commit on this pr ☺️

AlexAndBear: Hm now I think I understand what you are trying to archive 🤔 

Give me a second

AlexAndBear: I don't know if this is the correct way to archive the desired look, as we would limit the passwords space to the right always at the same position where the options button starts

🤔🤔🤔

JammingBen: Yeah that's true... hm. Well then, probably not worth investing more time into making this "beautiful". Just a nitpick anyways 🤷

AlexAndBear: I leave this question open to my superior 😁

A second review would anyway pretty welcome

kulmann: Not worth the refactoring... would be a three-column-grid with (at least) two lines, where on line 2 the columns 2 and 3 would be merged. Most of all I see the issue that in advanced mode the role icon comes from within the role select (= one component in the modal) whereas in the simple mode it's two different elements (the icon and the span). So in advanced mode it would again be different layouting or we'd need to break the icon from the role select in advanced mode out of the role select component. Meh. I share the dislike of pixel-based design/layout, but in this case I agree that the effort is not worth it. 🙈

kulmann (approved):
Love it! 😍

JammingBen (approved):

Issue #10728, opened 2024-04-05
[a11y] 9.1.4.10 Break content


tschiebel:
audit: https://infinite.owncloud.com/s/PnaAukHGGsLnrMM

https://bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-1-4-10-inhalte-brechen-um

### Personal 

- [ ] With a browser window width of 1280 CSS pixels and a zoom magnification of
400%, the page content does not break. This means that all information and
functions are not available without users having to scroll horizontally.
![image](https://github.com/owncloud/web/assets/132444655/9b38ef20-1531-4bf8-9f67-bc6cb751f622)

### Shares 

- [ ] If the browser window width is set to 1280 CSS pixels and the zoom magnification is set to
400%, only two columns are displayed in the table: "Name" and "Actions".
"Actions". Other table columns are no longer accessible to users. To
ensure that all table columns are visible even at higher zoom magnification, the table should either be
the table should either be designed in such a way that it adapts to the available
width or offer the option of scrolling horizontally through the table columns.
scroll horizontally through the table columns.
![image](https://github.com/owncloud/web/assets/132444655/49246a1c-7919-478c-80f1-df46ae213a7a)

### Account 

- [x] If the browser window width is set to 1280 CSS pixels and the zoom magnification is set to
400%, the content at the bottom of the page is cut off and is no longer accessible.
no longer accessible.
![image](https://github.com/owncloud/web/assets/132444655/8ca38acd-70db-4529-be1e-ff182b353b36)

pascalwengerter:
Assigning @tbsbdr here since I need a decision on how to handle the tables on small screens

tbsbdr:
- remove text from buttons "New" and "Upload" for like 1280@400% 

![Image](https://github.com/user-attachments/assets/c868f7e0-f425-4d64-87ab-811c579f6de7)


- **Shared** and **Spaces** should not scroll horizontally (lets make it like in personal, i.e. stack and remove)


![Image](https://github.com/user-attachments/assets/aece5ae5-9928-41ec-bddc-2b9a89ce0b5a)


![Image](https://github.com/user-attachments/assets/f674ecc3-198a-499d-b285-4dade2708db1)

kulmann:
@tbsbdr at 400% zoom on a 14 inch macbook the file list is not visible at all 🙈 see screenshot:

[...]

PR #11313, opened 2024-08-01
fix: horizontal resource table scroll

JammingBen:
## Description
Prevents resource tables from being horizontally scrollable for a11y reasons. Instead, information that is available elsewhere will get hidden on smaller displays.

## Related Issue

- Works towards https://github.com/owncloud/web/issues/10728

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/bugfix-horizontal-table-scroll
+++ b/changelog/unreleased/bugfix-horizontal-table-scroll
@@ -1,0 +1,6 @@
+Bugfix: Prevent horizontal table scroll
+
+We've prevented resource tables from being horizontally scrollable for a11y reasons. Instead, information that is available elsewhere will get hidden on smaller displays.
+
+https://github.com/owncloud/web/pull/11313
+https://github.com/owncloud/web/issues/10728
--- a/packages/design-system/src/components/OcDrop/OcDrop.vue
+++ b/packages/design-system/src/components/OcDrop/OcDrop.vue
@@ -291,6 +291,7 @@ export default defineComponent({
 }
 
 .oc-drop {
+  max-width: 100%;
   width: 300px;
 
   li a:focus {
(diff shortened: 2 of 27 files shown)


kulmann:
ftr I changed the PR description from `fixes` to `works towards` because I found another issue. We need to check first if that needs to be fixed as well...

kulmann:
Can we make sure that all action buttons above the file list are one row and break into the next line from the beginning? Also, line break only kind of works if I shrink the browser width even further, see screenshot (where you can actually see both of my issues):

Screenshot 2024-08-05 at 05 49 20

kulmann:
Can we have some margin-bottom on the filters?

Screenshot 2024-08-05 at 05 52 07

kulmann:
Spaces overview still scrolls horizontally:

Screenshot 2024-08-05 at 05 53 37

JammingBen:
> Can we make sure that all action buttons above the file list are one row and break into the next line from the beginning? 

Yes I can do that, but selection will be jumpy then because we add new vertical space that wasn't there before.

> Also, line break only kind of works if I shrink the browser width even further, see screenshot (where you can actually see both of my issues):

It's fine with the smallest pre-setting of the browser, no? We should have a limit IMO, nobody uses a screen <320px. If we really want to support that, I feel like this is a separate story.

> Can we have some margin-bottom on the filters?

I discussed that with @tbsbdr : it's not that easy because it's just a `flex-wrap`. So it would need a hand-crafted solution.

In general I thought the goal of this issue is to make it a11y-conform. Making things actually look nice on small screens is certainly possible, but we would need to allocate a lot more time for that.

> Spaces overview still scrolls horizontally:

I can't reproduce that 🤔  What's your vertical screen size?

image

JammingBen:
@kulmann I fixed the horizontal scroll for production build, can you please check if the spaces page is still scrollable?

For the breadcrumbs, as discussed, let's not solve this via a new line because it's too jumpy. This could be solved via a context menu, but not in the scope of this.

Issue #11335, opened 2024-08-07
Actions for disabled resources can be triggered via keyboard


JammingBen:
## Describe the bug
Actions for disabled resources can be triggered via keyboard when navigating via tab and then pressing enter.

## Steps to reproduce
1. Upload a resource
2. Disable that resource (can be done via code by returning `true` here: https://github.com/owncloud/web/blob/master/packages/web-pkg/src/components/FilesList/ResourceTable.vue#L549)
3. Tab through the table
4. As soon as you reach the pencil icon behind the resource name, press enter

## Expected behavior
The action should not be focusable.

## Actual behavior
The action can be focussed and performed. The same goes for the quick actions.

PR #11342, opened 2024-08-08
fix: keyboard actions for deactivated resources

JammingBen:
## Description
Prevents actions for disabled resources from being accessible via keyboard navigation. This is achieved by either disabling them (checkboxes) or hiding them (rename action, quick actions).

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11335

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


AlexAndBear (changes requested):
Unfortunately this disables the context menu in the tiles view for disabled spaces, actions like enable/delete life in there, and therefore should be accessible.

As you are already on it, the show members button will still be shown in the tiles view on a disabled space, clicking on that button opens the sidebar and ends in an endless loading animation. We might need to disable the show members button, but I think seeing members was possible on a disabled space, before switching to sharing ng 🤔

JammingBen:
> Unfortunately this disables the context menu in the tiles view for disabled spaces, actions like enable/delete life in there, and therefore should be accessible.

Good catch, should be fixed now.

> As you are already on it, the show members button will still be shown in the tiles view on a disabled space, clicking on that button opens the sidebar and ends in an endless loading animation. We might need to disable the show members button, but I think seeing members was possible on a disabled space, before switching to sharing ng 🤔

IMO this is a server issue, listing members should be possible for disabled spaces. I'll create an issue in the oCIS repo for that 👍

AlexAndBear:
Thank you, I took the opportunity, to disable the edit description action when space is disabled and don't load previews, since we don't have access to those files when disabled

AlexAndBear (approved):

AlexAndBear (changes requested):
you commited some vitest files by accident

AlexAndBear (approved):

Issue #10770, opened 2024-04-12
[Web & ocis] Make proper use of share permissions instead of roles


JammingBen:
With Sharing NG, roles and permissions are coming from the server. While there is a predefined set of available permissions, roles are simply a "bucket" of permissions. In the future we even may have custom roles per instance, which makes it impossible to rely on roles.

Therefore Web should switch its focus on permissions and only use/display role information if the server returns a matching role for a given set of permissions.

## Userstory
> As an organisation I dont want that fileversions are shared, so that informations are not accidentally revealed.

### Acceptance criteria

* Make sure all available permissions are included in `GraphSharePermission` (https://github.com/owncloud/web/blob/master/packages/web-client/src/helpers/share/types.ts#L4). Check against ocis to verify this.
* Rename `GraphSharePermission` to `DriveItemPermission`
* Remove all role checks (e.g. `share.role.id !== GraphShareRoleIdMap.SpaceManager`) in favour of permission checks
* Remove `GraphShareRoleIdMap`(?) We might implement something similar for tests though? Now sure...
* Re-think the hard-coded keys of the `spaceRoles` property in the `SpaceResource` interface. These names can break when the server adds new roles or renames existing.
* Make sure that permissions are being displayed in case a set of permissions doesn't have a matching role coming from the server.
  * How do we display the permissions without cluttering the UI? We probably shouldn't list all permissions coming from the server. @tbsbdr

PR #11364, opened 2024-08-15
refactor: remove share role checks

JammingBen:
## Description
Use share permissions instead of roles since roles are a very loose concept and just act as a container for a specific set of share permissions.

This affects 2 main scenarios:

* A `SpaceResource` doesn't have properties like `manager`, `editor` anymore. The same goes for methods like `isManager`. Instead, there is the `members` property that holds information about all members. When checking if the current user is allowed to do certain actions, we need check for specific permissions now. E.g. when sharing, don't check for `isManager` but for `canShare`.
* Shares (in Graph terms: "permissions") will also have permissions (in Graph terms: "permission actions") that need to be checked instead of roles. Hence the `GraphShareRoleIdMap` is being removed.

You may notice a discrepancy in our naming and the Graph naming. IMO we should align... but that's a another story, I guess. Because ideally, I would also rename `Space` to `Drive`. It's just so confusing all the time.

There is one exception where Web still needs to assume and work with roles. That is when listing the managers of a space. Here we imply that if the user can remove space members, they are a space manager.

## Related Issue

- refs https://github.com/owncloud/web/issues/10770

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersPanel.vue
+++ b/packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersPanel.vue
@@ -68,7 +68,15 @@ export default defineComponent({
     const availableRoles = computed<ShareRole[]>(() => {
       const permissionsWithRole = unref(spaceMembers).filter((p) => !!p.roleId)
       const roleIds = [...new Set(permissionsWithRole.map((p) => p.roleId))]
-      return roleIds.map((r) => sharesStore.graphRoles[r]).filter(Boolean)
+      return roleIds
+        .map((r) => sharesStore.graphRoles[r])
+        .filter(Boolean)
+        .sort((a, b) => {
+          // sort roles by amount of permissions (most likely translates to manager > editor > viewer)
+          const permissionsA = a.rolePermissions.flatMap((r) => r.allowedResourceActions)
+          const permissionsB = b.rolePermissions.flatMap((r) => r.allowedResourceActions)
+          return permissionsB.length - permissionsA.length
+        })
     })
 
     const membersWithoutRole = computed<SpaceMember[]>(() => {
--- a/packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersPanel.vue
+++ b/packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersPanel.vue
@@ -9,102 +9,75 @@
       <div v-if="!filteredSpaceMembers.length">
         <h3 class="oc-text-bold oc-text-medium" v-text="$gettext('No members found')" />
       </div>
-      <div
-        v-if="filteredSpaceManagers.length"
-        class="oc-mb-m"
-        data-testid="space-members-role-managers"
-      >
-        <h3 class="oc-text-bold oc-text-medium" v-text="$gettext('Managers')" />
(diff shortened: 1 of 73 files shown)


packages/web-pkg/src/composables/actions/spaces/useSpaceActionsEditReadmeContent.ts
```diff
@@ -73,10 +78,6 @@ export const useSpaceActionsEditReadmeContent = () => {
           return false
         }
 
-        if (resources[0].disabled) {
-          return false
-        }
-
```
JammingBen, line 79: This is already being checked in `canEditReadme`.

packages/web-client/src/helpers/share/functions.ts
```diff
@@ -151,9 +150,7 @@ export function buildIncomingShareResource({
     sharePermissions,
     outgoing: false,
     canRename: () => driveItem['@client.synchronize'],
-    canDownload: () =>
-      sharePermissions.includes(GraphSharePermission.readBasic) &&
-      !shareRoles.some((shareRole) => shareRole.id === GraphShareRoleIdMap.SecureViewer),
+    canDownload: () => sharePermissions.includes(GraphSharePermission.readContent),
```
kulmann, line 153: So do I get it right that `readContent` permission action is not given with a secure view share? Which permission action does the secure view share have for secure-viewing content?

packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersPanel.vue
```diff
@@ -9,103 +9,76 @@
...
         <h3 class="oc-text-bold oc-text-medium" v-text="$gettext('No members found')" />
       </div>
-      <div
-        v-if="filteredSpaceManagers.length"
-        class="oc-mb-m"
-        data-testid="space-members-role-managers"
-      >
-        <h3 class="oc-text-bold oc-text-medium" v-text="$gettext('Managers')" />
-        <members-role-section :members="filteredSpaceManagers" />
+      <div v-for="(role, i) in availableRoles" :key="i">
```
kulmann, line 12: The order of roles seems to be somewhat random. Seems to be stable per space, but from one space to another it's not the same. Any idea how to sort them in a meaningful and stable way?

Screenshot 2024-08-23 at 11 49 23
Screenshot 2024-08-23 at 11 49 31

JammingBen: Hmm I could at least sort by the amount of permissions a role has. That should then lead to Managers > Editors > Viewers.

kulmann: nice idea!

packages/web-client/src/helpers/share/functions.ts
JammingBen: Okay so the permissions of a secure view share are these:

```
* libre.graph/driveItem/path/read
* libre.graph/driveItem/children/read
* libre.graph/driveItem/basic/read
```

kulmann (approved):
Thank you for this! I can imagine that it has been a tedious chunk of work. 🤗

PR #11380, opened 2024-08-20
refactor: transform graph roles to object

JammingBen:
## Description
Refactors the unified graph role definitions from an array to an object because it simplifies the way how to work with those. When retrieving a role via its id (which is 99% of the use cases), it can now be done with the object index instead of looping through the array.

Works towards https://github.com/owncloud/web/issues/10770.

## Types of changes

- [x] Technical debt
- [ ] Tests
- [ ] Documentation


dschmidt (approved):
sweet 🥳

PR #11387, opened 2024-08-22
Disabled unified roles

fschade:
## Description
adds a new feature to the sharing sidebar panel which allows you to see the permissions of a role that has been disabled.

## Related Issue
- Fixes https://github.com/owncloud/ocis/pull/9727
- Part of https://github.com/owncloud/web/issues/10770

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
@@ -21,7 +21,7 @@
       </oc-button>
       <oc-contextual-helper
         v-if="isDisabledRole"
-        class="oc-ml-xs"
+        class="oc-ml-xs files-permission-actions-list"
         :list="existingSharePermissions.map((permission) => ({ text: $gettext(permission) }))"
       />
     </div>
--- a/packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
@@ -19,6 +19,11 @@
         <span class="oc-text-truncate" v-text="inviteLabel" />
         <oc-icon name="arrow-down-s" />
       </oc-button>
+      <oc-contextual-helper
+        v-if="isDisabledRole"
+        class="oc-ml-xs"
+        :list="existingSharePermissions.map((permission) => ({ text: $gettext(permission) }))"
+      />
     </div>
     <oc-drop
       v-if="availableRoles.length > 1"
@@ -81,15 +86,21 @@ import { useAbility, useUserStore } from '@ownclouders/web-pkg'
 import { Resource } from '@ownclouders/web-client'
 import { useGettext } from 'vue3-gettext'
 import { ShareRole } from '@ownclouders/web-client'
+import { $gettext } from '@ownclouders/web-pkg/src/router/utils'
 
 export default defineComponent({
   name: 'RoleDropdown',
   components: { RoleItem },
   props: {
-    existingRole: {
+    existingShareRole: {
       type: Object as PropType<ShareRole>,
       required: false,
       default: undefined
+    },
+    existingSharePermissions: {
+      type: Array as PropType<string[]>,
+      required: false,
+      default: () => []
     },
     domSelector: {
       type: String,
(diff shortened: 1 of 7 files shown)


packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
```diff
@@ -81,16 +86,22 @@ import { useAbility, useUserStore } from '@ownclouders/web-pkg'
 import { Resource } from '@ownclouders/web-client'
 import { useGettext } from 'vue3-gettext'
 import { ShareRole } from '@ownclouders/web-client'
+import { $gettext } from '@ownclouders/web-pkg/src/router/utils'
```
JammingBen, line 89: That's something IntelliJ loves to do for some reasons, although not needed.

```suggestion
```

packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
```diff
@@ -183,6 +216,7 @@ export default defineComponent({
   },
 
   methods: {
+    $gettext,
```
JammingBen, line 219: Same here.

```suggestion
```

packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
```diff
@@ -19,6 +19,11 @@
         <span class="oc-text-truncate" v-text="inviteLabel" />
         <oc-icon name="arrow-down-s" />
       </oc-button>
+      <oc-contextual-helper
+        v-if="isDisabledRole"
+        class="oc-ml-xs files-permission-actions-list"
+        :list="existingSharePermissions.map((permission) => ({ text: $gettext(permission) }))"
```
JammingBen, line 25: We still need add the list of permissions to the `additionalTranslations.ts` file, otherwise they don't land in Transifex. You can use the values from the `GraphSharePermission` enum as keys.

fschade: thanks for the heads up, done, but looks a bit loose to me... something that is easy to forget, isnt it?

i thing something like

```js
// pseudo code
existingSharePermissions.map((permission) => ({ text: additionalTranslations.knownSharePermissions[permission] }))
```

or similar won't work as expected? right?

JammingBen: > thanks for the heads up, done, but looks a bit loose to me... something that is easy to forget, isnt it?

There are only a few cases, but it can be overseen quite easily, yeah :/

> i thing something like
> 
> ```js
> // pseudo code
> existingSharePermissions.map((permission) => ({ text: additionalTranslations.knownSharePermissions[permission] ?? $gettext(permission) }))
> ```
> 
> or similar won't work as expected? right?

Correct, this again wouldn't land in Transifex. The extractor only looks for strings inside of `$gettext()`.

packages/web-runtime/src/helpers/additionalTranslations.ts
```diff
@@ -29,5 +30,19 @@ export const additionalTranslations = {
...
+    $gettext('libre.graph/driveItem/children/create'),
+    $gettext('libre.graph/driveItem/basic/read'),
+    $gettext('libre.graph/driveItem/path/read'),
+    $gettext('libre.graph/driveItem/quota/read'),
+    $gettext('libre.graph/driveItem/content/read'),
+    $gettext('libre.graph/driveItem/children/read'),
+    $gettext('libre.graph/driveItem/deleted/read'),
+    $gettext('libre.graph/driveItem/path/update'),
+    $gettext('libre.graph/driveItem/deleted/update'),
+    $gettext('libre.graph/driveItem/standard/delete')
```
JammingBen, line 46: @tbsbdr This is how the permissions are currently listed if a share/space role has been disabled on the server. Is this okay like this (for now), or should we replace it with more "meaningful" texts?

tbsbdr: you don't have a screenshot, do you?

JammingBen: Looks like this for the "Secure view" role e.g. (the margins of our contextual helper are a bit off):

image

JammingBen: As discussed: we leave these permissions untranslated (for now). As soon as we have a doc link, we'll add it to the context helper so users can make sense of it (see https://github.com/owncloud/web/issues/11392).

JammingBen (approved):

Issue #11360, opened 2024-08-14
[web] Personal: Show all share-receivers in shares panel


tbsbdr:
# Description

## User Stories

* > As a share receiver I want to know who else has access to the share so that I know who can access it.

## Value
- transparency
- compliance

## Acceptance Criteria

- Applies only, if the transparent sharing role is set: https://github.com/owncloud/ocis/issues/9701 (check needs to be done via permissions though)
- In `shares`-> `shared with me`: show sharing panel for share-receiver with all other share-receivers (same like in Spaces)

*Technical challange:* How to get the unified role permissions of the current share inside a share space?

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

PR #11393, opened 2024-08-23
feat: add permissions to share spaces

JammingBen:
## Description
Since share spaces are virtual, they and their permissions can't be retrieved from the server. Instead, we have to load the permissions via the underlying share. This change does exactly that and persists the permissions for the current user within the space object so they can be accessed and checked when needed.

Also makes sure that share spaces are being added to the spaces store so they're not just temporary objects.

## Related Issue

- Works towards https://github.com/owncloud/web/issues/11360

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-app-files/src/services/folder/loaderSpace.ts
```diff
@@ -33,10 +44,13 @@ export class FolderLoaderSpace implements FolderLoader {
   }
 
   public getTask(context: TaskContext): FolderLoaderTask {
-    const { router, clientService, resourcesStore, userStore } = context
-    const { webdav } = clientService
+    const { router, clientService, resourcesStore, userStore, spacesStore, sharesStore } = context
+    const { webdav, graphAuthenticated: graphClient } = clientService
```
kulmann, line 48: Didn't we have the issue of the different clients losing the access token reactivity when destructuring the clientService? 🤔

JammingBen: This function gets called every time the loader performs, so the token would need to expire within the milliseconds between here and the function call 😅

packages/web-pkg/src/composables/spaces/useGetMatchingSpace.ts
```diff
@@ -63,12 +62,15 @@ export const useGetMatchingSpace = (options?: GetMatchingSpaceOptions) => {
       shareName = resource.name
     }
 
-    return buildShareSpaceResource({
-      driveAliasPrefix,
-      id: resource.remoteItemId,
-      shareName,
-      serverUrl: configStore.serverUrl
-    })
+    const shareSpace =
```
kulmann, line 65: can be an inline return

packages/web-app-files/src/services/folder/loaderSpace.ts
```diff
@@ -102,4 +132,77 @@ export class FolderLoaderSpace implements FolderLoader {
...
+    }
+
+    const allPermissions: string[] = []
+    permissions.forEach((permission) => {
+      if (permission['@libre.graph.permissions.actions']) {
+        allPermissions.push(...permission['@libre.graph.permissions.actions'])
+        return
+      }
+      const role = sharesStore.graphRoles[permission.roles[0]]
+      if (role) {
```
AlexAndBear, line 190: rather have an early return here, which translates to 'continue' in a regular for each

AlexAndBear (changes requested):

packages/web-app-files/src/services/folder/loaderSpace.ts
JammingBen: Why? 😅 The if-block has 2 simple lines of code, so it wouldn't enhance readability or save up anything.

AlexAndBear: For me it indeed improves readability 🤭

AlexAndBear (approved):

PR #11421, opened 2024-08-27
chore: remove font-awesome dependency

JammingBen:
## Description
As far as I can see, the dependency was only used to display the attribution in the browser console. However, this attribution has been removed with their version `6.6.0`.

Also, according to https://fontawesome.com/license/free it's sufficient to have the attribution embedded in the downloaded icons (which we have).

Superseds https://github.com/owncloud/web/pull/11418.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/package.json
+++ b/packages/web-runtime/package.json
@@ -7,7 +7,6 @@
   "dependencies": {
     "@casl/ability": "6.7.1",
     "@casl/vue": "^2.2.2",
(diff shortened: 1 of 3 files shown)


kulmann (approved):
Cool, thank you! 💪 We had the lib only for displaying the attribution in the browser console, you got that right. 👏

PR #11464, opened 2024-08-28
chore: clean up mono repo dependencies

JammingBen:
## Description
* Removes `dependencies` and `peerDependencies` from root `package.json` since it should only hold `devDependencies`. "Real" dependencies are shipped via `web-runtime` or the respective packages.
* Removes dependencies from the root `package.json` if they are not needed there. Generally, only 2 types of dependencies are needed in the root file:
  * the ones needed for CLI commands that are ran from the root directory.
  * dev-dependencies that can be avoided to be added to each package.
* Removes some unused dependencies.
* Cleans up a bit of our own dependency usage in the mono repo (e.g. removes the runtime dependency from the files app).
* Creates a clear separation between dev-, peer- and regular dependencies in all our packages.

## Open questions/tasks

I'm going to tackle those in separate PRs:

* ~Go through each package and check if the separation in dev-, peer- and regular dependency is correct. E.g. `web-pkg` has the runtime dependency `@toast-ui/editor` as dev-dependency. This seems wrong, it should either be a regular dependency, or additionally a peer dependency with `web-runtime` shipping it as regular dependency.~
* The files app has itself as dependency to enable imports via `web-app-files/...`. IMO this should be removed because it results in the file app being added recursively in `node_modules`. Imports within the files app should be relative.
* ^ the same goes for the `admin-settings`.
* ^ the same goes for the `webfinger`.
* The files app has the search app as dependency because it's importing one component from it. This is bad.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


kulmann (approved):
Thanks for the cleanup ❤️

PR #11515, opened 2024-09-04
Add default actions extension point

kulmann:
## Description
This PR adds another extension point for the default actions of files (as used in resource table / tiles, etc). With this e.g. the new `Open in Notebooks` jupyter extension can be connected to the left click in the files list.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-pkg/src/composables/actions/files/useFileActions.ts
```diff
@@ -240,45 +248,39 @@ export const useFileActions = () => {
...
-    return options.omitSystemActions ? null : unref(systemActions).filter(filterCallback)[0]
+    const extensionActions = unref(defaultActions).filter(filterCallback)
+    if (extensionActions.length) {
+      return extensionActions[0]
+    }
+
+    if (options.omitSystemActions) {
+      return null
+    }
+    return unref(systemActions).filter(filterCallback)[0]
```
JammingBen, line 269:
```suggestion
    return unref(systemActions).find(filterCallback)
```

JammingBen (approved):
Not for the scope of this, but what I noticed: shortcuts should probably be `<a>` links too, no? At least for a11y reasons.

kulmann:
> Not for the scope of this, but what I noticed: shortcuts should probably be `<a>` links too, no? At least for a11y reasons.

It should be, but that would mean that we need to read and evaluate the .url files contents upon listing them, doesn't it?

Issue #10578, opened 2024-03-11
[Performance] File selection is extremely slow


dschmidt:
## Describe the bug
Selecting files should be instant, especially in the tiles view it's super slow.

## Steps to reproduce
1. Upload 500 photos into one folder
2. Switch to tiles view
3. Try to select files (-> selected = checkbox is ticked)

## Expected behavior
Clicking a resource instantly selects it.

## Actual behavior
With 500 tiles it takes multiple seconds to select a resource on my machine.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

AlexAndBear:
Good news 🥳 

A combination of:
- https://github.com/owncloud/web/pull/11508
- https://github.com/owncloud/web/pull/11631

fixes the issue:

https://github.com/user-attachments/assets/90090fb2-e15b-4695-8d01-c2528c4faaee

PR #11518, opened 2024-09-04
Select all checkbox tiles view

AlexAndBear:
## Description
image

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcCheckbox/OcCheckbox.vue
+++ b/packages/design-system/src/components/OcCheckbox/OcCheckbox.vue
@@ -85,14 +85,6 @@ export default defineComponent({
       default: false
     },
     /**
-     * Is the label of the Checkbox visually hidden?
-     **/
-    hideLabel: {
-      type: Boolean,
-      required: false,
-      default: false
-    },
-    /**
      * Size of the Checkbox. Valid values are `small`, `medium` and `large`.
      * If not specified, defaults to `medium`
      */
@@ -132,7 +124,6 @@ export default defineComponent({
     },
     labelClasses() {
       return {
-        'oc-invisible-sr': this.hideLabel,
         'oc-cursor-pointer': !this.disabled
       }
     },
--- a/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
@@ -6,7 +6,7 @@
         size="large"
         :disabled="resources.length === disabledResources.length"
         :label="$gettext('Select all')"
-        :hide-label="true"
+        :label-hidden="true"
         :model-value="areAllResourcesSelected"
         @click.stop="toggleSelectionAll"
       />
@@ -77,7 +77,7 @@
             <oc-checkbox
               v-if="!isLocationPicker && !isFilePicker"
               :label="getResourceCheckboxLabel(resource)"
-              :hide-label="true"
+              :label-hidden="true"
               size="large"
               class="oc-flex-inline oc-p-s"
               :disabled="!isSpaceResource(resource) && isResourceDisabled(resource)"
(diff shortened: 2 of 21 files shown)


packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -1,46 +1,59 @@
...
+          </oc-list>
+        </oc-drop>
+      </div>
     </div>
     <oc-list class="oc-tiles oc-flex">
       <li
         v-for="resource in resources"
         :key="resource.id"
         class="oc-tiles-item has-item-context-menu"
+        v-memo="[resources]"
```
AlexAndBear, line 53: @JammingBen @kulmann I added the v-memo directive, so the list gets only re-rendered when the resources do change (which is not the case when a resource gets selected) which makes the the selection quite fast with 500 tiles and fixes https://github.com/owncloud/web/issues/10578.

I couldn't spot any bad side effects

JammingBen: Nice! Can/Should we use it for the files table as well? 🤔

packages/design-system/src/components/OcCheckbox/OcCheckbox.vue
```diff
@@ -84,14 +84,6 @@ export default defineComponent({
...
       default: false
     },
-    /**
-     * Is the label of the Checkbox visually hidden?
-     **/
-    hideLabel: {
-      type: Boolean,
-      required: false,
-      default: false
-    },
```
JammingBen, line 94: Why did you remove this, didn't you just add it recently?

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -363,10 +380,31 @@ export default defineComponent({
       return resource.processing === true
     }
 
+    const disabledResources: ComputedRef<Array<Resource['id']>> = computed(() => {
```
JammingBen, line 383: Can we name this `disabledResourceIds`? Otherwise it suggests full resources here. Then you could also remove the type annotation `ComputedRef<Array<Resource['id']>>` and let TS infer the type.

JammingBen (commented):
Something messes with the right-click on tiles, it only works the first time for me.

packages/design-system/src/components/OcCheckbox/OcCheckbox.vue
AlexAndBear: I added labelHidden prop, which is more a11y compatible than the implementation behind the hideLabel prop. I adjusted the other occurrences

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
AlexAndBear: Sure this was just copy paste

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
AlexAndBear: From my observations, the table doesn't get rerendered again when props change but the list in the resource tiles component did, so I don't think it's necessary

AlexAndBear: > Something messes with the right-click on tiles, it only works the first time for me.

Aaaaaaaand that is the v-memo side effect urghs

JammingBen:
> Something messes with the right-click on tiles, it only works the first time for me.

This is still an issue it seems like.

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
AlexAndBear: Removed the v-memo, this is not working correctly and scope creep, need to investigate further 🕵️

JammingBen (approved):

Issue #11163, opened 2024-07-09
[deps] Upgrade eslint to v9


JammingBen:
We're currently running v8. v9 brings a breaking change in how to set up the config (see https://eslint.org/docs/latest/use/configure/migration-guide), so there is a bit of work to do to get this working.

PR #11519, opened 2024-09-04
chore: migrate eslint config to be compliant with v9

JammingBen:
## Description
Removes the old `eslint.json` in favor of the new `config.eslint.js` config. See the [migration guide](https://eslint.org/docs/latest/use/configure/migration-guide) for more details.

I needed to manually add some TS rules so it keeps behaving like it was before. This could be due to the new flat configs for `.ts` and `.vue` files.

Afterwards I ran the linter. It auto-fixed some issues (removed some unnecessary ignore-comments, refactored `let` to `const` if possible).

Next we should release a new version of the `eslint-config` package.

## Related Issue

- closes https://github.com/owncloud/web/issues/11163

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


tests/e2e/.eslintrc.json
JammingBen, line 1: Not sure why this package extended our config with its own custom rules. I guess this is not necessary?

tests/e2e/eslint.config.js
kulmann, line 1: Did you try if this file can be removed? Does it pick the root level eslint config then?

kulmann (approved):
🚀

Issue #11478, opened 2024-08-29
User does not get logged out when token renewal fails


JammingBen:
## Describe the bug
The user does not get logged out when token renewal fails. As a result, they can still navigate the UI, but all requests will fail.

## Steps to reproduce
1. Login with a user
2. Wait and make sure your device goes into standby
3. Wake up your device and check the UI

## Expected behavior
The user should be logged out if no active refresh token present.

## Actual behavior
They are not.

Kinda hard to debug... maybe manually deactivating the token renewal worker could work for finding a solution.

kulmann:
cc @hodyroff this is what you described in the chat recently.

PR #11526, opened 2024-09-05
fix: logout when token renewal fails

JammingBen:
## Description
Adds a logout when the token renewal fails for some reason. This fixes the issue with the user being able to access the Web UI without a valid token, leading to a broken state.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11478

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/services/auth/authService.ts
+++ b/packages/web-runtime/src/services/auth/authService.ts
@@ -350,6 +350,11 @@ export class AuthService implements AuthServiceInterface {
     this.authStore.clearUserContext()
   }
 
+  public async getRefreshToken() {
+    const user = await this.userManager.getUser()
+    return user?.refresh_token
+  }
+
   private handleDelegatedTokenUpdate(event: MessageEvent) {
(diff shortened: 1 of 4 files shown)


packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -350,6 +350,11 @@ export class AuthService implements AuthServiceInterface {
     this.authStore.clearUserContext()
   }
 
+  public async getRefreshToken() {
+    const user = await this.userManager.getUser()
+    return user.refresh_token
```
kulmann, line 355: Is it guaranteed that we have a user here? Or should we return `user?.refresh_token`

kulmann (approved):

packages/web-runtime/src/services/auth/authService.ts
JammingBen: Better be safe, yes!

Issue #11544, opened 2024-09-09
Remove internal link


tbsbdr:
# Description

## User Stories

* > As a user I want a clear way to create a pointer for colleagues.

## Acceptance Criteria

- remove Internal Link from drop menu

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

## Visual

![screenshot_003117](https://github.com/user-attachments/assets/8df040f6-543b-47f8-a16c-9fee29d73e9f)

PR #11553, opened 2024-09-09
feat: remove internal link type

JammingBen:
## Description
Removes the internal link type since the permanent link makes it obsolete. Existing internal links still resolve correctly to ensure backwards compatibility.

Also removes the action in favor of the new action for copying permanent links.

## Related Issue

- Closes https://github.com/owncloud/web/issues/11544

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


JammingBen:
@tbsbdr @kulmann There is one edge case when the user doesn't have the permission to create public links. In the past, creating internal links was still possible. With this change though, those users can't create links at all, which is a bit tricky because of the "Copy link" quick action. We can easily check if the resource as a link share and then decide whether to display the quick action or not. But we can't check if the resource has a *quick link*. And since the quick action is designed to only copy quick links, this results in an issue when a resource has a normal link, but not a quick link.

We could change to quick action to factor in regular links as well. Or simply ignore this edge case.

tbsbdr:
I'd opt to make the permalink the default quickaction link

kulmann:
> I'd opt to make the permalink the default quickaction link. we already talked about it some time ago. what do you think @kulmann ?

default was in always use the permalink?!

If I remember correctly I advocated to kill the quick link concept entirely and instead just use any first link that can be found on the resource. If there is none -> permalink.

Since this PR is about removing the internal link type, a good step in between is to copy the permalink when the user isn't allowed to create a public link. Not so sure about always copying the permalink as the only option for the quick action. 🤔

tbsbdr:
Lets always use the perma link in the quickaction. ok @JammingBen @AlexAndBear @kulmann ?

JammingBen:
> Lets always use the perma link in the quickaction. ok @JammingBen @AlexAndBear @kulmann ?

What about the action in the right-click context menu? We have the same problem here. Should it also just copy the permanent link?

kulmann:
> > Lets always use the perma link in the quickaction. ok @JammingBen @AlexAndBear @kulmann ?
> 
> What about the action in the right-click context menu? We have the same problem here. Should it also just copy the permanent link?

IMO yes. Maybe it makes sense to rename the `Share` menu item to `Share or link` or something like that, because everything is in the same panel anyway.

kulmann:
> Okay, so to get this straight: The only way to create a public link will be via the sidebar in the future - no more (quick-) action for that, right? I like that, in fact that's what I was suggesting a few week ago. I just want to make sure we are all on the same page here, because I can certainly imagine that some users won't like it.

Yep, discussed with Tobi yesterday, that  we want to try exactly what you're describing. 👍

AlexAndBear (approved):
Absolutely stunning 🤩

PR #11560, opened 2024-09-10
bugfix(ocm): exclude the local instance in the inviter select

fschade:
## Description
this fixes a bug where the local instance was listed in the inviter select.

## Motivation and Context
reduce confusion by removing the current (users instance) from the inviter select.

### the current user is on https://ocis.ocm.i2:10200/
Screenshot 2024-09-10 at 10 03 20

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-ocm/src/views/IncomingInvitations.vue
+++ b/packages/web-app-ocm/src/views/IncomingInvitations.vue
@@ -35,7 +35,7 @@
                 <span class="option" v-text="domain" />
               </div>
             </template>
-            <template #no-options> No institutions found with this name </template>
+            <template #no-options> No institutions found with this name</template>
             <template #selected-option="{ full_name, domain }">
               <div class="options-wrapper oc-text-break">
                 <strong class="oc-mr-s oc-text-break" v-text="full_name" />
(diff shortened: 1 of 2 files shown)


packages/web-app-ocm/src/views/IncomingInvitations.vue
```diff
@@ -155,7 +155,32 @@ export default defineComponent({
...
+        // also check the port, multiple instances can run on the same host but not on the same port...
+        configStorePort === providerSchemaPort,
+      ]
+
+      switch (true) {
+        case !considerations.length:
+          return false
+        default:
+          return considerations.every((c) => c)
+      }
```
JammingBen, line 173: I think you don't need to switch statement since `considerations.length` is always 2, no?

```suggestion
      return considerations.every((c) => c)
```

AlexAndBear (approved):

JammingBen (commented):
Your linter did something weird 🤔

fschade:
> Your linter did something weird 🤔

fixed... strange

packages/web-app-ocm/src/views/IncomingInvitations.vue
AlexAndBear: I thought you omitted the switch ?  @fschade :D

JammingBen (approved):
Nice, thanks for looking into this! 🚀

Issue #11546, opened 2024-09-09
add date range filter


tbsbdr:
# Description

## User Stories

* > As a user I want to select a daterange so taht I can see activities for a specific daterange.

## Acceptance Criteria
- add dateranges to modified filter chip 
- click on apply filters for daterange
- chip shows daterange dates ex `2.4.2024 - 4.5.2.2024`

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

## Visual
![image](https://github.com/user-attachments/assets/878bdcb7-283d-475d-aa73-e8c61695f0d3)

PR #11565, opened 2024-09-11
feat: add custom date range filter

JammingBen:
## Description
Adds a filter for custom date ranges. This is needed for the soon to come activities app.

The filter stores the custom dates in a query param as milliseconds because that's a nice format for URLs. That means the implementing party might need to convert it to an actual date when working with the filter.

image

image

## Related Issue

- Closes https://github.com/owncloud/web/issues/11546

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/design-system/src/components/OcDatepicker/OcDatepicker.vue
+++ b/packages/design-system/src/components/OcDatepicker/OcDatepicker.vue
@@ -14,7 +14,7 @@
 </template>
 
 <script lang="ts">
-import { computed, defineComponent, onMounted, PropType, ref, unref, watch } from 'vue'
+import { computed, defineComponent, PropType, ref, unref, watch } from 'vue'
 import { useGettext } from 'vue3-gettext'
 import { DateTime } from 'luxon'
 export default defineComponent({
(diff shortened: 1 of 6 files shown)


JammingBen:
I temporarily added the new filter to the search result page for testing. This needs to be reverted before merging though as the search currently can't deal with custom date ranges.

AlexAndBear:
image
The apply button looks deactivated

AlexAndBear (approved):
LGTM,
look and feel is top notch.

You could use the min-date prop on the from date picker, so a warning will be shown when the from date is after the to date. 

As well I am not too sure about the x button, it's super handy, but we don't have it elsewhere in our drop menus, so it would create an inconsistency 🤔

JammingBen:
> You could use the min-date prop on the from date picker, so a warning will be shown when the from date is after the to date.

Good idea, fixed 👍 

> As well I am not too sure about the x button, it's super handy, but we don't have it elsewhere in our drop menus, so it would create an inconsistency 🤔

Tbh the main reason I did this is because the layout of back button - title - close button looks good (see right sidebar). No strong opinion though.

PR #11571, opened 2024-09-12
[tests-only][full-ci] report tracings of failing tests only

saw-jan:
## Description
Currently, we get the tracing reports of all the tests that have been retried regardless of pass or fail. And with keycloak tracing is enabled which causes all tracing to be reported for all tests.

Screenshot from 2024-09-13 10-02-17

Report only the tracings of failed tests. With this PR, now we should get the tracing reports of failed tests only.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1829,7 +1829,7 @@ def e2eTestsOnKeycloak(ctx):
                          "BASE_URL_OCIS": "ocis:9200",
                          "HEADLESS": "true",
                          "RETRY": "1",
-                         "REPORT_TRACING": "true",
+                         "REPORT_TRACING": "with-tracing" in ctx.build.title.lower(),
                          "KEYCLOAK": "true",
                          "KEYCLOAK_HOST": "keycloak:8443",
                      },
--- a/tests/e2e/cucumber/environment/index.ts
+++ b/tests/e2e/cucumber/environment/index.ts
@@ -5,10 +5,13 @@ import {
   setWorldConstructor,
   ITestCaseHookParameter,
   AfterAll,
-  After
+  After,
+  Status
 } from '@cucumber/cucumber'
 import pino from 'pino'
 import { Browser, chromium, firefox, webkit } from '@playwright/test'
+import path from 'path'
+import fs from 'fs'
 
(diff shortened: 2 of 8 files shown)


.drone.star
```diff
@@ -1829,7 +1829,7 @@ def e2eTestsOnKeycloak(ctx):
                          "BASE_URL_OCIS": "ocis:9200",
                          "HEADLESS": "true",
                          "RETRY": "1",
-                         "REPORT_TRACING": "true",
+                         "REPORT_TRACING": "with-tracing" in ctx.build.title.lower(),
```
ScharfViktor, line 1832: sorry I don't get it. That means we only get tracing for keycloak test is we set `with-tracing` to the PR header?

tests/e2e/cucumber/environment/index.ts
```diff
@@ -145,9 +154,35 @@ AfterAll(async () => {
   if (state.browser) {
     await state.browser.close()
   }
+
+  if (config.reportTracing) {
+    // move failed tracing reports
+    const failedDir = path.dirname(config.tracingReportDir) + '/failed'
```
ScharfViktor, line 160: should we change path here?
https://github.com/owncloud/web/pull/11571/files#diff-6a968d4b7901c8526c8d9a86d1f434737dbfe2b00da21a7d3041ed4e3d58d62eR1512-R1513

to `reports/e2e/playwright/tracing/failed`

.drone.star
saw-jan: we get tracing if there's with-tracing in the title or tracings are enabled if a scenario is retried after failing.

tests/e2e/cucumber/environment/index.ts
saw-jan: All failed tracings will be moved to the original path after all tests are executed

saw-jan: So that means there will be no change in the tracing paths at the end

ScharfViktor (approved):
👍

PR #11584, opened 2024-09-16
fix: idp logout issues when token renewal fails

JammingBen:
## Description
Adds a silent signIn retry when token renewal fails for some reasons. If the retry fails again, then force the user to logout. This is the "old" behavior from `stable-8.0` for the most part, expect for the forced logout.

We might want to force the logout in other scenarios as well, e.g. when the webfinger app receives a `401` status code from the server. This might solve the webfinger issues we have on newer instances, although I can't confirm that.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11478
- Fixes https://github.com/owncloud/ocis/issues/10038

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/services/auth/authService.ts
+++ b/packages/web-runtime/src/services/auth/authService.ts
@@ -145,7 +145,7 @@ export class AuthService implements AuthServiceInterface {
             this.handleAuthError(unref(this.router.currentRoute), { forceLogout: true })
           }
 
-          // retry silent signin once, force logoout if it fails
+          // retry silent signin once, force logout if it fails
           this.userManager.signinSilent().catch(handleExpirationError)
         })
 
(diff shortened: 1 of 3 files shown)


kulmann (approved):
💯

packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -140,8 +140,13 @@ export class AuthService implements AuthServiceInterface {
...
       if (!this.userManager.areEventHandlersRegistered) {
         this.userManager.events.addAccessTokenExpired((...args): void => {
-          console.error('AccessToken Expired：', ...args)
-          this.handleAuthError(unref(this.router.currentRoute))
+          const handleExpirationError = () => {
+            console.error('AccessToken Expired：', ...args)
+            this.handleAuthError(unref(this.router.currentRoute), { forceLogout: true })
+          }
+
+          // retry silent signin once, force logoout if it fails
```
AlexAndBear, line 148:
```suggestion
          // retry silent signin once, force logout if it fails
```

AlexAndBear (approved):

Issue #10724, opened 2024-04-05
[a11y] 9.1.3.1h Labeling of form elements can be determined programmatically


tschiebel:
audit: https://infinite.owncloud.com/s/PnaAukHGGsLnrMM

https://bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-1-3-1h-beschriftung-von-formularelementen-programmatisch-ermittelbar

Programmatic linking of labels with form elements takes place with
"label" elements via the "for" attribute or the inclusion of the labeled form element in the
form element in the "label" element. If labels are not marked with the "label" element
element, a label for the associated form element should be created in another way
other way (e.g. via the "aria-labelledby" attribute).

### Personal: Partially fulfilled.

- [ ] When focusing the slider "Tile size" in the menu "Customization options for the
Show file list customization options" menu, the accessible name is not displayed because the
programmatic linking using the "for"/"id" attributes does not work. The
corresponding "id" attribute is missing in the "input" element.
![image](https://github.com/owncloud/web/assets/132444655/c8e3db5b-6c75-4b39-b50f-c19b9a384e37)

### Spaces: Partially fulfilled.

- [ ]  The "Filter members" field in the "Members" section is not accessible as it is not programmatically
is not programmatically linked to a label. Although a "label"
element is present, it does not contain any text. A placeholder is not a suitable
replacement for a label, as it is not output in many cases. 
![image](https://github.com/owncloud/web/assets/132444655/10d81574-5284-42b8-b4c3-ae58e7a5a8a7)

- [ ] See commentary on Personal

Info: Shares, Deleted Files and Account: Fulfilled

Issue #10729, opened 2024-04-05
[a11y] 9.1.4.11 Sufficient contrast of graphics and graphic control elements


tschiebel:
audit: https://infinite.owncloud.com/s/PnaAukHGGsLnrMM

https://bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-1-4-11-kontraste-von-grafiken-und-grafischen-bedienelementen-ausreichend

Note: Information-bearing graphics and graphic control elements as well as their states should
states should have a contrast of at least 3:1 to adjacent colors.

### Personal

- [ ] The non-activated checkboxes (frames) do not fulfill the contrast requirements
for graphic control elements, as the current contrast value of 1.3:1
is lower than the required value of 3:1.

- [ ]  The "Search" input field (frame) in the "Enable" area does not meet the contrast requirements for
contrast requirements for graphic control elements, as the current contrast value of
contrast value of 1.2:1 is lower than the required value of 3:1.

### Shares

- [ ] See commentary on Personal.

### Spaces
It is likely that we will have a similar problem with checkboxes on this page as on pages 1 and 2.
as on pages 1 and 2. In the test, the contrasts were OK,
However, this was due to the fact that dark background images were used.

### Deleted Files

- [ ] The "Search" input field (frame) does not fulfill the contrast requirements for graphic
control elements, as the current contrast value of 1.2:1 is lower than the required
required value of 3:1.

### Account

- [ ]  In the "Change password" modal dialog, all input field frames (frames)
do not meet the contrast requirements for graphical control elements, as the current contrast value of
contrast value of 1.2:1 is lower than the required value of 3:1.

- [ ] See comment on Personal.

pascalwengerter:
@tbsbdr this needs an update of the theming variables? Appearently, the input-background color, input-border color and "general" background-color don't fulfill the necessary contrast ratio

JammingBen:
Has been fixed via the linked PRs.

PR #11591, opened 2024-09-17
fix(a11y): contrast, checkbox focus and space member search label

JammingBen:
## Description
* Fixes the issue where the current focus would not change when navigating the resource table/tiles with the keyboard.
* Fixes the border color contrast for focused inputs. The border colors for the inputs in general need a color token change and therefore need a change in ocis (see https://github.com/owncloud/ocis/pull/10076).
* Redesigns the space member search in the right sidebar so it has a label. Also moves the search from the top of the sharing panel to the members section where all members are listed.

image

## Related Issue

- refs https://github.com/owncloud/web/issues/10729
- fixes https://github.com/owncloud/web/issues/10724

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/design-system/src/components/_OcTableRow/_OcTableRow.vue
+++ b/packages/design-system/src/components/_OcTableRow/_OcTableRow.vue
@@ -1,7 +1,6 @@
 <template>
   <tr
     ref="observerTarget"
-    tabindex="-1"
     @click="$emit('click', $event)"
     @contextmenu="$emit('contextmenu', $event)"
     @dragstart="$emit('dragstart', $event)"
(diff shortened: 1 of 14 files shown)


packages/design-system/src/components/_OcTableRow/_OcTableRow.vue
```diff
@@ -1,7 +1,6 @@
 <template>
   <tr
     ref="observerTarget"
-    tabindex="-1"
```
JammingBen, line 4: I don't see a reason to set `tabindex="-1"` here. A `tr` can't be focused and we don't have a scenario where we want to focus it programmatically.

kulmann (approved):
Awesome enhancements! 😍

Issue #11602, opened 2024-09-18
Redesign link share sidebar section


JammingBen:
We want to redesign the link share sidebar section because of recent changes (new permanent link, no more quick link).

* find a new place for the permanent link so it's always visible
* add a link icon to the permanent link
* align link design with user shares
* rename link section title to something like "Public available"
* rewrite contextual helper for public sharing
* don't show "Add link" twice

cc @tbsbdr

PR #11606, opened 2024-09-19
feat: redesign link sharing panel in right sidebar

JammingBen:
## Description
Aligns the design of the link sharing section with the people sharing section. Also refactors the code so both sections use the same structure (`LinkShares` > `ListItem` > `EditDropdown`).

Additionally, moves the permanent link to the top right of the sharing panel.

image

image

image

## Related Issue

- closes https://github.com/owncloud/web/issues/11602

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


tbsbdr:
nice! could you add more spacing to the icons? already fulfilled https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html but its more convenient to have more space.
screenshot_003150

packages/web-app-files/src/components/SideBar/Shares/Links/ListItem.vue
```diff
@@ -31,22 +31,23 @@
           v-if="linkShare.hasPassword"
           v-oc-tooltip="$gettext('This link is password-protected')"
           name="lock-password"
-          class="oc-files-file-link-has-password"
+          class="oc-files-file-link-has-password oc-mr-xs"
```
tbsbdr, line 34: I think I would make sense to have a ~xtra-small~ small margin here to have even spacing between the icons
screenshot_003159

tbsbdr (dismissed):

packages/web-app-files/src/components/SideBar/Shares/Links/ListItem.vue
JammingBen: I'm pretty sure it's even 🤔  All elements have `xs` margin, resulting in a total margin of `s` between the icons. It appears to be different in that case because the 3-dots-icon is so narrow compared to the other icons.

tbsbdr: ah , that thing, you are right 🙈

kulmann (approved):
love it 😍

PR #11607, opened 2024-09-19
Allow extensions to announce contentType to AppWrapper, make saving images possible

pascalwengerter:
## Description
Necessary for https://github.com/owncloud/web-extensions/pull/43

### DIFF
--- a/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
+++ b/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
@@ -36,7 +36,9 @@ import {
   computed,
   onMounted
 } from 'vue'
+import toNumber from 'lodash-es/toNumber'
 import { DateTime } from 'luxon'
+import { dirname } from 'path'
 import { useTask } from 'vue-concurrency'
 import { useGettext } from 'vue3-gettext'
 import { onBeforeRouteLeave, useRouter } from 'vue-router'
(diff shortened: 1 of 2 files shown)


dschmidt (commented):
Why do we need this? Can't the extension just use resource.mimeType directly? 

I would like to keep the api surface minimal

dschmidt:
Ah, it's the other way round - the extension needs to tell the mimetype to oC Web because we are only creating a new file?

pascalwengerter:
> Ah, it's the other way round - the extension needs to tell the mimetype to oC Web because we are only creating a new file?

No, it's updating an existing file

pascalwengerter:
> Why do we need this? Can't the extension just use resource.mimeType directly?
> 
> I would like to keep the api surface minimal

Open for suggestions regarding the approach. We could check the `resource.mimeType` (for jpeg/png), but then it's an implicat thing happening in the AppWrapper rathern than the application announcing what it wants to do (albeit the naming/API isn't ideal yet)

pascalwengerter:
@dschmidt do you have a better suggestion with regards to API design? Or @JammingBen maybe?

JammingBen:
Hmmm, I can't really think of a non-ugly solution here. You could give apps the possibility to define a custom save task and call that if given. Though it would be a lot of code duplication in this specific case... so I guess I'm fine with this approach. Maybe add a comment to the `AppWrapper` property to explain how this is used?

PR #11614, opened 2024-09-20
feat: soothe right sidebar panel transitions

JammingBen:
## Description
Soothe right sidebar transitions when opening and closing sub panels.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Visual
(Added by @tbsbdr )

Before and after for better comparison:

https://github.com/user-attachments/assets/94b59897-2de7-40b9-90f6-4ca4cd4d62f2

### DIFF
--- a/changelog/unreleased/enhancement-soothe-sidebar-transitions
+++ b/changelog/unreleased/enhancement-soothe-sidebar-transitions
@@ -1,0 +1,5 @@
+Enhancement: Soothe right sidebar panel transitions
+
+The panel transitions of the right sidebar are a bit smoother now.
+
+https://github.com/owncloud/web/pull/11614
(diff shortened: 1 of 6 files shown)


kulmann (approved):
I have to say that I'm not super hyped about the animation itself. But it is a lot more peaceful like this, so let's try that, definitely an improvement. 🚀 
Another 👍 for fixing the glitch with the sidebar header 😅

JammingBen:
> I have to say that I'm not super hyped about the animation itself. But it is a lot more peaceful like this, so let's try that, definitely an improvement.

Do you have something in mind that you'd like better? It's about 1:1 the animation from e.g. the iOS settings, which I find to be pretty much perfect in terms of subtle but also noticible.

kulmann:
> Do you have something in mind that you'd like better? It's about 1:1 the animation from e.g. the iOS settings, which I find to be pretty much perfect in terms of subtle but also noticible.

Ooooh now we're talking. With your latest change, the root panel is not in a static position anymore, but gets pushed out a little bit to the left. That looks smooth, calm and awesome now 💪

tbsbdr:
I like it a lot, thanks!

PR #11620, opened 2024-09-20
fix: force logout after token expiry retry (backport)

JammingBen:
## Description
This restores some of the behavior we had before https://github.com/owncloud/web/pull/10881.

It is a partial backport of https://github.com/owncloud/web/pull/11584.

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-webfinger/src/views/Resolve.vue
+++ b/packages/web-app-webfinger/src/views/Resolve.vue
@@ -62,7 +62,7 @@ export default defineComponent({
       } catch (e) {
         console.error(e)
         if (e.response?.status === 401) {
-          return authService.handleAuthError(unref(route))
+          return authService.handleAuthError(unref(route), { forceLogout: true })
         }
         hasError.value = true
       }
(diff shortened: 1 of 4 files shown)


packages/web-app-webfinger/src/views/Resolve.vue
```diff
@@ -62,7 +62,7 @@ export default defineComponent({
       } catch (e) {
         console.error(e)
         if (e.response?.status === 401) {
-          return authService.handleAuthError(unref(route))
+          return authService.handleAuthError(unref(route), { forceLogout: true })
```
JammingBen, line 65: This change here is not on master. I'm not sure if we should add it, but it might be the solution to the webfinger issues. AFAICT this now should behave like it was pre v8.0.2.

dev/docker/apps/jupyter/config.json.dist
```diff
@@ -0,0 +1,5 @@
+{
+  "config": {
```
AlexAndBear, line 2: accidental commit ?

dev/docker/apps/jupyter/manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+  "entrypoint": "jupyter.js"
```
AlexAndBear, line 2: accidental commit ?

AlexAndBear (changes requested):

dev/docker/apps/jupyter/config.json.dist
JammingBen: 🤦

AlexAndBear (approved):

kulmann (approved):

PR #11631, opened 2024-09-23
feat(perf): improve preview loading

JammingBen:
## Description
* Adds a limit of parallel preview requests to prevent the server from being overrun with requests
* Cancels ongoing and queued preview requests when they are aborted (e.g. when leaving the page while previews are loading)
* Adds lazy loading for space images
* Prevents the image resource from being fetched for each space
* Uses caching for space images

Also introduces a global `loadPreview` method and removes the visibility observers in each view. Previews are now loaded event-based instead, which should result in a better performance overall.

## Related Issue

- refs https://github.com/owncloud/ocis/issues/10054

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


packages/web-app-files/src/components/Search/List.vue
```diff
@@ -108,7 +108,13 @@
           :fields-displayed="['name', 'size', 'tags', 'mdate']"
           :resource-dom-selector="resourceDomSelector"
           @file-click="triggerDefaultAction"
-          @row-mounted="rowMounted"
+          @item-visible="
+            loadPreview({
+              space: getMatchingSpace($event),
+              resource: $event,
+              ignoreAlmostEmptyTxtFiles: true
```
kulmann, line 115: Huh, does this feature still exist? I thought we killed it.

JammingBen: We ignore almost empty text files in the file lists apparently. Should I kill it? 😂

kulmann: Yeah I remember when we built the feature. But the added value is minimal while adding unnecessary complexity. So I thought we killed it already. If not I'd vote for removing the feature from the code base.

packages/web-pkg/src/components/SideBar/Spaces/Details/SpaceDetails.vue
```diff
@@ -167,16 +142,36 @@ export default defineComponent({
...
+          return
+        }
+
+        const imageResource = buildSpaceImageResource(unref(resource))
+        spaceImage.value = await loadPreview({
+          space: unref(resource),
+          resource: imageResource,
+          dimensions: ImageDimension.Tile,
+          processor: ProcessorType.enum.fit,
+          cancelRunning: true,
```
kulmann, line 158: Really an edge case, but couldn't it happen now that you show space details from the space overview while previews are still loading and then that would kill all the remaining preview requests in the space overview?

JammingBen: `cancelRunning` would only cancel preview requests within this component. The space overview, while also using the `useLoadPreview` composable, has a different "instance" of it. So technically there are 2 separate tasks.

kulmann: Aha! nice, thanks for the explanation 👍

kulmann (approved):
So awesome 😍

Issue #11381, opened 2024-03-08
Change wording of Role Descriptions


kulmann:
## Describe the bug
OnlyOffice can write documents from a public link share with "Can upload" role.

## Steps to reproduce
1. As "admin", create `asdf/test.docx` via OnlyOffice
2. As "admin", create a public link to the folder asdf with `Can upload` role (which states permissions `view, download and upload`)
3. As anonymous, receive and follow the public link
4. As anonymous, open `asdf/test.docx` via OnlyOffice
5. As anonymous, write some content into the file and leave the document again
6. As "admin", see that the changes have been written to disk (e.g. by downloading the file after the lock has been released)

## Expected behavior
"anonymous" should have a read only view or at least an error message when trying to save.

## Actual behavior
"anonymous" can edit and save the document.

kulmann:
> @kulmann So I guess this would be a web ticket now? On the server side we don't have any descriptive texts for the public link types AFAICS.

Seems like it, yes. Transferred it to web now.

PR #11661, opened 2024-09-26
feat: adjust link type descriptions and remove "Uploader" type

JammingBen:
## Description
As per title.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11381

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/enhancement-remove-uploader-link-type
+++ b/changelog/unreleased/enhancement-remove-uploader-link-type
@@ -1,0 +1,6 @@
+Enhancement: Remove link type "Uploader"
+
+The link type "Uploader" has been removed because it's almost the same as "Editor". Existing links of this type still work though.
+
+https://github.com/owncloud/web/pull/11661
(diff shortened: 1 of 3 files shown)


JammingBen:
@tbsbdr Some annotations to https://github.com/owncloud/web/issues/11381#issuecomment-2301724340:

> Adds `edit` to `Can upload`

This would not be correct since editing is currently not allowed for this link role.

> Note: In general I'd prefer to remove the Uploader role at all, as it adds little value for the user, but if I understand @micbar correctly this would be expensive due to a migration necessity.

We could give it the same treatment like the internal link, meaning the uploader role wouldn't be available for setting anymore, but existing links with this role still exist and work.

JammingBen:
As discussed with @tbsbdr : We remove the "Uploader" link role while keeping compatibility with existing links with this role.

kulmann (approved):

PR #11665, opened 2024-09-26
[full-ci] chore: bump version to v10.3.0

kulmann:
bump version to v10.3.0


JammingBen (approved):

JammingBen (changes requested):
Ah, please update the versions in `web-pkg` and `web-client` as well so we can release them together.

JammingBen (approved):

Issue #11678, opened 2024-09-30
`X-Request-ID`s missing in error messages from web workers


kulmann:
## Describe the bug
Using `copy & paste` uses the `paste` web worker.

## Steps to reproduce
1. Set the personal space quota to `0.1` (100 MB)
2. Upload files so that the quota is maxed out
3. Create a project space `test`
4. Upload a "large" file into the space `test` (file size that doesn't fit into the remaining personal space quota)
5. Copy the file
6. Navigate into the personal space
7. Paste the file

## Expected behavior
The error message in the ui shows all available `X-Request-ID`s

## Actual behavior
The error message in the ui shows `X-Request-ID: undefined` (see screenshot)

The `MessageEvent` doesn't contain the response headers, thus there is no `X-Request-ID` to be extracted. https://github.com/owncloud/web/blob/f172060798ca6a03d0061c6c490397c7b41665df/packages/web-pkg/src/composables/webWorkers/pasteWorker/usePasteWorker.ts#L37

<img width="406" alt="Screenshot 2024-09-30 at 11 53 02" src="https://github.com/user-attachments/assets/de759b73-5ec0-4eaf-b997-1a01e7b87c06">

PR #11684, opened 2024-09-30
fix: error parsing in delete- and paste-worker

JammingBen:
Errors in workers need to be constructed manually in the main thread since they can't be serialized and passed from the worker to the main thread.

This also fixes the issue with the missing request ids in failing copy, move and delete requests.

fixes https://github.com/owncloud/web/issues/11678


AlexAndBear (approved):
🥇

kulmann (approved):

Issue #11677, opened 2024-09-30
Tags are not displayed on shared resources


ScharfViktor:
after debugging tag.feature:178 in https://github.com/owncloud/ocis/pull/10172 I found bug

use https://ocis.ocis-web.master.owncloud.works/
```
ocis | 6.4.0
web 10.3.0
```

Steps: 
- admin creates space `s1` and folder `f1` inside the space
- admin adds `tag1` and `tag2` to folder `f1`
- admin shares folder `f1` to einstein with edit permission
- einstein opens `f1`

Expected: einstein sees tags `tag1` and `tag2` to folder `f1`
Actual: no tags
<img width="1776" alt="Screenshot 2024-09-30 at 10 18 44" src="https://github.com/user-attachments/assets/19e38039-f9a2-448c-bb03-79540d8ca8a4">

PR #11703, opened 2024-10-02
fix: add missing resource tags on 'Shared with me' page

JammingBen:
Tags are currently not available on the Graph API, hence we need to manually take them from the webdav response.

fixes https://github.com/owncloud/web/issues/11677

### DIFF
--- a/packages/web-pkg/src/components/SideBar/FileSideBar.vue
+++ b/packages/web-pkg/src/components/SideBar/FileSideBar.vue
@@ -383,13 +383,17 @@ export default defineComponent({
 
         // shared resources look different, hence we need to fetch the actual resource here
         try {
-          let fullResource = await clientService.webdav.getFileInfo(props.space, {
+          const webDavResource = await clientService.webdav.getFileInfo(props.space, {
             path: resource.path
           })
 
-          // make sure props from the share (=resource) are available on the full resource as well
-          fullResource = { ...fullResource, ...resource }
-          loadedResource.value = fullResource
+          // make sure props from the share (=resource) are available on the merged resource
+          const mergedResource = {
+            ...webDavResource,
+            ...resource,
+            tags: webDavResource.tags // tags are always [] in Graph API, hence take them from webdav
+          }
+          loadedResource.value = mergedResource
         } catch (error) {
           loadedResource.value = resource
           console.error(error)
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -383,13 +383,14 @@ export default defineComponent({
...
+          const webDavResource = await clientService.webdav.getFileInfo(props.space, {
             path: resource.path
           })
 
-          // make sure props from the share (=resource) are available on the full resource as well
-          fullResource = { ...fullResource, ...resource }
-          loadedResource.value = fullResource
+          // make sure props from the share (=resource) are available on the merged resource
+          const mergedResource = { ...webDavResource, ...resource }
+          mergedResource.tags = webDavResource.tags // tags don't exist in Graph API yet, hence take them from webdav
```
kulmann, line 392: I don't get it... why are the tags not present from spreading the webdav resource and the share resource into a new object? 🙈

JammingBen: Because tags on a share resource (= a resource from the Graph API) are always `[]`. So it overwrites any potential tags from the `webDavResource`.

It's quite messy that we need to merge resources from both APIs with different data here :( But I guess that's the easiest "fix" until we fully switch to Graph.

AlexAndBear: IHMO ...resource is overwriting the field, no ?

AlexAndBear: @JammingBen did you try with lodash merge, IDK but it could be potentially smarter ... ?

kulmann: Ah, I didn't know that the graph resource also has the `tags` field with an empty array... maybe you can add that  to the comment. :-)

packages/web-pkg/src/components/SideBar/FileSideBar.vue
```diff
@@ -383,13 +383,14 @@ export default defineComponent({
...
-          let fullResource = await clientService.webdav.getFileInfo(props.space, {
+          const webDavResource = await clientService.webdav.getFileInfo(props.space, {
             path: resource.path
           })
 
-          // make sure props from the share (=resource) are available on the full resource as well
-          fullResource = { ...fullResource, ...resource }
-          loadedResource.value = fullResource
+          // make sure props from the share (=resource) are available on the merged resource
+          const mergedResource = { ...webDavResource, ...resource }
```
kulmann, line 391:
```suggestion
          const mergedResource = { ...webDavResource, ...resource, tags: webDavResource.tags }
```

Would do that as a one-liner :-)

AlexAndBear (approved):

PR #11706, opened 2024-10-02
Escape html in ActivitiesPanel and Notificiations views

### LINKED ISSUE #11705: displayname of the user can contain HTML tags and the string is not filtered in some views
## Describe the bug
The string from 'first and lastname'  of a user are shown unfiltered in the notifications
Luckily only the admin can set that data

## Steps to reproduce
1. as admin set the first and lastname of `einstein` to `<a href="http://jankari.tech">einstein</a>`
2. as `einstein` share a folder to `marie`
3. as `marie` check the notifications

## Expected behavior
the HTML string should be encoded as e.g. the sharing dialog
grafik

## Actual behavior
a link is rendered in the notifications
grafik

## Setup

```console
Version 	6.4.0
Web client version 	10.2.0

```

AlexAndBear:
## Related Issue

- Fixes #11705 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-escape-html-characters-in-activities-and-notification-view
+++ b/changelog/unreleased/bugfix-escape-html-characters-in-activities-and-notification-view
@@ -1,0 +1,7 @@
+Bugfix: Escape HTML characters in activities and notification view
+
+We've fixed a bug where HTML characters were not escaped in the activities and notification view.
+This could lead to potential XSS attacks.
+
+https://github.com/owncloud/web/pull/11706
+https://github.com/owncloud/web/issues/11705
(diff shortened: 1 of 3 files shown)


packages/web-app-files/package.json
```diff
@@ -15,6 +15,7 @@
     "axios": "1.7.7",
     "design-system": "workspace:@ownclouders/design-system@*",
     "email-validator": "^2.0.4",
+    "escape-html": "^1.0.3",
```
kulmann, line 18: Please check if we can use `escape` from `lodash-es` instead of installing a new lib.

JammingBen (approved):

Issue #11698, opened 2024-10-01
Move `web-app-importer` to the web-extensions repo


kulmann:
# Description
The `web-app-importer` app currently is part of the monorepo because we didn't have anything else properly set up at the time of app development. It is not a core functionality and as such doesn't belong in the web monorepo. We want to move the entire app to a different repo (either its own repo or into the `web-extensions` repo, preferably the latter if nothing speaks against it) and release it like all our other non-core apps, because it makes kubernetes deployments easier to have the app assets available as docker image and then use the app config via the ocis `apps.yaml` file.

## User Stories

* > As a devops person, I want to have the `web-app-importer` app assets available as docker image so that I can use the usual app config mechanism.

## Value
Easy deployment.

## Acceptance Criteria
- Move `web-app-importer` into the `web-extensions` repo
- Release it like all the other apps in that repo
- Make sure this works for the current master of ocis and web / the latest rolling release
- We currently don't want to publish the app in the app store. It seems to serve an edge case only.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

JammingBen:
Has been solved by the linked PRs.

PR #11709, opened 2024-10-02
feat!: remove importer from default apps

JammingBen:
BREAKING CHANGE: removes `importer` from the default web apps. The app is now available via the [web-extensions repository](https://github.com/owncloud/web-extensions). See https://github.com/owncloud/web-extensions/pull/61.

refs https://github.com/owncloud/web/issues/11698


dev/docker/ocis.web.config.json
```diff
@@ -28,18 +28,7 @@
...
-          "image/x-ms-bmp"
-        ]
-      }
-    },
-    {
-      "id": "importer",
-      "path": "web-app-importer",
-      "config": {
-        "companionUrl": "https://host.docker.internal:9200/companion"
+        "mimeTypes": ["image/tiff", "image/bmp", "image/x-ms-bmp"]
```
kulmann, line 31: Not your code, but we have these three mime types hardcoded for thumbnails and for the preview app by now (see e.g. the `mimeTypes.ts` file in the preview app). So we can remove the `preview` app from the external apps section again and just add `preview` to the string-array under the `apps` key. Could you do that here? Only dev stack, so it doesn't need changelog anyway.

kulmann (commented):
Can't we go back to upstream `uppy` now with companion being removed from the web repo? Or is it still needed in the runtime for the `importer` app to work correctly?

JammingBen:
> Can't we go back to upstream `uppy` now with companion being removed from the web repo? Or is it still needed in the runtime for the `importer` app to work correctly?

It's still needed in the runtime to work unfortunately :(

kulmann (approved):
Thanks for taking care 🥇

PR #11726, opened 2024-10-06
Bugfix: Dropdown menu links not working in Safari

### LINKED ISSUE #11718: Main menu does not work in safari
## Describe the bug
Clicking on an item in the main menu (3-by-3 grid at top left) does not open the desired page when using safari.

Likely reason: The menu disappears on mouse down, but the link would only be clicked on mouse up.

## Steps to reproduce
1. Open OCIS in Safari (I am using the admin user created during installation).
2. Click on the main menu (3-by-3 grid at top left. 
3. The menu opens with 4 icons (Files, Text Editor, App Store, Admin Settings)
4. Click on any item in the menu.

## Expected behavior
The clicked-on page opens and the menu disappears.

## Actual behavior
The menu disappears, but the page is not changed.

## Setup
I used a fresh install of OCIS 6.5.0 with Web client version 10.3.0. I used `sudo -u ocis ocis init --config-path /containers/ocis/config --admin-password "XXXXX" --insecure true` to configure the server.
The server runs on a machine in my local network and is behind a nginx reverse proxy with the recommended additional settings:
```
        # OIDC Tokens in headers are quite large and can exceed default limits of reverse proxies
        proxy_buffers 4 256k;
        proxy_buffer_size 128k;
        proxy_busy_buffers_size 256k;

        # Disable checking of client request body size
        client_max_body_size 0;
``` 
## Additional context
- The problem only ocurs with safari [I currently use Version 16.6 (18615.3.12.11.2)]. With chrome everything works as expected.
- I do not see any errors in the safari developer console, except a failure to load `/ocs/v2.php/apps/notifications/api/v1/notifications/sse`, which I guess is not related.
- Open the pages by entering the respective url in the browser's menu bar, works. E.g. going to `admin-settings/general` loads the  "Admin Settings". The menu also does not work on these pages.

In safari, the menu disappears immediately when I press the mouse button. In chrome, the menu stays open when I press the mouse button. The menu item is only selected/clicked when I release the mouse button. I am unsure if the menu is closed on mouse up or only when the new page is loaded.

AlexAndBear:
## Related Issue

- Fixes #11718 
## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcDrop/OcDrop.vue
+++ b/packages/design-system/src/components/OcDrop/OcDrop.vue
@@ -138,7 +138,10 @@ export default defineComponent({
     }
 
     const onFocusOut = (event: FocusEvent) => {
-      const focusLeft = event.relatedTarget && !unref(drop).contains(event.relatedTarget as Node)
+      const focusLeft =
+        event.relatedTarget &&
+        !unref(drop).contains(event.relatedTarget as Node) &&
+        !(event.relatedTarget as HTMLElement)?.classList?.contains('tippy-box')
       if (focusLeft) {
         // close drop when the focus leaves it
         hide()
@@ -171,7 +174,6 @@ export default defineComponent({
     position() {
       this.tippy.setProps({ placement: this.position })
     },
-
     mode() {
       this.tippy.setProps({ trigger: this.triggerMapping })
     }
@@ -268,6 +270,7 @@ export default defineComponent({
     // note: needed so that the box shadow from `oc-box-shadow-medium` doesn't get suppressed
     padding: 8px;
   }
+
   li.oc-menu-item-hover {
     a,
     .item-has-switch,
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -138,7 +138,8 @@ export default defineComponent({
     }
 
     const onFocusOut = (event: FocusEvent) => {
-      const focusLeft = event.relatedTarget && !unref(drop).contains(event.relatedTarget as Node)
+      const tippyBox = unref(drop).closest('.tippy-box')
```
AlexAndBear, line 141: focusOut event was triggered before the click event and was not bubbling the event up when the hide function was triggered. 
In Chrome/FF the related event was the a tag or button itself, on Safari it was the tippy box. 

Everything is still working fine

kulmann (approved):
Awesome findings! Thanks for fixing this ❤️

PR #11730, opened 2024-10-07
chore: prepare web-test-helpers package for publish

JammingBen:
Prepares the `web-test-helpers` package to be published under the name `@ownclouders/web-test-helpers`.

Because we currently don't publish the ODS, it can't be provided as peer dependency, so we need to ship it with the bundle.

Also, this introduces a circular dependency since the ODS imports the test-helpers and vice versa. An easy way would be to use the published version of the test-helpers in the ODS. However, I don't know if this is really an issue since I don't see any problems and pnpm officially supports this.

This PR is best reviewed by its commits. One commit contains the adjusted imports, hence the big number of changed files. The other one contains the changes that were needed to publish the package.


packages/web-test-helpers/package.json
```diff
@@ -1,16 +1,62 @@
 {
-  "name": "web-test-helpers",
-  "private": true,
+  "name": "@ownclouders/web-test-helpers",
+  "version": "0.1.0",
```
kulmann, line 3: What about versioning for publishing? Same strategy as for web-pkg and web-client I'd say... what do you think?

kulmann (approved):
awesome, thank you! 💪

packages/web-test-helpers/package.json
JammingBen: Yep I agree 👍  Now that you approved the changes, let me create a `0.0.1` to test the CI, then I'll add the correct version number.

Issue #10739, opened 2024-04-05
[a11y] 9.4.1.2 Name, role, value available


tschiebel:
audit: https://infinite.owncloud.com/s/PnaAukHGGsLnrMM

https://bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-4-1-2-name-rolle-wert-verfuegbar

General notes:

- [x] The "Collapse/expand sidebar" button does not communicate its status to non-visual users. status to non-visual users. To improve this, the "aria-expanded" attribute should be used. should be used.
![image](https://github.com/owncloud/web/assets/132444655/f0a71025-c5fe-484a-b524-af03a94041b9)
Fixed via https://github.com/owncloud/web/pull/11386

### Personal

- [x] The use of "aria-expanded" attributes together with "span" elements
is not permitted. Further information on this can be found at
https://www.w3.org/TR/wai-aria-1.2/#aria-expanded. The attribute should be removed
be removed.
![image](https://github.com/owncloud/web/assets/132444655/b278d407-2915-4036-ad2f-f22923e05451)

- [x] #11375
is located in a "div" wrapper. It is not recommended to use the "aria-describedby" attribute together with
attribute together with "div" elements, as the output cannot be guaranteed by screen readers.
screen readers cannot be guaranteed.
![image](https://github.com/owncloud/web/assets/132444655/8e0db016-e01b-48c0-80f9-99d1bd311cd3)

- [x] In the tile view, the label of the sort button is provided by the sort value, for example
sorting value, for example "A-Z". The visual label
"Sort by" is not linked to the element, however, and using the "aria-describedby
the "aria-describedby" attribute does not help to understand the purpose of the button.
understand the purpose of the button. It is recommended to implement the element using the HTML "select" element
element and link it programmatically to the visual label,
which was realized by the "label" element.
![image](https://github.com/owncloud/web/assets/132444655/a2408b1c-f0aa-468d-80a0-38b9fb89af36)
- tbsbdr: problem is not understood, postponed.

- [x] In the mobile (responsive) view, the label of the menu button corresponds to the name of the current page. This could be confusing for some users
confusing for some users and it is not clear which page they are on.
Even if the menu is open, the active page (e.g. via the "aria-current" attribute) is not displayed.
attribute) is not displayed.
![image](https://github.com/owncloud/web/assets/132444655/c0911ba8-3c29-4ac4-b113-f49af7e9d0fd)
- tbsbdr: problem is not understood, postponed.

[...]

PR #11747, opened 2024-10-09
fix: various small a11y issues 

JammingBen:
* transforms the select dropdown in the tiles view to a select input to be a11y compliant.
* hides the visible label in the `OcPageSize` component for screen readers since the combobox has its own label.
* removes a duplicated DOM id in quota select
* sets `aria-current` props in mobile nav
* moves the `aria-labelledby` property in modals to the modal body where `role="dialog"` sits.
* removes the `aria-expanded` property on the div that surrounds the share actions because it's invalid on a div-element. This was happening because the access details drop was placed in a wrong position in the DOM.

fixes https://github.com/owncloud/web/issues/10739


kulmann (approved):
Awesome, thank you! 🥇

Issue #11634, opened 2024-09-23
Test coverage for the app store


kulmann:
# Description
The app store has no test coverage at the moment. Needs to change.

## User Stories

* > As a product owner, I want to be certain that crucial functionality of the app store app works so that I can release with confidence.

## Acceptance Criteria
- Cover the most crucial parts of the app store app with unit- and e2e-tests.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues

AlexAndBear:
You want me to takeover here ?

kulmann:
> You want me to takeover here ?

Yes, that would be great, thank you. ❤️ do you want to pair on making a test plan?

AlexAndBear:
That would be awesome 💖

AlexAndBear:
e2e tests are done :)

dj4oC:
## Implementation prompt

**Problem:** The app store e2e journeys and a first batch of component unit tests (PR #11664, PR #11752) are merged, but the pinia stores (`apps.ts`, `repositories.ts`) and the `AppList.vue` overview view still have no unit test coverage.

**Acceptance criteria:**
- `useAppsStore` (`loadApps`, `getById`) has unit tests covering successful load, per-repository fetch failure (caught and logged, returns `[]` for that repo), and sorting of merged results by name.
- `useRepositoriesStore` (`setRepositories`) has a unit test covering state update.
- `AppList.vue` has unit tests covering: no-content message shown when `filteredApps` is empty, search filter term reflected into the `filter` route query on input, and tile rendering for each item in `filteredApps`.

**Files:**
- `packages/web-app-app-store/src/piniaStores/apps.ts` — add `packages/web-app-app-store/tests/unit/piniaStores/apps.spec.ts`
[...]

PR #11752, opened 2024-10-10
App store unit tests (first batch)

kulmann:
Adding a few unit tests for the app store... far from "done", but since I need to switch to a different task I probably won't add more coverage in the next few days. So trying to get the current state into master.

## Issues
Works towards https://github.com/owncloud/web/issues/11634

### DIFF
--- a/packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
+++ b/packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
@@ -1,5 +1,10 @@
 import LayoutContainer from '../../src/LayoutContainer.vue'
-import { defaultComponentMocks, defaultPlugins, mount, nextTicks } from 'web-test-helpers'
+import {
+  defaultComponentMocks,
+  defaultPlugins,
+  mount,
+  nextTicks
+} from '@ownclouders/web-test-helpers'
 import { useAppsStore } from '../../src/piniaStores'
 
 const selectors = {
--- a/packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
+++ b/packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
@@ -1,0 +1,52 @@
+import LayoutContainer from '../../src/LayoutContainer.vue'
+import { defaultComponentMocks, defaultPlugins, mount, nextTicks } from 'web-test-helpers'
+import { useAppsStore } from '../../src/piniaStores'
+
+const selectors = {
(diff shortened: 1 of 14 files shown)


packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (!keepLoadingApps) {
+        return res()
+      }
+      return setTimeout(() => res(), 500)
+    })
+  )
+
+  const mocks = {
+    ...defaultComponentMocks(),
+    loadApps
```
JammingBen, line 39: Do you really need to provide `loadApps` here?

packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+describe('LayoutContainer', () => {
+  it('shows a loading spinner while apps are loading from repositories', () => {
+    const { wrapper } = getWrapper({ keepLoadingApps: true })
+    expect(wrapper.find(selectors.loadingSpinner).exists()).toBeTruthy()
+    expect(wrapper.find(selectors.routerView).exists()).toBeFalsy()
+  })
+  it('renders the router view when loading apps is done', async () => {
+    const { wrapper } = getWrapper({})
+    await nextTicks(2)
```
JammingBen, line 18: Hmm this feels a bit brittle, I guess you wait for the 500 ms to pass here? Maybe you can return the promise in the wrapper and await it here?

JammingBen (commented):
A few small remarks, also `web-test-helpers` need to be changed to `@ownclouders/web-test-helpers`. Rest LGTM 👍

packages/web-app-app-store/tests/unit/LayoutContainer.spec.ts
kulmann: I don't want to let the 500ms pass. I want to stay in that timeframe so that a loading state is still active. Maybe still better to not do that with a timeout.

kulmann: Ah lol, sorry for my confusion. I thought it was the other test. The other test passes a `keepLoadingApps: true` to the wrapper. Only that one has the 500ms timeout. This one here doesn't have a loading delay.

JammingBen (approved):

Issue #11741, opened 2024-10-09
Publish web-pkg with implementation


JammingBen:
We currently publish only the types of `web-pkg`. Up until now, this was sufficient for app development since the runtime provides the actual implementation for apps.

However, this means we can't unit test app components that use functionality of `web-pkg`, since the unit tests run standalone and don't have the runtime behind it. As a result, we need to publish the package with its actual implementation.

PR #11754, opened 2024-10-10
chore(web-pkg): prepare publish with implementation

JammingBen:
Prepares `web-pkg` for a publish with the actual implementation, not just the types. This is needed for unit tests in external apps, since the runtime is not present in this context to provide the actual implementation.

Also moves the peer dependencies to regular dependencies because the app that uses this package should not need to provide all the peers. The ODS is currently an exception to this until we have it published via npm again.

closes https://github.com/owncloud/web/issues/11741

### DIFF
--- a/packages/web-pkg/README.md
+++ b/packages/web-pkg/README.md
@@ -2,8 +2,6 @@
 
 `web-pkg` is a package that provides utilities, most importantly a variety of components and composables, that can be useful when developing apps and extensions for ownCloud Web.
 
-It's important to note that this package solely provides types because `web-pkg` is intended solely for Web extension development. The actual implementation of the utilities is provided by the Web runtime automatically.
-
 ## Installation
 
 Depending on your package manager, run one of the following commands:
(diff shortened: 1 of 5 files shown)


packages/web-pkg/README.md
```diff
@@ -2,8 +2,6 @@
 
 `web-pkg` is a package that provides utilities, most importantly a variety of components and composables, that can be useful when developing apps and extensions for ownCloud Web.
 
-It's important to note that this package solely provides types because `web-pkg` is intended solely for Web extension development. The actual implementation of the utilities is provided by the Web runtime automatically.
```
kulmann, line 5: Should we maybe add a sentence here, that `web-pkg` needs to be a `devDependency`? That's not a new information here, but still, would be good to have it.

kulmann (approved):

packages/web-pkg/README.md
kulmann: that would be great! 💪

PR #11761, opened 2024-10-11
chore: move peers to regular dependencies

JammingBen:
Moves peers of `web-client` and `web-test-helpers` to regular dependencies where it makes sense so users of these packages are not forced to provide them.


kulmann (approved):
👍

PR #11763, opened 2024-10-11
chore: prepare design-system for publish

JammingBen:
Prepares the design-system to be published under `@ownclouders/design-system`.

Also introduces proper namespaces for the package (e.g. `components`, `composables` etc). So a component would be imported like this:

```
import { OcButton } from '@ownclouders/design-system/components'
```

Furthermore it declares the peer dependencies as regular ones for the same reasons as described in https://github.com/owncloud/web/pull/11761.


kulmann (approved):
awesome 🤓

PR #11777, opened 2024-10-15
chore: fix ods build and mark it external in other packages

JammingBen:
`design-system`:
* Fixes the published exports since esm and cjs were mixed up.
* Stubs `webfontloader` and removes it from the external deps to avoid loading fonts in a unit testing context. This would only be a problem when using the ODS standalone, which basically noone does.
* Removes `vue-select` from the external deps because there seems to be something wrong with their package type.

`web-pkg` and `web-test-helpers`:
* Removes the pre-bundling of the design-system and mark it external since it's now published.

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -51,23 +51,23 @@
     "linkDirectory": false,
     "exports": {
       ".": {
-        "default": "./dist/design-system.js",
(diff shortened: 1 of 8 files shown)


kulmann (approved):
Thank you for digging through this... everything coming together now, that's awesome! 🤩

PR #11782, opened 2024-10-16
chore: fix unmet peer warnings when using extension-sdk and eslint-config

JammingBen:
* `extension-sdk`: Aligns the extension-sdk version with the one of the mono repo and its packages. Also adds a version range to the `sass` dependency so users of this package don't need to provide the pinned version exactly.
* `eslint-config`: Inscreases the minimum version of `@babel/eslint-parser` to `7.25.8` because this ensures compatibility with eslint v9. The previous, older version didn't do that.


kulmann (approved):
💪

PR #11795, opened 2024-10-18
fix: sharing label for locked files

JammingBen:
Fixes an issue where a locked file would be falsely labelled as shared in the right sidebar.

refs https://github.com/owncloud/enterprise/issues/6915


kulmann (approved):
Nice 💪

Issue #11803, opened 2024-10-21
[QA] files appear in parent folder during web upload


jnweiger:
## Steps to reproduce
1. prepare a folder "5000p" with 5000 PNG images on a local machine
2. log in to an ocis instance via web ui, create a "sub" folder
3. upload "5000p" into "sub"
4. while the upload is in progress, use another browser window to
5. visit "sub" -> the folder "5000p" is there. OK
6. refresh this view a few times. Only the folder "5000p" is there. OK
7. visit "5000p" -> a file listing appears. OK
8. click back to "sub" -> some files appear there, next to "5000p". BAD
9. refresh. The files disappear, only "5000p" remains.
10. This can be repeated as long as the upload is in progress. Okayish.
11. Afterwards, no glitches occur

## Expected behavior
All files only appear in the correct folder, where they really, are, regardless what the users click in the web ui

## Actual behavior
Clicking the parent "sub" folder via breadcrumbs, some files appear there during upload, that should not.

https://github.com/user-attachments/assets/52fccb87-15b5-4d84-a429-28ec94944d0d

PR #11824, opened 2024-10-28
fix: files wrongly appearing in the file list

JammingBen:
Fixes the issue with files wrongly appearing in the file list during an upload. This was happening because of SSE since a check for the current folder was missing.

fixes https://github.com/owncloud/web/issues/11803


AlexAndBear (approved):
👍

kulmann (approved):
💪

Issue #11787, opened 2024-10-17
AppWrapper doesn't clear ancestor metadata


kulmann:
## Describe the bug
As found out in https://github.com/owncloud/web/pull/11775#issuecomment-2419192308
When opening a template file in a project space while having the sidebar open, the shares panel still shows the project space members, despite the new file being in the personal space. This happens because the ancestors and therefore the loaded space members don't get cleared. 
At the moment this is an edge case, because usually the context stays identical when navigating from the files app into an app.

## Steps to reproduce
1. Create a space "templates"
2. Upload a file "example.dotx" to the space
3. Open the right sidebar
4. Left-click on it to trigger the action for creating a document from a template

## Expected behavior
The right sidebar has the correct context of the personal space root, where the new document is being created.

## Actual behavior
The right sidebar still has the "templates" space as context.

We need to clear ancestor metadata.

PR #11831, opened 2024-10-29
fix: clear ancestor data in app wrapper

JammingBen:
This fixes an issue where the right sidebar would display wrong shares when being inside a space and opening a file from another space (e.g. via the search results).

fixes https://github.com/owncloud/web/issues/11787


kulmann (approved):
LGTM 💪  Thank you for fixing this!

PR #11843, opened 2024-11-01
Remove oc10 from the dev docs, restructure

pascalwengerter:
## Description
See PR title and code changes

## Motivation and Context
oC10 isn't supported as of web version ~`7.x.x`, and we're sitting at `v11.0.0` now. Let's remove the mentions and maybe work on the dev doc structure in the process


docs/getting-started.md
```diff
@@ -97,13 +97,29 @@ integration libraries.
...
-- [Running with oCIS as backend]({{< ref "backend-ocis.md#running-web" >}})
-- [Running with ownCloud as backend]({{< ref "backend-oc10.md#running-web" >}})
+### Running Web
+
+- Start bundling web with a watcher by running `pnpm build:w`
+
+### Setting up oCIS
+
+- Setup oCIS by following the [setup instructions](https://owncloud.dev/ocis/getting-started/)
+- Start oCIS with local links to your bundled web frontend and config by running `WEB_ASSET_CORE_PATH=../../web/dist WEB_UI_CONFIG_FILE=../../web/dist/config.json OCIS_INSECURE=true IDM_CREATE_DEMO_USERS=true ./bin/ocis server` (and make sure to adjust paths as necessary)
```
kulmann, line 115: This guide is not about setting up a production ocis with web, but about setting up a working dev stack. Shouldn't we just explain the docker-compose stack from the monorepo then? I'd like to refrain from describing an ocis binary with web setup (that can also be needed, but it's more of an edge case...)

kulmann (changes requested):

PR #11849, opened 2024-11-05
Create spaces via admin settings

### LINKED ISSUE #11771: Allow creating spaces in admin settings
Currently, spaces can be disabled and deleted in two ways:
1. via the files app
2. via the admin settings (batch mode)

New spaces can only be created in the files app.

What is the rationale behind this? Why do we have two places ways to do the same thing? Why can we not create spaces in the admin settings?

Can we merge the batch management into the files app? Or should we add support for creating spaces to the admin settings as well?

In the future I think admins may want to decide which storage provider should host a new space ... but even that might be possible to do in the files app...

AlexAndBear:
## Description
Screenshot 2024-11-05 at 13 04 15

## Related Issue

- Fixes #11771 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-admin-settings/src/views/Spaces.vue
+++ b/packages/web-app-admin-settings/src/views/Spaces.vue
@@ -13,6 +13,13 @@
       :batch-action-items="selectedSpaces"
       :show-view-options="true"
     >
+      <template #topbarActions="{ limitedScreenSpace }">
+        <create-space
+          v-if="hasCreatePermission"
+          :show-label="!limitedScreenSpace"
+          @space-created="(space) => spaceSettingsStore.upsertSpace(space)"
+        />
+      </template>
       <template #sideBarHeader>
         <space-info
           v-if="selectedSpaces.length === 1"
@@ -71,7 +78,9 @@ import {
   useSpaceActionsRestore,
   useSpaceActionsEditQuota,
   AppLoadingSpinner,
-  useSharesStore
+  useSharesStore,
+  useAbility,
+  CreateSpace
 } from '@ownclouders/web-pkg'
 import { call, SpaceResource } from '@ownclouders/web-client'
 import { computed, defineComponent, onBeforeUnmount, onMounted, ref, unref } from 'vue'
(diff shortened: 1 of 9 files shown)


packages/web-pkg/src/components/AppBar/CreateSpace.vue
```diff
@@ -2,14 +2,14 @@
   <oc-button
     id="new-space-menu-btn"
     key="new-space-menu-btn-enabled"
-    v-oc-tooltip="$gettext('Create a new space')"
+    v-oc-tooltip="showTooltip ? $gettext('New space') : undefined"
```
kulmann, line 5: When you show a label:
- aria-label must not be present because the button label already says it all
- tooltip also can be removed in that case

When you're not showing a label:
- aria-label and tooltip are supposed to be the same content (And I prefer the longer `Create a new space` tooltip over the shorter one actually).


No other combination allowed a11y-wise :-)

AlexAndBear: `Create a new space` wording would clash with the other create buttons (e.G New User) in the admin settings, adjusted the other remarks.

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -81,10 +89,12 @@ import { useGettext } from 'vue3-gettext'
 import { useSpaceSettingsStore } from '../composables'
 import { storeToRefs } from 'pinia'
 import { Quota } from '@ownclouders/web-client/graph/generated'
+import CreateSpace from '@ownclouders/web-pkg/src/components/AppBar/CreateSpace.vue'
```
JammingBen, line 92:
```suggestion
import { CreateSpace } from '@ownclouders/web-pkg'
```

AlexAndBear: Good point 👍 , but wrong place

JammingBen (approved):
LGTM 👍

PR #11862, opened 2024-11-06
[full-ci]enable vips for thumbnails

ScharfViktor:
After talk with @rhafer we enabled vips for web CI see here https://github.com/owncloud/ocis/pull/10310

we can use it in in webCI and don't use in ocisCi in this way we can check both options 

@micbar @kulmann Is that okay?

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1031,7 +1031,7 @@ def cacheOcisPipeline(ctx):
                 rebuildBuildArtifactCache(ctx, "ocis", "ocis")
     else:
         steps = checkForExistingOcisCache(ctx) + \
-                buildOcis() + \
+                buildOcis(enableVips = True) + \
                 cacheOcis()
     return [{
         "kind": "pipeline",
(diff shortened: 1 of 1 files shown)


kulmann:
> After talk with @rhafer we enabled vips for web CI see here [owncloud/ocis#10310](https://github.com/owncloud/ocis/pull/10310)
> 
> we can use it in in webCI and don't use in ocisCi in this way we can check both options
> 
> @micbar @kulmann Is that okay?

Don't like it too much honestly... I'd prefer to switch everything to libvips and have a single e2e thumbnail test with the "old" thumbnailer in a nightly CI run.

ScharfViktor:
> Don't like it too much honestly... I'd prefer to switch everything to libvips and have a single e2e thumbnail test with the "old" thumbnailer in a nightly CI run.

we can switch everything to libvips - it's no problem
about how to have a single e2e thumbnail test with the "old" thumbnailer -  we need think how

ScharfViktor:
now we run e2e test with libvips by default. in the nightly CI we run tests with the "old" thumbnailer

kulmann (approved):

Issue #11870, opened 2024-11-07
Retry preview loading on 425 status code


JammingBen:
## Describe the bug
Requests for preview images that are in post-processing and return a 425 status code never get retried. As a result, the images will never appear in such case until the user reloads the page.

## Steps to reproduce
1. Set `POSTPROCESSING_DELAY=5s` in oCIS
2. Open a space
3. Upload a space image

## Expected behavior
The space image is being fetched at some point after the post-processing is finished

## Actual behavior
The space image is never loaded.

PR #11874, opened 2024-11-08
fix: preview loading retry after 425 status code

JammingBen:
Adds a retry mechanism for preview requests when the images could not be loaded due to postprocessing.

fixes https://github.com/owncloud/web/issues/11870

### DIFF
--- a/packages/web-pkg/tests/unit/services/previewService.spec.ts
+++ b/packages/web-pkg/tests/unit/services/previewService.spec.ts
@@ -87,7 +87,7 @@ describe('PreviewService', () => {
       })
       expect(preview).toEqual(undefined)
     })
-    it('retries when the server returns a 429 status code', async () => {
+    it.each([425, 429])('retries when the server returns a %s status code', async (status) => {
       const supportedMimeTypes = ['image/png']
       const { previewService, clientService } = getWrapper({
         supportedMimeTypes,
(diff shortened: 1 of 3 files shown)


AlexAndBear (approved):

packages/web-pkg/tests/unit/services/previewService.spec.ts
```diff
@@ -87,30 +87,33 @@ describe('PreviewService', () => {
...
       expect(preview).toEqual(undefined)
     })
-    it('retries when the server returns a 429 status code', async () => {
-      const supportedMimeTypes = ['image/png']
-      const { previewService, clientService } = getWrapper({
-        supportedMimeTypes,
-        version: '1'
-      })
+    it.each([425, 429])(
+      'retries when the server returns a 425 or 429 status code',
```
kulmann, line 91: You can interpolate the status into the string, see e.g. https://github.com/owncloud/web/blob/ac3e2cfb0c1fb92b25be136744fe0b79bb33a24a/packages/web-pkg/tests/unit/utils/urlJoin.spec.ts#L17

This improves readability and debugging of the test runs IMO, because the descriptions are much more specific.

kulmann:
```suggestion
      'retries when the server returns a %s status code',
```
should work like this, but please try that before committing it ;-)

kulmann (approved):

Issue #11883, opened 2024-11-11
User can see hidden files in the preview app on Shared with me page


ScharfViktor:
ocis 7.0.0-rc.1+22d80256d
web 11.0.2

Steps: 
- `admin` shares folder image files to `einstein`
- `einstein` opens one image file in preview app and tries to scroll through all the images of the shares jail

Expected: `einstein` can scroll through all the images of the shares jail
Actual: 

https://github.com/user-attachments/assets/9d3e1427-c358-4e87-8ca9-8a796cf60086

## another strange thing: 

- admin share folder `f1` with two image files
- `einstein` opens first image file

Actual:  he cannot scroll through two images

- `einstein` opens second image file

Actual:  he can scroll through two images

https://github.com/user-attachments/assets/8a11a931-3b3b-4ee9-8b8f-45fc165a5f9b

ScharfViktor:
I can confirm that it's work but I noticed several another problem: 


1. user cannot open concrete image
Steps: 
- admin shares several image files to einstein
- einstein opens Shares with me page and tries to open second (or last image)

Actual: 
first image opens

https://github.com/user-attachments/assets/0e25c307-a363-468f-87b5-829d525f014e



2. hidden files in the preview app
- einstein hiddes shared file
- einstein opens preview app

Actual: hidden file is in the app


3. secure view files in the preview app
- admin shares image file with secure view role to einstein
- einstein opens any image file in the preview app

Actual: secure view files in the preview app

JammingBen:
I fixed issues 1. & 3, via https://github.com/owncloud/web/pull/11947.

Issue 2 is a known issue for now, respecting the filters would take much more effort.

ScharfViktor:
can confirm that 1. & 3 are fixed
I renamed issue name to case 2

PR #11893, opened 2024-11-13
fix: preview app image navigation on shared with me page

JammingBen:
fixes https://github.com/owncloud/web/issues/11883


packages/web-pkg/src/composables/appDefaults/useAppFolderHandling.ts
```diff
@@ -29,22 +29,39 @@ export function useAppFolderHandling({
...
-    resourcesStore.clearResourceList()
+
     try {
       context = unref(context)
-      const space = unref(context.space)
 
+      if (context.routeName === 'files-shares-with-me') {
+        // FIXME: this is a somewhat hacky solution to load the shared with me files.
+        // ideally we should check if there is a current folder and if not, use the
+        // folder loader to load files. unfortunately, it currently lives in the files app.
```
JammingBen, line 51: As described, this is a bit of a hacky solution. My original idea was to check the presence of a current folder. If none is given, we can assume that the user is on some page where there is no actual space (Shared with me page, search results, etc). In that case we basically want the same logic as the respective folder loader for this page. However, the folder loader currently lives in the files app... Would it maybe make sense to register folder loaders via extension? @dschmidt What do you think?

dschmidt: The idea of the folder loaders was definitely to make them reusable from app contexts, but I'm not sure they really fit our concept of extensions. 

At least I currently don't see a use case for third party folder loaders...

Can't we just move the folder loaders to web-pkg and use them here?

AlexAndBear (approved):

dschmidt (approved):

packages/web-pkg/src/composables/appDefaults/useAppFolderHandling.ts
dschmidt: We had a quick chat: for the upcoming release we want this quick fix in, as it's easy to backport to the stable branch.
In the long run we think it's best to move the folder loaders to web-pkg.
Making them extensions would be nice for consistency reasons and we can come up with theoretic use cases for that (i.e. when apps register new routes/views which should be reloadable ... but it seems a bit artificial at this point and we suggest to avoid the effort until it's actually needed)

PR #11946, opened 2024-11-21
Multiple fixes found when running web with CERNBox

diocas:
The fixes are described in the changelog.
Most of the cases are error handling of things that your server returns but ours might not. But, for example, the lack of parent ID should be treated by Web, since in the code is marked as optional, so I would argue it is not correct to just crash if not set. Same for TUS support, if not configured it just crashed to upload, which should not.

For the preview service, I think that when the object was created, the store was not loaded yet. So, whatever config I returned in the capabilities, the listing of files would always render all the hardcoded mime types.
But let me know if you have a proper way/different way of solving the issue.

More to come!


LukasHirt:
Thank you for your work! Fixes have been picked up in separate PRs and last one is waiting to be merged now. Closing this PR.

PR #11960, opened 2024-11-25
fix: preview app folder loading

JammingBen:
fixes https://github.com/owncloud/web/issues/8932


packages/web-app-preview/src/App.vue
```diff
@@ -336,32 +341,24 @@ export default defineComponent({
...
-          // with share space resources, we don't have an underlying space, so match the file id
-          if (this.filteredFiles[i].remoteItemId === this.fileId) {
-            this.activeIndex = i
-            return
-          }
-
-          this.activeIndex = 0
-          continue
-        }
+        const filterAttr = isLocationSharesActive(this.$router, 'files-shares-with-me')
```
dschmidt, line 346: urghs 🙈

dschmidt: worst part for me is that it's in the app not in the composables.. can we somehow reaonsably move it there?

JammingBen: Hmm I don't see how since this is preview app logic. IMO the only real solution would be to dig into this id-stuff and see if we can align the behaviors somehow.

JammingBen: The problem is that the `remoteItemId` is also available for search resources on the search result page, but in this case we still need to check for `fileId`. So it's for some reason really bound to that one route :(

PR #12040, opened 2024-12-16
fix: hidden shares in preview app

LukasHirt:
We've fixed the issue with displaying hidden files when user open a file in preview app in "Shared with me" list. The `q_share-visibility` query param is now passed to the app so that if a user opens a file while listing "visible shares", the preview app will allow moving only to another visible files. When the `q_share-visibility` query param is set to `hidden` (i.e. user is listing "hidden shares"), the preview app will allow moving only to another hidden files.

## Description

We've fixed the issue with displaying hidden files when user open a file in preview app in "Shared with me" list.
The `q_share-visibility` query param is now passed to the app so that if a user opens a file while listing "visible shares", the preview app will allow moving only to another visible files.
When the `q_share-visibility` query param is set to `hidden` (i.e. user is listing "hidden shares"), the preview app will allow moving only to another hidden files.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11883

## Motivation and Context

Display only what the user would expect to see.

## How Has This Been Tested?

- test environment: local & 🤖 
- test case 1: added unit tests
- test case 2: manual test in chrome

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-preview-app-files-visibility
+++ b/changelog/unreleased/bugfix-preview-app-files-visibility
@@ -1,0 +1,8 @@
+Bugfix: Preview app files visibility
+
+We've fixed the issue with displaying hidden files when user open a file in preview app in "Shared with me" list.
+The `q_share-visibility` query param is now passed to the app so that if a user opens a file while listing "visible shares", the preview app will allow moving only to another visible files.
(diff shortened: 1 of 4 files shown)


saw-jan (approved):
LGTM.
Confirmed the fix

Issue #12021, opened 2024-12-06
Disable the 'Paste here' when user `cut` and `paste`  in the same folder


2403905:
## Describe the bug
Disable the `Paste here` button when the user tries to `cut` and `paste` the file or folder in the same folder

## Steps to reproduce
On video

https://github.com/user-attachments/assets/b405dc5e-a014-46f4-baf5-64755c9a1a8d

## Expected behavior
Disalo `Paste here`

## Actual behavior
The folder or file has renamed

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

PrajwolAmatya:
The button is disabled, but the bug is reproducible when pasting with `Ctrl + V`.

cc @LukasHirt

PrajwolAmatya:
The button is disabled, but the bug is reproducible when pasting with `Ctrl + V`.

cc @LukasHirt

PR #12044, opened 2024-12-18
fix: disable paste button in same folder

LukasHirt:
## Description

We've fixed the state of the "paste files" action when copied resources are from the same folder. The button will be disabled in such case and a tooltip with explanation message displayed.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/12021

## Motivation and Context

Prevent overwrite action.

## How Has This Been Tested?

- test case 1: added unit tests
- test case 2: try to paste resources in chrome into same folder

Snímek obrazovky 2024-12-18 v 9 51 41

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
@@ -158,14 +158,12 @@
     <div
       v-if="showPasteHereButton"
       id="clipboard-btns"
-      v-oc-tooltip="
-        uploadOrFileCreationBlocked ? $gettext('You have no permission to paste files here!') : ''
-      "
+      v-oc-tooltip="pasteHereButtonTooltip"
       class="oc-button-group"
-      :class="{ disabled: uploadOrFileCreationBlocked }"
+      :class="{ disabled: isPasteHereButtonDisabled }"
     >
       <oc-button
-        :disabled="uploadOrFileCreationBlocked"
+        :disabled="isPasteHereButtonDisabled"
         class="paste-files-btn"
         @click="pasteFileAction"
       >
@@ -270,6 +268,7 @@ export default defineComponent({
     const sharesStore = useSharesStore()
     const route = useRoute()
     const language = useGettext()
+    const { $gettext } = language
 
     const clipboardStore = useClipboardStore()
     const { clearClipboard } = clipboardStore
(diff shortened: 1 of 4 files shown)


LukasHirt:
@saw-jan it seems that tests are failing because when trying to click on the breadcrumbs the tooltip of the paste actions is overflowing it. Any idea why the paste action would even get hovered when the button is not target of the test step? I.e. the locator is on a different element than the action. Have you seen something like this already?

image

saw-jan:
>@saw-jan it seems that tests are failing because when trying to click on the breadcrumbs the tooltip of the paste action is overflowing it. Any idea why the paste action would even get hovered when the button is not target of the test step? I.e. the locator is on a different element than the action. Have you seen something like this already?

Just reviewing it and the test failure. let me have a look at the test. paste button should not be hovered

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -401,6 +400,36 @@ export default defineComponent({
...
+      return unref(uploadOrFileCreationBlocked) || unref(isPastingIntoSameFolder)
+    })
+
+    const pasteHereButtonTooltip = computed(() => {
+      if (unref(uploadOrFileCreationBlocked)) {
+        return $gettext('You have no permission to paste files here!')
+      }
+
+      if (unref(isPastingIntoSameFolder)) {
+        return $gettext('You cannot paste resources into the same folder where you copied them.')
```
saw-jan, line 427:
```suggestion
        return $gettext('You cannot paste resources into the same folder!')
```

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -401,6 +400,36 @@ export default defineComponent({
...
+      )
+    })
+
+    const isPasteHereButtonDisabled = computed(() => {
+      return unref(uploadOrFileCreationBlocked) || unref(isPastingIntoSameFolder)
+    })
+
+    const pasteHereButtonTooltip = computed(() => {
+      if (unref(uploadOrFileCreationBlocked)) {
+        return $gettext('You have no permission to paste files here!')
```
saw-jan, line 423:
```suggestion
        return $gettext('You have no permission to paste files here.')
```

saw-jan:
> > @saw-jan it seems that tests are failing because when trying to click on the breadcrumbs the tooltip of the paste action is overflowing it. Any idea why the paste action would even get hovered when the button is not target of the test step? I.e. the locator is on a different element than the action. Have you seen something like this already?
> 
> Just reviewing it and the test failure. let me have a look at the test. paste button should not be hovered

Can I push the test fix in this PR? the paste button appear just where the cut button was and the mouse point will be on the same position that's why the tooltip was visible. I can push the fix here

LukasHirt:
> Can I push the test fix in this PR?

Yes please. Thanks a lot!

saw-jan (approved):
:sparkles:

Issue #11536, opened 2024-09-06
Details tab cannot be closed when viewing a video and zoomin


individual-it:
## Describe the bug
There are cases while viewing a video when the "Details" tab cannot be closed

## Steps to reproduce
1. open the details tab of a video
2. while keeping the details open, view the video
3. zoom the browser (Ctrl + mouse scroll)

## Expected behavior
The X button should still be visible

## Actual behavior
the X button disappears

## Additional context
seen on https://infinite.owncloud.com/

https://github.com/user-attachments/assets/fbd20154-527b-469b-9a79-016468d16419

PR #12045, opened 2024-12-18
fix: handle sidebar resizing

LukasHirt:
## Description

Fixed various issues with sidebar resizing. Close button is no longer pushed away, resize event is added, and correct breakpoint is used.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/11536

## Motivation and Context

Resilient layout 💪 

## How Has This Been Tested?

- test environment: chrome
- test case 1: Resize browser window
- test case 2: Zoom in & out in browser

https://github.com/user-attachments/assets/ca766a4f-d370-499a-b421-a1a0dabe9d29

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/SideBar/SideBar.vue
+++ b/packages/web-pkg/src/components/SideBar/SideBar.vue
@@ -104,6 +104,7 @@ import {
   defineComponent,
   nextTick,
   onBeforeUnmount,
+  onMounted,
   PropType,
   ref,
   unref,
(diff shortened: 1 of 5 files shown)


packages/web-pkg/src/components/SideBar/SideBar.vue
```diff
@@ -203,11 +204,21 @@ export default defineComponent({
...
 
-    const fullWidthSideBar = computed(() => window.innerWidth <= 1024)
+    const windowWidth = ref(window.innerWidth)
+
+    const fullWidthSideBar = computed(() => unref(windowWidth) <= 960)
     const backgroundContentEl = computed(() => {
       return unref(appSideBar)?.parentElement?.querySelector('div') as HTMLElement
     })
 
+    const onResize = () => {
```
saw-jan, line 214: maybe we can think of debouncing the function.

saw-jan:
I built this branch and I see that space is allocated for sidebar even though it is not opened
 :x: | opened
--|--
Screenshot from 2024-12-19 15-47-16 | Screenshot from 2024-12-19 15-47-20

packages/web-pkg/src/components/SideBar/SideBar.vue
LukasHirt: That might cause again "laggy" reactivity. Resize does not happen that often so I think it is fine not to have it debounced.

LukasHirt:
> I built this branch and I see that space is allocated for sidebar even though it is not opened

Right 🤦 fixed by applying the adjusted width only when the sidebar is open.

saw-jan (approved):
:+1:

PR #12105, opened 2025-01-16
feat: handle editor-less extensions

LukasHirt:
## Description

When an app does not expose any routes, mark it as editor-less so that we do not try to load editor actions for such extension.

## Motivation and Context

Extension without editor does not break the app.

## How Has This Been Tested?

- test case 1: manually add extension without any routes and open sidebar
- test case 2: extend existing unit tests

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


LukasHirt:
Let's post-pone this as I might be using the wrong extension type to achieve what I need...

LukasHirt:
I double checked it and I do not see any other way to achieve create file action without automatically creating an editor for it.

saw-jan (approved):
:+1:

Issue #12108, opened 2025-01-17
Preview feature makes the app unresponsive


nabim777:
## Describe the bug
The UI freezes or becomes unresponsive when using the Preview feature.

## Steps to reproduce
1. Upload an image `test.png`
2. Open the image `test.png`
3. Using the sidebar, navigate to the `actions` section
4. When a user clicks the preview and closes the image

## Expected behavior
The preview should work without affecting the UI.

## Actual behavior
The UI freezes or becomes unresponsive after closing the image

https://github.com/user-attachments/assets/45db7bae-b4ef-46df-85d5-4cd74f4e7d15

## Tested on
Infinite Scale 7.1.0-rc.1 Community
ownCloud Web UI 11.1.0

saw-jan:
^ @LukasHirt

LukasHirt:
> When a user clicks the preview and closes the image

By this, you mean that the user clicks on the "Preview" action in the sidebar?

![Image](https://github.com/user-attachments/assets/29c5b15b-b994-44ed-856a-3bdf3b2d27ca)

saw-jan:
yeah. looks like it

PR #12110, opened 2025-01-17
fix(web-pkg): hide sidebar editor action when the editor is already opened

LukasHirt:
## Description

When an editor is already opened and the user clicks on the Open in ... action, it pushes the same editor route into the history causing a bug where the user needs to click twice on the close button to actually close the editor. This change hides the editor action if an editor with matching route name is already opened which prevents this behaviour.

## Related Issue

- refs https://github.com/owncloud/web/issues/12108

## Motivation and Context

No confusion with re-opening the active editor.

## How Has This Been Tested?

- test case 1: added unit test
- test case 2: open an image in the preview app and open sidebar

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/changelog/unreleased/enhancement-hide-active-editor-action.md
+++ b/changelog/unreleased/enhancement-hide-active-editor-action.md
@@ -1,0 +1,6 @@
+Enhancement: Hide active editor action
+
+We've hidden the sidebar file action of a currently opened editor. This prevents confusion where user might try to re-open the current editor.
+
+https://github.com/owncloud/web/pull/12110
+https://github.com/owncloud/web/issues/12108
--- a/packages/web-pkg/src/composables/actions/files/useFileActions.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActions.ts
@@ -41,6 +41,7 @@ import { Resource, SpaceResource } from '@ownclouders/web-client'
 import { storeToRefs } from 'pinia'
 import { useEmbedMode } from '../../embedMode'
 import { RouteRecordName } from 'vue-router'
+import { isLocationActive } from '../../../router/utils'
 
 export const EDITOR_MODE_EDIT = 'edit'
 export const EDITOR_MODE_CREATE = 'create'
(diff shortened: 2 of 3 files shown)


nabim777:
I have tested this bug-fix PR and found that it introduced the following changes:

- The **preview** button has been replaced with **Open in Collabora.**

## Bug found in this PR

 @LukasHirt  When we open the image in Collabora and try to close it, the UI becomes unresponsive:

[Screencast from 01-17-2025 05:02:09 PM.webm](https://github.com/user-attachments/assets/6060da6c-2025-4cef-bff2-ad1935212713)

saw-jan:
> I have tested this bug-fix PR and found that it introduced the following changes:
> 
> * The **preview** button has been replaced with **Open in Collabora.**
> 
> ## Bug found in this PR
> @LukasHirt When we open the image in Collabora and try to close it, the UI becomes unresponsive:
> 
>  Screencast.from.01-17-2025.05.02.09.PM.webm

This should be fixed by https://github.com/owncloud/web/pull/12112

nabim777:
> > I have tested this bug-fix PR and found that it introduced the following changes:
> > 
> > * The **preview** button has been replaced with **Open in Collabora.**
> > 
> > ## Bug found in this PR
> > @LukasHirt When we open the image in Collabora and try to close it, the UI becomes unresponsive:
> > Screencast.from.01-17-2025.05.02.09.PM.webm
> 
> This should be fixed by #12112

Yes, it has been fixed by [PR](https://github.com/owncloud/web/pull/12112), and I have tested it.

LukasHirt:
This will need some more work as the `routeName` is not always defined leading into wrong behaviour of hiding the action even outside of the editor. Since the regression has been fixed in https://github.com/owncloud/web/pull/12112, I would suggest postponing this into next release.

LukasHirt:
@nabim777 could you please retest this pr? I fallback to the app name when routeName is undefined which is the same approach as in other places of the code base. All should work as expected now.

nabim777:
> @nabim777 could you please retest this pr? I fallback to the app name when routeName is undefined which is the same approach as in other places of the code base. All should work as expected now.

Hi @LukasHirt, 
It's working as expected :white_check_mark: , `preview` is hidden in the sidebar when the editor is already open.

Issue #12039, opened 2024-12-16
Password Protected Folders


kobergj:
New menu point below `New Folder`  called `New Password Protected Folder`
* On click:
  * Open Dialog that allows setting `Foldername`  and `Password` 
  * OPTIONAL: Allow configuring the permissions on the public link
  * Create a folder in the users personal space `.PasswordProtectedFolders/projects/<space_name>/<path>/<to>/<folder>`
  * Create a public link to this folder with the given password an view permissions (or the configured permissions)
  * Create a `.url` (?) file that points to the public link
  * Either use `.ppf.url` extension or a custom one

* On clicking the `.url` file:
  * Open the `.url` file in a splash screen.

* On delete of `.url` file the public link and the hidden folder will also be deleted.
* No other actions available on the `.url` file.

LukasHirt:
Rough estimate:

- create new app: 2
- extend extensions system to allow passing create folder actions: 5
- displaying dialog with folder name, password and link permissions: 3
- creating the hidden folder, .ppf.url file and public link: 5
- open folder in a popup: 3
- delete .ppf.url file together with the hidden folder and the public link: 2
- hide file extension and actions other than delete on the .ppf.url file: 3

LukasHirt:
@dj4oC you showed us in the call opening the "splash screen" from some other action. Could you pls let me know what action it was? I tried to find it in the UI but has not been successful.

dj4oC:
<img width="1912" alt="Screenshot 2024-12-10_New" src="https://github.com/user-attachments/assets/bc32bb07-d426-48e4-a412-c9e993acd3ad" />
<img width="1912" alt="Screenshot 2024-12-10_New2" src="https://github.com/user-attachments/assets/b9ff1ca1-6eb8-468a-bf80-3bf36bea277f" />
![Screenshot 2024-11-15 at 14 22 23](https://github.com/user-attachments/assets/04962c15-6e94-4da0-9325-e89b4569e4b4)
![Screenshot 2024-11-11 at 15 23 04](https://github.com/user-attachments/assets/6ac33ded-75cc-4a4d-8399-0eb55228aba4)

dj4oC:
> @dj4oC you showed us in the call opening the "splash screen" from some other action. Could you pls let me know what action it was? I tried to find it in the UI but has not been successful.

The three dots when you have an document open as shown in the last screenshot

LukasHirt:
> OPTIONAL: Allow configuring the permissions on the public link

[...]

PR #12146, opened 2025-01-30
fix(password-protected-folders): create folder in the personal space

LukasHirt:
## Description

Create the hidden folder in users personal space

## Related Issue

- refs https://github.com/owncloud/web/issues/12039

## Motivation and Context

Other users cannot access the folder without the pass

## How Has This Been Tested?

- test case 1: Create a password protected folder in users personal space
- test case 2: Create a password protected folder in created space

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-password-protected-folders/src/composables/useCreateFileHandler.ts
+++ b/packages/web-app-password-protected-folders/src/composables/useCreateFileHandler.ts
@@ -1,7 +1,6 @@
 import { Resource, SpaceResource, urlJoin } from '@ownclouders/web-client'
 import { SharingLinkType } from '@ownclouders/web-client/graph/generated'
 import { useClientService, useResourcesStore, useSharesStore } from '@ownclouders/web-pkg'
-import { unref } from 'vue'
 
 export const useCreateFileHandler = () => {
   const clientService = useClientService()
@@ -10,13 +9,15 @@ export const useCreateFileHandler = () => {
 
   const createFileHandler = async ({
     fileName,
-    space,
+    personalSpace,
+    currentSpace,
     currentFolder,
     password,
     type
   }: {
     fileName: string
-    space: SpaceResource
+    personalSpace: SpaceResource
+    currentSpace: SpaceResource
     currentFolder: Resource
     password: string
     type: SharingLinkType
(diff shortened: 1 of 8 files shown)


packages/web-app-password-protected-folders/src/composables/useCreateFileHandler.ts
```diff
@@ -26,20 +27,22 @@ export const useCreateFileHandler = () => {
     }
 
     const folderPath = '/.' + fileName
+    const folder = await clientService.webdav.createFolder(personalSpace, { path: folderPath })
```
saw-jan, line 30: does this mean if a protected folder is created in any project space, the folder (hidden folder) will be created in the creator's personal space?

saw-jan: Also if I create `test1` folder in my personal space. and try to create `test1` in project space, it doesn't allow creation of `test1` in the project space.

LukasHirt: Done

saw-jan: @LukasHirt 
should the user be able to create the same name protected folder in the project spaces?
Suppose, I have manually created `.test` folder in my personal space, and now I want to create a protected folder `test`  in the project space. Should this be allowed? Currently, it doesn't. Wonder if this is expected behavior.

Screenshot from 2025-02-11 17-00-52

saw-jan: If expected then looks good :+1:

LukasHirt: It is not optimal solution but I think in this case it would be fine to have this limitation and simply force the user to rename it. The hidden folder name needs to currently match the .psec file name as it is used when deleting the file and folder together. Meaning adding some suffix to the folder name would break the deletion then. So in favor of keeping this feature technically simple, I would say forcing the user to have unique name is fine. What do you say @kobergj

kobergj: Isn't the folder in the personal space under `.PasswordProtectedFolders/projects/<space_name>`? So that would mean a user needs to create the exact same folder structure (including the space name) to achieve a name clash?

LukasHirt: 🤦 and of course I missed this requirement... you are right. I will look into this asap.

saw-jan: hmm, I thought something changed internally in the requirements and I didn't ask about it while reviewing. should have just asked it.

LukasHirt: Fixed

saw-jan (approved):
implementation of this PR works. :+1:
folders are created inside `.PasswordProtectedFolders/projects/{space-name}` dir

Issue #12150, opened 2025-02-03
Navigating into space and back to personal space crashes the app


LukasHirt:
## Describe the bug

When navigating into a created space and then back to users personal space, the app crashes.

## Steps to reproduce
1. Open "Spaces"
2. Open any space (create one first if there isn't any)
3. Open "Personal"

## Expected behavior

Resources of personal space are loading and showing up

## Actual behavior

App crashes and prints an error into the console:

![Image](https://github.com/user-attachments/assets/bde683ec-08c6-42b4-ab3d-77d1831156c7)

LukasHirt:
Using Prio 2 as it does not seem to be affecting the current stable release.

PR #12179, opened 2025-02-13
fix(files): check files view presence in upload dropzone init

LukasHirt:
## Description

When initializing the uppy upload dropzone, first check whether the target files view element is present to prevent the app from crashing due to the non-existing target.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/12150

## Motivation and Context

Non-crashing app

## How Has This Been Tested?

- test case 1: navigate into any space and then back to personal space

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
@@ -444,12 +444,15 @@ export default defineComponent({
 
     watch(
       canUpload,
-      () => {
-        if (unref(canUpload)) {
-          uppyService.useDropTarget({ targetSelector: '#files-view' })
-        } else {
+      (canUpload) => {
+        const el = document.getElementById('files-view')
+
+        if (!el || !canUpload) {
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -444,12 +444,15 @@ export default defineComponent({
 
     watch(
       canUpload,
-      () => {
-        if (unref(canUpload)) {
-          uppyService.useDropTarget({ targetSelector: '#files-view' })
-        } else {
+      (canUpload) => {
```
mzner, line 447: We don't need to check if it is a ref first?

mzner (approved):

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
LukasHirt: The watcher callback param is already "unrefed".

Issue #12228, opened 2025-02-27
Cannot use email as user name in the user creation form: "User name cannot contain special characters"


skatsubo:
## Describe the bug
Email addresses are not allowed as user name in the user creation form: UI gives an error "User name cannot contain special characters". Though email-as-user-name is permitted in backend.

## Steps to reproduce
1. Launch "Create user" in Administration Settings > Users.
2. Enter email address in "User name" field.

![Image](https://github.com/user-attachments/assets/d77185bf-4512-4ced-b82b-b8d42a195c6f)

## Expected behavior
When entering email as user name in the user creation form it should be accepted without errors.

## Actual behavior
When entering email as user name in the user creation form it gets rejected with error "User name cannot contain special characters".

## Additional context

Workaround 1. As the username validation is performed only in `CreateUserModal.vue`, but not performed in `EditPanel.vue`, my current UI-based workaround is: 
1. Create a user without `@` in user name, e.g. replacing `@` with underscore: `sk_domain.tld`.
2. Edit the user and put `@` back in user name, e.g. `sk_domain.tld -> sk@domain.tld`.

Workaround 2. Alternatively, auto-provisioning also works with email addresses as user names.

The issue seems to be caused by extra/redundant escaping in https://github.com/owncloud/web/pull/9703/files#diff-e4578202fcddaeb2086e30f2ed4fa01d7fc7944c8d185928937afa26eeb835b9R144

PR #12229, opened 2025-02-27
fix (web-app-admin-settings): allow email as user name in the user creation form

skatsubo:
## Description
Fix extra/redundant escaping in user name filtering in the user creation form to allow entering email address as user name.

This PR removes that extra escaping and adds a test case covering email address as user name. Thus aligning UI filtering with backend filtering which allows emails as user names.

## Related Issue
- Fixes https://github.com/owncloud/web/issues/12228

## Motivation and Context
Currently user name regex filter in the user creation modal rejects email addresses. It emits an error "User name cannot contain special characters". This seems to be caused by redundant escaping in the regex pattern. 

## How Has This Been Tested?
- Tested manually in Firefox, Chrome, Safari.
- Added a test case for CreateUserModal to cover email-as-user-name scenario.

image

## Types of changes

- [ ] Technical debt
- [x] Tests
- [ ] Documentation


LukasHirt (approved):
🚀

Issue #12183, opened 2025-02-14
[OCM] Users can invite themselves to their own federated connection


2403905:
## Describe the bug

When a user generates a token for federated connection and pastes it to their own accept invitation field, then a federated connection is created with the user itself.

Related issue [https://github.com/owncloud/ocis/issues/11004](https://github.com/owncloud/ocis/issues/11004)

<img width="899" alt="Image" src="https://github.com/user-attachments/assets/82146ee0-4916-4487-97ea-a5ecd0316b27" />

## Expected behavior
Deny to accept the federated connection from the same instance.

## Actual behavior
A clear and concise description of what happened.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

mzner:
@2403905 I'm not able to reproduce this bug any more. Is this still relevant or should we close the ticket?

2403905:
It was fixed on the backend. But I think we should have some validation on web and disallow pressing the `Accept invitation` button.

PR #12328, opened 2025-03-19
bugfix: disallow user from inviting themselves

mzner:
Disable accept invitation button when user is using their own generated token link.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/12183

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-app-ocm/src/composables/useInviteTokens/useInviteTokens.ts
```diff
@@ -0,0 +1,105 @@
...
+        }),
+        ...(t.description && { description: t.description })
+      })
+    })
+  } catch (error) {
+    throw new Error(error)
+  }
+}
+
+async function generateToken(description: string) {
```
LukasHirt, line 55: Why not have this inside the `useInviteTokens` composable so that you do not need to have the top level variable?

LukasHirt (commented):
I think it would be nice if there would be some tooltip on the button to explain why it's disabled. But only if it would be easy to implement.

mzner:
> I think it would be nice if there would be some tooltip on the button to explain why it's disabled. But only if it would be easy to implement.

good point. will do.  👍

LukasHirt (approved):

PR #12588, opened 2025-05-21
[full-ci] use same hostname for both ocis instances in docker compose setup

saw-jan:
## Description
Use the same hostname for both ocis instances:
```diff
host.docker.internal:9200
- ocis.owncloud.test:10200
+ host.docker.internal:10200
```

This reduces the need for adding many hosts in the system while using the `docker-compose.yml` setup.

Fix: `collabora/code:25.04.2.1.1` was not working, so replaced it with `collabora/code:24.04.10.2.1` (same as in the CI)

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/dev/docker/ocis.idp.config.yaml
+++ b/dev/docker/ocis.idp.config.yaml
@@ -11,23 +11,23 @@ clients:
       - https://host.docker.internal:9201/
       - https://host.docker.internal:9201/oidc-callback.html
       - https://host.docker.internal:9201/oidc-silent-redirect.html
-      - https://ocis.owncloud.test:10200/
-      - https://ocis.owncloud.test:10200/oidc-callback.html
-      - https://ocis.owncloud.test:10200/oidc-silent-redirect.html
-      - https://ocis.owncloud.test:10201/
-      - https://ocis.owncloud.test:10201/oidc-callback.html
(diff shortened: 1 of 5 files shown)


LukasHirt (approved):

LukasHirt:
> Fix: collabora/code:25.04.2.1.1 was not working, so replaced it with collabora/code:24.04.10.2.1 (same as in the CI)

Why are there managed two different versions?

saw-jan:
>Why are there managed two different versions?

The version was updated by renovate bot: https://github.com/owncloud/web/commit/eefe52c39ed2bbafb0c5c9bb18f1bc9bcd33bdbc

saw-jan:
> > The version was updated by renovate bot
> 
> I meant it in the sense of why actually have two places where the version has to be specified? Couldn't we have the same for local development and for CI? Then an update that breaks something like the changed version here could be already caught in the CI of the bump PR.

Ahh, thanks. The reason is that for CI, we define images in `.drone.star` file and for local dev setup we have `docker-compose.yml`. we cannot have single source of images for both files as they are different and `.drone.star` doesn't support importing/including files.
From the look of the renovate config, one thing we can do is make image definition in .drone.star accessible to renovate.

PR #12656, opened 2025-06-03
Update web docs

mmattel:
This PR updates the documentation for the web repo. It was driven to add content for translations...

* Fix the make steps
* Fix hugo (was broken)
* Add/fix docs about docs
* Add translation related stuff
* Text fixes

Commands tested locally - worked

@kobergj FYI

### DIFF
--- a/docs/translations/translations.md
+++ b/docs/translations/translations.md
@@ -28,7 +28,7 @@ The following files are managed via `make` commands in all folders containing tr
 `make l10n-read` --> `template.pot` and\
 `make l10n-write` --> `translations.json`,
 
-The `config` files must be created manually and define the Transifex setup for each folder. Here is an example how this file lookes like:
+The `config` files must be created manually and define the Transifex setup for each folder. Here is an example how this file looks like:
 
 ```
 [main]


.gitignore
```diff
@@ -5,6 +5,7 @@ release/*
 
 build/version.json
 
+# docs
```
LukasHirt: Could this be removed?

docs/.gitignore
```diff
@@ -0,0 +1,2 @@
+# exclude the hugo directory
+hugo/
```
LukasHirt: There is also `/hugo` in the root `.gitignore`. I would like to have only one of these - no preference which one.

docs/translations/translations.md
```diff
@@ -0,0 +1,90 @@
...
+  .tx			(dir)
+    config		(file)
+```
+
+The following files are managed via `make` commands in all folders containing translations, see section [Translation Process](#translation-process) for more details.
+
+`make l10n-read` --> `template.pot` and\
+`make l10n-write` --> `translations.json`,
+
+The `config` files must be created manually and defines the Transifex setup for each folder. Here is an example how this file lookes like:
```
LukasHirt: I am not 100% sure that this change is correct but it feels right.

```suggestion
The `config` files must be created manually and define the Transifex setup for each folder. Here is an example how this file lookes like:
```

docs/translations/translations.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+## Translation Definitions
+
+Translations are defined in 4 levels:
+
+1. The source text such as `$gettext('Administration Settings')`.
+2. For each language, the minimum translation completion percentage required for processing.\
+This can either be the individual setting in the config file or the global setting behind `make l10n-pull`, the latter takes precedence.
+3. The languages that will appear in the language selector are defined in `/gettext.config.cjs`.
```
LukasHirt: This is not 100% right. The gettext config works behind the scene. It has an effect on how the translations are read/written by the make commands which makes them available to the app itself but it does not affect the language selector in the UI. The language selector is only using the `/packages/web-runtime/src/defaults/languages.ts`. Sorry for any confusion.

docs/translations/translations.md
```diff
@@ -0,0 +1,90 @@
...
+4. The user-facing text in the selector identifies a language, which is defined in\
```
LukasHirt: See comment above - it does not only affect the text but what is actually available in the selector.

docs/translations/translations.md
```diff
@@ -0,0 +1,90 @@
...
+You will get a '403, permission_denied: You do not have permission to perform this action.' otherwise. If you need to push to Transifex manually but lack the proper write permissions, request or trigger a sync which is described via the above link.
+{{< /hint >}}
+
+Note that only the command `make l10n-write` will create changes that can be committed, following files are excluded by `.gitignore`:
+```
+**/l10n/locale
+**/l10n/template.pot
+```
+
+If you want to remove all extractions and translations like to have a clean translation base, use the following command **BEFORE** starting with `make l10n-read`:
```
LukasHirt: > If you want to remove all extractions and translations like

Is the "like" word there on purpose? Feels a bit strange when reading it. But maybe just lack of English skills on my side 😁

LukasHirt (changes requested):

.gitignore
mmattel: fixed

docs/.gitignore
mmattel: fixed

docs/translations/translations.md
mmattel: fixed

docs/translations/translations.md
```diff
@@ -0,0 +1,88 @@
...
+  .tx			(dir)
+    config		(file)
+```
+
+The following files are managed via `make` commands in all folders containing translations, see section [Translation Process](#translation-process) for more details.
+
+`make l10n-read` --> `template.pot` and\
+`make l10n-write` --> `translations.json`,
+
+The `config` files must be created manually and define the Transifex setup for each folder. Here is an example how this file lookes like:
```
LukasHirt:
```suggestion
The `config` files must be created manually and define the Transifex setup for each folder. Here is an example how this file looks like:
```

PR #12691, opened 2025-06-11
bugfix: testing throws stderr

mzner:
When running tests, stderr was being printed to the console, which can clutter the output and make it harder to read. This has been fixed so that stderr is no longer printed during test runs.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/tests/unit/components/Filters/DateFilter.spec.ts
+++ b/packages/web-pkg/tests/unit/components/Filters/DateFilter.spec.ts
@@ -96,6 +96,7 @@ describe('DateFilter', () => {
     })
     describe('apply button', () => {
       it('is not clickable without dates entered', async () => {
+        console.error = vi.fn()
         const { wrapper } = getWrapper()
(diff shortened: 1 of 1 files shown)


LukasHirt:
Personally, I am not a fan of test specific conditions and of suppressing errors. Shouldn't we fix the error instead?

mzner:
> Personally, I am not a fan of test specific conditions and of suppressing errors. Shouldn't we fix the error instead?

I don't see something to fix. It is just console logging when user selects invalid dates, I honestly don't know what's the point of a console log in the first place. We either show the user an error or we collect this data somewhere.

What do you suggest?

LukasHirt:
If it's intended that the test selects the invalid date, I would like the log to be silenced by the test instead of in the component. Then we do not get it logged when expected but still get it logged when not expected.

PR #12702, opened 2025-06-12
Update docs build process

mmattel:
References: https://github.com/owncloud/owncloud.github.io/pull/100 (Update docs build process and images used)

This PR is based on changes made in the referenced PR.

* Update the hugo image/version to be used for the CI.
The images must match and maintaining it is a manual process.
* Updating the use of `make` inside a makefile to be compliant with the make documentation
* Update the targets for building docs as we now can use the targets from the cloned repo (see the referenced PR)
We no longer need a local hugo installation but use docker images which are defined via the cloned repo.
This eases maintenance a lot
* Update the building documentation accordingly

This PR MUST be merged AFTER the referenced PR gets merged !


.drone.star
```diff
@@ -8,7 +8,10 @@ OC_CI_BAZEL_BUILDIFIER_IMAGE = "owncloudci/bazel-buildifier:latest"
 OC_CI_DRONE_ANSIBLE_IMAGE = "owncloudci/drone-ansible:latest"
 OC_CI_DRONE_SKIP_PIPELINE_IMAGE = "owncloudci/drone-skip-pipeline:latest"
 OC_CI_GOLANG_IMAGE = "owncloudci/golang:1.24"
-OC_CI_HUGO_IMAGE = "owncloudci/hugo:0.115.2"
+
+# hugo needs to be the same as in owncloud.github.io 
+OC_CI_HUGO_IMAGE = "hugomods/hugo:base-0.129.0"
```
LukasHirt: Using the `_IMAGE` will cause this to be picked up by Renovate bot and update PRs will be automatically created. That might not be necessarily desired now since it needs to be synced with the other repo? If it should not be picked up by Renovate, it's enough to rename it to `OC_CI_HUGO`.

mmattel: I updated the naming from `OC_CI_HUGO_IMAGE` --> `OC_CI_HUGO_IMAGE_STATIC`.
Keeping image in the name but ending it differently. We keep being alinged to the other image var names.

docs/building.md
```diff
@@ -24,12 +24,8 @@ Note that the documentation is located in the `docs/` folder.
 
 ### Install Hugo
 
-- Install [hugo](https://gohugo.io/getting-started/installing/)\
-{{< hint warning >}}
-It is **VERY** important to have the same hugo version installed as defined in the [ocis](https://github.com/owncloud/ocis/) repo.
-The relevant file/command to look at is `cat .bingo/Variables.mk | grep HUGO`. At the time of writing, the version is `v0.123.7`. Using any other version will break building docs in the web repo.
-<!-- the required/correlated theme version is defined in: https://github.com/owncloud/owncloud.github.io/blob/main/Makefile -->
-{{< /hint >}}
+- There is no need to install Hugo, the framework that generates the documentation. You can install the required version for other projects. Hugo is used via a predefined container that is downloaded if it does not exist.
```
LukasHirt: What do you think about replacing this section with new section `Prerequisites` and only mention the need for docker? It feels a bit odd to me reading "Install Hugo" and then immediately "There is no need to install Hugo...".

PR #12744, opened 2025-06-24
chore: refacot web-admin-members-role-section to use script setup

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersRoleSection.vue
+++ b/packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersRoleSection.vue
@@ -1,7 +1,7 @@
 <template>
   <ul class="oc-list">
     <li
-      v-for="(m, index) in members"
+      v-for="(m, index) in props.members"
       :key="index"
       class="oc-flex oc-flex-middle oc-mb-s"
       data-testid="space-members-list"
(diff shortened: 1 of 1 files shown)


packages/web-app-admin-settings/src/components/Spaces/SideBar/MembersRoleSection.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <ul class="oc-list">
     <li
-      v-for="(m, index) in members"
+      v-for="(m, index) in props.members"
```
LukasHirt, line 4: Is the `props.` needed? I thought it works without it.

mzner: They append it if it is missing but it is confusing, and I believe they give a warning. We can destructure members from props and then we can remove `props`, but then it is clear where is it coming from. What do you think?

LukasHirt: I'm fine with leaving it as is. Thanks for the context. :+1:

Issue #12853, opened 2025-07-24
[QA] Run E2E tests only related to changed packages


LukasHirt:
**JIRA:** [#OCISDEV-220](https://kiteworks.atlassian.net/browse/OCISDEV-220)

## Summary

Instead of running tests for every package/extension we have in `/packages` in every PR, we want to add tags to them to assert when they should run. The tags should be then filtered by changed files in the PR.

## Requirements

- new tags matching package/extension can be added
- CI checks for changed files and runs only E2E tests with tags matching the changed package/extension
- if no such tag is present in a test, it needs to run every time
- if `[full-ci]` tag is added in the PR title, these tags are ignored and everything is run

PR #12865, opened 2025-07-28
OCISDEV-220: Check how to fail e2e tests in web-app-ocm

deyankiteworks:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/12853

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-ocm/src/views/App.vue
+++ b/packages/web-app-ocm/src/views/App.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="sciencemesh">
+  <!-- <div class="sciencemesh">
     <div class="oc-flex oc-flex-column sciencemesh-wrapper">
       <div class="oc-flex sciencemesh-top">
         <div id="sciencemesh-invite" class="oc-width-1-2">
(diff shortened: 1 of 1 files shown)


tests/e2e/cucumber/features/admin-settings/web-packages.txt
saw-jan, line 1: @deyankiteworks @LukasHirt I am introducing this approach to define the dependent web packages right within the test suite in a file called `web-package.txt`. And we will use this information to determine the test run for that suite.
What are your thoughts on it?

saw-jan: And we also need your help @LukasHirt to better determine the relation between the web packages and the test suites.

LukasHirt: I think the file is fine. I'll try to prepare something to make the relationships clear.

saw-jan: THank you @LukasHirt

LukasHirt (approved):
@saw-jan do you think we should add some information about this into the dev docs?

PR #12894, opened 2025-08-07
bugfix(web-runtime): [OCISDEV-240] fixes vue missing prop label warning

mzner:
## Related Issue

- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-240

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/pages/account.vue
+++ b/packages/web-runtime/src/pages/account.vue
@@ -222,6 +222,8 @@
               <oc-select
                 :model-value="emailNotificationsValues[option.id]"
                 :options="option.singleChoiceValue.options"
+                :label="emailNotificationsValues[option.id].displayValue"
+                :label-hidden="true"
                 :clearable="false"
                 option-label="displayValue"
                 @update:model-value="(value) => updateSingleChoiceValue(option.name, value)"


packages/web-runtime/src/pages/account.vue
```diff
@@ -222,6 +222,7 @@
               <oc-select
                 :model-value="emailNotificationsValues[option.id]"
                 :options="option.singleChoiceValue.options"
+                :label="$gettext('Email Notifications Intervals')"
```
LukasHirt: Since the options are coming from the bundle, I would say the label should not be hardcoded. We do not know if there will or won't be a new option in the future which would then use the same label.

mzner: Should we just make it optional then?

LukasHirt: It's needed for a11y

mzner:
@LukasHirt Please take another look. 👍🏻

PR #12939, opened 2025-08-20
chore(web-app-pkg): [OCISDEV-256] refactor resource-table component to use script setup

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/views/spaces/GenericTrash.vue
+++ b/packages/web-app-files/src/views/spaces/GenericTrash.vue
@@ -28,7 +28,6 @@
           :are-paths-displayed="true"
           :resources="paginatedResources"
           :are-resources-clickable="false"
-          :are-thumbnails-displayed="false"
           :header-position="fileListHeaderY"
           :sort-by="sortBy"
           :sort-dir="sortDir"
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/views/spaces/GenericTrash.vue
```diff
@@ -28,7 +28,6 @@
           :are-paths-displayed="true"
           :resources="paginatedResources"
           :are-resources-clickable="false"
-          :are-thumbnails-displayed="false"
```
LukasHirt, line 31: If this prop is dropped, is the logic to hide thumbnails done somewhere else or it's no longer necessary to hide thumbnails in trash?

mzner: I looked into the project, but it is not used. I believe it is no longer necessary. In the `<resource-table>`, where it is passed to, it is not used there. Same for PaddingX which is set to a fixed value in the `<resource-table>`

LukasHirt: I just checked it. Thumbnails are still hidden in trash so the logic must live somewhere else.

LukasHirt (approved):

PR #12964, opened 2025-08-21
chore(web-app-pkg): [OCISDEV-256] refactor resource-tile component to use script setup

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTile.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTile.vue
@@ -8,7 +8,6 @@
       'oc-tile-card-disabled': isResourceDisabled && !isProjectSpaceResource(resource),
       'state-trashed': isResourceDisabled && isProjectSpaceResource(resource)
     }"
-    @contextmenu="$emit('contextmenu', $event)"
   >
     <div v-if="isHidden" class="oc-tile-card-lazy-shimmer"></div>
     <template v-else>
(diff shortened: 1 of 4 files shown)


packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -8,7 +8,6 @@
       'oc-tile-card-disabled': isResourceDisabled && !isProjectSpaceResource(resource),
       'state-trashed': isResourceDisabled && isProjectSpaceResource(resource)
     }"
-    @contextmenu="$emit('contextmenu', $event)"
```
mzner, line 11: @LukasHirt I don't see that this is being used/triggered anywhere. Also, I don't think we should be using custom listeners on native elements. What do you think?

LukasHirt (approved):

packages/web-pkg/src/components/FilesList/ResourceTile.vue
LukasHirt: Sorry, I missed this before approving the PR. It actually is used here: https://github.com/owncloud/web/blob/2a6cd31703e0a7a868ce990d5417cf9619c182e4/packages/web-pkg/src/components/FilesList/ResourceTiles.vue#L48

In general, components from either design system or web-pkg can be used also outside of the Web repo as they are intended to be used by extensions so even if it wouldn't be used, I would be a bit hesitant to drop it. If we would, I would consider doing it as a breaking change. Also, I am not sure what you mean by custom event as `contextmenu` is part of the `Element` events? https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event

Could you please make a PR to bring this back? I am a bit worried we could have same situation as in https://github.com/owncloud/web/pull/12929

PR #13108, opened 2025-09-16
bugfix(web-runtime): [OCISDEV-267] fix progress bar not being translated when user changes language

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/layouts/Application.vue
+++ b/packages/web-runtime/src/layouts/Application.vue
@@ -203,28 +203,44 @@ export default defineComponent({
 
     const progressBarExtensionId = 'com.github.owncloud.web.runtime.default-progress-bar'
     const progressBarExtensionPointId = 'app.runtime.global-progress-bar'
-    const defaultProgressBarExtension: CustomComponentExtension = {
-      id: progressBarExtensionId,
-      type: 'customComponent',
-      extensionPointIds: [progressBarExtensionPointId],
-      content: LoadingIndicator,
-      userPreference: {
-        optionLabel: $gettext('Default progress bar')
+
+    function registerProgressBarExtension() {
+      const defaultProgressBarExtension: CustomComponentExtension = {
+        id: progressBarExtensionId,
+        type: 'customComponent',
+        extensionPointIds: [progressBarExtensionPointId],
+        content: LoadingIndicator,
+        userPreference: {
+          optionLabel: $gettext('Default progress bar')
+        }
       }
-    }
-    extensionRegistry.registerExtensions(toRef([defaultProgressBarExtension] satisfies Extension[]))
-    const progressBarExtensionPoint: ExtensionPoint<CustomComponentExtension> = {
-      id: progressBarExtensionPointId,
-      extensionType: 'customComponent',
-      multiple: false,
-      defaultExtensionId: defaultProgressBarExtension.id,
-      userPreference: {
-        label: $gettext('Global progress bar'),
-        description: $gettext('Customize your progress bar')
+      extensionRegistry.registerExtensions(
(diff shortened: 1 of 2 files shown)


packages/web-runtime/src/layouts/Application.vue
```diff
@@ -203,28 +203,44 @@ export default defineComponent({
...
+        id: progressBarExtensionPointId,
+        extensionType: 'customComponent',
+        multiple: false,
+        defaultExtensionId: defaultProgressBarExtension.id,
+        userPreference: {
+          label: $gettext('Global progress bar'),
+          description: $gettext('Customize your progress bar')
+        }
       }
+      extensionRegistry.registerExtensionPoints(toRef([progressBarExtensionPoint]))
```
LukasHirt, line 230: I'm quite worried what re-registering extension point will do. If there are any other progress bar extensions, are they persisted or are they lost due to this?

packages/web-runtime/src/layouts/Application.vue
```diff
@@ -203,28 +203,44 @@ export default defineComponent({
...
+      return progressBarExtensionPoint
     }
-    const extensionPoints = computed<ExtensionPoint<Extension>[]>(() => [progressBarExtensionPoint])
-    extensionRegistry.registerExtensionPoints(extensionPoints)
+
+    const progressBarExtensionPoint = ref(registerProgressBarExtension())
+
+    watch(
+      () => $gettext('Global progress bar'),
```
LukasHirt, line 237: If it's needed to depend on the locale, wouldn't it be nicer to use `current` from the `useGettext` composable?

packages/web-runtime/src/layouts/Application.vue
```diff
@@ -203,28 +203,44 @@ export default defineComponent({
...
+      () => {
+        extensionRegistry.unregisterExtensions([progressBarExtensionId])
+        extensionRegistry.unregisterExtensionPoints([progressBarExtensionPoint.value.id])
```
LukasHirt, line 240: Same worry as in the register function... I do not think other progress bars will survive this.

LukasHirt (commented):
So I take it using `composable` for the extension was not enough? I fear this is not the right approach though.

1. This fix would most probably apply only to this specific progress bar but wouldn't affect other ones.
2. I think re-registering the extension point will break the system (see specific comments on the LOC).

packages/web-runtime/src/layouts/Application.vue
mzner: current is just a string, so it doesn't trigger reactivity when it changes. I thought the same, but didn't work, unfortunately. :(

packages/web-runtime/src/layouts/Application.vue
mzner: good points. I have to check..I'm not sure. Do you have any other ideas in general? I couldn't come up with anything better. 🙈

LukasHirt: If the computed prop did not work, I have no other idea from top of my head. I would need to dig deeper.

mzner: > If the computed prop did not work, I have no other idea from top of my head. I would need to dig deeper.

Nope, it didn't work either. Already tried that. I will try to find something else and we can talk tomorrow. 👍🏻

LukasHirt:
Closing as the other PR has been merged.

PR #13110, opened 2025-09-16
feat(web-app-external): [OCISDEV-305] include mobile in the query params

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-app-external/src/App.vue
```diff
@@ -66,6 +66,7 @@ const appProviderService = useAppProviderService()
 const { makeRequest } = useRequest()
 
 const viewModeQuery = useRouteQuery('view_mode')
+const isMobileWidth = inject<Ref<boolean>>('isMobileWidth')
```
LukasHirt: Per the ticket, we should also include `navigator.userAgentData?.mobile`

packages/web-app-external/src/App.vue
```diff
@@ -111,10 +112,12 @@ const loadAppUrl = useTask(function* (signal, viewMode: string) {
 
     const fileId = props.resource.fileId
     const baseUrl = urlJoin(configStore.serverUrl, capabilityStore.filesAppProviders[0].open_url)
+    const isOnlyOfficeApp = unref(appName)?.toLowerCase().includes('onlyoffice')
```
LukasHirt: Please, sync with @2403905 if this should be hardcoded or if the app registry will somehow expose whether the app has a mobile view. @mzner or do you think we should always send it and leave it out to the app to either ignore it or use it?

2403905: To me, the Web shouldn't care what app we use. Just send the `mobile=1`. The collaboration service should resolve it.

PR #13132, opened 2025-09-23
chore(web-app-files, web-pkg, web-runtime): [CISDEV-287] ui improvements

mzner:
- change public shared upload text.
 - disable sidebar button.
 - add web and backend versions.

## Related Issue

- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-287

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-pkg/src/helpers/versions.ts
```diff
@@ -0,0 +1,20 @@
+import { CapabilityStore } from '../composables/index'
+
+export const getWebVersion = (): string => {
```
LukasHirt, line 3: We already have such helpers here https://github.com/owncloud/web/blob/master/packages/web-runtime/src/container/versions.ts#L4 Any reason why not use that one?

packages/web-runtime/src/components/Topbar/TopBar.vue
```diff
@@ -123,11 +123,15 @@ export default {
...
+    const isFilesPublicUpoad = computed(() => {
+      return unref(router.currentRoute).name === 'files-public-upload'
+    })
+
     const isSideBarToggleVisible = computed(() => {
       return authStore.userContextReady || authStore.publicLinkContextReady
     })
     const isSideBarToggleDisabled = computed(() => {
-      return isRuntimeRoute(unref(router.currentRoute))
+      return isRuntimeRoute(unref(router.currentRoute)) || isFilesPublicUpoad.value
```
LukasHirt: In general, I think it's better to use either only `unref` or `.value`. Just for consistency.

packages/web-pkg/src/helpers/versions.ts
mzner: We don't export/expose web-runtime. I haven't seen a package importing from web-runtime. But I could make the one in web-runtime use the one from web-pkg. What do you think?

LukasHirt: Yes, having it then in web-pkg and import it into runtime sounds good to me.

packages/web-app-files/src/views/FilesDrop.vue
```diff
@@ -83,6 +83,10 @@
                   {{ $gettext('Learn more about ownCloud') }}
                 </oc-button>
               </template>
+              <div class="versions oc-pb-m oc-pl-m oc-text-xsmall oc-text-muted">
```
LukasHirt: This will be now visible only on `> s` devices. I think it would make sense to make it visible also on mobile devices. But only the version. Not the whole parent div with logo and slogan.

packages/web-app-files/src/views/FilesDrop.vue
```diff
@@ -139,6 +146,10 @@ const { getInternalSpace } = useGetMatchingSpace()
 useUpload({ uppyService })
 
 const resourcesStore = useResourcesStore()
+const capabilityStore = useCapabilityStore()
+
+const webVersion = computed(() => getWebVersion())
```
LukasHirt: I guess this one does not need to be a computed property...

PR #13170, opened 2025-10-07
bugfix(web-design): [OCISDEV-126] set initial focus to false

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/design-system/src/components/OcModal/OcModal.vue
+++ b/packages/design-system/src/components/OcModal/OcModal.vue
@@ -288,7 +288,7 @@ const initialFocusRef = computed<FocusTargetOrFalse>(() => {
     return focusTrapInitial as FocusTargetOrFalse
   }
 
-  return () => unref(ocModalInput)?.$el || unref(ocModal)
+  return () => false
 })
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -288,7 +288,7 @@ const initialFocusRef = computed<FocusTargetOrFalse>(() => {
     return focusTrapInitial as FocusTargetOrFalse
   }
 
-  return () => unref(ocModalInput)?.$el || unref(ocModal)
+  return () => false
```
LukasHirt, line 291: The focus should still be sent somewhere in the dialog. With this change, the focus is sent to body. That results in "Skip to main" button being the next element which will be focused.

LukasHirt (changes requested):

PR #13192, opened 2025-10-13
bugfix(web-pkg): [OCISDEV-418] fetch file on mount

### LINKED ISSUE #13174: Refreshing a deleted public link for a file stuck at infinite loading state
**JIRA:** [OCISDEV-378](https://kiteworks.atlassian.net/browse/OCISDEV-378)
---
### Describe the bug

When a public link for a file is deleted, refreshing that link gives infinite loading screen without any details

### Steps to reproduce

```textarea
1. Upload or create a file => Let's say Admin uploads a file `orange.png`
2. Admin creates a public link for `orange.png` file with password
3. Admin opens that public link in an incognito window => resource loads correctly
4. Admin deletes that public link
5. Admin refreshes( or hard refreshes) the public link opened in step 3
```

### Expected behavior

Some useful information about the deleted link like `Resource not found` or 

Image

### Actual behavior

The refreshed page gives infinite loading screen without any information/details

### Setup

```markdown
- Web version: `ownCloud Web UI 12.1.0`
- oCIS version: `Infinite Scale 7.3.0-rc.1 Community`
- browser: Firefox
```

### Additional context

https://github.com/user-attachments/assets/9bee6d97-c299-4c2d-bf5a-c5895ff868e6

mzner:
## Related Issue

- Fixes https://github.com/owncloud/web/issues/13174

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...


packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -468,7 +468,8 @@ const hasAutoSave = computed(() => {
 })
 
 let autosaveIntervalId: ReturnType<typeof setInterval> = null
-onMounted(() => {
+onMounted(async () => {
+  await loadResourceTask.perform()
```
LukasHirt: This is already happening on line 369, isn't it?

mzner: Nope because of the if condition.

LukasHirt: But that's why there's the `else` as well, right? Or am I missing something?

mzner: I was confused as well. :D

The else is for this condition "if (unref(fileSizeLimit) && toNumber(unref(resource).size) > unref(fileSizeLimit))"

we need the loadResourceTask.perform because it calls the getFileInfo. Or we can call the getFileInfo directly and based on it, show the error, if you prefer that approach.

LukasHirt: Ok, now I get it...

But then this change would get around the intended approach:

1. runtime should preferably not care about what extension is loaded so doing "X" if extension "Y" is matched feels wrong;
2. loading of the resource is delegated to the extension (web-app-preview in this case) so we should look into the logic there instead and find out where the issue is.

One of the main thing I would really want to avoid is having the resource loaded several times which I am worried would happen if we mix delegated resource loading in the extension and also handle this in mounted hook in the wrapper.

CHANGELOG.md
```diff
@@ -9,12 +9,9 @@ Summary
 -------
 
 * Bugfix - Fix copying public link and password on Safari: [#13177](https://github.com/owncloud/web/pull/13177)
```
LukasHirt, line 11: `CHANGELOG.md` should not be touched manually.

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -362,6 +365,20 @@ const loadFileTask = useTask(function* (signal) {
   }
 }).restartable()
 
+watch(
+  () => appStore.error,
+  () => {
+    if (appStore.error) {
+      loadingError.value = new Error(
+        $gettext('The resource could not be located, it may not exist anymore.')
```
LukasHirt, line 373: Should this be somehow reset as well or is that handled already?

LukasHirt, line 373: What would you say about getting the message from the app instead of hardcoding it? Just to make sure we're open to different types of errors and don't force the same message for them.

packages/web-pkg/src/composables/piniaStores/app.ts
```diff
@@ -0,0 +1,16 @@
+import { defineStore } from 'pinia'
+import { ref } from 'vue'
+
+export interface AppError {
+  message: string
+  status?: number
+}
+
+export const useAppStore = defineStore('app', () => {
```
LukasHirt, line 9: Just out of curiosity: why did you use store for this instead of e.g. event from the app like is done with the `update:resource` event?

mzner: It was the most the made sense to me given that this is a global error and not originated from a component per se.

mzner:
closed in favor of https://github.com/owncloud/web/pull/13223

PR #13198, opened 2025-10-14
bugfix(web-app-activities, web-app-files, web-runtime): [OCISDEV-420] fix incorrect translations

mzner:
## Related Issue

- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-420

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-activities/src/components/ActivityItem.vue
+++ b/packages/web-app-activities/src/components/ActivityItem.vue
@@ -4,7 +4,7 @@
       <oc-avatar :width="36" :user-name="activity.template.variables?.user?.displayName" />
       <span class="oc-ml-s" v-text="activity.template.variables?.user?.displayName" />
     </div>
-    <div>activity unknown</div>
+    <div>{{ $gettext('Activity unknown') }}</div>
     <div class="oc-text-truncate">
       <resource-list-item v-if="resource" :resource="resource" :is-resource-clickable="false" />
       <div
--- a/packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
@@ -2,7 +2,7 @@
   <div id="oc-files-file-link" class="oc-position-relative">
     <div class="oc-flex oc-flex-middle">
       <h3 class="oc-text-bold oc-text-medium oc-m-rm" v-text="$gettext('Public links')" />
-      <oc-contextual-helper v-if="helpersEnabled" class="oc-pl-xs" v-bind="viaLinkHelp" />
+      <oc-contextual-helper v-if="helpersEnabled" class="oc-pl-xs" v-bind="shareViaLinkHelp" />
     </div>
     <p v-if="!directLinks.length" class="files-links-empty oc-mt-m" v-text="noLinksLabel" />
     <ul
@@ -51,7 +51,11 @@
       <hr class="oc-my-m" />
       <h4 class="oc-text-bold oc-text-medium oc-m-rm">
         {{ indirectLinksHeading }}
-        <oc-contextual-helper v-if="helpersEnabled" class="oc-pl-xs" v-bind="indirectLinkHelp" />
+        <oc-contextual-helper
+          v-if="helpersEnabled"
+          class="oc-pl-xs"
+          v-bind="shareViaIndirectLinkHelp"
+        />
       </h4>
       <div
         class="files-links-indirect-list"
@@ -95,7 +99,7 @@ import {
   UpdateLinkOptions,
   useRouter
 } from '@ownclouders/web-pkg'
-import { shareViaLinkHelp, shareViaIndirectLinkHelp } from '../../../helpers/contextualHelpers'
+import { useContextualHelpers } from '../../../composables/contextualHelpers/useContextualHelpers'
 import { isSpaceResource, LinkShare } from '@ownclouders/web-client'
 import ListItem from './Links/ListItem.vue'
 import { Resource, SpaceResource } from '@ownclouders/web-client'
@@ -128,6 +132,8 @@ const { linkShares } = storeToRefs(sharesStore)
 
 const configStore = useConfigStore()
 const { options: configOptions } = storeToRefs(configStore)
+
+const { shareViaLinkHelp, shareViaIndirectLinkHelp } = useContextualHelpers()
 
 const { actions: createLinkActions } = useFileActionsCreateLink()
 const createLinkAction = computed<FileAction>(() =>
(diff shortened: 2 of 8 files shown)


packages/web-app-files/src/helpers/contextualHelpers.ts
```diff
@@ -1,10 +1,12 @@
 import { omit } from 'lodash-es'
 import { ContextualHelperData } from '@ownclouders/design-system/helpers'
 import { ConfigStore } from '@ownclouders/web-pkg'
+import { useGettext } from 'vue3-gettext'
 
 // just a dummy function to trick gettext tools
 function $gettext(msg: string) {
-  return msg
+  const { $gettext: getText } = useGettext()
```
LukasHirt: What do you think about making a composable out of these helpers? Personally, I do not like to mix vue-specific code with regular ts code.

packages/web-app-files/src/composables/contextualHelpers/useContextualHelpers.ts
```diff
@@ -0,0 +1,142 @@
+import { computed, unref, type MaybeRef } from 'vue'
+import { omit } from 'lodash-es'
+import { ContextualHelperData } from '@ownclouders/design-system/helpers'
+import { ConfigStore } from '@ownclouders/web-pkg'
+import { useGettext } from 'vue3-gettext'
+
+export const useContextualHelpers = (configStore: MaybeRef<ConfigStore>) => {
```
LukasHirt: Why is the `configStore` a parameter instead of using the pinia composable here?

packages/web-app-files/src/components/SideBar/Shares/SpaceMembers.vue
```diff
@@ -140,9 +142,7 @@ const filteredSpaceMembers = computed(() => {
 const helpersEnabled = computed(() => {
   return unref(configStore.options.contextHelpers)
 })
-const spaceAddMemberHelp = computed(() => {
-  return shareSpaceAddMemberHelp({ configStore: unref(configStore) })
-})
+const spaceAddMemberHelp = shareSpaceAddMemberHelp
```
LukasHirt: Let's use it directly instead of wrapping it in the extra `const`?

packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
```diff
@@ -282,32 +277,38 @@ const showSpaceMembers = computed(() => {
...
 
-function isSpaceMemberDeniable(collaborator: CollaboratorShare) {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function isSpaceMemberDeniable(_collaborator: CollaboratorShare) {
   // FIXME: currently not supported by sharing NG
   return false
 }
 
-function isShareDeniable(collaborator: CollaboratorShare) {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
LukasHirt, line 309: Why did it work before this?

packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -274,12 +276,9 @@ const helpersEnabled = computed(() => {
   return unref(configOptions).contextHelpers
 })
 
-const viaLinkHelp = computed(() => {
-  return shareViaLinkHelp({ configStore })
-})
-const indirectLinkHelp = computed(() => {
-  return shareViaIndirectLinkHelp({ configStore })
-})
+const viaLinkHelp = shareViaLinkHelp
```
LukasHirt: Let's use it directly instead of wrapping it in the extra `const`?

packages/web-app-files/src/components/SideBar/Shares/FileLinks.vue
```diff
@@ -274,12 +276,9 @@ const helpersEnabled = computed(() => {
...
+const indirectLinkHelp = shareViaIndirectLinkHelp
```
LukasHirt: Let's use it directly instead of wrapping it in the extra `const`?

packages/web-app-files/src/helpers/contextualHelpers.ts
LukasHirt: Would it be safe to completely remove this file? If not, I would at least revert this specific change and keep it as it was.

packages/web-app-files/src/composables/contextualHelpers/useContextualHelpers.ts
mzner: Just a preference instead of having the file depend on it -- it is more flexible when it is being passed. Can switch if you prefer. What do you think?

packages/web-app-files/src/components/SideBar/Shares/FileShares.vue
mzner: it still does work, just doesn't look nice in the IDE, I mean the IDE complains.

packages/web-app-files/src/helpers/contextualHelpers.ts
mzner: I would say so. I searched the project and replaced where it has been used, but never say never. I will bring it back. :D

packages/web-app-files/src/composables/contextualHelpers/useContextualHelpers.ts
LukasHirt: I usually choose simplicity over flexibility. I.e. if a composable is dependent on a store which will not be different if it's called from component A or component B, I don't want to always have to pass it instead of having it once in the composable directly. If it would be some data that might change depending on the place where the composable is called, then parameters would be of course the way to go for me as well.

However, I don't want to impose my thinking on you so if you prefer it like this, I am fine with leaving it like it is.

packages/web-app-files/src/helpers/contextualHelpers.ts
LukasHirt: Up to you. web-app-files is not published anywhere and is "only" bundled directly with Web, so if it's not used in this repo, there shouldn't be any other place that would use it. If you decide to keep this, it would be great to mark the exported code as deprecated using jsdocs though to make it clear it's not supposed to be used anymore.

packages/web-app-files/src/composables/contextualHelpers/useContextualHelpers.ts
mzner: fair point. 👍🏻

packages/web-app-activities/src/components/ActivityItem.vue
```diff
@@ -4,7 +4,7 @@
       <oc-avatar :width="36" :user-name="activity.template.variables?.user?.displayName" />
       <span class="oc-ml-s" v-text="activity.template.variables?.user?.displayName" />
     </div>
-    <div>activity unknown</div>
+    <div>{{ $gettext('activity unknown') }}</div>
```
LukasHirt: Since you're touching this and it will get synced to transifex, could you please double check whether it is correct behaviour that it starts with lowercase `a`? Feels a bit odd. Just to make sure that we push the correct string into transifex.

PR #13236, opened 2025-10-23
feat(web-pkg): [OCISDEV-388]: disallow renaming file to the same name it already has

### LINKED ISSUE #11785: [OCISDEV-388] Renaming a resource to its current name fails without meaningful error message
## Is your feature request related to a problem? Please describe.
When attempting to rename a file or resource using the same name it already has, the operation fails as expected — however, the error message provided is not informative and does not explain why the operation failed. It just gives 

Image
which is very hard to understand for normal user

## Describe the solution you'd like
Renaming a file to its current name should either:
- Succeed silently (no-op), or
- Fail with a clear and specific error message (e.g., "Source and destination names are the same. No action taken.")

## Setup
```
- Web version: `ownCloud Web UI 12.1.0`
- oCIS version: `Infinite Scale 7.3.0-rc.1 Community`
- browser: Firefox
```

mzner:
## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/11785
- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-388

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
@@ -51,6 +51,10 @@ export const useFileActionsRename = () => {
       return $gettext('The name cannot be empty')
     }
 
+    if (resource.name === newName) {
+      return $gettext('Choose a new name to be able to rename the file')
+    }
+
     if (/[/]/.test(newName)) {
       return $gettext('The name cannot contain "/"')
     }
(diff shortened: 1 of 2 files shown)


LukasHirt (changes requested):
Let's wait with this for the comment in Jira to be answered. This doesn't feel like good ux.

PR #13268, opened 2025-11-05
[tests-only] [full-ci] test: [OCISDEV-417] add a11y playwright tests

LukasHirt:
## Description

Setup new Playwright a11y tests and add a11y reporter. Accessibility checks are added into existing E2E tests. Docs are added to explain the new accessibility tests.

## Motivation and Context

Migrating tests from cucumber to playwright.

## How Has This Been Tested?

- test environment: macos
- test case 1: run playwright tests
- test case 2: run cucumber tests
- test case 3: let CI pass

## Types of changes

- [ ] Technical debt
- [x] Tests
- [ ] Documentation


packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
```diff
@@ -5,10 +5,11 @@
       :type="componentType"
       v-bind="componentProps"
       :class="[action.class, 'action-menu-item', 'oc-py-s', 'oc-px-m', 'oc-width-1-1']"
-      :aria-label="componentProps.disabled ? action.disabledTooltip?.(actionOptions) : ''"
+      :aria-label="componentProps.disabled ? action.disabledTooltip?.(actionOptions) : null"
```
mzner, line 8: Should we be using a label instead of null?

packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
mzner, line 4: Is there a difference between this and `:title`?

tests/e2e-playwright/reporters/a11y.ts
```diff
@@ -0,0 +1,125 @@
...
+class A11yReporter implements Reporter {
+  private results: A11yTestResult[] = []
+  private outputFile: string
+  private startTime: number = 0
+
+  constructor(options: { outputFile?: string } = {}) {
+    this.outputFile = options.outputFile || 'a11y-report.json'
+  }
+
+  onBegin(_config: FullConfig, _suite: Suite): void {
```
mzner, line 47: Not that important, but what do you think about using `onStart` instead of `onBegin`?

packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
LukasHirt: We need the label to be gone if the button is not disabled and with `''` it is unfortunately persisted on the element (even though with no value). Using `null` makes sure that the label is gone.

packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
LukasHirt: `tooltip` is not available to screen readers.

tests/e2e-playwright/reporters/a11y.ts
LukasHirt: We cannot use arbitrary name here because these methods are overwriting the ones from the reporter class.

LukasHirt: https://playwright.dev/docs/test-reporters#custom-reporters

PR #13381, opened 2025-12-11
chore(web-pkg): [OCISDEV-510] ui improvements

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTable.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTable.vue
@@ -172,6 +172,7 @@
       <oc-icon
         v-oc-tooltip="item.disabled ? $gettext('Disabled') : $gettext('Enabled')"
         :name="item.disabled ? 'stop-circle' : 'play-circle'"
+        :accessible-label="`${item.disabled ? $pgettext(`used by screen reader to announce spaces's status`, 'Space is disabled') : $pgettext(`used by reader to announce spaces's status`, 'Space is enabled')}`"
         size="small"
         fill-type="line"
       />


packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -172,6 +172,7 @@
       <oc-icon
         v-oc-tooltip="item.disabled ? $gettext('Disabled') : $gettext('Enabled')"
         :name="item.disabled ? 'stop-circle' : 'play-circle'"
+        :accessible-label="`Space is ${item.disabled ? $gettext('Disabled') : $gettext('Enabled')}`"
```
LukasHirt: The complete string should be translated.

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -172,6 +172,7 @@
       <oc-icon
         v-oc-tooltip="item.disabled ? $gettext('Disabled') : $gettext('Enabled')"
         :name="item.disabled ? 'stop-circle' : 'play-circle'"
+        :accessible-label="`${$pgettext('use by reader to announce spaces\'s  status', 'Space is')} ${item.disabled ? $gettext('Disabled') : $gettext('Enabled')}`"
```
LukasHirt: Use [interpolation](https://jshmrtn.github.io/vue3-gettext/functions.html#interpolation-1) instead as some languages might use different format, order, etc.

Also, is this really about disabled space? I would like to play this safe and would suggest double checking whether a resource can be disabled while it's parent space is enabled...

LukasHirt (changes requested):

PR #13385, opened 2025-12-12
chore(web-runtime): [OCISDEV-538] ui improvements

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-runtime/src/components/Topbar/Notifications.vue
+++ b/packages/web-runtime/src/components/Topbar/Notifications.vue
@@ -28,6 +28,7 @@
         <span
           v-if="!notifications.length"
           class="oc-notifications-no-new"
+          tabindex="0"
           v-text="$gettext('Nothing new')"
         />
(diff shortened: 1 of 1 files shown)


packages/web-runtime/src/components/Topbar/Notifications.vue
```diff
@@ -28,6 +28,8 @@
         <span
           v-if="!notifications.length"
           class="oc-notifications-no-new"
+          tabindex="0"
+          role="button"
```
LukasHirt: Why is it now a button if it has no action?

mzner: That was a mistake while testing. good catch

Issue #13403, opened 2025-12-22
[QA] e2e-playwright test is flaky


saw-jan:
Builds:
- https://drone.owncloud.com/owncloud/web/56593/10/12
- https://drone.owncloud.com/owncloud/web/56615/9/10

```feature

  1) [chromium] › tests/e2e-playwright/specs/internalLink.spec.ts:67:3 › internal link share › opening a link with internal role 

    Error: frame.evaluate: Execution context was destroyed, most likely because of a navigation

       at ../../e2e/support/objects/a11y/actions.ts:47

      45 |   const { page, include } = args
      46 |
    > 47 |   const a11yResult = await new AxeBuilder({ page })
         |                      ^
      48 |     .withTags(a11yRuleTags)
      49 |     .include(include)
      50 |     .analyze()
        at AxeBuilder.runPartialRecursive (/var/www/owncloud/web/node_modules/.pnpm/@axe-core+playwright@4.11.0_playwright-core@1.56.1/node_modules/@axe-core/playwright/dist/index.mjs:283:34)
        at AxeBuilder.analyze (/var/www/owncloud/web/node_modules/.pnpm/@axe-core+playwright@4.11.0_playwright-core@1.56.1/node_modules/@axe-core/playwright/dist/index.mjs:210:28)
        at Module.analyzeAccessibilityConformityViolations (/var/www/owncloud/web/tests/e2e/support/objects/a11y/actions.ts:47:22)
        at Accessibility.getAccessibilityConformityViolations (/var/www/owncloud/web/tests/e2e/support/objects/a11y/index.ts:17:12)
        at Accessibility.getSevereAccessibilityViolations (/var/www/owncloud/web/tests/e2e/support/objects/a11y/index.ts:21:24)
        at Public.open (/var/www/owncloud/web/tests/e2e/support/objects/app-files/page/public.ts:25:5)
        at Module.openPublicLink (/var/www/owncloud/web/tests/e2e-playwright/steps/ui/public.ts:23:3)
        at /var/www/owncloud/web/tests/e2e-playwright/specs/internalLink.spec.ts:68:5

    Error Context: reports/e2e/internalLink-internal-link-c2b73-g-a-link-with-internal-role-chromium/error-context.md
```

a11y report: https://cache.owncloud.com/public/owncloud/web/56593/a11y/a11y-report.json

```json
{
  "summary": {
    "totalTests": 4,
    "totalViolations": 4,
    "totalPasses": 92,
[...]

PR #13407, opened 2025-12-23
chore(test): fix flaky a11y test

saw-jan:
## Description
Fix a11y e2e tests

## Related Issue
- Fixes https://github.com/owncloud/web/issues/13403

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/tests/e2e/support/objects/app-files/page/public.ts
+++ b/tests/e2e/support/objects/app-files/page/public.ts
@@ -32,15 +32,15 @@ export class Public {
       ),
       this.#page.goto(url)
     ])
-    // check a11y for password protected link or error message
-    // other cases are covered in other page navigations
     if (
-      (await this.#page.locator(passwordProtectedPublicLinkForm).isVisible()) ||
-      (await this.#page.locator(publicLinkErrorMessage).isVisible())
+      !(await this.#page.locator(passwordProtectedPublicLinkForm).isVisible()) &&
+      !(await this.#page.locator(publicLinkErrorMessage).isVisible())
     ) {
-      const a11yObject = new objects.a11y.Accessibility({ page: this.#page })
-      await a11yObject.getSevereAccessibilityViolations('body')
+      // wait for redirection to complete
+      await this.#page.waitForURL('**/public/**')
     }
+    const a11yObject = new objects.a11y.Accessibility({ page: this.#page })
+    await a11yObject.getSevereAccessibilityViolations('body')
   }
 
   async authenticate({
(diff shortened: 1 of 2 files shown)


tests/e2e/support/objects/app-files/page/public.ts
```diff
@@ -19,7 +23,24 @@ export class Public {
...
+      this.#page.waitForResponse(
+        (res) =>
+          res.url().includes('/public-files/') &&
+          res.request().method() === 'PROPFIND' &&
+          res.status() >= 207
+      ),
+      this.#page.goto(url)
+    ])
+    // check a11y for password protected link or error message
+    // other cases are covered in other page navigations
```
LukasHirt: This feels rather confusing to me. Why in some cases we would check it here and in others elsewhere? What happens if new pages are added? Do we add the check here or do we add it somewhere else?

tests/e2e/support/objects/app-files/link/index.ts
```diff
@@ -15,9 +15,13 @@ export class Link {
     const startUrl = this.#page.url()
     const url = await po.createLink({ ...args, page: this.#page })
 
+    let linkName = 'Unnamed link'
```
LukasHirt: Could we use `const linkName = args.name ?? 'Unnamed link'` to make it shorter?

tests/e2e/support/objects/app-files/page/public.ts
saw-jan: My understanding here is, when we navigate using the public link, we have 4 routes:
1. Enter password page
2. Link Error page
3. files page (for authenticated user) - link without password
4. shared-with-me page (for internal links) - link without password

Since files (3) and shared-with-me (4) pages navigation will also have a11y checks sooner or later once we increase the checks, I added checks for 1 and 2 only which also fixed the flakiness

LukasHirt: Basically, I am not a fan of the mixed automatic and manual approach. What would you say about either having all routes automatically checked here (would require some more digging regarding the flakiness) or always manually? I would very much like to stay consistent.

Personally, I would prefer to have it "automatically" done in this function as in general I like not having to take care of every route manually. Of course I am now talking only about this basic check of the full page and no specific checks after certain actions or of specific components...

saw-jan: The falkiness here is the page redirection. Sometimes, the a11y check starts before the complete redirection which causes the change in page object. I couldn't find any smart wait for this case other than waiting explicitly. But let me check more options.

saw-jan: I have added the wait for the navigation, and it should eliminate the flakiness

LukasHirt: Thank you @saw-jan!

PR #13409, opened 2025-12-26
fix: [OCISDEV-554] duped call for public share access with p/w

### LINKED ISSUE #13363: [OCISDEV-554] Password protected share gets authenticated twice
### Describe the bug

When you try to access to a public share and enter the password, the request is sent twice

### Steps to reproduce

```textarea
1. Create a password protected public share
2. Try to access the share from another browser
3. Enter the password. It doesn't matter if it's correct or not.
...
```

### Expected behavior

Only one request to access the password is made

### Actual behavior

2 requests are made. 

Image

### Setup

```markdown
- Web version:
- oCIS version:
- browser:

OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

### Additional context

_No response_

existentialcoder:
## Description
- Every time a new shared link is accessed with a password, two redundant calls are made to list the files
- The first one is a watcher listener and a second is to verify password
- This PR makes sure the watcher triggers the network call only during password verification.

## Related Issue

- Fixes #13363 

## Motivation and Context
Removed a duplicated call for the files-list when accessing a password protected shared link
 

## How Has This Been Tested?
Unit and manual tests

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-prevent-duplicate-call-for-password-protected-public-links.md
+++ b/changelog/unreleased/bugfix-prevent-duplicate-call-for-password-protected-public-links.md
@@ -1,0 +1,6 @@
+Bugfix: Prevent duplicate call for password protected public links
+
+We've fixed an issue where the password protected public link was being called twice.
+
+https://github.com/owncloud/web/issues/13363
+https://github.com/owncloud/web/pull/13409


LukasHirt (changes requested):
Could you please add unit tests to verify that the change prevent the second call?

existentialcoder:
> Could you please add unit tests to verify that the change prevent the second call?

Trying to get an idea here. @LukasHirt 

The authService is actually [mocked](https://github.com/owncloud/web/blob/c742de510f115cfe3fb3847152f57e1d706d0f3c/packages/web-runtime/tests/unit/pages/resolvePublicLink.spec.ts#L10) in the test and therefore the watcher won't be triggered. 

Can the authService test file instead include this test since the changes are added there? TIA

LukasHirt:
Yes, if there are more duplicate calls we should handle those as well. If the fix for it is simple, I would include it in this PR. If it's a bit more complex, I would suggest separate PR.

existentialcoder:
> Yes, if there are more duplicate calls we should handle those as well. If the fix for it is simple, I would include it in this PR. If it's a bit more complex, I would suggest separate PR.

Yes can be taken separately. As you mentioned this needs a separate test file for index.ts. Should we take this PR without the test?

LukasHirt (approved):

packages/web-app-files/src/views/spaces/DriveResolver.vue
```diff
@@ -60,10 +62,11 @@ const fileId = computed(() => {
   return queryItemAsString(unref(fileIdQueryItem))
 })
 
-const getSpaceResource = async (): Promise<SpaceResource> => {
+const getSpaceResource = (): SpaceResource => {
   const spaceValue = unref(space)
+  const allSpaces = spacesStore.spaces
   try {
-    return (await clientService.webdav.getFileInfo(spaceValue)) as SpaceResource
+    return allSpaces.find(space => space.id === spaceValue.id)
```
LukasHirt: I think the previous behaviour here actually makes sense. If we open different space, it's nice to actually get it updated in case there has been any change. Otherwise we'd be "stuck" with old data.

LukasHirt (changes requested):
Revert the latest change please.

packages/web-app-files/src/views/spaces/DriveResolver.vue
existentialcoder: Makes sense. Will revert

PR #13415, opened 2025-12-29
fix(web-pkg): [OCISDEV-553] filter only personal trashed spaces

LukasHirt:
## Description

Instead of filtering all trashed spaces, we now filter only personal trashed spaces. This fixes the issue where listing spaces with "Include disabled" option toggled did not show disabled spaces.

## Motivation and Context

Disabled project spaces are again listable.

## How Has This Been Tested?

- test environment: macos v26.2, chrome v143.0.7499.170
- test case 1: open UI with light user
- test case 2: toggle "Include disabled" option with regular user in spaces list

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-pkg/tests/unit/composables/piniaStores/spaces.spec.ts
+++ b/packages/web-pkg/tests/unit/composables/piniaStores/spaces.spec.ts
@@ -206,14 +206,23 @@ describe('spaces', () => {
         }
       })
     })
-    it('should filter out trashed spaces', async () => {
+    it('should filter out trashed personal spaces', async () => {
       await new Promise<void>((resolve, reject) => {
         try {
           getWrapper({
             setup: async (instance) => {
               const spaces = [
                 mock<SpaceResource>({ id: '1' }),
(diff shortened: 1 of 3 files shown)


packages/web-pkg/tests/unit/composables/piniaStores/spaces.spec.ts
```diff
@@ -238,7 +247,7 @@ describe('spaces', () => {
                 },
                 expect.anything()
               )
-              expect(instance.spaces.length).toBe(2)
+              expect(instance.spaces.length).toBe(4)
```
mzner, line 250: Should that be 2 or 4? I can only see two with trashed state.

LukasHirt: Only space with ID `'2'` gets filtered out so normally two spaces would be in the `spaces` array but the call to the graph API is happening twice due to different queries. The test does not care about that though and always returns the same mocked response. That's why the number of spaces here is doubled.

PR #13426, opened 2025-12-30
[full-ci] feat: [OCISDEV-541] add crash page

LukasHirt:
## Description

We've added a crash page to the application. This page is displayed when the application encounters an error that it cannot recover from. The first such error we are catching is when spaces loading fails.

## Motivation and Context

Handle errors that the app cannot recover from instead of showing the UI in a "stuck" state.

## How Has This Been Tested?

- test environment: macos v26.2, chrome v143.0.7499.170
- test case 1: trigger API error when loading spaces

image

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation


LukasHirt:
@nirajacharya2 can you please review the new tests? THX

LukasHirt:
> both tests are not working for me locally

`- navigating to "/crash?errorCode=RUNTIME-BOOTSTRAP-040", waiting until "load"`

This error is strange as it should use the base URL and not only path...

tests/e2e-playwright/steps/ui/crashPage.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function openCrashPage({
+  actorsEnvironment,
+  errorCode
+}: {
+  actorsEnvironment: ActorsEnvironment
+  errorCode: string
+}) {
+  const { page } = actorsEnvironment.getActor({ key: 'Alice' })
+  return page.goto(`${config.baseUrl}/crash?errorCode=${errorCode}`)
```
ghost: it will be better to pass user to function. if we logged in with brian this test would have failed
```suggestion
  errorCode: string
  user: string
}) {
  const { page } = actorsEnvironment.getActor({ key: user })
  return page.goto(`${config.baseUrl}/crash?errorCode=${errorCode}`)
```

tests/e2e-playwright/specs/runtime/crashPage.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await page.route('**/me/drives*', (route) => {
+      route.abort('failed')
+    })
+    await page.reload()
+
+    await ui.expectCrashPageToBeVisible({ actorsEnvironment })
+    await ui.logOutUser({ actorsEnvironment, stepUser: 'Alice' })
+  })
+
+  test('the crash page has not got any accessibility violations', async () => {
```
ghost:
```suggestion
  test('the crash page does not have any accessibility violations', async () => {
```

tests/e2e/support/objects/crash-page/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export class CrashPage {
+  #page: Page
+
+  constructor({ page }: { page: Page }) {
+    this.#page = page
+  }
+
+  getIsVisible() {
+    return expect(this.#page.locator(selectors.page).first()).toBeVisible()
+  }
```
ghost: according to the function i expect it to return something but it returns void.

LukasHirt: I renamed it to `assertVisibility`.

tests/e2e-playwright/steps/ui/crashPage.ts
LukasHirt: I did slightly different behaviour where I am passing the `page` directly. Saves one extra param and when doing multiple steps, we get the page from the actor only once instead of inside every step. Let me please know whether this change is fine.

packages/web-runtime/src/pages/crash.vue
```diff
@@ -0,0 +1,98 @@
...
+    <div class="card">
+      <h1 class="title">
+        {{
+          $pgettext(
+            'Title of the crash page displayed when the application crashes',
+            'Application crashed'
+          )
+        }}
+      </h1>
+      <p v-if="errorCode !== ''" class="error-code">
```
mzner, line 13: do we need to improve accessibility for this one? I guess some users won't be able to understand the connection between the error and the code. What do you think?

LukasHirt: For now, the error is there only for us. I.e. if we get a bug report, we know the code and know right away into which part of the code we should look into. In the future, we might consider adding some documentation where the codes could be listed and users/admins could orient themselves but as since we haven't got any error codes correlating with errors that the users could "fix" themselves I don't see too much value in doing that now.

Issue #13112, opened 2025-09-16
[QA] replace all cucumber test with playwright


nirajacharya2:
We want to move all tests from cucumber to playwright. The necessary setup is done in this PR https://github.com/owncloud/web/pull/12941. Now we need to migrate all test feature file by file.

- a11y
    - ~[ ] [a11y.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/a11y/a11y.feature)~
- admin-settings
    - [x] [general.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/admin-settings/general.feature) @anon-pradip https://github.com/owncloud/web/pull/13453
    - [x] [groups.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/admin-settings/groups.feature) @anon-pradip https://github.com/owncloud/web/pull/13472
    - [ ] [spaces.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/admin-settings/spaces.feature) @anon-pradip https://github.com/owncloud/web/pull/13488
    - [ ] [users.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/admin-settings/users.feature) @anon-pradip  https://github.com/owncloud/web/pull/13511
- app-provider
    - [x] [lock.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/app-provider/lock.feature) @nabim777  https://github.com/owncloud/web/pull/13505
    - [ ] [officeSuites.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/app-provider/officeSuites.feature)
    - [ ] [secureView.feature](https://github.com/owncloud/web/blob/140881e6b436d774d0dc0c92ac2483960850a030/tests/e2e/cucumber/features/app-provider/secureView.feature)
[...]

PR #13445, opened 2026-01-05
[tests-only] [full-ci] migrate gdprExport.feature to playwright test

Salipa-Gurung:
## Description
Migrated `gdprExport.feature` test to playwright only test.
## Related Issue

- Part of: https://github.com/owncloud/web/issues/13112

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [x] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/tests/e2e-playwright/playwright.config.ts
+++ b/tests/e2e-playwright/playwright.config.ts
@@ -9,6 +9,9 @@ const reportsDir = path.resolve(__dirname, '../../', config.reportDir)
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
+  // Set default test timeout to 60 seconds
+  timeout: config.timeout * 1000,
+
   // Look for test files in the following directory, relative to this configuration file.
   testDir: 'specs',
 


tests/e2e-playwright/playwright.config.ts
```diff
@@ -9,6 +9,9 @@ const reportsDir = path.resolve(__dirname, '../../', config.reportDir)
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
+  // Set default test timeout to 60 seconds
+  timeout: 60000,
```
ghost:
```suggestion
  timeout: config.timeout,
```

tests/e2e-playwright/specs/user-settings/gdprExport.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      },
+      browser: browser
+    })
+    await setAccessAndRefreshToken(usersEnvironment)
+    // Given "Admin" creates following users using API
+    //   | id    |
+    //   | Alice |
+    await api.userHasBeenCreated({ usersEnvironment, stepUser: 'Admin', userToBeCreated: 'Alice' })
+    // And "Alice" logs in
+    await ui.logInUser({ usersEnvironment, actorsEnvironment, stepUser: 'Alice' })
```
saw-jan, line 27: setup steps
```suggestion
    await setAccessAndRefreshToken(usersEnvironment)
    await api.userHasBeenCreated({ usersEnvironment, stepUser: 'Admin', userToBeCreated: 'Alice' })
    await ui.logInUser({ usersEnvironment, actorsEnvironment, stepUser: 'Alice' })
```

tests/e2e/support/objects/account/actions.ts
```diff
@@ -66,10 +66,7 @@ export const requestGdprExport = async (args: { page: Page }): Promise<void> =>
         resp.url().endsWith('.personal_data_export.json') &&
         resp.status() === 207 &&
         resp.request().method() === 'PROPFIND' &&
-        resp.text().then((text) => text.includes('HTTP/1.1 200 OK')),
-      // generating GDPR report can take a while
-      // so we need to increase the timeout to 60 seconds
-      { timeout: config.timeout * 1000 }
+        resp.text().then((text) => text.includes('HTTP/1.1 200 OK'))
```
saw-jan: hmm, i wonder why this change?

Salipa-Gurung: playwright test has a default timeout of 30 seconds, specifying a timeout here does not extend the test duration, so this 60000 ms timeout is ignored.

saw-jan: Good 👍, but we probably do not need to remove the timeout here. because the playwright action timeout and the playwright config timeout are two different things.
action timeout is for that action (an action must resolve withint that timeout), but config timeout is for a test scenario (a test scenario including all the actions must resolve within that timeout)

Salipa-Gurung: reverted the change.

saw-jan: hmm, but reverting will not work. you need to keep this timeout as well as add overall timeout (test scenario timeout) in playwright config

Salipa-Gurung: I've increased test scenario timeout to 70s for GDPR test only.

tests/e2e-playwright/specs/user-settings/gdprExport.spec.ts
saw-jan: set the default test timeout to `config.timeout` in the playwright config and then the following:
```suggestion
  test.setTimeout((config.timeout + 10) * 1000)
  let actorsEnvironment: ActorsEnvironment
  const usersEnvironment = new UsersEnvironment()

  test.beforeEach(async ({ browser }) => {
    actorsEnvironment = new ActorsEnvironment({
```

tests/e2e/support/objects/account/actions.ts
saw-jan: :+1: but we also need to set default timeout to `config.timeout`

tests/e2e-playwright/specs/user-settings/gdprExport.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { test } from '@playwright/test'
+import { config } from './../../../e2e/config.js'
+import { ActorsEnvironment, UsersEnvironment } from '../../../e2e/support/environment'
+import { setAccessAndRefreshToken } from '../../helpers/setAccessAndRefreshToken'
+import * as api from '../../steps/api/api'
+import * as ui from '../../steps/ui/index'
+
+test.describe('GDPR export', { tag: '@predefined-users' }, () => {
+  test.setTimeout((config.timeout + 10) * 1000)
+  let actorsEnvironment: ActorsEnvironment
```
saw-jan, line 9: maybe we can remove this timeout. this way, the behaviour will be like cucumber e2e
```suggestion
test.describe('GDPR export', { tag: '@predefined-users' }, () => {
  let actorsEnvironment: ActorsEnvironment
```

saw-jan (approved):

PR #13458, opened 2026-01-07
bugfix(web-app-admin-settings): [OCISDEV-474] fix showing hint to select a user and details when user is already selected

mzner:
We have fixed a bug in the admin-settings/users page where a hint for selecting a user and user's details were showing at the same time while a user has already been selected.

## Related Issue

- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-474

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-admin-settings/tests/unit/components/Users/SideBar/DetailsPanel.spec.ts
+++ b/packages/web-app-admin-settings/tests/unit/components/Users/SideBar/DetailsPanel.spec.ts
@@ -1,5 +1,6 @@
 import { User } from '@ownclouders/web-client/graph/generated'
 import DetailsPanel from '../../../../../src/components/Users/SideBar/DetailsPanel.vue'
+import UserInfoBox from '../../../../../src/components/Users/SideBar/UserInfoBox.vue'
 import { PartialComponentProps, defaultPlugins, shallowMount } from '@ownclouders/web-test-helpers'
 
 const defaultUser = { displayName: 'user', memberOf: [] } as User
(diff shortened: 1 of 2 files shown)


packages/web-app-admin-settings/tests/unit/components/Users/SideBar/DetailsPanel.spec.ts
```diff
@@ -25,15 +25,15 @@ describe('DetailsPanel', () => {
...
-    it('should have 0 users if no users are given', () => {
+    it('should be true if no users are given', () => {
       const { wrapper } = getWrapper({
         props: { user: null, users: [] }
       })
-      expect((wrapper.vm as any).noUsers).toEqual(0)
+      expect((wrapper.vm as any).noUsers).toBeTruthy()
     })
-    it('should have users if users are given', () => {
+    it('should be fals users if users are given', () => {
```
LukasHirt:
```suggestion
    it('should be false if users are given', () => {
```

packages/web-app-admin-settings/tests/unit/components/Users/SideBar/DetailsPanel.spec.ts
```diff
@@ -25,15 +25,15 @@ describe('DetailsPanel', () => {
   })
 
   describe('computed method "noUsers"', () => {
-    it('should have 0 users if no users are given', () => {
+    it('should be true if no users are given', () => {
       const { wrapper } = getWrapper({
         props: { user: null, users: [] }
       })
-      expect((wrapper.vm as any).noUsers).toEqual(0)
+      expect((wrapper.vm as any).noUsers).toBeTruthy()
```
LukasHirt, line 33: Since the previous test was also done like this, I leave this up to you whether you think it would be worth it and if you would like to tackle it here but I would say having the test actually checking what is rendered in the panel instead of the value of the computed prop would be much more useful. Knowing what the value is does not really tell me what effect it has on the UI.

mzner: Definitely makes sense! Please take another look.

PR #13462, opened 2026-01-08
bugfix(web-app-admin-settings): [OCISDEV-568] fixed missing group members in admin settings

mzner:
We have fixed the issue in which group members were not shown when user selects a group in the admin settings and trying to view its members.

## Related Issue

- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-568

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-admin-settings/src/views/Groups.vue
+++ b/packages/web-app-admin-settings/src/views/Groups.vue
@@ -42,7 +42,7 @@
                   v-model.trim="filterTerm"
                   :label="$gettext('Search')"
                   autocomplete="off"
-                  @enterKeyDown="filterGroups"
+                  @enter-key-down="filterGroups"
                 />
                 <oc-button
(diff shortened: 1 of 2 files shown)


mzner:
> Add a unit test for this please.

I thought about it but I couldn't think of a good unit test to add. The MembersPanel component needs this injection to be passed correctly. In our unit tests, this is all mocked, so even if I pass it in the unit test or don't the actual component will not show the real case scenario. Do you have a better idea or did I miss something? I guess an E2E would make more sense. What do you think?

LukasHirt (approved):
Sorry for not responding to your question earlier but I see you found the solution. Thank you!

PR #13486, opened 2026-01-16
feat(design-system, web-pkg): add explicit explanation to enable/disable sync functionality to avoid confusing the user

### LINKED ISSUE #10962: Disable/enable sync confudsed users who only use web
Precondition: user `einstein` uses only web client
Steps: 
- `admin` share folder `f1` with content to user `einstein`
- `einstein` `opens shares`->`f1` and can see folder content
- `einstein` disables sync -> he can still see folder content 

Actual: For a user who uses only web client `disable/enable sync` looks like something that's not needed or doesn't work

Expected: provide to users some clues. @tbsbdr what do you think

@butonic please add your remarks
CC @AlexAndBear

mzner:
add explicit explanation to enable/disable sync functionality to avoid confusing the user

 - add tooltip function to Action interface
 - remove aria-label when aria-label value is null so that it does not show in the html with empty value

## Related Issue

- Fixes https://github.com/owncloud/web/issues/10962

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
+++ b/packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
@@ -1,7 +1,7 @@
 <template>
   <li v-oc-tooltip="componentProps.disabled ? action.disabledTooltip?.(actionOptions) : ''">
     <oc-button
-      v-oc-tooltip="showTooltip || action.hideLabel ? action.label(actionOptions) : ''"
+      v-oc-tooltip="showTooltip || action.hideLabel ? tooltipText : ''"
       :type="componentType"
       v-bind="componentProps"
       :class="[action.class, 'action-menu-item', 'oc-py-s', 'oc-px-m', 'oc-width-1-1']"
@@ -9,7 +9,8 @@
       data-testid="action-handler"
       :size="size"
       justify-content="left"
-      :title="action.label(actionOptions)"
+      :title="tooltipText"
+      :aria-describedby="uniqueId"
       v-on="componentListeners"
     >
       <oc-img
@@ -51,11 +52,14 @@
         v-text="action.shortcut"
       />
     </oc-button>
+    <span :id="uniqueId" class="oc-invisible-sr">
+      {{ tooltipText }}
+    </span>
   </li>
 </template>
 
 <script lang="ts" setup>
-import { computed, unref } from 'vue'
+import { computed, unref, useId } from 'vue'
 import { Action, ActionOptions, useConfigStore } from '../../composables'
 import { useGettext } from 'vue3-gettext'
 import { storeToRefs } from 'pinia'
@@ -96,6 +100,16 @@ const componentType = computed<string>(() => {
   }
   console.warn('ActionMenuItem: No handler, route or href callback found in action', action)
   return 'button'
+})
+
+const uniqueId = useId()
+
+const tooltipText = computed<string>(() => {
+  if (action?.tooltip) {
+    return `${action.label(actionOptions)} - ${action.tooltip(actionOptions)}`
+  }
+
+  return action.label(actionOptions)
 })
 
 const ariaLabel = computed<string | null>(() => {
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsEnableSync.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsEnableSync.ts
@@ -12,7 +12,7 @@ import { useMessages, useConfigStore, useResourcesStore } from '../../piniaStore
 export const useFileActionsEnableSync = () => {
   const { showMessage, showErrorMessage } = useMessages()
   const router = useRouter()
-  const { $gettext, $ngettext } = useGettext()
+  const { $gettext, $ngettext, $pgettext } = useGettext()
 
   const clientService = useClientService()
   const loadingService = useLoadingService()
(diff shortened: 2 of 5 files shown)


packages/web-pkg/src/composables/actions/files/useFileActionsEnableSync.ts
```diff
@@ -82,6 +82,11 @@ export const useFileActionsEnableSync = () => {
       name: 'enable-sync',
       icon: 'check',
       handler: (args) => loadingService.addTask(() => handler(args)),
+      tooltip: () =>
+        $pgettext(
+          'Explanation tooltip for the enable sync action in files shared with me and spaces',
+          'Only affects desktop and mobile clients. You can toggle this setting here, but it will only take effect on those platforms.'
```
LukasHirt, line 88: Best to check with @DeepDiver1975 whether mobile clients are really affected or if it's only desktop...

LukasHirt, line 88: What would you say about being a bit more explicite what the settings actually do? Then maybe the mention that it does not affect web client might be redundant because it might be clear enough from the functionality description.

packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
```diff
@@ -51,11 +52,14 @@
         v-text="action.shortcut"
       />
     </oc-button>
+    <span :id="uniqueId" class="oc-invisible-sr">
+      {{ tooltipText }}
```
LukasHirt, line 56: Won't this result in the action label being read twice in case there is no custom tooltip?

packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
```diff
@@ -98,6 +102,16 @@ const componentType = computed<string>(() => {
   return 'button'
 })
 
+const uniqueId = useId()
+
+const tooltipText = computed<string>(() => {
+  if (action?.tooltip) {
+    return `${action.label(actionOptions)} - ${action.tooltip(actionOptions)}`
```
LukasHirt, line 109: I wonder whether we need to include the action label here or if we can rely only on the tooltip...

DeepDiver1975 (commented):
Thanks for tackling #10962, @mzner — the confusion around enable/disable sync for web-only users is real, and adding an explanation is the right direction. Reviewed as maintainer; a few points to address before this is mergeable.

### Accessibility — label likely announced twice (blocking)

Same concern @LukasHirt already raised. The hidden description span is rendered unconditionally and `aria-describedby` is always set on the button:

```vue
<span :id="uniqueId" class="oc-invisible-sr">
  {{ tooltipText }}
</span>
```
```js
const tooltipText = computed<string>(() => {
  if (action?.tooltip) {
    return `${action.label(actionOptions)} - ${action.tooltip(actionOptions)}`
  }
  return action.label(actionOptions)   // <- falls back to just the label
})
```

For the large majority of actions (which have no `tooltip`), `tooltipText` is exactly `action.label(...)`. So the button ends up with accessible **name** = label and accessible **description** = the same label, and screen readers will announce the label twice for every action menu item. Suggested fix:

- Only render the description span and set `aria-describedby` when `action.tooltip` is present (gate both on it), and
- in the description expose only the tooltip/explanation text, not `label + " - " + tooltip` — the label is already the accessible name, so the description should add only the new information (this also addresses @LukasHirt's "do we need the label here" point).

That keeps behavior unchanged for all existing actions and adds the explanation only where one is defined.

### Open product question still unresolved

@LukasHirt asked whether mobile clients are actually affected or only desktop, pending @DeepDiver1975's confirmation. The string currently asserts "desktop and mobile clients" — please get that confirmed before merge, since it's user-facing copy. If only desktop is affected the message is inaccurate. (Per @LukasHirt's other suggestion, describing what the setting *does* may also let you drop the explicit "does not affect web" framing.)

### Changelog missing (requested)

This is a user-facing change but there's no `changelog/unreleased/` fragment. Please add one following `changelog/TEMPLATE`, e.g. `changelog/unreleased/enhancement-explain-enable-disable-sync.md`:

```
Enhancement: Explain the enable/disable sync action

We've added an explanatory tooltip to the "Enable sync" and "Disable sync"
actions, clarifying which clients they affect, to avoid confusing users who
only use the web client.

https://github.com/owncloud/web/issues/10962
https://github.com/owncloud/web/pull/13486
```
(I held off pushing this fragment myself since the exact wording depends on resolving the desktop/mobile question above.)

### Tests missing

No unit tests were added. `ActionMenuItem.spec.ts` already exists — please cover the new `tooltipText` behavior: (a) an action without `tooltip` produces no extra description / no double labelling, and (b) an action with `tooltip` exposes the explanation. This also guards the a11y fix above against regression.

### Smaller notes (non-blocking)

- `OcButton.vue` `:aria-label="ariaLabel || null"` is a good change — drops the empty `aria-label=""` from the DOM, and no existing OcButton snapshots reference an empty aria-label, so nothing breaks there.
- `tooltip?(options?: T): string` on the `Action` interface is reasonable and consistent with `label`/`disabledTooltip`.
- The string literal is duplicated across `useFileActionsEnableSync.ts` and `useFileActionsDisableSync.ts` — fine to leave inline, just keep both in sync if the copy changes.

Summary: good idea, small footprint, but I'd like the double-announce a11y issue fixed, the desktop/mobile copy confirmed, a changelog fragment added, and a test before merging. Close once those are addressed.

LukasHirt:

PR #13500, opened 2026-01-20
feat: add instance switcher

LukasHirt:
## Description

Added a list of users' instances to the user menu. When multiple instances are enabled in oCIS, the user can switch between instances by clicking on the instance in the list. The list displays maximum 3 instances. If there are more instances, the user can click on a button to open the instances modal. Within the modal, the user can see all instances and switch between them by clicking on the instance or a "Switch" button.

## Motivation and Context

Users are able to switch between their instances.

## How Has This Been Tested?

- test environment: macos v26.2, chrome v144.0.7559.61
- test case 1: open user menu and switch instance
- test case 2: open instances modal and switch instance

Snímek obrazovky 2026-01-22 v 11 58 47

Snímek obrazovky 2026-01-22 v 11 57 40

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-runtime/src/composables/instances/useInstances.ts
+++ b/packages/web-runtime/src/composables/instances/useInstances.ts
@@ -1,0 +1,72 @@
+import { useModals, useUserStore } from '@ownclouders/web-pkg'
+import { storeToRefs } from 'pinia'
+import { computed, unref } from 'vue'
+import { useGettext } from 'vue3-gettext'
+import InstancesModal from '../../components/InstancesModal.vue'
+
+/** Limits the number of instances displayed in the user menu list */
+const INLINE_INSTANCES_LIMIT = 3
+
+export function useInstances() {
+  const { dispatchModal } = useModals()
+  const { $pgettext } = useGettext()
+
+  const userStore = useUserStore()
+  const { user } = storeToRefs(userStore)
+
+  const currentInstance = computed(() => window.location.origin)
+
+  const instances = computed(() => {
+    if (unref(user).instances.length < 1) {
+      return []
+    }
+
+    return unref(user)
+      .instances.map((instance) => ({
+        ...instance,
+        active: instance.url === unref(currentInstance)
+      }))
+      .toSorted((a, b) => {
+        if (a.active !== b.active) {
+          return a.active ? -1 : 1
+        }
+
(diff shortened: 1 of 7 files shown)


packages/web-runtime/src/composables/instances/useInstances.ts
```diff
@@ -0,0 +1,72 @@
+import { useModals, useUserStore } from '@ownclouders/web-pkg'
+import { storeToRefs } from 'pinia'
+import { computed, unref } from 'vue'
+import { useGettext } from 'vue3-gettext'
+import InstancesModal from '../../components/InstancesModal.vue'
+
+/** Limits the number of instances displayed in the user menu list */
+const INLINE_INSTANCES_LIMIT = 3
```
mzner, line 8: Is there a reason why we are limiting it to three? What if they have 6, which ones of these 6 are we showing?

packages/web-runtime/src/composables/instances/useInstances.ts
```diff
@@ -0,0 +1,72 @@
...
+  const { dispatchModal } = useModals()
+  const { $pgettext } = useGettext()
+
+  const userStore = useUserStore()
+  const { user } = storeToRefs(userStore)
+
+  const currentInstance = computed(() => window.location.origin)
+
+  const instances = computed(() => {
+    if (unref(user).instances.length < 1) {
```
mzner, line 20: Not that big of a difference, but was curious why `not === 0 or <=0` ? Can we have negative instances?

packages/web-runtime/src/composables/instances/useInstances.ts
LukasHirt: The user menu should not expand too much so that it becomes huge. That's why there is the show all button that triggers the modal which can then have whatever size is needed. Their order is handled in the composable. Instances are ordered by `Active` -> `Primary (alphabetically)` -> `Guest (alphabetically)`.

packages/web-runtime/src/composables/instances/useInstances.ts
LukasHirt: Simply to write less chars as `< 1` is the shortest 😁

Issue #5847, opened 2021-09-28
Set meta prop "theme-color"


kulmann:
The new Safari on macOS decorates the tab bar with the color from `<meta name="theme-color" content="#375f7E">`

![CleanShot 2021-09-27 at 21 34 40](https://user-images.githubusercontent.com/3532843/135050080-b56865d2-f914-4455-9691-817bd3397128.png)

Check if we can use the brand color from the currently loaded theme there.

cc @michaelstingl

pascalwengerter:
Perhaps not working, see https://css-tricks.com/meta-theme-color-and-trickery/#custom-properties

Shu12388y:
@kulmann can you assign this issue to me. I want to work on this issue.

dschmidt:
Can probably be added here: https://github.com/owncloud/web/blob/8723750b07ce49b04a9d0a3c46ebc064eedbe2f2/packages/web-runtime/src/composables/head/useHead.ts

kulmann:
> Can probably be added here: https://github.com/owncloud/web/blob/8723750b07ce49b04a9d0a3c46ebc064eedbe2f2/packages/web-runtime/src/composables/head/useHead.ts

I was just about to post the very same link 😆 Thank you @dschmidt !

Shu12388y:
@kulmann I have added the meta prop "theme-color"

PR #13509, opened 2026-01-21
enhancement: dynamic theme-color meta tag from theme brand color

priyanshuharshbodhi1:
## Description

Dynamic [theme-color](cci:7://file:///home/priyanshu/repos/Resume_Projects/owncloud-web/changelog/unreleased/enhancement-dynamic-theme-color:0:0-0:0) meta tag using the brand color from the loaded theme via `getComputedStyle()`.

## Related Issue

- Fixes https://github.com/owncloud/web/issues/5847

## Motivation and Context

Safari uses `<meta name="theme-color">` to style the tab bar. The hardcoded value didn't match custom themes.

## How Has This Been Tested?

- test environment: Local dev server, Safari macOS
- test case 1: Verified meta tag contains resolved hex color (not CSS variable)
- test case 2: Inspected `<head>` to confirm [theme-color](cci:7://file:///home/priyanshu/repos/Resume_Projects/owncloud-web/changelog/unreleased/enhancement-dynamic-theme-color:0:0-0:0) meta tag is present

## Types of changes

### DIFF
--- a/packages/web-runtime/src/composables/head/useHead.ts
+++ b/packages/web-runtime/src/composables/head/useHead.ts
@@ -1,4 +1,4 @@
-import { computed, unref, onMounted, ref } from 'vue'
+import { computed, unref } from 'vue'
 import { useHead as _useHead } from '@vueuse/head'
 import {
   useCapabilityStore,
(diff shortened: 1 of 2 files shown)


packages/web-runtime/src/composables/head/useHead.ts
```diff
@@ -15,6 +15,17 @@ export const useHead = () => {
 
   const favicon = computed(() => currentTheme.value.logo.favicon)
 
+  // Get brand color from CSS variable set by theme
+  const themeColor = ref('#375f7E')
+  onMounted(() => {
```
LukasHirt: Using this hook will not respect switching theme in the account settings.

packages/web-runtime/src/composables/head/useHead.ts
```diff
@@ -15,6 +15,17 @@ export const useHead = () => {
...
+    const brandColor = getComputedStyle(document.documentElement)
```
LukasHirt: Why use this instead of

```vue
const themeColor = computed(
  () => currentTheme.value.designTokens.colorPalette['swatch-brand-default']
)
```

This would give you the exact value of the theme color and you do not need to do the `getComputedStyle`.

LukasHirt (changes requested):

packages/web-runtime/src/composables/head/useHead.ts
ubermensch19: Great suggestion, switched to using computed() with the theme store directly.

priyanshuharshbodhi1:
@LukasHirt  PTAL!

LukasHirt (approved):
Awesome, thank you! Tested on macos v26.2.

PR #13512, opened 2026-01-22
fix: disable rename action in password-protected folder view

### LINKED ISSUE #12365: After renaming or deleting a password-protected folder from inside it, the changes are not reflected
### Describe the bug

When a password protected folder is renamed from inside it (using the 3 vertically aligned dots in the breadcrumb i.e., action menu), it shows newly renamed name in the breadcrumb and  we delete it. It says and shows that the folder is deleted but when we come outside the password protected folder, the folder is neither renamed nor deleted. <b> This is for both Personal and Project Space </b>

### Steps to reproduce

```textarea
1. Admin creates a password protected folder "mainFolder"
2. Admin opens and unlocks the password protected folder "mainFolder"
3. Admin renames "mainFolder" to "mainFolder123" using `context menu (three vertical dots) in breadcrumb`
4. Admin deletes "mainFolder123" using `context menu (three vertical dots) in breadcrumb`
5. Admin navigates outside the "mainFolder" by closing the password protected folder modal.

Same happens for:
1. navigates inside "mainFolder"
2. `rename "mainFolder" -> "mainFolder123" => renamed inside the password protected folder but not outside it
3. `close the password protected folder model`
4. `navigate inside "mainFolder"`
5. `delete "mainFolder123"`
```

### Expected behavior

- When "mainFolder" is renamed from inside it, the new name is also reflected when navigated outside the password protected folder.
- When "mainFolder" is deleted after renaming it from inside it, folder is deleted when navigated outside the password protected folder.

### Actual behavior

- When "mainFolder" is renamed from inside it, the new name is not reflected when navigated outside the password protected folder.
- When "mainFolder" is deleted after renaming it from inside it, folder is not deleted when navigated outside the password protected folder.

### Setup

```markdown
- Web version: ownCloud Web UI 11.3.1
- oCIS version: Infinite Scale 7.1.1+7146fefb3 Community
- browser:

OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

### Additional context

https://github.com/user-attachments/assets/570a0f8d-b859-4af6-b6f1-08ccfe91fa7a

ubermensch19:
## Description
Disables the **Rename** action in the breadcrumb context menu when viewing a password-protected folder (public link context).

## Related Issue
Fixes #12365

## Motivation and Context
When renaming a folder from inside a password-protected folder view, the action operates in the public link context. This causes the rename to create a duplicate folder instead of renaming the original.

The fix aligns with:
- The existing pattern in `useFileActionsDelete.ts` which already disables delete in public links
- The README documentation stating "Moving or renaming a link file is prevented by the webUI"

## How Has This Been Tested?
- [x] Unit tests pass (17 tests)
- [x] Manual: Verified rename option is hidden inside PPF view
- [x] Manual: Verified rename still works in normal folder views

## Types of changes

### DIFF
--- a/changelog/unreleased/bugfix-disable-rename-in-ppf
+++ b/changelog/unreleased/bugfix-disable-rename-in-ppf
@@ -1,0 +1,8 @@
+Bugfix: Disable rename action in password-protected folder view
+
+We've fixed a bug where renaming a folder from inside a password-protected
+folder view would create a duplicate folder instead of renaming the original.
+The rename action is now hidden in public link contexts, which aligns with
+the documented behavior that link files should not be manipulated from this view.
+
+https://github.com/owncloud/web/issues/12365
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
@@ -1,5 +1,9 @@
 import { isSameResource } from '../../../helpers/resource'
-import { isLocationTrashActive, isLocationSharesActive } from '../../../router'
+import {
+  isLocationTrashActive,
+  isLocationSharesActive,
+  isLocationPublicActive
+} from '../../../router'
 import { Resource } from '@ownclouders/web-client'
 import { dirname, join } from 'path'
 import { WebDAV } from '@ownclouders/web-client/webdav'
@@ -216,6 +220,11 @@ export const useFileActionsRename = () => {
       handler,
       isVisible: ({ resources }) => {
         if (isLocationTrashActive(router, 'files-trash-generic')) {
+          return false
+        }
+        // Disable rename in public link context (e.g., password-protected folders)
+        // See https://github.com/owncloud/web/issues/12365
+        if (isLocationPublicActive(router, 'files-public-link')) {
           return false
         }
         if (
--- a/packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
+++ b/packages/web-pkg/src/composables/actions/files/useFileActionsRename.ts
@@ -222,8 +222,7 @@ export const useFileActionsRename = () => {
         if (isLocationTrashActive(router, 'files-trash-generic')) {
           return false
         }
-        // Disable rename in public link context (e.g., password-protected folders)
-        // See https://github.com/owncloud/web/issues/12365
+
         if (isLocationPublicActive(router, 'files-public-link')) {
           return false
         }


LukasHirt (changes requested):
> The rename action is now hidden in public link contexts, which aligns with the documented behavior that link files should not be manipulated from this view.

Could you please point me where is this documented?

DeepDiver1975 (commented):
Thanks for tackling #12365, @ubermensch19 — the diagnosis (rename inside a PPF operating in the public-link context and diverging from the outer space) is correct, the changelog fragment is well-formed, and the delete-action analogy is a reasonable place to look. A few concerns before this can go in, though.

**1. The condition over-disables rename in every public link, not just the PPF-from-inside case.**
`isLocationPublicActive(router, 'files-public-link')` is true for the *entire* public-link file listing, for all resources and all recipients. Public links can be created with edit/write permissions, where renaming files and folders inside the link is a legitimate, working operation. This PR removes the Rename action for those users entirely. Note that the two sibling actions you'd expect to mirror — `useFileActionsDelete` and `useFileActionsMove` (cut) — deliberately *keep* `files-public-link` enabled (their guards `return false` only when the location is *not* public/spaces/...). So this change makes rename inconsistent with delete and move in public links, which is the opposite of the alignment the description claims.

**2. The README citation is about the link file, not folder contents.**
The README line ("Moving or renaming a link file is prevented by the webUI") refers to the `.psec` *link file* that represents the PPF — not to arbitrary files/folders viewed inside a public link. Blocking rename for the whole public-link view is a broader behavior change than that sentence supports.

**3. This treats a symptom rather than the bug.**
#12365 is a state-consistency bug: a rename inside the PPF succeeds on the server but the change isn't reflected in the outer space, and the UI shows a misleading state. Hiding the action avoids the misleading UI in the PPF case, but at the cost of disabling legitimate renames elsewhere, and it leaves the underlying propagation issue unaddressed. If the intent is specifically "don't allow renaming the PPF's own root folder from inside it," the guard should be scoped to that (e.g. the resource being the current/root folder of the public link), not the whole public-link context.

**4. Tests — the description and the diff disagree.**
The PR description says "Unit tests pass (17 tests)" and lists added unit tests, but the diff modifies only the source file and the changelog — `useFileActionsRename.spec.ts` is not changed. The existing `isVisible` test cases don't exercise a `files-public-link` router context, so there is no regression coverage asserting the new behavior (nor a test guarding against the over-disable above). Please add `isVisible` cases that mock the router location for both public-link and non-public contexts.

**5. CI is red.**
`continuous-integration/drone/pr` is currently failing and the PR is `BLOCKED` — needs to go green before merge regardless of the above.

Given the over-disable in editable public links (1–2), the symptom-vs-root-cause question (3), and the missing tests (4), I'm requesting changes rather than approving. Happy to re-review once the guard is scoped to the actual PPF-root case (or the underlying propagation is fixed) and tests cover it.

🤖 Automated review by Claude Code review agent (maintainer: @DeepDiver1975)

LukasHirt:

PR #13545, opened 2026-01-28
[full-ci] feat(web-client): strip base path

LukasHirt:
## Description

Added a new logic to the WebDAV client to strip the prefix from the path if it is present. Previously, only `/dav/` prefix would be stripped when parsing the response. Now, it will strip all the parts of the base remote URL. For example, if the base remote URL is `[https://example.com/my/nested/path/`](https://example.com/my/nested/path/%60), it will strip the `/my/nested/path/dav/` prefix.

## Motivation and Context

Support deployments where base server URL is using nested paths.

## How Has This Been Tested?

- test environment: chrome, macos
- test case 1: run oC on a subpath

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/web-client/src/webdav/client/parsers.ts
+++ b/packages/web-client/src/webdav/client/parsers.ts
@@ -3,6 +3,7 @@ import { XMLParser } from 'fast-xml-parser'
 import { WebDavResponseResource, WebDavResponseTusSupport } from '../../helpers'
 import { urlJoin } from '../../utils'
 import { DavErrorCode } from '../constants'
+import { join, normalize } from 'path'
 
 export const parseTusHeaders = (headers: Headers) => {
   const result: WebDavResponseTusSupport = {}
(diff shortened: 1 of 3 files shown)


packages/web-client/src/webdav/client/parsers.ts
```diff
@@ -25,12 +26,13 @@ export const parseTusHeaders = (headers: Headers) => {
   return result
 }
 
-export const parseMultiStatus = async (xmlBody: string) => {
+export const parseMultiStatus = async (xmlBody: string, remoteBasePath: string) => {
   const parseFileName = (name: string) => {
     const decoded = decodeURIComponent(name)
-    if (name?.startsWith('/dav/')) {
+    const prefix = normalize(join(remoteBasePath, 'dav'))
```
mzner, line 32: @LukasHirt  We are only removing when it contains 'dav'. How are we removing any other prefixes?

LukasHirt: It joins the `dav` string with `remoteBasePath` so all of that gets removed.

mzner: > It joins the `dav` string with `remoteBasePath` so all of that gets removed.

So `dav` has to be there?

LukasHirt: Yes.

PR #13628, opened 2026-03-13
feat: add vault and drive to breadcrumbs

mzner:
## Description

  Breadcrumbs in the Projects and GenericSpace views now reflect the current vault/drive context. When the user has vault access, the
  root breadcrumb shows "Vault" or "Drive" depending on the active scope. For the personal space, a parent breadcrumb is prepended.
  Users without vault access see the original labels unchanged.

  ## Related Issue

  - Fixes https://kiteworks.atlassian.net/browse/OCISDEV-711

  ## Motivation and Context

  With the introduction of vault mode, users need visual feedback in the breadcrumb navigation to understand whether they are browsing
  within the vault or drive scope. Without this, the navigation context is unclear when switching between modes.

  ## How Has This Been Tested?

  - test environment: local dev server with oCIS backend, vault permissions enabled
  - test case 1: User with vault access sees "Drive" breadcrumb on default scope
  - test case 2: User with vault access sees "Vault" breadcrumb on vault scope
  - test case 3: User without vault access sees "Spaces" in Projects view
  - test case 4: User without vault access sees only space name in personal space
  - test case 5: Personal space with vault access shows "Drive > {space name}" or "Vault > {space name}"
  - test case 6: Unit tests pass for all breadcrumb combinations in both Projects.spec.ts and GenericSpace.spec.ts

  ## Types of changes

  - [ ] Technical debt
  - [ ] Tests
  - [ ] Documentation

### DIFF
--- a/packages/web-app-files/src/views/spaces/GenericSpace.vue
+++ b/packages/web-app-files/src/views/spaces/GenericSpace.vue
@@ -129,6 +129,7 @@ import {
 import {
   ResourceTransfer,
   TransferType,
+  useAbility,
   useConfigStore,
   useExtensionRegistry,
   useFileActions,
@@ -189,6 +190,7 @@ interface Props {
 const { space = null, item = null, itemId = null } = defineProps<Props>()
 
 const router = useRouter()
+const { can } = useAbility()
 const userStore = useUserStore()
 const { $gettext, $ngettext } = useGettext()
 const openWithDefaultAppQuery = useRouteQuery('openWithDefaultApp')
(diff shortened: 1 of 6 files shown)


packages/web-app-files/src/views/spaces/GenericSpace.vue
```diff
@@ -261,7 +261,7 @@ const breadcrumbs = computed(() => {
   if (isProjectSpaceResource(unref(space))) {
     rootBreadcrumbItems.push({
       id: uuidV4(),
-      text: $gettext('Spaces'),
+      text: unref(route).params.scope === 'vault' ? $gettext('Vault') : $gettext('Drive'),
```
LukasHirt: Has this been clarified whether we really need to include the `Drive`? Most instances will be deployed without vault and in such case this feels like unnecessary string.

LukasHirt (changes requested):
Missing changelog item.

packages/web-app-files/src/views/spaces/GenericSpace.vue
mzner: We will need to check with David/Julian.

PR #13659, opened 2026-04-09
chore: generate web docs using github actions

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,44 @@
+name: documentation
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+permissions:
+  contents: write
+
+jobs:
+  documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: setup go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version: "1.25"
+
+      - name: docs-generate
+        run: make docs-generate
+
+      - name: docs-copy
+        run: make docs-copy
+
+      - name: docs-hugo-drone-prep
+        run: make docs-hugo-drone-prep
+
+      - name: docs-build
+        run: cd hugo && hugo
+
+      - name: publish
+        if: github.event_name != 'pull_request'
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: docs-generate
+        run: make docs-generate
+
+      - name: docs-copy
+        run: make docs-copy
+
+      - name: docs-hugo-drone-prep
+        run: make docs-hugo-drone-prep
+
+      - name: setup hugo
```
LukasHirt: This is not necessary. The makefile uses docker for docs build/serve. Since you already gall `make docs-generate` you can simply call `make docs-build`.

LukasHirt (changes requested):

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,55 @@
...
+        if: github.event_name != 'pull_request'
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs/hugo/content/web
+          publish_branch: docs
+          commit_message: "Automated documentation update [skip ci]"
+          keep_files: false
+
+      - name: list and remove temporary files
```
LukasHirt: I wonder whether we need this. Every run is independent and is destroyed after it ends so I don't really see the value in clean up. I know this was already running in drone but I guess we do not need to bring this over?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: make docs-generate
+
+      - name: docs-copy
+        run: make docs-copy
+
+      - name: docs-hugo-drone-prep
+        run: make docs-hugo-drone-prep
+
+      - name: docs-build
+        run: cd hugo && hugo
```
LukasHirt, line 34: This can be replaced with calling the makefile build.

LukasHirt (changes requested):

mmattel:
@LukasHirt as already made Matteo aware, there is no need to push that forward anymore because a  required backend repo will not be updated to GH actions. This means that the current state of the dev docs on the web stays as it is and no new content can be added. The dev docs will be migrated into the Antora framework and when finished, we can remove the doc component here (and all the other contributing repos).

.github/workflows/documentation.yml
mzner: Given the new information, if there is nothing we have missed or you have a different opinion, shall we close this PR?

LukasHirt:
I'll close this PR then.

PR #13667, opened 2026-04-14
bugfix(web-app-files): [OCISDEV-775] shared with does not show members

mzner:
We've fixed an issue where the list of external users in the "Shared with" section was not displayed.

## Related Issue

- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-775

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests
- [ ] Documentation

## Open tasks:

- [ ] ...

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
@@ -282,18 +282,9 @@ const shareOwnerDisplayName = computed(() => share.sharedBy.displayName)
 
 const externalShareDomainName = computed(() => {
   if (unref(isExternalShare)) {
-    let decodedId = ''
+    const [, serverUrl] = share.sharedWith.id.split('@')
 
-    try {
-      decodedId = atob(share.sharedWith.id)
-    } catch (_) {
-      decodedId = share.sharedWith.id
-    }
-
-    const [, serverUrl] = decodedId.split('@')
-    const domain = new URL(serverUrl).hostname
-
-    return domain
+    return serverUrl
   }
 
   return null
--- a/packages/web-app-files/tests/unit/components/SideBar/Shares/Collaborators/ListItem.spec.ts
+++ b/packages/web-app-files/tests/unit/components/SideBar/Shares/Collaborators/ListItem.spec.ts
@@ -37,7 +37,7 @@ const getShareMock = ({
 }: Partial<CollaboratorShare> = {}): CollaboratorShare => ({
   id: '1',
   sharedWith: sharedWith || {
-    id: 'ZWluc3RlaW5AaHR0cHM6Ly93d3cubG9yZW0uY29t',
+    id: 'ZWluc3RlaW5AaHR0cHM6Ly93d3cubG9yZW0uY29t@www.lorem.com',
     displayName: 'einstein'
   },
   sharedBy: { id: '2', displayName: 'marie' },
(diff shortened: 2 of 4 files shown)


packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
```diff
@@ -282,18 +282,9 @@ const shareOwnerDisplayName = computed(() => share.sharedBy.displayName)
 
 const externalShareDomainName = computed(() => {
   if (unref(isExternalShare)) {
-    let decodedId = ''
+    const [, serverUrl] = share.sharedWith.id.split('@')
 
-    try {
-      decodedId = atob(share.sharedWith.id)
```
LukasHirt, line 288: Is it safe to remove this? Are we sure that _all_ IDs will now be a plain string and not base64 encoded string?

packages/web-app-files/tests/unit/components/SideBar/Shares/Collaborators/ListItem.spec.ts
```diff
@@ -210,7 +210,7 @@ describe('Collaborator ListItem component', () => {
     it('should fallback to original ID if decoding fails', () => {
       const share = getShareMock({
         shareType: ShareTypes.remote.value,
-        sharedWith: { id: 'einstein@https://www.lorem.com', displayName: 'einstein' }
```
LukasHirt, line 213: Is it expected now that all IDs will omit protocol?

LukasHirt (commented):
Code-wise it looks good but I have no idea about the actual fix here. What was the reason that the list was not displayed? Are these changes safe since they're dropping two specific guardrails?

packages/web-app-files/src/components/SideBar/Shares/Collaborators/ListItem.vue
2403905: Yes, it is safe.

packages/web-app-files/tests/unit/components/SideBar/Shares/Collaborators/ListItem.spec.ts
mzner: @2403905 We also don't include any protocols anymore, correct? For example https.

2403905: https://github.com/cs3org/OCM-API/blob/bd94a9a104dc1b1d4a24daa87fcf068623d201c4/IETF-RFC.md?plain=1#L709

2403905: Yes, we are not including a protocol now

2403905 (approved):

packages/web-app-files/tests/unit/components/SideBar/Shares/Collaborators/ListItem.spec.ts
```diff
@@ -210,7 +210,7 @@ describe('Collaborator ListItem component', () => {
     it('should fallback to original ID if decoding fails', () => {
```
LukasHirt: I don't think this test is then needed anymore as it has been done specifically to verify the fallback which is no longer in place.

LukasHirt (changes requested):

LukasHirt (approved):

PR #13686, opened 2026-04-21
chore: open a PR on changelog

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Open tasks:

- [ ] ...

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -11,37 +11,13 @@ on:
       - stable-*
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   changelog:
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: checkout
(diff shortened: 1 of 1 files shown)


LukasHirt (commented):
Why not use the action from https://github.com/owncloud/reusable-workflows/blob/main/.github/workflows/calens.yml ?

LukasHirt (approved):

LukasHirt (changes requested):
Needs to have read permissions:

```sh
The workflow is not valid. .github/workflows/changelog.yml (Line: 16, Col: 3): Error calling workflow 'owncloud/reusable-workflows/.github/workflows/calens.yml@main'. The nested job 'changelog' is requesting 'contents: read', but is only allowed 'contents: none'.
```

LukasHirt (approved):

PR #13696, opened 2026-04-22
chore: [OCISDEV-736] add web release jobs to gh actions

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Open tasks:

- [ ] ...

### DIFF
--- a/.github/workflows/release-npm.yml
+++ b/.github/workflows/release-npm.yml
@@ -1,0 +1,82 @@
+name: Release NPM
+
+on:
+  push:
+    tags:
+      - 'babel-preset-v*'
+      - 'design-system-v*'
+      - 'eslint-config-v*'
+      - 'extension-sdk-v*'
+      - 'prettier-config-v*'
+      - 'tsconfig-v*'
+      - 'web-client-v*'
+      - 'web-pkg-v*'
+      - 'web-test-helpers-v*'
+
+jobs:
+  release-npm:
+    name: Publish NPM package
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
+        with:
+          version: 10.29.2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: 24.15.0
+          registry-url: 'https://registry.npmjs.org'
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Determine package
+        id: package
+        run: |
+          TAG="${{ github.ref_name }}"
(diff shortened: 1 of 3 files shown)


LukasHirt (commented):
Curious, what do you think about splitting it into several workflows? I think it might make it a bit easier… basically having web release workflow, docker workflow and npm workflow. Or maybe docker needs to be tied to release…

.github/workflows/release-web.yml
```diff
@@ -0,0 +1,58 @@
...
+  release-web:
+    name: Publish GitHub release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
```
LukasHirt: Do we need the full history here?

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: 24.15.0
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Determine package
```
LukasHirt, line 42: I guess we don't need to do it now but it might be interesting to see whether this could be handled via matrix. Might be more elegant.

LukasHirt (commented):
When publishing npm packages, we also do a release for them in GH. That seems to be missing now?

.github/workflows/release-web.yml
mzner: > Do we need the full history here?

it is needed there for the changelog/calens.

.github/workflows/release-npm.yml
mzner: You mean we should remove these two steps? Install and Determine?

LukasHirt: No, I mean the `matrix` option from GH actions: https://github.com/owncloud/web/blob/6d1d01e97e5b3403f0a05fe3de00baae7eab456c/.github/workflows/test.yml#L76

But let's not do it now. First we get it working and then we can think about optimising.

.github/workflows/release-web.yml
LukasHirt: But calens is not using the history but the actual files which are part of the cloned code even when only the specific commit is fetched.

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Publish to NPM
+        run: |
+          env "npm_config_//registry.npmjs.org/:_authToken=${NPM_TOKEN}" pnpm whoami
+          env "npm_config_//registry.npmjs.org/:_authToken=${NPM_TOKEN}" pnpm publish \
+            --no-git-checks --filter "@ownclouders/${{ steps.package.outputs.name }}" \
+            --access public --tag latest
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish GitHub release
```
LukasHirt: Shoot… I confused this with web-extensions repo. We don't do the releases of them here. I'm sorry, my bad. Can you remove it again? 🙊

LukasHirt (approved):

PR #13718, opened 2026-04-27
ci: drop calens prefix from app secrets

LukasHirt:


mzner (approved):

LukasHirt:
This is actually not correct - closing.

PR #13728, opened 2026-05-04
[full-ci] chore: [OCISDEV-790] add ci-full flag

mzner:
## Description

Introduces a `check-flags` job to the GitHub Actions test workflow that gates downstream jobs based on changed file paths and an opt-in `full-ci` flag in the PR title.

**Changes:**
- New `check-flags` job runs before all other jobs and produces three outputs:
  - `full-ci`: `true` when the PR title contains `full-ci` (case-insensitive)
  - `ci-relevant`: `true` when files relevant to CI (packages, tests, config files, etc.) were changed
  - `affected-suites`: the subset of test suites affected by the changed files, determined via filterTestSuitesToRun.js
- The `build` job is now skipped on PRs when no CI-relevant files changed and `full-ci` is not set
- The `e2e-playwright` job skips individual test suites when they are not in the affected suites list
- Two suites (`app-provider`, `ocm`) are marked `requires_full_ci: true` and only run when `full-ci` is explicitly requested
- `check-tests-passed` now also depends on `build`

## Related Issue
- Fixes https://kiteworks.atlassian.net/browse/OCISDEV-790

## Motivation and Context

CI runs on every PR are time-consuming and resource-intensive. Previously all jobs ran regardless of which files were changed. This change makes jobs conditional on the actual changes in the PR, while providing a `full-ci` opt-in in the PR title for cases where complete coverage is required.

## How Has This Been Tested?
- test environment: GitHub Actions (PR workflow)
- test case 1: PR with no CI-relevant file changes → `build` and `e2e-playwright` jobs are skipped
- test case 2: PR with changes to source packages → only affected test suites are executed
- test case 3: PR title includes `full-ci` → all jobs and suites run unconditionally
- test case 4: Push/tag event → all jobs run as before

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,86 @@ on:
   pull_request:
 
 jobs:
+  check-flags:
+    runs-on: ubuntu-latest
+    outputs:
+      full-ci: ${{ steps.pr-title.outputs.full-ci }}
+      ci-relevant: ${{ steps.paths.outputs.ci-relevant }}
+      affected-suites: ${{ steps.affected-suites.outputs.suites }}
+    steps:
+      - name: Check for full-ci flag in PR title
+        id: pr-title
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          if [[ "${PR_TITLE,,}" == *"full-ci"* ]]; then
+            echo "full-ci=true" >> $GITHUB_OUTPUT
+          else
+            echo "full-ci=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Checkout code
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        if: github.event_name == 'pull_request'
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: 24.15.0
+
+      - name: Detect changed files
+        id: paths
+        if: github.event_name == 'pull_request'
+        uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        with:
+          filters: |
+            ci-relevant:
+              - 'packages/**'
+              - '!packages/web-app-skeleton/**'
+              - 'polyfills/**'
+              - 'tests/**'
+              - '!tests/e2e/**'
+              - 'vite*.ts'
+              - 'tsconfig.json'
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -214,7 +314,7 @@ jobs:
           retention-days: 7
 
   check-tests-passed:
-    needs: [e2e-playwright]
+    needs: [build, e2e-playwright]
```
LukasHirt: Is this change needed? Tests already wait for build.

LukasHirt (changes requested):
In drone, we were using `filterTestSuitesToRun.js` to find what tests should run. If you find the current approach works better I am fine with not using that but I would:

1. check whether the result is the same
2. if it won't be used anymore, delete it

Also, if we go with this approach, the GH action needs to be whitelisted.

.github/workflows/test.yml
mzner: > Is this change needed? Tests already wait for build.

build is needed here because needs.*.result only evaluates direct dependencies. 
If build fails, e2e-playwright gets skipped, not failed, due to its if condition. Without build in check-tests passed's needs, the gate would only see e2e-playwright: skipped, which doesn't match failure or cancelled, and the check would pass despite a broken build.

LukasHirt (changes requested):
Seems like there is still some action needing to be whitelisted.

mzner:
> Seems like there is still some action needing to be whitelisted.

I'm waiting for your approval here first before asking Thomas to white list it.

LukasHirt:
> I'm waiting for your approval here first before asking Thomas to white list it.

It would be best to first see the CI running here to see whether it works as expected. I would approve the code as it is now but waiting for the successful run is necessary.

LukasHirt (approved):

mzner:
@PrajwolAmatya Could you please confirm the changes in this commit? Some of the latest changes we have in this commit "https://github.com/owncloud/web/commit/9bd7e30cc7aa616d3974b9c69dd48573256f1b6f" broke the ci cause the `filter test js file` was not updated.

tests/drone/filterTestSuitesToRun.js
```diff
@@ -5,14 +5,7 @@ import path from 'path'
...
 // paths that if changed will run all test suites
-const mandatoryPaths = [
-  'tests/e2e/',
-  'tests/e2e-playwright/',
-  'tests/drone/',
-  '.drone.star',
-  '.drone.env',
-  'package.json'
-]
+const mandatoryPaths = ['tests/e2e/', 'tests/drone/', '.drone.star', '.drone.env', 'package.json']
```
PrajwolAmatya:
```suggestion
const mandatoryPaths = ['tests/e2e/', 'tests/drone/', 'package.json']
```
This can also be removed. Other than that LGTM. 👍

PR #13769, opened 2026-05-11
feat(web-pkg, web-app-files, web-app-search): [OCISDEV-853] add vault:true token to search payload in vault context

mzner:
## Description
Implements vault search separation by appending the `vault:true` query token to the `<oc:pattern>` search payload when the user is in the vault context.

The token is injected in both **All files** and **Current folder** search scenarios:

```xml
<!-- All files -->
<oc:pattern>(name:"*foo*" OR content:"foo") vault:true</oc:pattern>

<!-- Current folder -->
<oc:pattern>(name:"*foo*" OR content:"foo") vault:true scope:&lt;space-id&gt;</oc:pattern>
```

## Related Issue
- Fixes https://jira.owncloud.com/browse/OCISDEV-853

## Motivation and Context
Without this token, searching in the vault context would return results from all spaces. The `vault:true` token ensures the backend scopes the search correctly to vault content only.

## How Has This Been Tested?
- test environment: oCIS local + Docker
- test case 1: Search in vault context — verify `vault:true` appears in the request payload and results are scoped to vault
- test case 2: Search outside vault context — verify `vault:true` is NOT included in the request payload

### DIFF
--- a/changelog/unreleased/enhancement-add-vault-to-search.md
+++ b/changelog/unreleased/enhancement-add-vault-to-search.md
@@ -1,0 +1,5 @@
+Enhancement: Add vault search separation
+
+We've implemented vault search separation by adding the `vault:true` query token to the `<oc:pattern>` search payload. The token is now included in both "All files" and "Current folder" search requests, ensuring vault content is correctly scoped in all search scenarios.
+
+https://github.com/owncloud/web/pull/13769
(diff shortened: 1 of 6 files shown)


LukasHirt (changes requested):
Could you add some small unit test confirming the change please?

LukasHirt (approved):

PR #13770, opened 2026-05-11
chore: add dependabot weekly update config

DeepDiver1975:
This PR adds/updates the Dependabot configuration for this repository.

Dependabot is configured to run weekly (Sunday at 22:00 UTC) with a limit of 5 open pull requests per ecosystem. Minor and patch updates are grouped into a single PR per ecosystem to reduce the number of concurrent update PRs, lowering maintainer overhead and CI load. Major version updates remain as individual PRs so they receive deliberate review.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,29 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: sunday
+      time: '22:00'
+      timezone: Etc/UTC
+    open-pull-requests-limit: 5
+    groups:
+      minor-and-patch:
+        update-types:
+          - minor
+          - patch
+
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: sunday
+      time: '22:00'
+      timezone: Etc/UTC
+    open-pull-requests-limit: 5
+    groups:
+      minor-and-patch:
+        update-types:
+          - minor
+          - patch


LukasHirt (changes requested):
We are using Renovate in this repository. This would then cause having two separate kinds of dependency update PRs. For this repo specifically, the interval can be configured in https://github.com/owncloud/web/blob/master/.renovaterc.json

It should be also possible to configure it globally either in the shared config in https://github.com/owncloud-ops/renovate-presets/blob/main/docker.json or via Renovate bot settings.

DeepDiver1975:
> It should be also possible to configure it globally either in the shared config in https://github.com/owncloud-ops/renovate-presets/blob/main/docker.json or via Renovate bot settings.

as long as there is noone maintaining this I vote against renovatebot .... :shrug:

LukasHirt:
I'm fine with moving back to dependabot. Especially for consistency because only some of frontend repos are using renovate and everything else uses dependabot. However, this will have a few implications that need to be taken care of or at least kept in mind:

1. Renovate automatically groups things like linters or packages belonging together. To persist this, we would need to manually create groups in dependabot.
2. Renovate does not truncate the changelog as dependabot does. It's a bit annoying with dependabot but it's nothing that cannot be worked around by simply opening the changelog of the dependency directly.
3. We need to check where else is renovate used. I know about https://github.com/owncloud/web-extensions and https://github.com/owncloud/web-app-skeleton but there might be more.

@DeepDiver1975 do you want to merge this PR right away or should I drop the renovate config from this repo first?

DeepDiver1975:
> 1. Renovate automatically groups things like linters or packages belonging together. To persist this, we would need to manually create groups in dependabot.

there is already some grouping in this pr - further groups can be defined within each project individually



> 2\. Renovate does not truncate the changelog as dependabot does. It's a bit annoying with dependabot but it's nothing that cannot be worked around by simply opening the changelog of the dependency directly.

I don't get this part - care to explain this more?




> 3\. We need to check where else is renovate used. I know about https://github.com/owncloud/web-extensions and https://github.com/owncloud/web-app-skeleton but there might be more.


PR should already be open there as well....




> 3\. 
> 
> 
> @DeepDiver1975 do you want to merge this PR right away or should I drop the renovate config from this repo first?

first bring dependabot to a level we are are comfortable with - maybe first address this even in a smaller/less frequently updated repo .... and drop renovate last

LukasHirt:
> 2. Renovate does not truncate the changelog as dependabot does. It's a bit annoying with dependabot but it's nothing that cannot be worked around by simply opening the changelog of the dependency directly.

> I don't get this part - care to explain this more?

If the changelog in the PR gets longer dependabot truncates it. Renovate doesn't and simply prints the full changelog. So when checking what the new version brings it's one extra step to go to the changelog directly. That's no obstacle but I still mentioned it just to make it known.

image

LukasHirt (approved):

LukasHirt:
Rebased to bring in the fix for the flaky E2E test.

PR #13796, opened 2026-05-19
fix(files-drop): Fix upload confirmation not visible at small viewports

awabcodes:
## Description

Fix a responsive design issue where the upload confirmation message was not visible on the file drop page.

Refactor the layout to use natural document flow instead of fragile height-dependent CSS

## Related Issue

- Fixes [OCISDEV-703](https://kiteworks.atlassian.net/browse/OCISDEV-703)

## Motivation and Context

The upload confirmation message was invisible at small browser window sizes when using file drop links. The UploadInfo panel was placed in a 1fr grid row whose height depended on a broken height: 100% chain from the root element. The oc-overflow hidden wrapper then silently clipped the collapsed panel. This caused users to be unable to confirm their upload succeeded

## How Has This Been Tested?

- test environment: Chrome on macOS,
- test case 1: Open a file drop link at small viewport, upload a file. confirmation message is now visible immediately
- test case 2: Resize from large to small viewport after upload. confirmation remains visible
- ...

## Open tasks:

- [ ] ...

### DIFF
--- a/changelog/unreleased/bugfix-files-drop-upload-info-not-visible.md
+++ b/changelog/unreleased/bugfix-files-drop-upload-info-not-visible.md
@@ -1,0 +1,5 @@
+Bugfix: Fix upload confirmation not visible on file drop page
+
+We've fixed a responsive layout issue where the upload confirmation message was not visible at small browser window sizes on the file drop page. The `UploadInfo` panel was being clipped due to broken CSS. The layout now uses natural document flow so the confirmation is always visible regardless of viewport size.
+
+https://github.com/owncloud/web/pull/13796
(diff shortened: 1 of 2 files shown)


LukasHirt (changes requested):
Change looks good. Just add changelog item please. Here's some hint on how to do that:

- readme: https://github.com/owncloud/web/blob/master/changelog/README.md
- template: https://github.com/owncloud/web/blob/master/changelog/TEMPLATE
- example of existing one: https://github.com/owncloud/web/blob/master/changelog/unreleased/enhancement-strip-webdav-prefix.md

mzner:
@awabcodes Could you please sign the commit?

PR #13805, opened 2026-05-22
chore: add claude to git ignore

mzner:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Open tasks:

- [ ] ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ release/*
 **/l10n/locale
 **/l10n/template.pot
 **/hugo
+*claude*
 
 build/version.json
 


LukasHirt (commented):
This could also ignore `CLAUDE.md` depending on developers config. I would say that file might be useful in the future. What do you think about not ignoring that specific file (in any subdir as well)?

mzner:
> This could also ignore `CLAUDE.md` depending on developers config. I would say that file might be useful in the future. What do you think about not ignoring that specific file (in any subdir as well)?

Yea make sense. I will close this PR. 👍🏻

PR #13844, opened 2026-06-05
fix(web-pkg): [OCISDEV-777] validate postMessage origin in embed mode…

awabcodes:
… modals

## Description
The embed mode modals (`SaveAsModal`, `ExportAsPdfModal`, `FilePickerModal`) registered global `window` `message` listeners and acted on incoming `owncloud-embed:*` messages without validating `event.origin`. Any page holding a reference to an authenticated ownCloud window (e.g. via `window.open`) could forge an `owncloud-embed:select`, `owncloud-embed:file-pick` or `owncloud-embed:cancel` message and trigger an authenticated file write in the victim's space (CSRF).

## Related Issue

- Fixes [OCISDEV-777](https://kiteworks.atlassian.net/browse/OCISDEV-777)

## How Has This Been Tested?

- test environment: local oCIS, chrome browser
- test case 1: manual - reproduced the report's PoC from a foreign origin via `window.open` + `postMessage(payload, '*')`; no file is created and the modal stays open (before the fix a file was written)
- ...

## Open tasks:

- [ ] ...

PR #13906, opened 2026-06-24
fix(web-app-external): enable Collabora Save As / export to storage

dj4oC:
## Problem

Office documents opened in Collabora Online can be **renamed** but **"Save As" / "Export As" silently does nothing** — the copy is never written back into oCIS storage.

Root cause: Collabora implements "Save As" as a **host-delegated** operation. When the user triggers it, Collabora posts a `UI_SaveAs` message and waits for the integration to reply with the target filename via `Action_SaveAs`; only then does it render to the chosen format and call WOPI `PutRelativeFile`. The app-provider (`web-app-external`) never opened that postMessage channel — it never sent `Host_PostmessageReady` and never handled `UI_SaveAs` (it only listened for `UI_Edit`). Rename works because Collabora performs it as a *native* WOPI `RenameFile` call that needs no host involvement.

## Fix

In `web-app-external/src/App.vue`:
- Send `Host_PostmessageReady` once the editor iframe has loaded — this is what makes Collabora start emitting its postMessage API (`App_LoadingStatus`, `UI_SaveAs`, …).
- Append `ui_defaults=SaveAsMode=group` to the editor URL so Collabora shows the grouped Save-As control that delegates the target filename to the host (the parameter is Collabora-specific and ignored by other app providers).
- Handle `UI_SaveAs`: prompt for the copy's name (the extension selects the export format) and reply with `Action_SaveAs`. Incoming messages are now origin-checked against the editor origin.

No server change is needed — the collaboration service already implements `PutRelativeFile`.

## Testing

- Unit: `web-app-external` tests pass; snapshot updated for the `ui_defaults` URL parameter.
- `check:types`, ESLint clean.
- Verified live against `ocis_full` + Collabora Online `25.04.7.3.1`:
  - `PutRelativeFile` writes the converted copy into the space (confirmed end-to-end: a real call created a `.pdf` next to the source; `PutFile: success` → `PutRelativeFileSuggested: success`, HTTP 200).
  - Sending `Host_PostmessageReady` on iframe load opens Collabora's full postMessage channel (`App_LoadingStatus` → `Document_Loaded`); without it Collabora stays silent. `CheckFileInfo` already returns `UserCanNotWriteRelative:false` + `PostMessageOrigin`.
  - The end-to-end `UI_SaveAs → Action_SaveAs → PutRelativeFile` chain is exercised by the office web e2e suite (`webOffice`), which drives Collabora's Save-As UI.

## Risk

Low. Behaviour is additive: a new postMessage handshake + an extra UI param. The existing `UI_Edit` handling is preserved (now also origin-checked). Non-Collabora app providers ignore the `ui_defaults` parameter.

### DIFF
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -110,6 +110,10 @@ const appName = computed(() => {
   return appProviderService.appNames.find((appName) => appName.toLowerCase() === lowerCaseAppName)
 })
 
+// The grouped Save-As control and the Host_PostmessageReady handshake are
+// Collabora-specific WOPI extensions; other app providers don't implement them.
+const isCollabora = computed(() => unref(appName) === 'Collabora')
+
 const appUrl = ref()
 const formParameters = ref({})
 const method = ref()
@@ -129,14 +133,17 @@ const appOrigin = computed(() => {
 // Collabora exposes "Save As" (export to another format, saved back to storage
 // via WOPI PutRelativeFile) as a host-delegated operation: it shows a grouped
 // Save-As control and posts a `UI_SaveAs` message, expecting the host to reply
-// with the target filename. Request that grouped control via `ui_defaults`.
-// The parameter is Collabora-specific and ignored by other app providers.
+// with the target filename. Request that grouped control via `ui_defaults`,
+// which is a semicolon-delimited list, so append rather than overwrite it in
+// case the server already decorates the URL with its own defaults.
 const withSaveAsUiDefaults = (rawUrl: string) => {
   try {
     const url = new URL(rawUrl)
-    if (!url.searchParams.has('ui_defaults')) {
-      url.searchParams.set('ui_defaults', 'SaveAsMode=group')
-    }
+    const existing = url.searchParams.get('ui_defaults')
+    url.searchParams.set(
+      'ui_defaults',
+      existing ? `${existing};SaveAsMode=group` : 'SaveAsMode=group'
+    )
     return url.href
   } catch {
     return rawUrl
@@ -155,14 +162,18 @@ const postToApp = (message: Record<string, unknown>) => {
 // The editor only emits its rich postMessage API (UI_SaveAs, App_LoadingStatus,
 // ...) once the host has announced itself with `Host_PostmessageReady`. Without
 // this handshake "Save As" silently does nothing. Send it as soon as the iframe
-// has loaded.
+// has loaded. This handshake is Collabora-specific; other providers don't expect it.
 const onIframeLoad = () => {
+  if (!unref(isCollabora)) {
+    return
+  }
   postToApp({ MessageId: 'Host_PostmessageReady', Values: {} })
 }
 
-// Reject empty names and any path separators before the name is sent on. The
-// server-side `PutRelativeFile` sanitizes the name as well, but validating here
-// gives immediate feedback and avoids a pointless round-trip on invalid input.
+// Reject empty names, path separators and the reserved "." / ".." names before
+// the name is sent on. The server-side `PutRelativeFile` sanitizes the name as
+// well, but validating here gives immediate feedback and avoids a pointless
+// round-trip on invalid input.
 const validateSaveAsFilename = (filename: string): string => {
   const trimmed = filename?.trim()
   if (!trimmed) {
@@ -171,6 +182,9 @@ const validateSaveAsFilename = (filename: string): string => {
   if (/[/\\]/.test(trimmed)) {
     return $gettext('The file name cannot contain "/" or "\\".')
   }
+  if (trimmed === '.' || trimmed === '..') {
+    return $gettext('The file name cannot be "." or "..".')
+  }
   return ''
 }
 
@@ -178,6 +192,10 @@ const validateSaveAsFilename = (filename: string): string => {
 // extension selects the export format) and reply with `Action_SaveAs`, which
 // makes the editor render to that format and PutRelativeFile it into the space.
 const onSaveAs = () => {
+  if (props.isReadOnly) {
+    showErrorMessage({ title: $gettext('Cannot save a copy: file is read-only') })
+    return
+  }
   dispatchModal({
     variation: 'passive',
     title: $gettext('Save a copy'),
@@ -261,7 +279,9 @@ const loadAppUrl = useTask(function* (signal, viewMode: string) {
       throw new Error('Error in app server response')
     }
 
-    appUrl.value = withSaveAsUiDefaults(response.data.app_url)
+    appUrl.value = unref(isCollabora)
+      ? withSaveAsUiDefaults(response.data.app_url)
+      : response.data.app_url
     method.value = response.data.method
 
     if (response.data.form_parameters) {
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -121,10 +121,12 @@ const subm: VNodeRef = ref()
 const appIframe = ref<HTMLIFrameElement>()
 
 // Origin of the editor (e.g. the Collabora host) used as the target origin when
-// posting messages into the iframe.
+// posting messages into the iframe. Resolved against the current origin so a
+// relative `app_url` (WOPI server co-hosted with the web app) still yields a
+// usable origin instead of permanently failing the postMessage origin check.
 const appOrigin = computed(() => {
   try {
-    return new URL(unref(appUrl)).origin
+    return new URL(unref(appUrl), window.location.origin).origin
   } catch {
     return ''
   }
@@ -151,12 +153,20 @@ const withSaveAsUiDefaults = (rawUrl: string) => {
 }
 
 // Post a WOPI postMessage into the editor iframe (messages are JSON strings).
-const postToApp = (message: Record<string, unknown>) => {
-  const target = unref(appIframe)?.contentWindow
-  if (!target || !unref(appOrigin)) {
-    return
+// Accepts an explicit target/origin so callers that captured them at the time
+// a user action started (e.g. opening the Save-As modal) don't re-read the
+// live refs, which may have moved on to a different resource by the time the
+// user confirms. Returns whether the message was actually sent.
+const postToApp = (
+  message: Record<string, unknown>,
+  target: Window | null | undefined = unref(appIframe)?.contentWindow,
+  origin: string = unref(appOrigin)
+): boolean => {
+  if (!target || !origin) {
+    return false
   }
-  target.postMessage(JSON.stringify({ ...message, SendTime: Date.now() }), unref(appOrigin))
+  target.postMessage(JSON.stringify({ ...message, SendTime: Date.now() }), origin)
+  return true
 }
 
 // The editor only emits its rich postMessage API (UI_SaveAs, App_LoadingStatus,
@@ -188,6 +198,11 @@ const validateSaveAsFilename = (filename: string): string => {
   return ''
 }
 
+// Guards against a second Save-As modal being dispatched while one is already
+// open (e.g. the editor's grouped Save-As control emitting `UI_SaveAs` twice
+// in quick succession).
+const isSaveAsModalOpen = ref(false)
+
 // Handle the editor's "Save As" request: ask the user for the copy's name (the
 // extension selects the export format) and reply with `Action_SaveAs`, which
 // makes the editor render to that format and PutRelativeFile it into the space.
@@ -196,6 +211,18 @@ const onSaveAs = () => {
     showErrorMessage({ title: $gettext('Cannot save a copy: file is read-only') })
     return
   }
+  if (unref(isSaveAsModalOpen)) {
+    return
+  }
+
+  // Capture the target iframe/origin now, at dispatch time, rather than
+  // re-reading the (reactive) refs in `onConfirm`: if the component gets
+  // reused for a different resource while the modal is open, the live refs
+  // would point at the new document by the time the user presses Save.
+  const target = unref(appIframe)?.contentWindow
+  const origin = unref(appOrigin)
+
+  isSaveAsModalOpen.value = true
   dispatchModal({
     variation: 'passive',
     title: $gettext('Save a copy'),
@@ -206,16 +233,24 @@ const onSaveAs = () => {
     onInput: (filename: string, setError: (error: string) => void) => {
       setError(validateSaveAsFilename(filename))
     },
+    onCancel: () => {
+      isSaveAsModalOpen.value = false
+    },
     onConfirm: (filename: string) => {
+      isSaveAsModalOpen.value = false
       // Guard again on confirm (defense-in-depth); onInput normally prevents
       // reaching here with an invalid name.
       if (validateSaveAsFilename(filename)) {
         return
       }
-      postToApp({
-        MessageId: 'Action_SaveAs',
-        Values: { Filename: filename.trim(), Notify: true }
-      })
+      const sent = postToApp(
+        { MessageId: 'Action_SaveAs', Values: { Filename: filename.trim(), Notify: true } },
+        target,
+        origin
+      )
+      if (!sent) {
+        showErrorMessage({ title: $gettext('Cannot save a copy: the editor is not available') })
+      }
     }
   })
 }
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -160,6 +160,20 @@ const onIframeLoad = () => {
   postToApp({ MessageId: 'Host_PostmessageReady', Values: {} })
 }
 
+// Reject empty names and any path separators before the name is sent on. The
+// server-side `PutRelativeFile` sanitizes the name as well, but validating here
+// gives immediate feedback and avoids a pointless round-trip on invalid input.
+const validateSaveAsFilename = (filename: string): string => {
+  const trimmed = filename?.trim()
+  if (!trimmed) {
+    return $gettext('The file name cannot be empty.')
+  }
+  if (/[/\\]/.test(trimmed)) {
+    return $gettext('The file name cannot contain "/" or "\\".')
+  }
+  return ''
+}
+
 // Handle the editor's "Save As" request: ask the user for the copy's name (the
 // extension selects the export format) and reply with `Action_SaveAs`, which
 // makes the editor render to that format and PutRelativeFile it into the space.
@@ -171,8 +185,19 @@ const onSaveAs = () => {
     hasInput: true,
     inputValue: props.resource.name,
     inputLabel: $gettext('File name'),
+    onInput: (filename: string, setError: (error: string) => void) => {
+      setError(validateSaveAsFilename(filename))
+    },
     onConfirm: (filename: string) => {
-      postToApp({ MessageId: 'Action_SaveAs', Values: { Filename: filename, Notify: true } })
+      // Guard again on confirm (defense-in-depth); onInput normally prevents
+      // reaching here with an invalid name.
+      if (validateSaveAsFilename(filename)) {
+        return
+      }
+      postToApp({
+        MessageId: 'Action_SaveAs',
+        Values: { Filename: filename.trim(), Notify: true }
+      })
     }
   })
 }
@@ -261,7 +286,14 @@ const determineOpenAsPreview = (appName: string) => {
 // Single handler for the editor's postMessage events. Only messages coming from
 // the editor's own origin are accepted.
 const onAppMessage = (event: MessageEvent) => {
-  if (unref(appOrigin) && event.origin !== unref(appOrigin)) {
+  // Fail closed: reject every message until the editor origin is known (i.e.
+  // `appUrl` has resolved to an absolute URL) and accept only messages from that
+  // exact origin. The listener is attached on mount, before the WOPI `open_url`
+  // POST resolves, so an empty `appOrigin` must reject rather than wave messages
+  // through. No legitimate editor message can arrive before `appUrl` is set (the
+  // iframe cannot have loaded yet), so this is strictly correct and also covers a
+  // backend returning a non-absolute `app_url`.
+  if (!unref(appOrigin) || event.origin !== unref(appOrigin)) {
     return
   }
   let message: { MessageId?: string }
--- a/packages/web-app-external/src/App.vue
+++ b/packages/web-app-external/src/App.vue
@@ -1,11 +1,13 @@
 <template>
   <iframe
     v-if="appUrl && method === 'GET'"
+    ref="appIframe"
     :src="appUrl"
     class="oc-width-1-1 oc-height-1-1"
     :title="iFrameTitle"
     allowfullscreen
     allow="camera; clipboard-read; clipboard-write"
+    @load="onIframeLoad"
   />
   <div v-if="appUrl && method === 'POST' && formParameters" class="oc-height-1-1 oc-width-1-1">
     <form :action="appUrl" target="app-iframe" method="post">
@@ -15,19 +17,32 @@
       </div>
     </form>
     <iframe
+      ref="appIframe"
       name="app-iframe"
       :src="appUrl"
       class="oc-width-1-1 oc-height-1-1"
       :title="iFrameTitle"
       allowfullscreen
       allow="camera; clipboard-read; clipboard-write"
+      @load="onIframeLoad"
     />
   </div>
 </template>
 
 <script lang="ts" setup>
 import { stringify } from 'qs'
-import { computed, inject, unref, nextTick, ref, watch, VNodeRef, onMounted, type Ref } from 'vue'
+import {
+  computed,
+  inject,
+  unref,
+  nextTick,
+  ref,
+  watch,
+  VNodeRef,
+  onMounted,
+  onBeforeUnmount,
+  type Ref
+} from 'vue'
 import { useTask } from 'vue-concurrency'
 import { useGettext } from 'vue3-gettext'
 
@@ -38,6 +53,7 @@ import {
   useCapabilityStore,
   useConfigStore,
   useMessages,
+  useModals,
   useRequest,
   useAppProviderService,
   useRoute,
@@ -67,6 +83,7 @@ const props = defineProps<Props>()
 const language = useGettext()
 const { $gettext } = language
 const { showErrorMessage } = useMessages()
+const { dispatchModal } = useModals()
 const capabilityStore = useCapabilityStore()
 const configStore = useConfigStore()
 const route = useRoute()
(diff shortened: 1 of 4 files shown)


kw-security:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/06332f87-653d-4cb0-8018-2c740b8822c3?refs=b48d42e03c2765ba2b3102f31c12caceba549148&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/06332f87-653d-4cb0-8018-2c740b8822c3/license?refs=b48d42e03c2765ba2b3102f31c12caceba549148&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/3cbab7ff-f737-4396-848e-40e030758732/pr-check/81972e8d-ac25-469e-9253-988aa8ffa79e?refs=b48d42e03c2765ba2b3102f31c12caceba549148&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

DeepDiver1975 (commented):
Reviewed as maintainer. The WOPI Save-As approach is correct and well-explained: Collabora delegates "Save As" to the host via `UI_SaveAs`/`Action_SaveAs`, the `Host_PostmessageReady` handshake is what opens that channel, and `PutRelativeFile` is already implemented server-side. The refactor of the old `catchClickMicrosoftEdit` into a single `onAppMessage` switch is a clean improvement (and now correctly registers/unregisters the listener and re-evaluates `openAsPreview` reactively rather than once at mount). The changelog fragment is present and follows `changelog/TEMPLATE`. CI is green. Nice work.

I have one requested change before approving, plus a test gap.

### 1. (requested) Origin check fails open while `appUrl` is unresolved

```js
const onAppMessage = (event: MessageEvent) => {
  if (unref(appOrigin) && event.origin \!== unref(appOrigin)) {
    return
  }
  ...
```

`appOrigin` is a computed over `appUrl`, which is `undefined` until the async `loadAppUrl` task resolves; on parse failure it also yields `''`. Because an empty `appOrigin` is falsy, the `&&` short-circuits and **the origin check is skipped entirely** — every origin's message is dispatched. The `message` listener is attached in `onMounted`, which fires while the WOPI `open_url` POST is still in flight (network latency, the 425 "still processing" path, a slow IdP). During that window any page holding a reference to this window can post `{"MessageId":"UI_SaveAs"}` (→ unsolicited "Save a copy" modal) or `{"MessageId":"UI_Edit"}` (→ forced view→write reopen of the user's own file). It also disables the check for the whole session if the backend ever returns a non-absolute `app_url`.

Impact is low (no cross-origin exfiltration — `postToApp` correctly fails closed on empty origin and never uses `targetOrigin:'*'`; the actions only affect the user's own file). But this is exactly the fail-open pattern the recent embed-mode CSRF fix (#13844, `security-embed-mode-postmessage-origin-validation.md`) set out to eliminate, so it's worth holding the same line here. The fix is to fail closed:

```js
if (\!unref(appOrigin) || event.origin \!== unref(appOrigin)) {
  return
}
```

No legitimate editor message can arrive before `appUrl` is set (the editor iframe can't have loaded yet), so rejecting until the origin is known is strictly correct and closes both the load-race window and the bad-`app_url` case.

### 2. (test gap) No unit coverage for the new handler

The only test delta is the two updated snapshots for the `ui_defaults` URL param. The new behavior — origin filtering in `onAppMessage`, `UI_SaveAs` → modal → `Action_SaveAs` payload, and `Host_PostmessageReady` on iframe load — has no unit test. The PR body credits the `webOffice` e2e suite for the end-to-end chain, which is great, but the origin-check logic (precisely the part flagged in #1) is cheap to assert at the unit level and would lock in the fail-closed behavior so it can't regress. Please add a small test that (a) drops a message from a foreign origin, (b) accepts `UI_SaveAs` from the editor origin and asserts the modal/`Action_SaveAs` payload, and (c) asserts `Host_PostmessageReady` is posted on `@load`.

### Minor (non-blocking)
- The modal accepts any filename and posts it straight into `Action_SaveAs.Values.Filename`. Server-side `PutRelativeFile` is responsible for sanitizing the name (strip path separators / leading dots), which I trust it does — but a small client-side guard against empty/`/`/`\` names in `onConfirm` would be reasonable defense-in-depth and avoids a pointless round-trip.

Happy to approve once the origin guard fails closed (#1); the test (#2) would ideally land in the same PR. Thanks\!

dj4oC:
Thanks for the thorough review @DeepDiver1975 — all addressed in ed7188f9a.

**1. (requested) Origin check now fails closed.** `onAppMessage` rejects every message until the editor origin is known and accepts only that exact origin:

```js
if (!unref(appOrigin) || event.origin !== unref(appOrigin)) {
  return
}
```

This closes the load-race window (the listener is attached on mount, before the `open_url` POST resolves) and the non-absolute `app_url` case. As you noted, no legitimate editor message can arrive before `appUrl` is set, so rejecting until the origin is known is strictly correct.

**2. (test gap) Unit tests added** in `app.spec.ts` for the new handler:
- a message from a foreign origin is dropped (no modal);
- a message arriving before the origin is known is dropped (the fail-closed path from #1);
- `UI_SaveAs` from the editor origin opens the modal and, on confirm, posts `Action_SaveAs` with the expected payload and `targetOrigin`;
- invalid file names are rejected;
- `Host_PostmessageReady` is posted on iframe `@load`.

**3. (minor) Client-side filename guard added.** The modal now validates via `onInput` (live, using the repo's `setError` pattern) and re-checks in `onConfirm`: empty names and any `/` or `\` are rejected before `Action_SaveAs` is posted. Server-side `PutRelativeFile` still sanitizes; this is defense-in-depth and avoids the pointless round-trip.

Local: `check:types`, `lint`, `check:format`, and `test:unit` for `web-app-external` (8 passing) are all green.

DeepDiver1975:
Re-reviewed `ed7188f9a` (changed code only). Both points from my previous review are resolved — thanks @dj4oC.

**1. Origin guard now fails closed.** `onAppMessage` is now `if (\!unref(appOrigin) || event.origin \!== unref(appOrigin)) return`, so messages are dropped until the editor origin is resolved to an absolute URL, instead of being waved through when `appOrigin` is empty/unparseable. This is the fix I asked for, and the inline comment correctly captures why (listener attached on mount before the WOPI `open_url` POST resolves). 

**2. Tests added.** The new `app.spec.ts` cases cover the security behavior directly — `ignores editor messages until the editor origin is known (fail closed)` exercises the unresolved-`appUrl` path, plus foreign-origin drop, the happy-path `Action_SaveAs` round-trip, the empty / path-separator filename rejection, and `Host_PostmessageReady`. Good coverage.

The added `validateSaveAsFilename` (reject empty + `/`/`\`, guarded on both `onInput` and `onConfirm`, `.trim()` before send) is sensible defense-in-depth on top of the server-side `PutRelativeFile` sanitization.

No remaining concerns on the changed code. CI is green across the board. LGTM from my side.

packages/web-app-external/src/App.vue
```diff
@@ -97,6 +114,93 @@ const appUrl = ref()
...
+
+// Collabora exposes "Save As" (export to another format, saved back to storage
+// via WOPI PutRelativeFile) as a host-delegated operation: it shows a grouped
+// Save-As control and posts a `UI_SaveAs` message, expecting the host to reply
+// with the target filename. Request that grouped control via `ui_defaults`.
+// The parameter is Collabora-specific and ignored by other app providers.
+const withSaveAsUiDefaults = (rawUrl: string) => {
+  try {
+    const url = new URL(rawUrl)
+    if (!url.searchParams.has('ui_defaults')) {
```
LukasHirt: **Bug (CONFIRMED):** `url.searchParams.has('ui_defaults')` returns `true` for any deployment where the server already sets `ui_defaults` (e.g. `UIMode=notebookbar`), so the `if` body is skipped and `SaveAsMode=group` is never added. Collabora's `ui_defaults` is a semicolon-separated list, not a single value — the correct fix is to append rather than set:

```ts
const existing = url.searchParams.get('ui_defaults')
url.searchParams.set(
  'ui_defaults',
  existing ? `${existing};SaveAsMode=group` : 'SaveAsMode=group'
)
```

With the current code the Save As feature is silently unavailable for a significant class of Collabora deployments, with no error or UI indication.

packages/web-app-external/src/App.vue
```diff
@@ -97,6 +114,93 @@ const appUrl = ref()
...
+  if (/[/\\]/.test(trimmed)) {
+    return $gettext('The file name cannot contain "/" or "\\".')
+  }
+  return ''
+}
+
+// Handle the editor's "Save As" request: ask the user for the copy's name (the
+// extension selects the export format) and reply with `Action_SaveAs`, which
+// makes the editor render to that format and PutRelativeFile it into the space.
+const onSaveAs = () => {
```
LukasHirt, line 209: **Bug (CONFIRMED):** `onSaveAs` has no `isReadOnly` guard, whereas `loadAppUrl` (line 221) already short-circuits with a user-visible error for read-only sessions. When `isReadOnly` is true, `withSaveAsUiDefaults` still adds `SaveAsMode=group` to the URL (see line 264), so Collabora may still render the Save-As control and emit `UI_SaveAs`. The user completes the modal, `Action_SaveAs` is posted fire-and-forget, the WOPI server rejects `PutRelativeFile`, the modal closes normally, and the user sees no error while no copy is created.

Add the guard at the top of `onSaveAs`:

```ts
if (props.isReadOnly) {
  showErrorMessage({ title: $gettext('Cannot save a copy: file is read-only') })
  return
}
```

packages/web-app-external/src/App.vue
```diff
@@ -97,6 +114,93 @@ const appUrl = ref()
...
+    return
+  }
+  target.postMessage(JSON.stringify({ ...message, SendTime: Date.now() }), unref(appOrigin))
+}
+
+// The editor only emits its rich postMessage API (UI_SaveAs, App_LoadingStatus,
+// ...) once the host has announced itself with `Host_PostmessageReady`. Without
+// this handshake "Save As" silently does nothing. Send it as soon as the iframe
+// has loaded.
+const onIframeLoad = () => {
```
LukasHirt, line 176: **Design issue (CONFIRMED):** `onIframeLoad` posts `Host_PostmessageReady` for every WOPI provider, not just Collabora. Combined with line 264 unconditionally calling `withSaveAsUiDefaults` on every provider's `app_url`, all non-Collabora WOPI editors receive an unsolicited `Host_PostmessageReady` message and have `ui_defaults=SaveAsMode=group` appended to their URLs on every load. While most providers ignore both, any future provider that interprets `UI_SaveAs` differently will incorrectly trigger the Collabora Save-As modal flow.

Both `withSaveAsUiDefaults` and the `Host_PostmessageReady` send should be gated on the provider being Collabora (e.g. checking `appName` or a capability flag). The inline comment on line 133 (`// The parameter is Collabora-specific`) acknowledges this but doesn't enforce it in code.

packages/web-app-external/src/App.vue
```diff
@@ -97,6 +114,93 @@ const appUrl = ref()
...
+  postToApp({ MessageId: 'Host_PostmessageReady', Values: {} })
+}
+
+// Reject empty names and any path separators before the name is sent on. The
+// server-side `PutRelativeFile` sanitizes the name as well, but validating here
+// gives immediate feedback and avoids a pointless round-trip on invalid input.
+const validateSaveAsFilename = (filename: string): string => {
```
LukasHirt, line 187: **Code quality (CONFIRMED):** `validateSaveAsFilename` implements its own empty-name and path-separator checks, but `web-pkg` already exports `getNameErrorMsg` (used by `useFileActionsCreateNewFile` and `useFileActionsRename`). Those validators also reject `.`, `..`, and other reserved names that `validateSaveAsFilename` currently allows. Save-As now silently permits names that the rename and create flows reject, and the validators will continue to diverge as the shared helper evolves.

Consider reusing `getNameErrorMsg` from `@ownclouders/web-pkg` instead of implementing a third copy.

packages/web-app-external/src/App.vue
```diff
@@ -179,20 +283,42 @@ const determineOpenAsPreview = (appName: string) => {
...
+    return
+  }
+  switch (message?.MessageId) {
+    case 'UI_Edit':
+      // switch to write mode when edit is clicked
+      if (determineOpenAsPreview(unref(appName))) {
+        loadAppUrl.perform('write')
+      }
+      break
+    case 'UI_SaveAs':
```
LukasHirt, line 367: **Plausible hazard:** No re-entrancy guard on `UI_SaveAs`. If Collabora emits it twice in quick succession (e.g. user double-clicks the grouped Save-As control before the host responds), `onSaveAs()` is called twice, two concurrent modals open, and if both are confirmed, two `Action_SaveAs` messages are posted, triggering two `PutRelativeFile` calls. A simple boolean flag or a check for an already-open modal (`useModals` exposes the modal stack) would prevent this.

packages/web-app-external/src/App.vue
```diff
@@ -97,6 +114,93 @@ const appUrl = ref()
 const formParameters = ref({})
 const method = ref()
 const subm: VNodeRef = ref()
+const appIframe = ref<HTMLIFrameElement>()
+
+// Origin of the editor (e.g. the Collabora host) used as the target origin when
+// posting messages into the iframe.
+const appOrigin = computed(() => {
```
LukasHirt, line 127: **Plausible hazard:** `appOrigin` is `''` for relative `app_url` values (the `URL` constructor throws for relative paths with no base, hitting the `catch` block). The origin guard at line 296 then drops *all* postMessages — including `UI_Edit` — for the entire session. Any deployment where the WOPI server returns a relative `app_url` would silently break the Edit-button flow. The `try/catch` in `withSaveAsUiDefaults` (line 135) already handles this for the URL transform, but `appOrigin` has no fallback for the same case.

packages/web-app-external/src/App.vue
```diff
@@ -97,6 +114,93 @@ const appUrl = ref()
...
+    variation: 'passive',
+    title: $gettext('Save a copy'),
+    confirmText: $gettext('Save'),
+    hasInput: true,
+    inputValue: props.resource.name,
+    inputLabel: $gettext('File name'),
+    onInput: (filename: string, setError: (error: string) => void) => {
+      setError(validateSaveAsFilename(filename))
+    },
+    onConfirm: (filename: string) => {
```
LukasHirt, line 239: **Plausible hazard (two related issues):**

1. **Silent no-op on null iframe:** `postToApp` returns `undefined` without error when `appIframe.contentWindow` is null (line 149). `onConfirm` doesn't inspect the return value, and `ModalWrapper` unconditionally closes the dialog after `onConfirm` resolves. If `loadAppUrl` restarts (e.g. triggered by the `watch` at line 324) between `UI_SaveAs` receipt and the user pressing Save, `appIframe` can be transiently null — `Action_SaveAs` is never sent, but the user sees the modal close normally and believes the copy was saved.

2. **Reactive-ref stale closure:** `onConfirm` reads `appOrigin` and `appIframe` at the moment the user presses Save (not at modal-dispatch time). In a SPA where the component is reused across resources without remounting, a resource-prop change while the modal is open replaces both refs with values for the new document. The user's Save press then dispatches `Action_SaveAs` to the wrong iframe with the filename typed for the old document.

For (1): `postToApp` could return a boolean indicating success, and `onConfirm` could show an error rather than silently closing.

packages/web-app-external/tests/unit/app.spec.ts
```diff
@@ -72,6 +74,117 @@ describe('The app provider extension', () => {
...
+
+    await iframe.trigger('load')
+
+    expect(postMessage).toHaveBeenCalledTimes(1)
+    const [payload, targetOrigin] = postMessage.mock.calls[0]
+    expect(targetOrigin).toBe(appOrigin)
+    expect(JSON.parse(payload as string)).toMatchObject({ MessageId: 'Host_PostmessageReady' })
+    wrapper.unmount()
+  })
+})
```
LukasHirt, line 289: **Test gap (CONFIRMED):** None of the new tests verify that `UI_Edit` is a no-op when `openAsPreview` is `false`. The old static `catchClickMicrosoftEdit` registration implicitly prevented this; now it's a conditional inside the `UI_Edit` switch case (App.vue line 308). If the guard is accidentally inverted, any authenticated Collabora instance can post `UI_Edit` to force a write-mode reload of any file regardless of configuration — and no existing test would catch it.

Please add a test along the lines of:
```ts
it('ignores UI_Edit when openAsPreview is false', async () => {
  // configure openAsPreview: false
  // dispatch UI_Edit from editor origin
  // assert loadAppUrl was not called with 'write'
})
```

packages/web-app-external/tests/unit/app.spec.ts
```diff
@@ -72,6 +74,117 @@ describe('The app provider extension', () => {
   })
 })
 
+describe('Collabora Save As postMessage handling', () => {
+  const successGet = { ok: true, status: 200, data: providerSuccessResponseGet }
+
+  const flushAppUrl = async (wrapper: any) => {
```
LukasHirt: **Plausible hazard:** `flushAppUrl` is a fixed loop of 5 `$nextTick()` calls. `flushPromises` from `@ownclouders/web-test-helpers` is already exported from the same entry point that `shallowMount` is imported from on line 2 (and used in the sibling `Redirect.spec.ts`). If `loadAppUrl` gains another `yield` (another async step), the hardcoded count silently falls short and all four `flushAppUrl`-gated tests become intermittently unreliable with no compile-time warning.

Suggestion: replace `flushAppUrl` with `await flushPromises()` throughout.

LukasHirt (changes requested):
Ran a multi-agent review (28 verifier agents, high effort) across the two changed files. Found **4 confirmed bugs / design issues** that warrant changes before merge, plus **6 plausible hazards** worth noting. Grouped by severity below.

---

### Confirmed (request changes)
1. **`withSaveAsUiDefaults` silently no-ops when `ui_defaults` already in the URL** — feature entirely broken for Collabora deployments where the server decorates the URL with its own `ui_defaults` (e.g. `UIMode=notebookbar`). `searchParams.has` returns `true`, the branch is skipped, `SaveAsMode=group` is never appended, Collabora never sends `UI_SaveAs`. Should append to the semicolon-delimited list instead of replacing.
2. **`onSaveAs` has no `isReadOnly` guard** — `loadAppUrl` already short-circuits with a visible error for read-only files, but `onSaveAs` dispatches the modal regardless. When the user confirms, `Action_SaveAs` is posted; WOPI rejects `PutRelativeFile` silently; the modal closes normally. User sees success, no copy is created.
3. **Collabora-specific protocol applied unconditionally to all WOPI providers** — `withSaveAsUiDefaults` and `onIframeLoad → Host_PostmessageReady` fire for every provider regardless of which editor is active. `ui_defaults=SaveAsMode=group` is appended to every `app_url`, and `Host_PostmessageReady` is posted on every iframe load. Should be gated on `appName` indicating Collabora.
4. **`validateSaveAsFilename` duplicates shared validator logic** — already diverges from `getNameErrorMsg` in `useFileActionsCreateNewFile`/`useFileActionsRename` (those also reject `.`, `..`, and reserved names). Adding a third implementation means they drift further over time.

### Confirmed (test gap)
5. **No test that `UI_Edit` is a no-op when `openAsPreview` is false** — the old static guard prevented this case entirely; now it's a conditional inside the switch. If the guard is accidentally inverted, any authenticated Collabora instance can force a write-mode reload of any file. No existing test would catch the regression.

### Plausible hazards (worth addressing)
6. **Relative `app_url` silently breaks `UI_Edit`** — `appOrigin` computes `''` for relative URLs; every postMessage (including `UI_Edit`) is then dropped, so the Edit button stops working for any deployment where the WOPI server co-hosts on the same origin as the web app.
7. **Silent `postToApp` null + modal dismissed as success** — if `loadAppUrl` restarts between `UI_SaveAs` receipt and the user pressing Save, `appIframe` can be transiently null; `postToApp` no-ops silently but `ModalWrapper` still closes the dialog, giving false positive feedback.
8. **`onConfirm` closure over live reactive refs** — in a SPA where the component is reused across resources without remounting, a resource-prop change while the modal is open rewrites `appUrl`/`appIframe`; the user's Save press dispatches `Action_SaveAs` to the new document.
9. **`UI_SaveAs` can double-fire** — no re-entrancy guard; rapid double-click of Collabora's export button opens two concurrent Save As dialogs and sends two `Action_SaveAs` messages, causing two `PutRelativeFile` calls.
10. **`flushAppUrl` hardcoded to 5 ticks instead of `flushPromises`** — if `loadAppUrl` gains another async step the count silently falls short, making tests intermittently unreliable.

packages/web-app-external/src/App.vue
LukasHirt: Fixed in fabe69253: `withSaveAsUiDefaults` now reads the existing `ui_defaults` value and appends `;SaveAsMode=group` to it instead of skipping the param entirely when it's already present.

packages/web-app-external/src/App.vue
LukasHirt: Fixed in fabe69253: `onSaveAs` now short-circuits with an error message when `props.isReadOnly` is true, before dispatching the modal.

packages/web-app-external/src/App.vue
LukasHirt: Fixed in fabe69253: added an `isCollabora` computed (`appName === 'Collabora'`) and gated both the `ui_defaults` request and the `Host_PostmessageReady` handshake behind it, so non-Collabora WOPI providers are unaffected. Updated/added unit tests to cover both the Collabora and non-Collabora paths.

packages/web-app-external/src/App.vue
LukasHirt: Looked into reusing `getNameErrorMsg`, but neither existing implementation is a drop-in fit here: `useFileActionsRename`'s version needs a `Resource` + parent-path context (this is a brand-new copy name, not a rename), and `useFileActionsCreateNewFile`'s version needs the current-folder `resources` list from `useResourcesStore` for duplicate-name detection, plus both composables pull in unrelated dependencies (modals, messages, embed-mode, apps store, client service) that this component doesn't otherwise need. Neither of them actually rejects `\`, which `validateSaveAsFilename` does and I don't want to lose.

Instead of a literal reuse, I closed the actual behavioral gap in fabe69253 by adding the `.`/`..` rejection to `validateSaveAsFilename`, so it now matches the other validators' character-level rules while staying a plain, dependency-free string function. Let me know if you'd still prefer extracting a shared pure helper — happy to do that as a follow-up if so, but felt out of scope for this PR.

packages/web-app-external/src/App.vue
LukasHirt: Fixed in b48d42e03: added an `isSaveAsModalOpen` guard that's set on dispatch and cleared in `onCancel`/`onConfirm`, so a second `UI_SaveAs` while the modal is open is ignored. Added a test for this.

packages/web-app-external/src/App.vue
LukasHirt: Fixed in b48d42e03: `appOrigin` now resolves `appUrl` against `window.location.origin` as the base (`new URL(unref(appUrl), window.location.origin).origin`), so a relative `app_url` resolves to the current origin instead of throwing and permanently disabling the postMessage origin check.

packages/web-app-external/src/App.vue
LukasHirt: Fixed both parts in b48d42e03:
1. `postToApp` now returns a boolean indicating whether the message was actually sent, and `onConfirm` shows an error instead of silently letting the modal close when it returns `false`.
2. `onSaveAs` now captures `appIframe`/`appOrigin` into local consts at the moment the modal is dispatched and passes them explicitly into `postToApp`, instead of re-reading the live refs inside `onConfirm` — so a resource change while the modal is open can no longer redirect the Save-As to the wrong document.

Added tests for both.

packages/web-app-external/tests/unit/app.spec.ts
LukasHirt: Fixed in fabe69253: added a test asserting `UI_Edit` does not call `loadAppUrl`/`makeRequest` again when `openAsPreview` is `false`.

packages/web-app-external/tests/unit/app.spec.ts
LukasHirt: Fixed in b48d42e03: replaced the hardcoded 5x `$nextTick()` loop with `flushPromises()` from `@ownclouders/web-test-helpers` (same helper used in the sibling `Redirect.spec.ts`) throughout the Collabora Save-As test block.

LukasHirt:
Re-reviewed after `fabe69253` and `b48d42e03`. All confirmed issues from my previous review are addressed, plus the plausible hazards I flagged:

**Confirmed (were blocking):**
1. `ui_defaults` append bug — fixed, now appends to the existing semicolon-delimited list instead of skipping.
2. Missing `isReadOnly` guard on `onSaveAs` — fixed.
3. Collabora-specific behavior applied to all providers — fixed, gated behind an `isCollabora` computed.
4. `validateSaveAsFilename` duplication — not a literal `getNameErrorMsg` reuse (neither implementation is a clean fit; see inline reply), but the character-level gap is closed by adding `.`/`..` rejection.
5. Test gap for `UI_Edit` no-op — fixed, test added.

**Plausible hazards (also addressed):**
6. Relative `app_url` permanently breaking the origin check — fixed via `window.location.origin` base.
7. No re-entrancy guard on `UI_SaveAs` — fixed.
8. Silent no-op / stale-closure risk on `postToApp`/`onConfirm` — fixed, `postToApp` now returns success and the target/origin are captured at dispatch time.
9. Hardcoded 5-tick `flushAppUrl` — replaced with `flushPromises()`.

Local `check:types`, `lint`, `check:format`, and `test:unit` for `web-app-external` are all green (13 passing in `app.spec.ts`).

LukasHirt (approved):

